PR #2, opened 2018-03-21
using args for xu daemon

sangaman:
This commit properly passes the command line arguments for `xunion` to
the `Config` constructor in a format that will override existing
configuration parameters.

### DIFF
--- a/bin/xunion
+++ b/bin/xunion
@@ -2,16 +2,18 @@
 
 const XUnion = require('../lib/xunion');
 const commandLineArgs = require('command-line-args');
+const util = require('../lib/utils/util');
 
 const argDefinitions = [
-  { name: 'peerport', type: Number },
-  { name: 'rpcport', type: Number },
-  { name: 'database', type: String },
+  { name: 'p2p.port', type: Number },
+  { name: 'rpcPort', type: Number },
+  { name: 'db.database', type: String },
   { name: 'xudir', type: String },
   { name: 'lnddir', type: String },
 ];
 const args = commandLineArgs(argDefinitions);
 
-const xUnion = new XUnion(args);
+const modifiedArgs = util.modifyArgs(args);
+const xUnion = new XUnion(modifiedArgs);
 
 xUnion.start();
--- a/lib/utils/util.js
+++ b/lib/utils/util.js
@@ -14,6 +14,26 @@ util.copyObject = function(source, dest) {
     }
 };
 
+util.modifyArgs = (args) => {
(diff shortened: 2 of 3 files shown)


bin/xunion
```diff
@@ -4,14 +4,31 @@ const XUnion = require('../lib/xunion');
...
+  { name: 'p2p.port', type: Number },
+  { name: 'rpcPort', type: Number },
   { name: 'database', type: String },
   { name: 'xudir', type: String },
   { name: 'lnddir', type: String },
 ];
 const args = commandLineArgs(argDefinitions);
 
-const xUnion = new XUnion(args);
+const argKeys = Object.keys(args);
```
moshababo: Rather export this processing to a generic util function.
Benefits:
- reusable
- testable
- easier to understand what it's doing

moshababo:
I merged my PR so you can use the util structure I've done.
So please sync your branch with master first.

bin/xunion
sangaman: @moshababo Thanks, done.

lib/utils/util.js
```diff
@@ -14,6 +14,26 @@ util.copyObject = function(source, dest) {
     }
 };
 
+util.modifyArgs = (args) => {
```
moshababo: 1. better to use a name which describes what is the modification that is being done.
1. lets stick in this file to functions or to arrow functions. I preferred functions because it's a more recognizable syntax, and it doesn't hint that the function is dependent on a `this` context - the main reason for using an arrow function.

sangaman: 1. What name do you propose? I have no better ideas.
2. The other function in the file throws a linter warning for an unexpected unnamed function which is why I went with a different approach. So we can name both of them, but I prefer the arrow functions because they are more concise.

sangaman: Some function name ideas: "argsToOptions" , "argsToConfig" , "unflattenArgs" , "structureArgs"

Either way I'll be documenting this method (and others) shortly, it's just hard to describe in a function name.

moshababo:
@sangaman 
- util file was changed to use arrow functions.
- I changed `modifyArgs` to `nestObject`. 
it might be easier and better to write it recursively, but it's not that important for now. 

Feel free to marge.

sangaman:
I undid the merge commit because it overwrote a couple of lines in `xunion.js` and just resolved the conflicts instead.

Re: making the nestObject method recursive, I agree that is a better approach but I'll leave it for a future commit.

PR #3, opened 2018-03-22
Miscellaneous

moshababo:


lib/db/db.js
```diff
@@ -31,15 +35,19 @@ class DB {
...
+      const {dialectName} = this.sequelize.connectionManager;
+      this.logger.info(`connected to database. host:${host} port:${port} database:${database} dialect:${dialectName}`);
     } catch (err) {
-      this.logger.error('unable to connect to the database: ', err);
+      this.logger.error('unable to connect to the database', err);
+      throw err;
     }
 
     // sync schemas with the database
+    const options = {logging: this.logger.info};
```
sangaman: I was getting weird (and very verbose) logging output when I tried passing in the logger previously, have you been able to test this at all to see what the output looks like?

lib/rpcserver/Service.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+class Service {
```
sangaman: Maybe call this class something like "Methods"? That seems a bit more descriptive.

lib/Logger.js
```diff
@@ -31,24 +32,31 @@ class Logger {
     winston.level = this.level;
   }
 
+  _bind() {
```
sangaman: The default linter rule objects to functions with dangling underscores. Can we just stick the bind statements at the end of the constructor? I'm assuming this will only be done once, and this also ensures that `_bind()` can't be called from any other classes.

lib/orderbook/orderbook.js
```diff
@@ -7,6 +7,12 @@ class OrderBook {
   constructor(db) {
     this.db = db;
     this.logger = Logger.global;
+    this._bind();
+  }
+
+  _bind() {
```
sangaman: Same comment as the other `_bind()` function.

lib/db/db.js
moshababo: it’s verbose even with the default console logger (we can reduce it if it’s disturbing).
Now it’s just using the logger’s color, timestamp, etc. Nothing major.

lib/rpcserver/Service.js
moshababo: Service is the usual name for this kind of layer. Methods sounds even more generic to me, but RpcMethods might work.

lib/Logger.js
moshababo: Yes, I'll change it.

lib/db/db.js
sangaman: OK cool, I thought I was seeing way more output when I set the logger although I did think it was very verbose either way. We can possibly reduce the db logging output in the future.

lib/rpcserver/Service.js
sangaman: I like RpcMethods.

PR #7, opened 2018-03-25
Upgrading winston logger to v3.0

moshababo:
`createLogger` is supported in winston only from v3.0.0-rc0.

### DIFF
--- a/lib/Logger.js
+++ b/lib/Logger.js
@@ -1,7 +1,7 @@
 const fs = require('fs');
 const path = require('path');
 const winston = require('winston');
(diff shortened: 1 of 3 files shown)


moshababo:
@sangaman Just noticed that the logs are now unformatted.

sangaman:
Good catch, I just used the example from winston's master github readme not realizing it was for a new version. I think it makes sense to start using 3.0 though. I fixed a couple of things to make our code play nicely with 3.0 and tested locally that both the console and file logging looks good, take a quick look and merge when you get a chance.

PR #8, opened 2018-03-25
win/mac/linux compatibility

moshababo:


lib/Config.js
```diff
@@ -1,21 +1,35 @@
...
-      const homeDir = process.env.HOME;
-      this.xuDir = `${homeDir}/.xunion/`;
-      this.lndDir = `${homeDir}/.lnd/`; // default lnd directory location
+    const platform = os.platform();
+    switch (platform) {
+      case 'win32': { // windows
+        const homeDir = process.env.LOCALAPPDATA;
+        this.xuDir = `${homeDir}/XUnion/`;
+        this.lndDir = `${homeDir}/Lnd/`; // default lnd directory location
+        this.lndRpcProto = '../rpc.proto';
```
sangaman: I'm wondering here if it would make sense to make the path to rpc.proto configurable, maybe we'd even want to have its default location to be in the xunion directory? The relative paths things seems tricky and may give us trouble in the future.

lib/XUnion.js
```diff
@@ -20,7 +20,7 @@ class XUnion {
     try {
       this.db = new DB(this.config.db);
       this.orderBook = new OrderBook(this.db);
-      this.lndClient = new LndClient(this.config.lndDir);
+      this.lndClient = new LndClient(this.config.lndDir, this.config.lndRpcProto);
```
sangaman: Maybe just pass in the `config` object in its entirety and then read the properties we need in the LndClient constructor?

lib/Config.js
moshababo: I agree. Just wanted to follow the existing situation for now, and make it work. 
We should also change it name, to make it clear it's the LND proto. 

If we're moving it to xunion directory, should the user copy it manually to there, or we should provide some kind of an installation process?
In other words - what's going to create the `xuDir` directory and content?

lib/XUnion.js
moshababo: Definitely.

lib/Config.js
sangaman: Agree with changing the name of the file, `lndrpc.proto` should be fine.

In thinking of this some more, a concern is that if we upgrade to work with a newer lnd interface, we want to be sure that the file gets overwritten. If we put it in a config directory, it wouldn't get overridden when the latest code is pulled down.

Ultimately, we want to be able to load the `.proto` file regardless . On Linux currently if I add the `/bin` folder to my path and try to run `xunion` from outside the xunion directory, it can't find the file. Maybe we could use an environment variable?

Or a setup script might be the way to go, which will copy the file over to the config directory and overwrite if necessary. This might be something useful to have regardless - it could create the xunion user and database automatically for example.

It doesn't have to be fixed in this PR necessarily unless it can be a quick fix, we can merge this and then replace the logic later on.

moshababo: After a second thought, I think that this file should be a part of the code base and not part of xunion directory, unless we can think of a scenario where we want to enable the user to override it, which I don’t see happening.

The relative path issue shouldn’t be a big deal. we will adjust it to different environments in a few iterations. it’s usually supported out of the box in most of Nodejs modules.

PR #10, opened 2018-03-26
Config args

moshababo:
to enable command line args to override config:
* pass args object to config constructor
* override config with args regardless of xunion.conf file

### DIFF
--- a/lib/XUnion.js
+++ b/lib/XUnion.js
@@ -8,12 +8,13 @@ const P2P = require('./p2p/P2P');
 const P2PServer = require('./p2p/P2PServer');
 
 class XUnion {
-  constructor() {
+  constructor(args) {
+    this.args = args;
     this.logger = Logger.global;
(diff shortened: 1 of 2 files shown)


sangaman:
Thanks, I had done this as part of #2 but I must've merged it incorrectly so the changes weren't preserved.

lib/XUnion.js
```diff
@@ -8,12 +8,13 @@ const P2P = require('./p2p/P2P');
...
 class XUnion {
-  constructor() {
+  constructor(args) {
+    this.args = args;
     this.logger = Logger.global;
   }
 
   async start() {
-    this.config = new Config();
+    this.config = new Config(this.args);
```
sangaman: Can you move `this.config = new Config(this.args);` to the constructor? That way we don't need to make `args` a class property. Otherwise looks good to squash and merge.

PR #12, opened 2018-03-28
db schema & automation

moshababo:
basic db schema & automation. 

In the next PR i'll separate the schemas to different files, and improve the gulp automation.


moshababo:
Currencies and Pairs tables id is string and not incremented integer so it will be recognized globally. Just an initial solution for now.

lib/db/DB.js
```diff
@@ -16,19 +19,61 @@ class DB {
...
   define() {
-    this.Order = this.sequelize.define('order', {
-      id: { type: Sequelize.INTEGER, primaryKey: true, autoIncrement: true },
-      peerId: Sequelize.INTEGER,
-      invoice: Sequelize.STRING,
-      quantity: Sequelize.DECIMAL(14, 8),
-      price: Sequelize.DECIMAL(9, 8),
+    // P2P
+
+    this.Peer = this.sequelize.define('peers', {
```
sangaman: We can simply define the model as `peer` as it will automatically get pluralized.

lib/db/DB.js
```diff
@@ -16,19 +19,61 @@ class DB {
...
     });
 
-    this.Peer = this.sequelize.define('peer', {
-      id: { type: Sequelize.INTEGER, primaryKey: true, autoIncrement: true },
-      nodeKey: { type: Sequelize.STRING, unique: true },
-      ipv4: Sequelize.STRING(15),
-      port: Sequelize.SMALLINT,
+    // DOB
+
+    this.Currency = this.sequelize.define('currencies', {
```
sangaman: `currency` will be pluralized (I checked to make sure it properly pluralizes words ending in -y)

lib/db/DB.js
```diff
@@ -16,19 +19,61 @@ class DB {
...
+      id: { type: DataTypes.STRING, primaryKey: true },
+    });
+
+    //
+
+    this.Pair = this.sequelize.define('pairs', {
```
sangaman: `pair` will be pluralized

lib/enums.js
```diff
@@ -0,0 +1,11 @@
+const enums = exports;
```
sangaman: Maybe we can move enums into our utils folder? Eventually I was thinking a constants file can go in that folder as well.

lib/db/DB.js
```diff
@@ -16,19 +19,61 @@ class DB {
...
+      id: { type: DataTypes.STRING, primaryKey: true },
```
sangaman: What are your thoughts here regarding making `id` a string? Would an example id be like `BTCLTC`? I'm thinking we can just make this an auto-incrementing `int` and get the currency pairs from `baseCurrency` and `quoteCurrency`.

lib/db/DB.js
```diff
@@ -16,19 +19,61 @@ class DB {
...
+    this.Pair.belongsTo(this.Currency, {
+      foreignKey: 'quoteCurrency',
+    });
+    const derivePairId = (pair) => {
+      pair.id = `${pair.baseCurrency}/${pair.quoteCurrency}`; // eslint-disable-line no-param-reassign
+    };
+    this.Pair.beforeBulkCreate(pairs => pairs.forEach(pair => derivePairId(pair)));
+    this.Pair.beforeCreate(pair => derivePairId(pair));
+
+    //
```
sangaman: Drop this empty comment.

lib/db/DB.js
```diff
@@ -16,19 +19,61 @@ class DB {
...
+
+    this.Order = this.sequelize.define('orders', {
```
sangaman: `order` will be pluralized.

lib/db/DB.js
```diff
@@ -16,19 +19,61 @@ class DB {
...
+      id: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
+      pairId: { type: DataTypes.STRING, allowNull: false },
```
sangaman: If we change the pairId to an int as I suggested above, we'd want to change this as well.

lib/db/DB.js
```diff
@@ -16,19 +19,61 @@ class DB {
...
+      makerPeerId: { type: DataTypes.INTEGER, allowNull: true },
```
sangaman: Maybe just call this `peerId`? The "maker" portion is implied.

lib/db/DB.js
```diff
@@ -16,19 +19,61 @@ class DB {
...
+      side: {
+        type: DataTypes.ENUM, values: Object.values(enums.orderSides), allowNull: true,
```
sangaman: Rather than specify a side, we can do what some other exchanges out there do which is to have a negative quantity indicate a sell while a positive quantity indicates a buy. One less column for us to track.

lib/db/DB.js
```diff
@@ -16,19 +19,61 @@ class DB {
...
+      },
+      quantity: DataTypes.DECIMAL(14, 8),
+      price: DataTypes.DECIMAL(14, 8),
+    });
+    this.Order.belongsTo(this.Pair, {
+      foreignKey: 'pairId',
+    });
+    this.Order.belongsTo(this.Peer, {
+      foreignKey: 'makerPeerId',
```
sangaman: If we change makerPeerId, we'd want to change it here too.

lib/enums.js
```diff
@@ -0,0 +1,11 @@
...
+
+enums.swapProtocols = {
+  LND: 'LND',
+  RAIDEN: 'RAIDEN',
+};
+
+enums.orderSides = {
```
sangaman: We may be able to drop `orderSides` entirely if we use the convention of negative quantities meaning sells, positive meaning buys.

lib/db/DB.js
moshababo: I prefer to see in the code what I want to get and not be depended on magic tricks, which are most confusing.

lib/db/DB.js
moshababo: same as above

lib/db/DB.js
moshababo: same as above

lib/enums.js
moshababo: `utils` folder will grow to contain more files - utils for specific domains. 

I don't think it should be mixed with enums/constants, which should have they own folder eventually.

lib/db/DB.js
moshababo: it's just a separation between the tables until i'll move it to separate files.

lib/db/DB.js
moshababo: `currencies` and `pairs` should be recognized globally (p2p network), and not by when they were added to a local node.  using the currency string symbol might be the easiest way to do so.

lib/db/DB.js
moshababo: same as above

lib/db/DB.js
moshababo: Yes, i'll fix

lib/db/DB.js
moshababo: @kilrau 
What do you say?

lib/db/DB.js
moshababo: Will be fixed

lib/db/DB.js
```diff
@@ -16,19 +19,61 @@ class DB {
...
+    this.Pair.beforeCreate(pair => derivePairId(pair));
+
+    //
+
+    this.Order = this.sequelize.define('orders', {
+      id: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
+      pairId: { type: DataTypes.STRING, allowNull: false },
+      peerId: { type: DataTypes.INTEGER, allowNull: true },
+      side: {
+        type: DataTypes.ENUM, values: Object.values(enums.orderSides), allowNull: true,
```
moshababo: @kilrau 

sangaman suggested:

> Rather than specify a side, we can do what some other exchanges out there do which is to have a negative quantity indicate a sell while a positive quantity indicates a buy. One less column for us to track.

orderSides: {BUY,SELL}

What do you say about it?

lib/db/DB.js
sangaman: Ok, either way works so I will defer to you, I was mainly basing off the approach in the sequelize api docs & examples.

lib/db/DB.js
sangaman: Makes perfect sense for `currencies` having the ticker string be the only identifier, and I see what you mean for `pairs`. After thinking about this a bit, I think the best approach is probably to have `baseCurrency` and `quoteCurrency` act as a composite primary key. That's the only data we really need to track, and it'll make it simpler to perform queries like "select all orders with BTC as base currency".

See [here](https://gist.github.com/lucasscariot/5b8747fbc8a6948a805c646fae4ceef8) for an example of a composite primary key in sequelize.

sangaman:
@moshababo I added a few more responses to my earlier comments above, thanks.

lib/db/DB.js
kilrau: Good suggestion @sangaman ! As discussed, let's go with positive quantity for buy and negative quantity for sell!

lib/db/DB.js
kilrau: Let's do it - positive quantity for buy and negative quantity for sell.

lib/constants/enums.js
```diff
@@ -0,0 +1,11 @@
+const enums = exports;
+
+enums.swapProtocols = {
+  LND: 'LND',
+  RAIDEN: 'RAIDEN',
+};
+
+enums.orderSides = {
```
kilrau: Needs change

lib/db/DB.js
moshababo: I went with your approach. Better to be compatible with the documentation ;)

PR #13, opened 2018-03-30
swap protocols config

moshababo:
Basic swapProtocols config

* I disabled the `no-else-return` eslint rule because I think it enforces a bad practice
* lndClient obviously need more connectivity checks - later on
* feel free to fix this PR as you wish and merge it yourself (i'll be away for a few days).


moshababo:
@sangaman 
should I merge this?

PR #16, opened 2018-04-03
improving p2p robustness

sangaman:
This is a set of changes to improve peer to peer functionality.
Connecting to a peer first attempts to resolve the hostname. If a
connection attempt is made to an ip address and port that is already
connected, the connection attempt is aborted. Bad data received from a
peer no longer crashes the daemon, instead a warning is logged.

### DIFF
--- a/lib/rpcserver/RpcMethods.js
+++ b/lib/rpcserver/RpcMethods.js
@@ -42,8 +42,8 @@ class RpcMethods {
     this.p2p.broadcast('neworder', params);
   }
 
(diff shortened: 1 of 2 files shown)


lib/rpcserver/RpcMethods.js
```diff
@@ -43,7 +43,7 @@ class RpcMethods {
   }
 
   async connect(params) {
```
moshababo: can remove the `async` keyword

PR #17, opened 2018-04-09
basic orderBook

moshababo:
* basic orderBook infrastructure - matchingEngine, matchesProcessor
* unit / integration tests infrastructure (combined with gulp tasks)

The code is not fully tested yet (manually or via code). Just wanted verify the basic design.


lib/orderbook/OrderBook.js
```diff
@@ -1,23 +1,74 @@
...
+      await this.repository.addOrder(tsOrder);
+      this.logger.debug(`order added: ${JSON.stringify(tsOrder)}`);
+      if (order.peerId === null) {
+        this.broadcastOrder(order);
+      }
+    }
+  }
+
+  broadcastOrder() {
+    // TODO: broadcast non-matched ownOrder to peers
```
sangaman: We have a method that does this actually, `p2p.broadcast()`. It's naively used in the RPC `placeOrder` method currently, I figure you can probably remove it from there and instead call the broadcast method from this class (where it will appropriately only be called if an order is not instantly filled).

package.json
```diff
@@ -50,12 +53,15 @@
     "secp256k1": "^3.5.0",
     "sequelize": "^4.37.3",
     "toml": "^2.3.3",
-    "winston": "^3.0.0-rc3"
+    "winston": "^3.0.0-rc3",
+    "yargs": "^11.0.0"
```
sangaman: I was using that `command-line-args` package above for working with command line arguments. Ultimately we should only use one. Thoughts on this?

lib/utils/utils.js
```diff
@@ -61,3 +61,27 @@ utils.getPublicMethods = (obj) => {
...
+};
+
+utils.getGuid = () => {
+  // TODO: make it truly unique
+  function s4() {
+    return Math.floor((1 + Math.random()) * 0x10000)
+      .toString(16)
+      .substring(1);
+  }
+  return `${s4() + s4()}-${s4()}-${s4()}-${s4()}-${s4()}${s4()}${s4()}`;
```
sangaman: This method work, but it may ultimately be more robust to use something like https://www.npmjs.com/package/uuid to generate guids. I see you have a TODO here, so if you want to just leave it as is I can incorporate the uuid package in a separate PR.

.eslintrc.json
```diff
@@ -1,6 +1,8 @@
 {
     "extends": "airbnb-base",
     "rules": {
-        "no-else-return": 0
+        "no-else-return": 0,
+        "class-methods-use-this": 0,
+        "no-unused-expressions": 0
```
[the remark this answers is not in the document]
moshababo: class-methods-use-this - I’m not fond of this one because many times you refactor a method between using and not-using ‘this’.

no-unused-expressions - sometimes it’s necessary or preferred.

But I guess we can stick with them, and use the file/line rule exceptions. I’ll remove it.

lib/orderbook/MatchesProcessor.js
```diff
@@ -0,0 +1,34 @@
...
+  add(match) {
+    this.buffer.push(match);
+
+    if (match.maker.peerId) {
+      this.notifyPeer(match);
+    } else if (match.taker.peerId) {
+      this.executeSwap(match);
+    } else {
+      this.notifyClient(match);
+    }
```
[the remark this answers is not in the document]
moshababo: Important topic. Let’s continue this discussion on Slack.

In regards to notifying the client - The logic I used is indeed not self explanatory. It’s just a sketch. 
When both `match.maker.peerId` and `match.taker.peerId` are falsely, it means that it’s an internal match, so notifying the client is the main path of execution in the matches processing. 
We should definitely notify the client also on swaps, but in this case that's not the only thing to be done. I'll work on a better terminology later on.

package.json
moshababo: Sorry, forgot about it. 
I guess we can stick with `command-line-args`. I'll remove `yargs` usage.

lib/utils/utils.js
moshababo: Yes, it was a temp solution. I'll look into this lib and incorporate it from now.

moshababo:
In my last commit I changed `p2p` class design a bit.

We had a circular dependency situation - `orderBook` needs `p2p`, and vice versa.
Although technically it's not a big deal in JS, since we can just impose some sort of lazy loading scheme, it's usually a code smell. 

So one approach would be to make a mutual parent class (`Xud`, or some new one), but I thought it's not appropriate.
Other approach is to make one of them base component, which will be injected to the main component, that will register its callback methods.
Because that `p2p` incoming messages is outside of any synchronous flow, I decided to make it the base component, which other components register through an event emitter.

PR #21, opened 2018-04-17
Standardizing default rpcProtoPath

sangaman:
This fixes the default path for the rpc.proto file included in the
source on windows, and also restructures the config parameters related
to lnd.


moshababo:
Added a small fix. Please verify and merge.

In addition, it's also possible to move `lnd`/`raiden` config objects under `swapProtocols`.

moshababo:
@sangaman 
I'm merging this and I'll do a follow up in a new PR that i'll open soon.

PR #22, opened 2018-04-17
Using yargs with command modules

sangaman:
This commit replaces the command-line-args package with the more
powerful yargs package. Each command that can be called from xucli is
now its own module with a description of the parameters for that
command.


moshababo:
Great 👍

PR #24, opened 2018-04-17
utils - deepMerge

moshababo:
`utils.copyObject` had some issues with null & string properties.
Trying to fix it made the code too clumsy, so I used a different implementation instead.

I change its name to `deepMerge` because that's the standard name for this kind of util.


moshababo:
Yes, that's a good convention (skipping `undefined` but not `null`).

PR #26, opened 2018-04-18
Placeorder RPC call

sangaman:
This adds the `pairId` parameter to the `placeorder` rpc call. I also renamed the `rpcserver` folder to `rpc` which I think is shorter and more accurate, since I expect most of our work going forward will be around the various methods that are exposed and not the server itself.


moshababo:
* placeorder cli command - fixed an issue with one assertion, added a new one, and moved all of them to `RpcMethods.placeOrder`. The validation must be there anyway because that will be the main entrypoint to `xud`. So if they were left in the cli, they would be needed to get duplicated.
* placeorder cli command - updated metadata
* updated README (placeorder cli command, db initialization)
* improved db automation
* added error codes domain prefixes
* fixed some issues in orderBook

lib/rpc/RpcMethods.js
```diff
@@ -54,10 +55,18 @@ class RpcMethods {
...
   async placeOrder(params) {
     const { order } = params;
-    // temporary simple invoices until swaps are operational
-    const invoice = await this.lndClient.addInvoice(order.price * order.quantity);
-    order.invoice = invoice.payment_request;
-    await this.orderbook.addOrder(order);
+    assert(typeof order.price === 'number', 'price name must be a number');
+    assert(order.price > 0, 'price must be greater than 0');
+    assert(typeof order.quantity === 'number', 'quantity must be a number');
+    assert(order.quantity > 0, 'quantity must be not equal 0');
```
sangaman: @moshababo We accept negative quantity for sell orders, right?

moshababo: Yes, good catch, fix commited

Issue #27, opened 2018-04-20
 error: Error: ENOENT: no such file or directory, open '/home/sky/.lnd/tls.cert'


yyy999:
In running ./xud:

 error: Error: ENOENT: no such file or directory, open '/home/sky/.lnd/tls.cert'
How do I do it?

sangaman:
See #20. This error is thrown when [lnd](https://github.com/lightningnetwork/lnd) is not installed, but it can be disabled for testing purposes. I'll leave this issue open until we have a more informative error message.

yyy999:
Ok!
resolved!

Issue #20, opened 2018-04-16
xud init error: tls.cert file 


Chikel:
After installing repo dependencies and running a MySQL instance (+ creating an xud database), 
I receive the following error:

$ ./xud
2018-4-16 13:53:43 info: config loaded
2018-4-16 13:53:43 error: Error: ENOENT: no such file or directory, open '/Users/Chikel/Library/Application Support/Lnd/tls.cert'
    at Object.fs.openSync (fs.js:646:18)
    at Object.fs.readFileSync (fs.js:551:33)
    at new LndClient (/Users/Chikel/sandbox/xud/lib/lndclient/LndClient.js:18:24)
    at Xud.start (/Users/Chikel/sandbox/xud/lib/Xud.js:35:26)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
    at Function.Module.runMain (module.js:686:11)
    at startup (bootstrap_node.js:187:16)
    at bootstrap_node.js:608:3

Thanks!

moshababo:
You should either install `lnd` (https://github.com/lightningnetwork/lnd/blob/master/docs/INSTALL.md), or disable its client via your [TOML](https://github.com/toml-lang/toml) `.conf` file. Its expected location is described in our `README.md`, but no example is given yet for this configuration.
You should try to add this:

````
[swapProtocols.LND]
enable = false
````

PR #28, opened 2018-04-20
check if lnd tls.cert file exists

sangaman:
This improves the error handling around issues #20 and #27 when lnd is not installed.


moshababo:
I wasn't sure about the "silent error" approach. It might be preferred to throw exception, and let the user disable `lnd` if he wants to progress. This way there won't be any surprises.

PR #34, opened 2018-04-22
config and cli

moshababo:
* Fixing an existing issue with `rpcPort` (the name wasn't consistent, sometime it was `rpcport`), I ended up making it nested, as `rpc.port` (better approach since `rpc` will have more config, and this is our pattern anyway).
* Adding relevant description for the `./bin/xud` symlink (he's necessary if you want to create few `xud` instances). 
* Made it more clear that in the `connect` cli command, the args (host, port) are referring to the p2p server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Xud uses [MySQL](https://www.mysql.com/) or [MariaDB](https://mariadb.org/). You
 
 ## Command-Line Interface
 
-Spawning a new `xud` process
+Spawn a new `xud` process
 
 ```bash
 ~/xud/bin $ ./xud 
@@ -47,13 +47,13 @@ Options:
   -p, --p2p.port                                        [number] [default: 8885] 
 ```
 
-Interacting with an `xud` process, identified by his rpc port
+Interact with an `xud` process, identified by his `rpc` port
 ```bash
 ~/xud/bin $ ./xucli --help
 xucli [command]
 
 Commands:
-  xucli connect <p2p_host> [p2p_port]            connect to an xu node
+  xucli connect <host> [port]                    connect to an xu node
   xucli getinfo                                  get general info from the xud node
   xucli getorders                                get orders from the orderbook
   xucli getpairs                                 get orderbook's available pairs
--- a/README.md
+++ b/README.md
@@ -53,13 +53,13 @@ Interacting with an `xud` process, identified by his rpc port
 xucli [command]
 
 Commands:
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -38,38 +38,48 @@ Xud uses [MySQL](https://www.mysql.com/) or [MariaDB](https://mariadb.org/). You
...
 
 Commands:
-  xucli connect <host> [port]                   connect to an xu node
-  xucli getinfo                                 get general info from the xud node
-  xucli getorders                               get orders from the orderbook
-  xucli getpairs                                get orderbook's available pairs
-  xucli placeorder <pairId> <price> <quantity>  place an order
-  xucli shutdown                                gracefully shutdown the xud node
-  xucli tokenswap <identifier> <role>           perform a raiden token swap
+  xucli connect <p2p_host> [p2p_port]            connect to an xu node
```
sangaman: I actually think this may have been more clear as original, particularly to distinguish this from the `p2p.port` for `xud` above, which is the port on which to listen on for incoming connections.

lib/cli/commands/connect.js
```diff
@@ -1,21 +1,21 @@
 const command = require('../command');
 
-exports.command = 'connect <host> [port]';
+exports.command = 'connect <p2p_host> [p2p_port]';
```
sangaman: See above about keeping it as `<host> [port]`.

lib/cli/commands/connect.js
```diff
@@ -1,21 +1,21 @@
...
 exports.describe = 'connect to an xu node';
 
 exports.builder = {
-  host: {
-    type: 'number',
+  p2p_host: {
+    type: 'string',
   },
-  port: {
+  p2p_port: {
```
sangaman: We can add `description` properties here to clarify that these are the host and port of the remote peer we are connecting to.

lib/cli/commands/connect.js
```diff
@@ -1,21 +1,21 @@
...
     type: 'number',
     default: '8885',
   },
 };
 
 function callHandler(xuClient, argv) {
-  return xuClient.connect(argv.host, argv.port);
+  return xuClient.connect(argv.p2p_host, argv.p2p_port);
```
sangaman: See above about keeping it as `host` and `port`.

moshababo:
Fixed ;)
Didn't know you can add a description in the argument level.

sangaman:
Yup, the description shows up when you do `xucli connect --help`. Thanks.

PR #35, opened 2018-04-22
p2p order

moshababo:
* Fixed some inconsistencies in `P2P.addPeer` in regards to the peer's identifier. 
* Added some `host`/`address` nuances in `P2P.connect`


moshababo:
Sorry, i merged `config_cli` to be able to test it, but it made a mess in the Files changes tab.
Take a look only at the `P2P.js` file. That's the only relevant file.

moshababo:
Sorry, i merged `config_cli` to be able to test it, but it made a mess in the `Files changes` tab.
Take a look only at the `P2P.js` & `P2PServer.js` files. That's the only relevant changes.

lib/p2p/P2P.js
```diff
@@ -24,6 +24,19 @@ class P2P extends EventEmitter {
...
+   * Construct peer's id
+   * @param {string} host
+   * @param {string} address
+   * @param {number} port
+   * @returns {string}
+   */
+  static getPeerId(host, address, port) {
+    return !host || address === host
+      ? `${address}:${port}`
+      : `${host}[${address}]:${port}`;
```
sangaman: Why use the host name here? Would it not be the case here that if I tried to connect to the same remote node twice, once by host name and once by IP address, I'd get different peerIds each time? I'm also wondering if we should call this something other than "peerId" perhaps? I'm thinking that name conflicts with database columns. Maybe call this something like `peerString`?

lib/p2p/P2P.js
```diff
@@ -43,19 +56,23 @@ class P2P extends EventEmitter {
...
+
         const peer = net.connect({
-          host,
+          address,
           port,
         }, () => {
-          this.addPeer(peer);
-          const msg = `connected to peer ${peer.remoteAddress}:${peer.remotePort}`;
+          this.addPeer(peer, peerId);
+          const msg = `connected to peer ${peerId}`;
```
sangaman: Here we still have access to the `host` variable. Maybe we do something like:

```javascript
const msg = host ? `connected to peer (${host})${peerId}` : `connected to peer ${peerId}`
```

to display the host name if one is provided, otherwise just the ip address and port.

lib/p2p/P2P.js
sangaman: That all makes sense. I agree we don't need to keep the host name past the point of logging the initial connection. I made the keys for the `peers` object use ip address and port since that is essentially the unique identifying tuple for a peer, going with just ip address doesn't work for when multiple nodes are running on a single address which is the case with local testing.

Logging the host name in the "connected to" message would be nice, see my newest comment for how we could do that without having to put `host` in the peerId/peerString.

I think with that change and renaming `getPeerId` to `getPeerString`, as well as removing the host from the peer string, this should be good to merge. I'm looking forward to your wider scope followup PR on this.

moshababo: That solution was not feasible because there were more dependencies on that string (the one with the host), and what you're called `peerString` is basically a key (although it's a string), so a different solution was required.

moshababo:
Created a Peer object, to be able to retrieve separately `getKey` and `toString`.
Will help later on to sync data from/to DB too.

moshababo:
Added more code to complete the basic P2P order procedure.

lib/Config.js
```diff
@@ -42,6 +42,9 @@ class Config {
       database: 'xud',
       dialect: 'mysql',
       operatorsAliases: false,
+      dialectOptions: {
+        multipleStatements: true,
+      },
```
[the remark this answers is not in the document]
moshababo: Yes, if we're approaching our config file as anything that the user might want to override, then it shouldn't be there. I saw it merely as the central point for all config. 

I'll change it according to your advise.

tasks/db/init.js
```diff
@@ -19,16 +19,16 @@ gulp.task('db.init', async () => {
...
+      { address: '127.0.0.1', port: '8885' },
+      { address: '127.0.0.2', port: '8885' },
+      { address: '127.0.0.3', port: '8885' },
+      { address: '127.0.0.4', port: '8885' },
+      { address: '127.0.0.5', port: '8885' },
+      { address: '127.0.0.1', port: '8886' },
+      { address: '127.0.0.2', port: '8886' },
+      { address: '127.0.0.3', port: '8886' },
+      { address: '127.0.0.4', port: '8886' },
+      { address: '127.0.0.5', port: '8886' },
```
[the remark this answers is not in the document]
moshababo: Right.. didn't thought about. Will be changed.

Issue #32, opened 2018-04-22
Docker Containers


moshababo:
We need to set up a configurable docker container automation for all XUD service dependencies, in order to get a functioning dev environment without much hassle. 

To begin with, we need to it for the database dependency (MySQL).

Later, we'll want to add all other dependencies:
* LND
* Raiden
* Different blockchains clients

In addition, we might want to set up a container solution which includes XUD itself, for easy deployment in tests and in production.

itsbalamurali:
I've started working on this, for now, to begin with:

1. Dockerizing XUD & XUCLI.
2. Setting up docker-compose.yml for automating entire dependencies.

itsbalamurali:
Done as part of #58  in the future plan to add Raiden and streamline networking and storage between containers.

PR #36, opened 2018-04-23
Initial Docker setup 🐳

itsbalamurali:
The following things are done for now supporting issue #32 
- Dockerize XUD,
- Docker Compose file for XUD and MySQL Services.

TODO: work on incorporating Lightning and Raiden Networks into docker compose services to make things much easier.

@moshababo  please, suggest a way to create the database if not exists in gulp db:init task.


lib/Config.js
```diff
@@ -30,16 +30,36 @@ class Config {
...
+    }
+    if (process.env.DB_USER !== '') {
+      dbuser = process.env.DB_USER;
+    }
+    if (process.env.DB_NAME !== '') {
+      dbname = process.env.DB_NAME;
+    }
+    if (process.env.DB_PASSWORD !== null) {
+      dbpass = process.env.DB_PASSWORD;
+    }
```
sangaman: I'm hesitant to use environment variables here. That's one more source of configuration to keep track of, plus as is this would get overwritten by command line and config file parameters. Instead of environment variables, could you maybe put any docker-specific configuration into a `xud.conf` file that is included in the container?

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+  xu:
+    image: exchangeunion
+    build:
+      context: .
+    volumes:
+      - ./:/opt/xud
+    environment:
+      DB_HOST: "db"
+      DB_PORT: 3306
+      DB_USER: "root"
```
sangaman: Is running as root alright here? I'm not sure about best practices for docker containers.

itsbalamurali:
@sangamanI see few blockers here, currently, XUD currently is assumed to be running on the same machine as Lnd & Raiden as per the configuration. But whereas docker goes with one process per container. Maybe we should refactor the XUD to be decoupled as service itself.

docker-compose.yml
itsbalamurali: Well, the docker-compose file currently is only for dev environment. We need a init script to create other non-root users and database.

lib/Config.js
itsbalamurali: I've refactored it load from xud.conf now.

moshababo:
@itsbalamurali 
From what I just checked, it is possible to connect with `sequelize` to the database server in a way that the authentication is not applied to a specific database. And when the connection is opened, the database can be created by a simple raw query. 

I'll open a new task for that.

moshababo:
@itsbalamurali 
See #41 
You can take it yourself, or i'll do it later on this week. 
It should all be a part of the DB class init process.

In regards to the `db:init` task - since it's all about populating the database with some dummy values, perhaps the name I chose is not so good.

lib/db/DB.js
```diff
@@ -37,10 +34,16 @@ class DB {
...
+      const dboptions = { ...this.options };
+      dboptions.database = '';
+      const sequelize = new Sequelize(dboptions);
+      sequelize.query(`CREATE DATABASE IF NOT EXISTS ${this.options.database};`);
+      this.sequelize = new Sequelize(this.options);
       await this.sequelize.authenticate();
       const { host, port, database } = this.sequelize.config;
       const { dialectName } = this.sequelize.connectionManager;
       this.logger.info(`connected to database. host:${host} port:${port} database:${database} dialect:${dialectName}`);
+      this.importModels();
```
sangaman: Move back to constructor.

lib/db/DB.js
```diff
@@ -37,10 +34,16 @@ class DB {
 
   async init() {
     try {
+      const dboptions = { ...this.options };
+      dboptions.database = '';
+      const sequelize = new Sequelize(dboptions);
+      sequelize.query(`CREATE DATABASE IF NOT EXISTS ${this.options.database};`);
```
sangaman: This needs `await` before the query and authenticate lines since these are asynchronous methods that must get a response from the db. Also, you can do this without needing to create a new `dboptions` object, see my example below. I tested this locally and it worked.

```javascript
const initSequelize = new Sequelize('', this.sequelize.options.username, this.sequelize.options.password, this.sequelize.options);
await initSequelize.authenticate();
await initSequelize.query(`CREATE DATABASE IF NOT EXISTS ${this.sequelize.options.database};`);
```

lib/db/DB.js
```diff
@@ -37,10 +34,16 @@ class DB {
...
+      this.sequelize = new Sequelize(this.options);
```
sangaman: Move back to constructor.

xud.conf
```diff
@@ -0,0 +1,21 @@
...
+listen = true
+port = 8885
+
+[lnd]
+disable = false
+rpcprotopath = "lndrpc.proto"
+
+[raiden]
+disable = false
+port = 5001
```
sangaman: It looks like these are all the default values, right? If so, we shouldn't need to explicitly create or define a config file for the docker container, default values are hardcoded and loaded automatically.

docker-compose.yml
```diff
@@ -0,0 +1,34 @@
+version: '3.3'
+services:
+  xu:
+    image: exchangeunion
+    build:
+      context: .
+    volumes:
+      - ./:/opt/xud
+      - ./xud.conf:/home/node/.xud
```
sangaman: See my comment below on whether we actually need to create an xud.conf file.

.travis.yml
```diff
@@ -1,8 +1,13 @@
...
   - "node"
   - "8"
 
+services:
+  - docker
+
 script:
   - npm run lint
   - npm run test:unit
+  - docker build -t exchangeunion/xud .
```
sangaman: Add new line.

.dockerignore
```diff
@@ -0,0 +1,64 @@
...
+.yarn-integrity
+
+# dotenv environment variables file
+.env
+
+# Visual Studio Code config
+.vscode/
+
+#  Webstorm config
+.idea/
```
sangaman: Add a new line at the end of file here, just for consistency with other files in the repo.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+USER node
+
+# Create .xud directory for config volume flexibility
+WORKDIR $HOME/.xud
+
+#Switchback to main dir
+WORKDIR /opt/xud
+
+# Start Exchange Union Daemon
+ENTRYPOINT ["bin/xud"]
```
sangaman: Add new line.

docker-compose.yml
itsbalamurali: We need to be able to inject the MySQL container connection details. hence, I think this is required.

xud.conf
itsbalamurali: We need to be able to inject the MySQL and other containers connection details. hence, I think this is required.

itsbalamurali:
@sangaman yes we cloud use `supervisord` for that purpose but that not really a recommended approach.

lib/db/DB.js
itsbalamurali: Since `this.sequelize` is initiated and used in `this.importModels()` it is throwing an error even before reaching the init method. Cloud please guide me on this?

sangaman: The last two lines of the constructor should be:

```javascript
this.sequelize = new Sequelize(options);
this.importModels();
```

Then in `init()` you should start with the three lines from my snippet above. Let me know if that's still not working for you.

xud.conf
sangaman: I'm not sure what you mean, is this file being used by something other than xud? The connection details specified here are hardcoded into xud.

moshababo:
@itsbalamurali,

In addition to @sangaman comments,

* If you want to clean the `database` property from the `options` object, you can do something like this:
````
const {database, ...databaselessOptions} = this.options
````
This way you can still use the `database` value for the query, and no shallow clone is required.
Making the database property an empty string is risky, and deleting it is also not a good practice either. 
* You're creating a 2nd connection without closing the first one
* The 2nd connection might not be necessary. Did you try to [define](https://stackoverflow.com/questions/10453503/database-context-change?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa) the database context instead? we will not want a solution that requires 2 consecutive connections on **every** launch. so if this won't work, we'll want the first solution I suggested at #41

xud.conf
itsbalamurali: This file is used by the `xud` container if you could refer to docker-compose.yml for connecting to MySQL.

itsbalamurali:
Closing This due to new typescript migration.

Issue #30, opened 2018-04-22
TypeScript POC


moshababo:
We want to asses switching over the project to TypeScript (TS).
To begin with, we need to set up the TS build automation, making sure the project remains debuggable in IDEs, and to provide a few examples of converting files code to the appropriate TS equivalent.

Chikel:
I'd like to handle this issue.

moshababo:
Great..
Do you think you can make it by next week?

Chikel:
Sure thing.

sangaman:
This was resolved by #37.

PR #37, opened 2018-04-23
Feature/typescript integration

Chikel:
This PR addresses issue #30.
Here are the main changes:

- Typescript 2.8 was added to app dependencies.
- utils.js was selected as an example for TS migration.
- TS compilation output is located at dist folder.
- npm scripts were updated to match new compilation folder. 

My wallet address: 0x752D843173705D2fc5EED233A640FE348b2bEb8E


sangaman:
Do you want to see if this https://www.npmjs.com/package/tslint-config-airbnb works? We should be able to rename all files to `.ts` extensions even without refactoring all of them to use strongly typed variables. I don't want to lose linting on the lib folder altogether, so I'm hoping we can replace `eslint` with `tslint`. You can also update the `lint` script in `package.json`.

Chikel:
I can convert all files to `.ts` but this may lead to numerous compilation errors (although it will still be compiled to dist folder).

moshababo:
@Chikel, good work.

open issues:
* In dev mode (`npm run dev`), TS build should be triggered on every file change. That will make `nodemon` watch on `dist` effective.
* any sourcemap debugging solutions?

Chikel:
These changes were introduced:
1. tsc watches .ts changes and recompile to dist folder. nodemon is tracking the dist folder for changes.
2. source maps are added to compilation (tested on webstorm debugger).

If you'd like, I can open an additional PR for adding types and fixing all .ts errors.

moshababo:
Please do it on this PR, because I don't want to merge it with all these errors.
Just do the minimal technical adjustment to make it smoothly.. the appropriate TS design will be done later.

Chikel:
All TS errors have been resolved.
I have updated npm scripts to run the app only if there are no compilation errors (`&&` instead of `||`).
Some additional changes were introduced:

- Common.js syntax is replaced with TS import/export syntax.
- Interfaces were added to all classes.
- TSlint updates

lib/Logger.ts
```diff
@@ -1,9 +1,51 @@
...
+class Logger implements LoggerInterface {
+  static levels = {
+    ERROR: 0,
+    WARN: 1,
+    INFO: 2,
+    VERBOSE: 3,
+    DEBUG: 4,
+  };
+
+  static levelsVal = {
```
Chikel: Static properties were moved inside the class and the 'static' accessor was added.

lib/db/models/Pair.ts
```diff
@@ -1,14 +1,14 @@
...
   const Pair = sequelize.define('Pair', {
     id: { type: DataTypes.STRING, primaryKey: true },
     baseCurrency: { type: DataTypes.STRING, allowNull: false },
     quoteCurrency: { type: DataTypes.STRING, allowNull: false },
     swapProtocol: {
       type: DataTypes.ENUM, values: Object.values(enums.swapProtocols), allowNull: true,
     },
-  }, {
+  },                            {
     tableName: 'pairs',
```
Chikel: This indentation is due to TSlint align rule, can be turned off if not desired.

lib/lndclient/errors.ts
```diff
@@ -1,8 +1,11 @@
...
+import errorCodesPrefix from '../constants/errorCodesPrefix';
 
-const errors = exports;
+const codesPrefix = errorCodesPrefix.LND;
+const errors : any = {};
 
 errors.LND_IS_DISABLED = {
   message: 'lnd is disabled',
-  code: codesPrefix.concat(1),
+  code: codesPrefix.concat('1'),
```
Chikel: Changing to import syntax disabled the ability to extract a property directly (LND).
I added line 3 for this matter.
Pay attention to line 8 --> now that TS recognize that `codesPrefix` is of type string, it cannot be concatenated by number 1. I changed it to string, make sure it doesn't affect any logic.

tslint.json
```diff
@@ -1,6 +1,10 @@
 {
   "extends": "tslint-config-airbnb",
   "rules": {
-    "variable-name": false
+    "variable-name": false,
+    "no-this-assignment": false,
+    "import-name": false,
+    "max-line-length": false,
+    "no-else-after-return": false
```
Chikel: These rules were turned off in order to keep your code intact. They may be turned on if you desire.

Chikel:
@moshababo , Merge conflicts fixed.

tslint.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "tslint-config-airbnb",
+  "rules": {
+    "variable-name": false,
+    "no-this-assignment": false,
+    "import-name": false,
+    "max-line-length": false,
+    "no-else-after-return": false,
+    "function-name": false
```
sangaman: Generally speaking I think it's good for us to limit the number of project-wide rule exceptions to best conform to common and popular styles.

`no-else-after-return` is one rule we'd agreed to discard earlier. After looking at these rules closely, I'm fine with keeping `variable-name": false` and `"import-name": false` because it allows us to keep our convention of upper case naming for classes and there seems to be a number of places where these apply.  For the others, I'd prefer to keep them out of the project-wide config, either disable them on a per-file or per-line basis or just edit the code to conform. `no-this-assignment` in particular is a useful rule to abide by, I think. We could also perhaps increase the limit for max-line-length, 150 characters fits easily on my default layout so personally I'm fine with `"max-line-length": [true, 150]`. Input from @moshababo would be good since we're gonna live with these rules for a while.

lib/lndclient/LndClient.ts
```diff
@@ -1,14 +1,31 @@
 /* eslint-disable camelcase */
```
sangaman: We should do a find all for "eslint-disable" in the codebase and remove these lines because they will no longer be needed.

Chikel: eslint related code removed.

lib/db/models/Peer.ts
```diff
@@ -1,10 +1,10 @@
-module.exports = (sequelize, DataTypes) => {
+export default (sequelize, DataTypes) => {
   const Peer = sequelize.define('Peer', {
     id: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
     address: { type: DataTypes.STRING(39), allowNull: false },
     port: { type: DataTypes.INTEGER, allowNull: false },
     pubKey: { type: DataTypes.STRING, allowNull: true, unique: true },
-  }, {
+  },                            {
```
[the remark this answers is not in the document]
Chikel: I disabled the align rule locally, let me know if you guys rather remove this rule globally.

Chikel:
CI will break due to tslint changes (`no-this-assignment`), see comment above.

package.json
```diff
@@ -9,11 +9,14 @@
...
-    "dev": "nodemon lib/Xud.js",
-    "lint": "eslint . ./**/*.js",
+    "test:unit": "npm run compile && mocha dist/test/unit --reporter spec",
+    "test:integration": "npm run compile && mocha dist/test/integration --reporter spec",
+    "start": "npm run compile && node dist/lib/Xud.js",
+    "dev": "npm-run-all -p compile:watch nodemon:watch",
+    "compile": "rm -rf ./dist && tsc",
+    "compile:watch": "tsc -w",
+    "nodemon:watch": "nodemon --watch dist -e js dist/lib/Xud.js",
+    "lint": "./node_modules/.bin/tslint ./**/*.ts",
```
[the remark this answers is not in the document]
Chikel: Correct, updated.
Let's handle the `no-this-asignment` rule that is mentioned below.

lib/Xud.ts
```diff
@@ -1,15 +1,37 @@
...
+import Config from './Config';
+import DB from './db/DB';
+import OrderBook from './orderbook/OrderBook';
+import LndClient from './lndclient/LndClient';
+import RaidenClient from './raidenclient/RaidenClient';
+import RpcServer from './rpc/RpcServer';
+import P2P from './p2p/P2P';
+import P2PServer from './p2p/P2PServer';
+
+interface XudInterface {
```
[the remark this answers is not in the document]
Chikel: @sangaman , I realized adding interfaces is redundant here. Class interface can be used in case we want several classes to implement the same methods / properties. There's no reason to create a distinct interface for each class. I removed the interfaces, class properties remains on top of the class declaration, before the constructor.

lib/Config.ts
```diff
@@ -64,7 +86,7 @@ class Config {
     if (!fs.existsSync(this.xuDir)) {
       fs.mkdirSync(this.xuDir);
     } else if (fs.existsSync(`${this.xuDir}xud.conf`)) {
-      const configText = fs.readFileSync(`${this.xuDir}xud.conf`);
+      const configText : any = fs.readFileSync(`${this.xuDir}xud.conf`);
```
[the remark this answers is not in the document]
Chikel: @sangaman you are correct, but toml's `parse` function receives string type.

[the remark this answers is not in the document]
sangaman: Ah yes good point, ok no worries we'll straighten this out later.

sangaman:
Alright, I squashed and rebased the commits here into a tidier history while preserving authorship, merging now.

kilrau:
Congrats, 50 XUC are yours: https://etherscan.io/tx/0x8e4debae442bf2cb0ff2a085d04a02632387c70ea6ba6ee87b00972ed6bf502e 🍾

PR #54, opened 2018-05-02
Refactoring raiden and lnd clients

sangaman:
This commit enforces several variable type restrictions and creates a shared base client class for the lnd and raiden clients. I ran some basic tests and the class inheritance seems to work properly.

### DIFF
--- a/lib/BaseClient.ts
+++ b/lib/BaseClient.ts
@@ -1,0 +1,27 @@
+import Logger from './Logger';
+
+enum ClientStatus {
+  DISABLED,
+  CONNECTION_VERIFIED,
+}
+
+abstract class BaseClient {
+  logger: Logger;
+  status!: ClientStatus;
+
+  constructor() {
+    this.logger = Logger.global;
+  }
+
+  setStatus(val: ClientStatus): void {
+    this.logger.info(`${this.constructor.name} status: ${ClientStatus[val]}`);
+    this.status = val;
+  }
+
+  isDisabled(): boolean {
+    return this.status === ClientStatus.DISABLED;
+  }
+}
+
+export default BaseClient;
+export { ClientStatus };
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -1,39 +1,35 @@
-import grpc from 'grpc';
+import grpc, { Metadata } from 'grpc';
 import fs from 'fs';
-import assert from 'assert';
 
 import Logger from '../Logger';
+import BaseClient, { ClientStatus } from'../BaseClient';
 import errors from './errors';
 
+interface LndClientConfig {
+  disable: boolean;
+  datadir: string;
+  rpcprotopath: string;
+}
+
 /** A class representing a client to interact with a running lnd instance. */
-class LndClient {
-
-  static statuses = {
-    DISABLED: 'DISABLED',
-    CONNECTION_VERIFIED: 'CONNECTION_VERIFIED', // assuming connection does not remain open
-  };
-
-  logger: any;
+class LndClient extends BaseClient{
   lightning: any;
-  meta: any;
-  status: any;
+  meta: Metadata | undefined;
 
   /**
    * Create an lnd client.
-   * @param {Object} options - The lnd configuration
+   * @param config - The lnd configuration
    */
-  constructor(options) {
-    const { disable, datadir, rpcprotopath } = options;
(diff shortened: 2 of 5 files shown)


lib/BaseClient.ts
```diff
@@ -0,0 +1,22 @@
+import Logger from './Logger';
+import enums from './constants/enums';
+
+class BaseClient {
+  logger: Logger;
+  status!: string;
```
moshababo: can switch `status` to be a TS enum

lib/BaseClient.ts
```diff
@@ -0,0 +1,22 @@
+import Logger from './Logger';
+import enums from './constants/enums';
+
+class BaseClient {
```
moshababo: should be an abstract class, assuming it's not meant to get instantiated directly

lib/constants/enums.ts
```diff
@@ -10,4 +10,9 @@ enums.orderingDirections = {
   ASC: 'ASC',
 };
 
+enums.statuses = {
```
moshababo: let's use TS enums (don't need to convert all enums in one go)

lib/lndclient/LndClient.ts
```diff
@@ -44,25 +41,16 @@ class LndClient {
...
-    this.status = val;
-  }
-
   /**
    * Return general information concerning the lightning node including it’s identity pubkey,
    * alias, the chains it is connected to, and information concerning the number of
    * open+pending channels.
    */
-  getInfo() {
+  getInfo(): Promise<any> {
```
moshababo: You explicitly declare the return type. I think it's the better way. Let's stick to that, and do it also in `BaseClient`.

lib/lndclient/LndClient.ts
```diff
@@ -1,39 +1,36 @@
...
-  status: any;
+  meta: Metadata | undefined;
 
   /**
    * Create an lnd client.
-   * @param {Object} options - The lnd configuration
+   * @param options - The lnd configuration
    */
-  constructor(options) {
+  constructor(options: LndClientConfig) {
```
moshababo: since the type is `LndClientConfig`, I think we should call the param `config` instead of `options`.

lib/raidenclient/RaidenClient.ts
```diff
@@ -1,41 +1,37 @@
...
+}
 
-  logger: any;
-  port: any;
-  status: any;
+class RaidenClient extends BaseClient {
+  port: number;
 
-  constructor(options) {
+  constructor(options: RaidenClientConfig) {
```
moshababo: same as above - might be better to call the param as `config`

lib/constants/enums.ts
moshababo: The enum name need to be better associated with `BaseClient`. maybe it should be called `clientConnectionStatuses`, or simply `connectionStatuses`. Merely `statuses` won't work since it's a global enum now.

sangaman:
@moshababo Thanks, just the kind of feedback I was looking for. I added return types to the `BaseClient` methods and addressed the other comments.

lib/BaseClient.ts
```diff
@@ -0,0 +1,25 @@
...
+    this.status = val;
+  }
+
+  isDisabled(): boolean {
+    return this.status === ClientStatus.DISABLED;
+  }
+}
+
+export default BaseClient;
+export enum ClientStatus {
```
moshababo: Let's stick to defining enums/interfaces before the class, for an easy read. 
If it's also for export, we'll do it in the end of the file, as usual (`export {ClientStatus}`)

moshababo: Also, let's stick with singular enum types from now (as you already did).

lib/BaseClient.ts
```diff
@@ -0,0 +1,25 @@
...
+abstract class BaseClient {
+  logger: Logger;
+  status!: ClientStatus;
+
+  constructor() {
+    this.logger = Logger.global;
+  }
+
+  setStatus(val: ClientStatus): void {
+    this.logger.info(`${this.constructor.name} status: ${val}`);
```
moshababo: instead of `${val}`, rather use `${ClientStatus[val]}` for better logging output (it will give you the enum value, `DISABLED`, instead of the key, `0`)

sangaman: Good point, I overlooked that the enum lost its string value.

PR #56, opened 2018-05-03
Adding types to config structure and rpc methods

sangaman:
These are some incremental changes towards refactoring to take advantage of TypeScript. I used types to define the object structure for our configuration and for some of the methods exposed by the RPC layer.


moshababo:
Good work.
One small thing - in `RpcMethods`, we should get rid of methods arguments being passed as one object. It's not necessary anymore in TS. and if it's like that because of how the RpcServer passes the args, i'd say we should fix that, but since we're gonna migrate to gRPC anyway, it can wait.

Issue #41, opened 2018-04-24
Create database


moshababo:
When authenticating through `sequelize` to the database server by specifying a specific database which does not exists, an exception is thrown. 

A possible solution would be to try to connect by specifying a specific database anyway, and if exception is thrown, reconnect without specifying one. Create the database with a raw query, and then either reconnect to the server or just define a new database context (should work in theory).

Other solution would be to always connect without specifying a specific database, create it if it does not exists, and then define the database context. If the database context trick doesn't work, and reconnecting is needed, then it's not a good solution.

itsbalamurali:
Would be closed as part of #36

itsbalamurali:
Closing this issue as this is done and merged as part of #58

PR #58, opened 2018-05-04
Dockerization & Create DB if not exists.

itsbalamurali:
This PR implements 

- Try to auto create the database if it does not exist, see issue #41.
- Dockerize XUD, see issue #32. 
- Docker Compose file to create the following services for xud to run.
  * `xud`
  * `mysql`
  * `btcd`
  * `lnd` 
 
I have introduced `host` & `port` config variables for the services in `xud.conf`. so that they can be run separately on individual machines (Service-oriented).


tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "strict": true,                           /* Enable all strict type-checking options. */
     "noEmitOnError": false,
     "noImplicitAny": false,                   /* Raise error on expressions and declarations with an implied 'any' type. */
-    "lib": ["ES2017"],
+    "lib": ["es2017"],
```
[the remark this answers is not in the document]
itsbalamurali: My vscode linter was throwing warning regarding the same and suggested to convert into lower case. as suggested will revert this and put it in a separate commit.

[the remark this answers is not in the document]
sangaman: @itsbalamurali I opened an issue for this with vscode, that's the same editor I am using. https://github.com/Microsoft/vscode/issues/49307

.travis.yml
```diff
@@ -1,8 +1,14 @@
...
   - "8"
 
+services:
+  - docker
+
 script:
   - npm run lint
   - npm run test:unit
+  - cd docker/xud && docker build -t exchangeunion/xud .
+
```
sangaman: End the file with a new line.

docker/docker-compose.yml
```diff
@@ -0,0 +1,123 @@
...
+  bitcoin:
+    driver: local
+
+  # litecoin volume is needed for maintaining blockchain persistence
+  # during ltcd container recreation.
+  litecoin:
+    driver: local
+
+networks:
+  xud:
```
sangaman: New line.

docker/xud/xud.conf
```diff
@@ -0,0 +1,24 @@
...
+[lnd]
+disable = false
+port = 10009
+hostname = "localhost"
+rpcprotopath = "lndrpc.proto"
+
+[raiden]
+disable = false
+port = 5001
+hostname = "localhost"
```
sangaman: New line (if we keep this file).

docker/xud/xud.conf
```diff
@@ -0,0 +1,24 @@
+lndDir = "~/.lnd"
```
sangaman: I know I brought this up in the last iteration of this PR, but I still don't understand why this file is needed if it does nothing but specify the default, hardcoded values. XUD does not require a config file to run (at least not currently), only for overriding default values.

lib/db/DB.ts
```diff
@@ -53,6 +53,11 @@ class DB {
 
   async init() {
     try {
+      const initSequelize = new Sequelize('', this.sequelize.options.username, this.sequelize.options.password, this.sequelize.options);
```
sangaman: Move this to a catch block to only fire when we get an error on accessing a non-existing database, see my last review comment above on how I'd approach this.

lib/lndclient/LndClient.ts
```diff
@@ -2,29 +2,35 @@ import grpc, { Metadata } from 'grpc';
...
   rpcprotopath: string;
 };
 
 /** A class representing a client to interact with a running lnd instance. */
-class LndClient extends BaseClient{
+class LndClient extends BaseClient {
   lightning: any;
   meta: Metadata | undefined;
+  port: number;
+  host: string;
```
sangaman: Good idea to make `host` and `port` for the lnd grpc service configurable, however I don't think we need to make them class properties. They're only used in the constructor to initialize the `lightning` property, so we can drop them here (but leave them in `LndClientConfig`).

lib/lndclient/LndClient.ts
```diff
@@ -2,29 +2,35 @@ import grpc, { Metadata } from 'grpc';
...
    * @param config - The lnd configuration
    */
   constructor(config: LndClientConfig) {
     super();
-    const { disable, datadir, rpcprotopath } = config;
+    const { disable, datadir, host, port, rpcprotopath } = config;
 
     this.logger = Logger.global;
+    this.port = port;
+    this.host = host;
```
sangaman: Don't need to set these on `this`.

lib/lndclient/LndClient.ts
```diff
@@ -34,7 +40,7 @@ class LndClient extends BaseClient{
       const lndCert = fs.readFileSync(`${datadir}tls.cert`);
       const credentials = grpc.credentials.createSsl(lndCert);
       const lnrpcDescriptor: any = grpc.load(rpcprotopath);
-      this.lightning = new lnrpcDescriptor.lnrpc.Lightning('127.0.0.1:10009', credentials);
+      this.lightning = new lnrpcDescriptor.lnrpc.Lightning(this.host + ':' + this.port, credentials);
```
sangaman: Can just use `host` and `port` here without `this`.

Also, although tslint doesn't enforce the [prefer-template](https://eslint.org/docs/rules/prefer-template) rule, I think we might as well stick to using it since that's how it is everywhere else in the code. So use `${host}:${port}` surrounded by backticks here instead of the string concatenation with the `+` operator.

package-lock.json
```diff
@@ -10,9 +10,9 @@
       "integrity": "sha512-BJ19rjnFFCeopEhbyK2Chg3Tq+o5xkjd6dtKxmFhfjwLH1Il2G7Ha4Jel2hpbyZL2Fh9/vrM9U0bpkANAL3pjA==",
       "dev": true,
       "requires": {
-        "@fimbul/ymir": "^0.6.0",
```
sangaman: I'm confused as to why there are changes to package-lock.json, particularly in the absence of changes to package.json. Are these necessary for the docker setup?

docker/xud/xud.conf
itsbalamurali: it's required when we are running in a Docker container for connectivity between MySQL and XUD containers. we actually mount this file into the container volume for xud to pickup `mysql`, `lnd` & `raiden` configurations. I'll add a new line.

docker/xud/xud.conf
itsbalamurali: Please refer to comment above.

package-lock.json
itsbalamurali: I had to update `grpc` package due to a protocol buffers loading bug. accidentally it entirely updated the project.

docker/xud/xud.conf
```diff
@@ -0,0 +1,24 @@
...
+rpcPort = 8886
+
```
moshababo: The structure you're using is a bit outdated. For example, `rpcPort` is now under `[rpc]`

docker/xud/xud.conf
moshababo: Perhaps just add the required config override, and leave the default values out.

lib/db/DB.ts
```diff
@@ -58,8 +58,22 @@ class DB {
       const { dialectName } = this.sequelize.connectionManager;
       this.logger.info(`connected to database. host:${host} port:${port} database:${database} dialect:${dialectName}`);
     } catch (err) {
-      this.logger.error('unable to connect to the database', err);
-      throw err;
+      if (err instanceof Sequelize.InvalidConnectionError) {
```
moshababo: Trying to run it, this returns `false` when the database doesn't exists. The error type is `SequelizeConnectionError`.

lib/db/DB.ts
```diff
@@ -58,8 +58,22 @@ class DB {
...
+        try {
+          const initSequelize = new Sequelize('', this.sequelize.options.username, this.sequelize.options.password, this.sequelize.options);
+          await initSequelize.authenticate();
+          await initSequelize.query(`CREATE DATABASE IF NOT EXISTS ${this.sequelize.options.database};`);
+          await initSequelize.query(`GRANT ALL PRIVILEGES ON ${this.sequelize.options.database}.* TO '${this.sequelize.options.username}'@'%'`);
+          await initSequelize.close();
```
moshababo: Any reason for not continuing with simply one connection, by defining a database context?

lib/db/DB.ts
```diff
@@ -58,8 +58,22 @@ class DB {
...
       this.logger.info(`connected to database. host:${host} port:${port} database:${database} dialect:${dialectName}`);
     } catch (err) {
-      this.logger.error('unable to connect to the database', err);
-      throw err;
+      if (err instanceof Sequelize.InvalidConnectionError) {
+        try {
+          const initSequelize = new Sequelize('', this.sequelize.options.username, this.sequelize.options.password, this.sequelize.options);
+          await initSequelize.authenticate();
+          await initSequelize.query(`CREATE DATABASE IF NOT EXISTS ${this.sequelize.options.database};`);
+          await initSequelize.query(`GRANT ALL PRIVILEGES ON ${this.sequelize.options.database}.* TO '${this.sequelize.options.username}'@'%'`);
```
moshababo: Is that necessary?

package-lock.json
sangaman: Yeah I think just revert these changes for now then as we discussed. We'll probably want to be updating dependencies, but no reason to mix it up with this PR.

docker/xud/xud.conf
sangaman: Agree with @moshababo, that's been my thinking all along and I believe that lines up with what we discussed, right? Keep the overridden config properties, certainly. But drop the lines that are repeating the default values. The main reason for doing this is so that, if our default values change in the future, we don't have to update this file as well.

docker/xud/xud.conf
sangaman: Yes, and can be dropped in this case because 8886 is the default.

lib/db/DB.ts
sangaman: Same for me.

lib/db/DB.ts
sangaman: I think I proposed this line earlier because it created the database without the proper permissions for me. I think it was because my `xunion` user didn't have the `GRANT` privilege, only `CREATE`. Without that, even this line doesn't work. I just tried this again commenting out this line and the database was created just fine, so I think we can indeed drop this `GRANT ALL PRIVILEGES` line.

lib/db/DB.ts
sangaman: I wasn't able to find a way how to do this.

moshababo: `USE [database]` doesn't work?

sangaman: Good call, this works for me:

```javascript

        this.sequelize = new Sequelize('', this.sequelize.options.username, this.sequelize.options.password, this.sequelize.options);
        await this.sequelize.query(`CREATE DATABASE IF NOT EXISTS ${this.sequelize.options.database};`);
        this.sequelize.query(`USE ${this.sequelize.options.database}`);
```

sangaman:
Looks good to me to squash and merge.

docker/README.md
```diff
@@ -0,0 +1,44 @@
+This document is written for people who are eager to do something with 
+the Exchange Union Daemon (`xud`). This folder uses `docker-compose` to
+package `xud`, `mysql`, `lnd` & `btcd` together to make deploying the two services as easy as
```
moshababo: fix "the **two** services"

docker/README.md
```diff
@@ -0,0 +1,44 @@
...
+typing a few commands. All configuration between `xud`, `mysql` and other containers are handled
+automatically by their `docker-compose` config file.
+
+The goal of this is that `docker-compose` to be able to spin up / lauch `xud` along with `lnd` & `Raiden` networks.
```
moshababo: instead of "along with `lnd` & `Raiden` networks", perhaps specify that it's for launching `xud` along with all its dependencies.

docker/docker-compose.yml
```diff
@@ -0,0 +1,88 @@
+version: '2.1'
+services:
```
moshababo: what was the reason for removing `ltcd`?

itsbalamurali: yeah, let us discuss this when we chat about `travis.yml`.

lib/db/DB.ts
```diff
@@ -58,8 +58,12 @@ class DB {
       const { dialectName } = this.sequelize.connectionManager;
       this.logger.info(`connected to database. host:${host} port:${port} database:${database} dialect:${dialectName}`);
     } catch (err) {
-      this.logger.error('unable to connect to the database', err);
-      throw err;
+      if (err instanceof Sequelize.ConnectionError) {
```
moshababo: This returns true for various of other errors, besides when database does not exist.
You should inspect it in more details for the right err code/msg.

lib/db/DB.ts
```diff
@@ -103,6 +107,18 @@ class DB {
...
   }
+
+  async dbCreate() {
+    try {
+      this.sequelize = new Sequelize('', this.sequelize.options.username, this.sequelize.options.password, this.sequelize.options);
+      await this.sequelize.authenticate();
+      await this.sequelize.query(`CREATE DATABASE ${this.sequelize.options.database};`);
+      await this.sequelize.query(`USE ${this.sequelize.options.database}`);
+    } catch (err) {
+      this.logger.error('unable to connect to the database', err);
```
moshababo: this will apply to a `CREATE DATABASE` error as well, which will end in a misleading error message.

lib/db/DB.ts
itsbalamurali: I am not sure how to achieve this but I've tried skimming through sequelize docs to check for any error code to capture but no luck. can you guide me on this?

lib/db/DB.ts
itsbalamurali: agree refactored them with proper error messages.

lib/db/DB.ts
moshababo: Sure.
I won't leave this PR open because of this, so let's talk about this today and finish it by tomorrow.

moshababo:
Open issues:
* better error inspection for when database does not exist
* travis docker integration

@itsbalamurali

PR #59, opened 2018-05-07
Lnd types

sangaman:
Updates to the latest lndrpc.proto file from the lnd master branch and generates a TypeScript declaration file using `ts-protoc-gen`. I didn't update any of the corresponding rpc methods (except to clear compilation errors) since we are migrating to grpc soon.


package.json
```diff
@@ -79,6 +79,7 @@
     "nodemon": "^1.17.2",
     "require-dir": "^1.0.0",
     "ts-node": "^6.0.2",
+    "ts-protoc-gen": "^0.6.0",
```
moshababo: I think we should add the script that is using that lib into our project, to be executable through `gulp` or `npm scripts`, so we'll know how to regenerate `lndrpc_pb.d.ts` after we're upgrading `lndrpc.proto`.
At the moment, I don't see this lib used through our project, so I assume you just used it to generate `lndrpc_pb.d.ts`, which in this case it doesn't need to be saved in our `package.json`, and can be used as a global npm package.
If that's the case, still, instead of removing it I think we should add the generating script to our automation.

itsbalamurali: Agree with @moshababo it would be a good idea to include it in npm scripts, maybe compile itself?

lndrpc.proto
```diff
@@ -28,13 +28,39 @@ package lnrpc;
 // The WalletUnlocker service is used to set up a wallet password for
```
itsbalamurali: I think we should be tracking the upstream versions of Lightning and Raiden networks we are developing against. Maybe Include the version of protobufs for now?

package.json
sangaman: Agreed, having a script was always my intention, I just generated this on windows and need to set this up on linux to get the command right. I'll work on it.

lndrpc.proto
sangaman: Yeah, for now I've just been trying to keep up to date with the latest since all projects involved are in rapid development. I was just of listing the targeted versions of Raiden & lnd in our readme/documentation and docker, do you have a better idea? I don't think Raiden has protobufs.

moshababo: Although it's a bit weird, maybe we should define these versions as constants in our project, and also log it when the lnd\raiden client is connecting. That way, in case of a compatibility problem, the user will know immediately what was the expected version.

Issue #39, opened 2018-04-24
Create node key


sangaman:
Use the https://www.npmjs.com/package/secp256k1 to create and save a key pair to uniquely identify an xud node and to sign/verify messages and orders. Look into how this key can be saved to disk in encrypted format.

moshababo:
What's the common practice of saving private keys as a part of the application data? 
Why plain disk file and not the database?
And if it's encrypted, where will you save the encryption key?

sangaman:
I'm most familiar with keys and other sensitive data being encrypted and saved to disk. Database is an option, but it's not something we'll be needing to query on any regular basis. I'm thinking it is decrypted on start and then kept in memory as long as xud is running. The encryption key wouldn't be saved, just remembered like a password.

kilrau:
Agree with Daniel's answer.

moshababo:
If it's not handled automatically then, what you describe means that each `xud` launch will require a password input, if the user have one, and if he don't then he'll create one and will get a new pubKey?

sangaman:
Yeah, I think launching on the first run would generate a new key and prompt the user to provide an optional encryption password. If the user provides a password, they'd need to enter it each time xud is started.

kilrau:
Close? @sangaman

sangaman:
I'll close when the encryption is merged.

kilrau:
Encryption part moved to Alpha: https://github.com/ExchangeUnion/xud/issues/184

PR #61, opened 2018-05-08
generate nodekey and persist to disk

sangaman:
This allows for creation of an ECDSA key pair (node key) for establishing node identity and signing messages. A node key is generated randomly and automatically if none exists. Generated node keys are saved to disk in the xu data folder and read on future starts. Methods to encrypt and decrypt stored keys on disk are included but currently are not used.

I didn't replace the placeholder node key in P2P.ts because the file is being replaced in a separate PR, but eventually we'll want to use this there.

This addresses #39.

### DIFF
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -1,3 +1,4 @@
+import fs from 'fs';
 import Logger from './Logger';
 import Config from './Config';
 import DB from './db/DB';
@@ -6,6 +7,7 @@ import LndClient from './lndclient/LndClient';
 import RaidenClient from './raidenclient/RaidenClient';
 import RpcServer from './rpc/RpcServer';
 import Pool from './p2p/Pool';
+import NodeKey from './nodekey/NodeKey';
 import dotenv from 'dotenv';
 
 /** Loads environment variables from the file .env */
@@ -14,17 +16,18 @@ dotenv.config();
 /** Class representing a complete Exchange Union daemon. */
 class Xud {
   logger: any;
-  config: any;
+  config: Config;
   db: any;
   lndClient: any;
   raidenClient: any;
   pool?: Pool;
   orderBook: any;
   rpcServer: any;
+  nodeKey!: NodeKey;
 
   /**
    * Create an Exchange Union daemon.
-   * @param {Object} args - Optional command line arguments to override configuration parameters.
+   * @param args Optional command line arguments to override configuration parameters.
    */
   constructor(args) {
     this.logger = Logger.global;
@@ -41,6 +44,9 @@ class Xud {
     this.logger.info('config loaded');
 
     try {
+      // TODO: wait for decryption of existing key or encryption of new key, config option to disable encryption
+      this.nodeKey = NodeKey.load(`${this.config.xudir}/nodekey.dat`);
+
       this.db = new DB(this.config.db);
       await this.db.init();
 
--- a/lib/nodekey/NodeKey.ts
+++ b/lib/nodekey/NodeKey.ts
@@ -1,0 +1,100 @@
+import CryptoJS from 'crypto-js';
+import { randomBytes } from 'crypto';
+import secp256k1 from 'secp256k1';
+import fs from 'fs';
+
+/**
+ * A class representing an ECDSA public/private key pair that identifies an XU node on the network
+ * and can sign messages to prove their veracity.
+ */
+class NodeKey {
+  private pubKeyStr: string;
+
+  constructor(private privKey: Buffer) {
+    const pubKey = secp256k1.publicKeyCreate(privKey);
+    this.pubKeyStr = pubKey.toString('hex');
+  }
+
+  /**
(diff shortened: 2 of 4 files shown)


lib/Xud.ts
```diff
@@ -39,6 +41,16 @@ class Xud {
     this.logger.info('config loaded');
 
     try {
+      // TODO: wait for decryption of existing key or encryption of new key, config option to disable encryption
```
moshababo: I think you should wrap this block with separate method, on `Xud` class, or even better - inside of `NodeKey` class, since you destined it to handle the persistence anyway.

You can create another static method called `get` or something, which receives as param the file path, and perhaps also the file name, and execute the same logic you currently put in this block, making the other two static methods potentially private.

lib/nodekey/NodeKey.ts
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * A class representing an ECDSA public/private key pair that identifies an xud node on the network
+ * and can sign messages to prove their veracity.
+ */
+class NodeKey {
+  private pubKeyStr: string;
+
+  constructor(private privKey: Buffer) {
+    this.privKey = privKey;
```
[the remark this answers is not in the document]
sangaman: Oops, I knew that actually.

lib/nodekey/NodeKey.ts
```diff
@@ -0,0 +1,95 @@
...
+      }
+
+      if (secp256k1.privateKeyVerify(buf)) {
+        return new NodeKey(buf);
+      } else {
+        throw new Error(`${path} does does not contain a valid ECDSA private key`);
+      }
+    } catch (err) {
+      if (err.code === 'ENOENT') {
+        return undefined;
```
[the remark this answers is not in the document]
sangaman: If I simply returned nothing, I got a warning because "not all code paths return a value." But I just realized if I change the return type to `NodeKey | void` it allows me to drop this statement, so I'll do that.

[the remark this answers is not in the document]
moshababo: I think you should always return `NodeKey`, and throw exception on error. Otherwise, we'll have to check if the object was initialized or not.

Why did you wanted to avoid throwing an exception if the error is specifically `ENOENT`?

lib/Xud.ts
sangaman: It felt a bit unnatural putting this all into `NodeKey`, also this logic will have to be expanded somewhat with waiting for a password to be provided, so I put this into its own method in `Xud` where I think it makes more sense as part of the initialization procedure. I figure we will want rpc methods to encrypt and decrypt the key eventually but we can discuss more later.

kilrau:
Did you make entering a password mandatory? @sangaman  I believe we have to do this, we'd store unencrypted otherwise!? Also, just regenerating keys when no password is set can lead to undesired behavior (someone whitelisted me, but I loose this status, because my `XUD_ID` changed).

sangaman:
@kilrau Prompting for a password is not actually part of this PR, mainly because I'd want to do that through an rpc method and we're going to be switching to grpc soon. As for whether it should be mandatory, I wasn't planning on it but we could do it that way. Bitcoin doesn't require encrypted wallets last I checked, and there's potentially a lot more at stake there. We can discuss.

sangaman:
Just rebased this after the latest changes to master. @moshababo could you see if the changes I made earlier look good to you?

lib/Xud.ts
```diff
@@ -41,6 +57,8 @@ class Xud {
     this.logger.info('config loaded');
 
     try {
+      this.loadNodeKey(`${this.config.xudir}/nodekey.dat`);
```
moshababo: It would be better to have the util function to handle only the getting of the instance, and to decouple it from the assigning.
`this.nodeKey = this.loadNodeKey(...)`

lib/nodekey/NodeKey.ts
```diff
@@ -0,0 +1,86 @@
+import CryptoJS from 'crypto-js';
+import { randomBytes } from 'crypto';
+import secp256k1 from 'secp256k1';
```
moshababo: What about creating a `crypto` layer/folder, to wrap up all usage in external libs?

sangaman: I'm not sure how this would be structured. All these modules are from third party dependencies outside this code module, would the `crypto` layer just import these and re-export them? I'm not sure where we'd be using this layer either. My preference would be to leave this alone for now until we have a better idea of how we're using cryptography outside of `NodeKey` and ecdsa sign/verify.

moshababo: That's fine, we can leave this out for now.

lib/Xud.ts
sangaman: When I did this, I realized it really made more sense for this to be in `NodeKey.ts` after all. So I put it there for now, and we'll see if anything needs to get rearranged when we do the rpc call to encrypt/decrypt a key.

PR #62, opened 2018-05-08
Refactoring Logger for TypeScript

sangaman:
Just some refactoring around the typescript migration for our `Logger` module. `@types/winston` is not up to date with version 3.0 which we're using, but it looks like it will be by the end of the month, so we can incorporate that then.

### DIFF
--- a/lib/Logger.ts
+++ b/lib/Logger.ts
@@ -4,48 +4,40 @@ import winston from 'winston';
 
 import utils from './utils/utils';
 
+enum Level {
+  ERROR = 'error',
+  WARN = 'warn',
+  INFO = 'info',
+  VERBOSE = 'verbose',
+  DEBUG = 'debug',
+}
+
+enum Context {
+  GLOBAL,
+}
+
+const contextFileMap = {
+  [Context.GLOBAL]: 'xud.log',
+};
+
 class Logger {
-  static levels = {
-    ERROR: 0,
-    WARN: 1,
-    INFO: 2,
-    VERBOSE: 3,
-    DEBUG: 4,
-  };
-
-  static levelsVal = {
-    [Logger.levels.ERROR]: 'error',
-    [Logger.levels.WARN]: 'warn',
-    [Logger.levels.INFO]: 'info',
-    [Logger.levels.VERBOSE]: 'verbose',
-    [Logger.levels.DEBUG]: 'debug',
-  };
-
-  static contexts = {
-    GLOBAL: 0,
-  };
-
-  static contextsFilename = {
-    [Logger.contexts.GLOBAL]: 'xud.log',
-  };
+  level: string;
+  logDir: string;
+  context: Context;
+  logger: any;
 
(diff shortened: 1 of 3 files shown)


lib/Logger.ts
```diff
@@ -4,48 +4,36 @@ import winston from 'winston';
 
 import utils from './utils/utils';
 
+enum Levels {
```
moshababo: Let's stick to singular TS enum naming (`Level`)

lib/Logger.ts
```diff
@@ -4,48 +4,36 @@ import winston from 'winston';
...
+  ERROR = 'error',
+  WARN = 'warn',
+  INFO = 'info',
+  VERBOSE = 'verbose',
+  DEBUG = 'debug',
+}
+
+enum Contexts {
```
moshababo: I think this should not be merely an enum, but:
1) an enum, named `Context`
2) An object mapping from `Context` to a string for it's file name

It's seems a bit redundant, since we're still not using it, but I plan to. it's just on a low priority (should probably open an issue for it). Even if we'll dump all contexts into one file for now, I think we should use them, in order to put a tag  in the beginning of the logging message. It will help us to format our logs messages better.

lib/Logger.ts
sangaman: Good call, just forgot to rename these.

lib/Logger.ts
sangaman: Having the contexts is fine, but why do we need a separate mapping? I think this accomplishes what we need it for, as this enum both tracks the possible context values as well as the files they map to. Or is there a reason we'd want to use the arbitrary number values (0, 1, 2...) assigned to enums by default?

moshababo: Yes, enum values should be arbitrary number values (0, 1, 2, ...), unless other values are needed, to be usable for some purpose (like in the `Level` enum). 
Technically speaking, what you suggest might work. But it's still weird to have an enum which defines contexts, that has as a value a file name. It won't make a good self-explanatory code in my opinion. 

Thinking about it now, it will probably also create a bug if you'll have 2 contexts which maps to the same file name (non distinct values).

sangaman: I see this as a similar use case to the `Level` enum where we want to map the enums to specific values. It's possible to have 2 contexts with the same value in the enum, e.g. both `GLOBAL` and `NOT_GLOBAL` = `xud.log`, but could be confusing if we are comparing context values with each other. I'll add the mapping, but I think down the road if it turns out we're not doing comparisons with `Context` values we could treat it as a single enum.

moshababo: We would like to use `Context` not merely for having the file name (although it does't exists in the code now), so in that case having non-distinct value will break it.

sangaman: Sounds good, see if the current approach looks good to you. I originally tried making the mapping a static class property rather than its own variable, but it wasn't getting properly initialized that way for some reason.

PR #63, opened 2018-05-09
Use typescript convention for cli commands

sangaman:
This is a mainly janitorial change to bring the `commands` used by yargs in line with the typescript style and convention used in the rest of the project.


moshababo:
We could perhaps try making this solution more OOP style, although it's a low priority task.

PR #66, opened 2018-05-11
Updating xud command arguments

sangaman:
This updates the `xud` command line arguments along with the readme. I renamed `xuDir` to `xudir` to be consistent with the lower case config/arg names.


sangaman:
Good idea, I went ahead and made those changes to the readme.

PR #67, opened 2018-05-12
Fixed SSL issue with LND

michael1011:
This fixes the SSL issue `ssl_transport_security.cc:989] Handshake failed with fatal error SSL_ERROR_SSL: error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure.` with newer versions of LND.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -6,6 +6,9 @@ import BaseClient, { ClientStatus } from'../BaseClient';
 import errors from './errors';
 import * as lndrpc from './lndrpc_pb';
 
+/** Fixes a SSL issue with newer versions of LND */
+process.env.GRPC_SSL_CIPHER_SUITES = 'HIGH+ECDSA';
+
 /**
  * The configurable options for the lnd client.
  */


michael1011:
@sangaman Sorry, I wasn't aware of that. Will do it as you described it :+1:

sangaman:
Yup no problem, we had only discussed it very recently to establish an approach to setting environment variables, this is the first place where it's a requirement.

lib/Xud.ts
```diff
@@ -38,6 +39,9 @@ class Xud {
     await this.config.load();
     this.logger.info('config loaded');
 
+    /** Loads environment variables from the file .env */
+    dotenv.config();
+
```
moshababo: Please move this to the top of the file, after the imports and before the class definition.

moshababo: Please move this to the top of the file, after the imports and before the class definition. it's should be called once per process.

moshababo:
@michael1011 
Yes, it's not common to do so, but since we merely using it to safety inject non-sensitive vars, it's not a problem.
Please also remove it from `.gitignore`.

Thanks for the initiative!

michael1011:
Should be done @moshababo :wink:

.env
```diff
@@ -0,0 +1 @@
+GRPC_SSL_CIPHER_SUITES='HIGH+ECDSA'
```
sangaman: Can you add a new line at the end of this file?

PR #68, opened 2018-05-12
Add cross-platform scripts

sangaman:
This commit uses the `cross-os` package to enable separate npm scripts based on OS platform. Currently the only script affected is `compile`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1082,6 +1082,12 @@
         "sha.js": "^2.4.8"
       }
     },
+    "cross-os": {
+      "version": "1.3.0",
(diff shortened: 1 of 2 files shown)


itsbalamurali:
Looks `cross-os` needs to be globally installed. Instead can we refer it from the node_modules?

sangaman:
@itsbalamurali It's not globally installed for me. Did you do `npm install` after checking out this branch?

moshababo:
Working fine for me, installed locally.

@itsbalamurali - after doing `npm install`, check that you have `cross-os` in `/node_modules/.bin`. it supposed to work out of the box.

PR #69, opened 2018-05-14
Packet handling

moshababo:
Packet handling, basic OrderBook integration, order types

* packet message structure currently contains `type` and `body`. `type` will eventually be contained in `header`, when more metadata will be added
* created `types` folder to contain all our type/interface/class object definitions that might cause circular dependencies
* moved order invoice creation into OrderBook, because not all incoming ownOrders will need one


lib/p2p/Pool.ts
```diff
@@ -1,8 +1,12 @@
+import assert from 'assert';
 import net, { Server, Socket } from 'net';
 import { EventEmitter } from 'events';
 import Peer from './Peer';
 import Hosts from './Hosts';
 import NetAddress from './SocketAddress';
+import Packet from './packets/Packet';
```
sangaman: Can this be imported from `./packets` below as `{ Packet }`?

lib/p2p/packets/packets.ts
```diff
@@ -0,0 +1,15 @@
+import PacketType from './PacketType';
+import OrderPacket from './OrderPacket';
+
+const fromRaw = (type:string, body:string) => {
```
sangaman: This file feels a bit out of place to me. I think the singular `packet` makes more sense, for consistency and since this only is representing/creating a single packet. But we already have a `Packet` class. I'm leaning towards getting rid of this file, and moving the `fromRaw` method to a static method on `Packet` (I checked and abstract classes can have static methods).

lib/p2p/packets/OrderPacket.ts
```diff
@@ -0,0 +1,35 @@
+import Packet from './Packet'
```
sangaman: Missing semicolon.

lib/p2p/packets/OrderPacket.ts
```diff
@@ -0,0 +1,35 @@
...
+import PacketType from './PacketType';
+import { OutgoingOrder } from '../../types';
+
+type OrderPacketBody = {
+  id: string;
+  pairId: string;
+  quantity: number;
+  price: number;
+  invoice: string;
+}
```
sangaman: Missing semicolon.

lib/p2p/packets/OrderPacket.ts
```diff
@@ -0,0 +1,35 @@
...
+class OrderPacket extends Packet {
+  public type: PacketType = PacketType.ORDER;
+
+  constructor(public body: OrderPacketBody) {
+    super();
+  }
+
+  static fromOutgoingOrder(order: OutgoingOrder) {
+    const { id, pairId, quantity, price, invoice } = order;
+    return new OrderPacket({ id, pairId, quantity, price, invoice } );
```
sangaman: Remove whitespace near end of line between `} )`.

lib/p2p/packets/packets.ts
```diff
@@ -0,0 +1,15 @@
...
+  switch(type) {
```
sangaman: Space after `switch`.

lib/p2p/packets/packets.ts
```diff
@@ -0,0 +1,15 @@
...
+    case PacketType.ORDER:
+      return OrderPacket.fromRaw(body);
+    default:
+      return null;
+  }
+}
```
sangaman: Missing semicolon.

lib/p2p/packets/packets.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default {
+  fromRaw
```
sangaman: Trailing comma

lib/p2p/packets/packets.ts
```diff
@@ -0,0 +1,15 @@
...
+}
```
sangaman: Semicolon

lib/orderbook/OrderBook.ts
```diff
@@ -82,37 +72,71 @@ class OrderBook {
     return this.repository.getOrders();
   }
 
-  async addOrder(order) {
+  async addOwnOrder(order: OwnOrder) {
```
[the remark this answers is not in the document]
moshababo: Good point, but i'm dealing with this area in my next PR (orderBook stuff)

lib/orderbook/OrderBookRepository.ts
```diff
@@ -73,3 +78,4 @@ class OrderbookRepository {
 }
 
 export default OrderbookRepository;
+export { Orders };
```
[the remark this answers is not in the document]
moshababo: I assume the dependency graph will support it. 

It’s a type which is relevant only for this class (OrderBookRepository), so as always it’s easier to be consumed together with it, and in this case I couldn’t predict a circular dependency situation, because these who consume this class (app-level stuff) will never be consumed by it. 

So basically, the db layer is still low enough to be able to export types.
We can argue that it’s better to put all type definitions in the `types` folder, but i’m not sure about it.

lib/p2p/Pool.ts
moshababo: yep, forgot to update that

lib/p2p/packets/packets.ts
moshababo: I’m not in favor with putting it on `Packet` as a static method, because it will create circular dependencies. 

What we have here is basically a simple function for mapping value to instance. I thought of doing it by using a simple object, but left it as it is for now, in case more logic will be needed. 

So if the name is the problematic thing here, we can either:
* export it as a stand-alone function 
* export it under `packet`, and turn `Packet` to `BasePacket`

moshababo:
Thanks, please see my replies.

In regards to the linting errors - for some reason none of the problems you described are caught by running `npm run lint`. We need to verify is something is different on your dev env, that you're getting these alerts.
In the meanwhile I fixed them by following your comments.

lib/p2p/packets/packets.ts
sangaman: Ah OK yeah I can envision the circular dependencies now that you mention it. I'd like your second bullet point, I think the naming will just be more clear. Also, on second look you should be able to declare a return type of `BasePacket` here.

sangaman:
@moshababo OK, let's try and figure out what's going on with the linting this week when we're both online. It looks like there are two more files that need new line EOF, and then some naming changes (see my comment) around `packets.ts`.

lib/p2p/packets/packets.ts
moshababo: Yes, I added the return type now. 

After trying what I suggested, I didn't found calling it `packet` very good. It result with something like this:
`const packet = packet.fromRaw(...)` 
which I didn't even wanted to fix by different naming, because I found `packet` to be the better term for the instance.

So instead, I ended up calling it `packetUtils`

lib/p2p/packets/Packet.ts
```diff
@@ -0,0 +1,9 @@
+import PacketType from './PacketType';
+
+abstract class Packet {
+  public abstract type: PacketType;
+  public abstract body: any;
+  public abstract toRaw() : string;
```
sangaman: Remove space between `toRaw() :`

lib/p2p/packets/packetUtils.ts
```diff
@@ -0,0 +1,16 @@
...
+    case PacketType.ORDER:
+      return OrderPacket.fromRaw(body);
+    default:
+      return;
+  }
+};
+
+export default {
+  fromRaw,
+}
```
sangaman: Missing semicolon

PR #70, opened 2018-05-14
Added config values for LND

michael1011:
I added values for _host_, _certificate_ and _macaroon_ file to the config. The default values stayed the same.


lib/lndclient/LndClient.ts
```diff
@@ -12,6 +12,9 @@ import * as lndrpc from './lndrpc_pb';
 type LndClientConfig = {
   disable: boolean;
   datadir: string;
+  certificate: string;
+  host: string;
+  macaroon: string;
```
sangaman: Add port and rename the path fields.

lib/Config.ts
```diff
@@ -64,6 +65,9 @@ class Config {
     this.lnd = {
       disable: false,
       datadir: lndDatadir,
+      certificate: path.join(lndDatadir, 'tls.cert'),
+      host: '127.0.0.1:10009',
+      macaroon: path.join(lndDatadir, 'admin.macaroon'),
```
sangaman: Add port and rename the path fields.

bin/xud
```diff
@@ -53,6 +53,18 @@ const { argv } = require('yargs')
       describe: 'Data directory for lnd',
       type: 'string',
     },
+    'lnd.certificate': {
+      describe: 'Certificate for lnd',
```
sangaman: Maybe call this `certpath` so it's more clear what this is used for just from the name. Also I think the description could be more clear as "Path to the SSL certificate for lnd."

bin/xud
```diff
@@ -53,6 +53,18 @@ const { argv } = require('yargs')
...
+      type: 'string',
+    },
+    'lnd.host': {
```
sangaman: I think we'd prefer to separate port and host by adding a `port` config option. Adding host/port for the lnd client is part of #58, actually, but lets keep that as part of this PR since it's more focused.

bin/xud
```diff
@@ -53,6 +53,18 @@ const { argv } = require('yargs')
...
+      describe: 'Host of the lnd gRPC interface',
+      type: 'string',
+    },
+    'lnd.macaroon': {
+      describe: 'Macaroon file for lnd',
```
sangaman: Likewise, this can be `macaroonpath` and the description can say it's the "Path to the admin macaroon for lnd."

michael1011: Do we really the admin macaroon? I think the only reason you are using it instead of the invoice macaroon is that you are using the `getinfo` method of LND? And it is just a placeholder for things to come later i suppose.

PR #71, opened 2018-05-15
Migrate JSON-RPC to GRPC 

itsbalamurali:
This PR implements Initial Skeleton of GRPC Server replacing existing JSON-RPC Server and a Rest API implemented as a proxy around GRPC service. Currently, the service doesn't have any kind of authentication mechanism.

@sangaman  & @moshababo  this PR is WIP, but please provide your feedback and suggestions on the same.


moshababo:
Sorry for the delay, will get to it tomorrow morning.

lib/Config.ts
```diff
@@ -61,6 +63,10 @@ class Config {
     this.rpc = {
       port: 8886,
     };
+    this.api = {
```
moshababo: * I think we should put `api` inside of `rpc` property, because it's not a separate module
* Perhaps should call it `webProxy`. `api` is too generic
* instead of `listen`, lets use `disable`, which is our usual pattern

lib/grpc/GrpcMethods.ts
```diff
@@ -1,13 +1,13 @@
...
+import { GrpcComponents } from './GrpcServer';
 import Logger from '../Logger';
 import Pool from '../p2p/Pool';
 import OrderBook, { Order } from '../orderbook/OrderBook';
 import LndClient from '../lndclient/LndClient';
 import RaidenClient, { TokenSwapPayload } from '../raidenclient/RaidenClient';
 
 /** Class containing the available RPC methods for Exchange Union */
-class RpcMethods implements RpcComponents {
+class GrpcMethods implements GrpcComponents {
```
moshababo: Let's call this class GrpcService, because that's the terminology in the protobuf

lib/grpc/GrpcMethods.ts
```diff
@@ -37,57 +35,62 @@ class RpcMethods implements RpcComponents {
...
   /**
-   * Connect to an XU node on a given host and port. See [[Pool.addOutbound]]
+   * Connect to an XU node on a given host and port.
    */
-  async connect(params) {
-    const peer = await this.pool.addOutbound(params.host, params.port);
-    return peer.statusString;
+  async connect(call, callback) {
+    const host = call.request.host;
+    const port = call.request.port;
```
moshababo: nothing major, but you can do it like this:
`const { host, port } = call.request;`

lib/grpc/google/api/annotations.proto
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2015, Google Inc.
```
moshababo: what's the usage for this file?

lib/grpc/google/api/http.proto
```diff
@@ -0,0 +1,291 @@
+// Copyright 2016 Google Inc.
```
moshababo: what's the usage for this file?

lib/grpc/http.proto
```diff
@@ -0,0 +1,291 @@
+// Copyright 2016 Google Inc.
```
moshababo: what's the usage for this file?

lib/xuclient/XUClient.ts
```diff
@@ -14,91 +14,95 @@ class XUClient {
     if (host) {
       this.host = host;
     }
+    this.port = port || 8886;
+    const xurpc: any = grpc.load(__dirname + '/xud.proto', 'proto');
```
moshababo: this causes the client to fail because you don't copy its `xud.proto` to the `dist` folder

lib/grpc/google/api/annotations.proto
itsbalamurali: They are required for the proxy/API gateway functionality. we define the endpoints inform of annotations.

lib/grpc/google/api/http.proto
itsbalamurali: same as above one.

lib/grpc/google/api/annotations.proto
moshababo: if so, please put them in a folder called `webProxy`, inside `grpc`, which can also contain `GrpcWebProxy.ts`. All the proxy stuff should be a separate sub module

lib/xuclient/XUClient.ts
moshababo: you duplicated the protobuf in`xuclient` folder, which makes sense to some degree, because the client actually should be detached from the project.
So while having the file there, you just need to copy to `dist` during the compile procedure.

If you want to use the `lib/grpc/xud.proto` one, then remove the duplication from `xuclient`, and let me know if you need help with targeting the file path.

itsbalamurali:
@moshababo  need your help in streamlining protobuf response message definitions.

moshababo:
@itsbalamurali can you elaborate more on this?

sangaman:
Closing in favor of #103 which rebases and builds on top of this branch.

Issue #29, opened 2018-04-21
RpcMethods.js getOrders() needs params


kilrau:
https://github.com/ExchangeUnion/xud/blob/master/lib/rpc/RpcMethods.js

*getOrders()*

An exchange will only be interested in this method to display some orders in their order book UI. Thus they probably only want to get the best 10 buy and sell orders. I think we need a parameter like an integer for the number of buy and sells (ordered by best price from top to bottom starting with the furthest sell, to the best sell, then starting with the best buy to the furthest buy). 0 returns all.

sangaman:
Number of orders to return is good, it should also take a `pair_id` parameter I'm thinking. I like 0 to indicate a return of all orders, maybe if nothing is passed in it should default to something, say 100 orders from each side of the book? I'll work on this soon, let me know if what I've said makes sense.

kilrau:
default all pairs, default 100 buy and 100 sell also cool

Chikel:
I can tackle this if you'd like.

michael1011:
@Chikel I have already done this in #78

moshababo:
solved by #78

PR #78, opened 2018-05-17
Added parameters to getOrders

michael1011:
Added the parameters `max_results` and `pair_id` to the RPC and CLI as described in #29.

### DIFF
--- a/bin/xud
+++ b/bin/xud
@@ -69,10 +69,6 @@ const { argv } = require('yargs')
       describe: 'Path of the admin macaroon for lnd',
       type: 'string',
     },
-    'lnd.rpcprotopath': {
-      describe: 'Path to lndrpc.proto file',
-      type: 'string',
-    },
     'raiden.disable': {
       describe: 'Disable raiden integration',
       type: 'boolean',
@@ -81,6 +77,11 @@ const { argv } = require('yargs')
     'raiden.port': {
       describe: 'Port for raiden REST service',
       type: 'number',
+    },
+    'orderbook.internalmatching': {
+      describe: 'OrderBook Internal Matching',
+      type: 'boolean',
+      default: undefined,
     },
   });
 
--- a/bin/xud
+++ b/bin/xud
@@ -9,6 +9,10 @@ const { argv } = require('yargs')
       type: 'number',
       alias: 'r',
     },
+    'rpc.defaultOrderAmount': {
+      describe: 'How many orders to return by default',
+      type: 'number',
+    },
     'p2p.port': {
       describe: 'Port to listen for incoming peers',
       type: 'number',
--- a/bin/xud
+++ b/bin/xud
@@ -9,10 +9,6 @@ const { argv } = require('yargs')
       type: 'number',
       alias: 'r',
     },
-    'rpc.defaultOrderAmount': {
-      describe: 'How many orders to return by default',
-      type: 'number',
-    },
     'p2p.port': {
       describe: 'Port to listen for incoming peers',
       type: 'number',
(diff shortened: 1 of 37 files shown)


lib/cli/command.ts
```diff
@@ -14,7 +14,7 @@ export default (argv: Arguments, callback: (XUClient, Arguments) => Promise<any>
 
   if (promise) {
     promise.then((result) => {
-      console.log(JSON.stringify(result));
+      console.log(JSON.stringify(result, null, 2));
```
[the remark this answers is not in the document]
michael1011: That was just for me to make the output of `xucli` more readable by adding an indent. And I left it because reading beautified JSON is much more fun.

bin/xud
```diff
@@ -9,6 +9,10 @@ const { argv } = require('yargs')
       type: 'number',
       alias: 'r',
     },
+    'rpc.defaultOrderAmount': {
+      describe: 'How many orders to return by default',
+      type: 'number',
+    },
```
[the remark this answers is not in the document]
michael1011: I am not sure. Third party applications and exchanges would use some kind of default amount anyway. If they specify it every time they request the orders or can configure it in the config of `xud` doesn't really matter. 

 But if minor details like this get added to the config it could get cluttered quickly.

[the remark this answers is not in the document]
moshababo: @sangaman any thought on that?

lib/orderbook/OrderBookRepository.ts
```diff
@@ -36,15 +46,25 @@ class OrderbookRepository {
     };
   }
 
-  async getOwnOrders() {
+  async getOwnOrders(maxResults?: number, pairId?: string) {
+    const whereClauseBuy: any = { quantity: { [Op.gt]: 0 }, peerId: { [Op.eq]: null } };
+    const whereClauseSell: any = { quantity: { [Op.lt]: 0 }, peerId: { [Op.eq]: null } };
+
+    if (pairId !== undefined && pairId !== '') {
```
moshababo: i think `if (pairId)` will be sufficient here

lib/orderbook/OrderBookRepository.ts
```diff
@@ -17,15 +17,25 @@ class OrderbookRepository {
     return this.models.Pair.findAll({ raw: true });
   }
 
-  async getOrders(): Promise<Orders> {
+  async getOrders(maxResults?: number, pairId?: string): Promise<Orders> {
+    const whereClauseBuy: any = { quantity: { [Op.gt]: 0 }, peerId: { [Op.ne]: null } };
+    const whereClauseSell: any = { quantity: { [Op.lt]: 0 }, peerId: { [Op.ne]: null } };
+
+    if (pairId !== undefined && pairId !== '') {
```
moshababo: i think if (pairId) will be sufficient here. in case empty string are getting here, we need to stop them beforehand.

lib/cli/commands/getorders.ts
```diff
@@ -2,12 +2,21 @@ import callback from '../command';
 import XUClient from '../../xuclient/XUClient';
 import { Arguments } from 'yargs';
 
-export const command = 'getorders';
+export const command = 'getorders [max_results] [pair_id]';
```
sangaman: Let's put `[pair_id]` first here. I see it being more useful to be able to specify a pair but leave the `max_results` as default, for example `getorders BTCLTC`. Would also need to update the example in the README.

lib/rpc/RpcMethods.ts
```diff
@@ -5,6 +5,7 @@ import Pool from '../p2p/Pool';
 import OrderBook, { Order } from '../orderbook/OrderBook';
 import LndClient from '../lndclient/LndClient';
 import RaidenClient, { TokenSwapPayload } from '../raidenclient/RaidenClient';
+import Config from '../Config';
```
sangaman: This line can be dropped because it's not being used in this file any longer.

PR #75, opened 2018-05-15
Db travis tests

itsbalamurali:
Resolves #73 The main objective of this PR is TravisCI's `MySQL` connectivity for DB integration tests but not integration tests themselves(which are failing currently and needs to be fixed).

PR #81, opened 2018-05-20
Database fixes

michael1011:
This pr fixes 

- the integration test failing if the database does not exist before running
- the error `Specified key was too long; max key length is 767 bytes` by enforcing the database charset `utf8`

and adds the TypeScript files in the tasks folder to `npm run lint`.

Edit: this fixes also the issues with Travis and the integration tests in #75

### DIFF
--- a/lib/db/DB.ts
+++ b/lib/db/DB.ts
@@ -112,7 +112,7 @@ class DB {
     try {
       this.sequelize = new Sequelize('', this.sequelize.options.username, this.sequelize.options.password, this.sequelize.options);
       await this.sequelize.authenticate();
-      await this.sequelize.query(`CREATE DATABASE ${this.sequelize.options.database};`);
+      await this.sequelize.query(`CREATE DATABASE ${this.sequelize.options.database} CHARACTER SET utf8 COLLATE utf8_general_ci;`);
       await this.sequelize.query(`USE ${this.sequelize.options.database}`);
     } catch (err) {
       this.logger.error('unable to create the database', err);
--- a/lib/db/DB.ts
+++ b/lib/db/DB.ts
@@ -18,13 +18,20 @@ class DB {
   models: any;
   logger: any;
   sequelize: Sequelize;
+  config: DBConfig;
 
   constructor(config: DBConfig) {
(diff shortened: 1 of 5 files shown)


lib/db/DB.ts
```diff
@@ -112,8 +119,13 @@ class DB {
     try {
       this.sequelize = new Sequelize('', this.sequelize.options.username, this.sequelize.options.password, this.sequelize.options);
       await this.sequelize.authenticate();
-      await this.sequelize.query(`CREATE DATABASE ${this.sequelize.options.database};`);
+      await this.sequelize.query(`CREATE DATABASE ${this.sequelize.options.database} CHARACTER SET utf8 COLLATE utf8_general_ci;`);
       await this.sequelize.query(`USE ${this.sequelize.options.database}`);
+      this.close();
+
+      // If the models aren't reinitialized the integration test fails
+      this.initSequelize();
```
moshababo: this doesn't seems to solve the problem when I test it on my env.
are you sure it makes your `test:int` pass on the first run, and not the second one?

lib/db/DB.ts
```diff
@@ -112,8 +119,13 @@ class DB {
     try {
       this.sequelize = new Sequelize('', this.sequelize.options.username, this.sequelize.options.password, this.sequelize.options);
       await this.sequelize.authenticate();
-      await this.sequelize.query(`CREATE DATABASE ${this.sequelize.options.database};`);
+      await this.sequelize.query(`CREATE DATABASE ${this.sequelize.options.database} CHARACTER SET utf8 COLLATE utf8_general_ci;`);
```
moshababo: do you know which key/column caused error before you applied this charset?

lib/db/DB.ts
michael1011: Did you run `npm run compile` after checking out my branch before you ran `npm run test:int`?

Because it works just fine here :thinking:

lib/db/DB.ts
moshababo: good to know 👍

lib/db/DB.ts
moshababo: you're right ;)
I didn't ran your branch properly. Now when I did, it works.

Thanks for finding the problem & the solution. We're gonna refactor this whole area (DB, tests, gulp), but it's good to have something which is already working.

PR #83, opened 2018-05-22
Fix lint script issues on linux

sangaman:
This addresses the mysterious recent issue of `npm run lint` not working on linux/mac. I've tested the new script on both windows and linux and it works as expected. @moshababo could you just sanity test this on mac?


sangaman:
This appears to be working on mac so I'll merge.

FYI, I got the idea for the solution from: https://stackoverflow.com/questions/36447910/how-to-lint-entire-folder-using-tslint .

Issue #80, opened 2018-05-19
Fresh install - problem with lndrpc.proto


offerm:
Installed xud for the first time. Tried to follow the instructions  on [https://github.com/ExchangeUnion/xud](url):

Started xud from bin directory without a config file resulted with:

> 2018-5-19 20:53:30 error: TypeError: Cannot read property 'ns' of null
    at Object.load (/home/offerm/github/xud/node_modules/grpc/index.js:146:29)
    at new LndClient (/home/offerm/github/xud/dist/lib/lndclient/LndClient.js:31:52)
    at Xud.<anonymous> (/home/offerm/github/xud/dist/lib/Xud.js:49:34)
    at Generator.next (<anonymous>)
    at fulfilled (/home/offerm/github/xud/dist/lib/Xud.js:4:58)
    at <anonymous>

Had to create a config file and change 

> rpcprotopath = "lndrpc.proto"

to

> rpcprotopath = "../lndrpc.proto"

michael1011:
You are right. This is an issue with the readme. `xud` should either be run in the root directory of the repository or with `npm start`.

offerm:
Yes. Starting from XUDHOME with npm is OK.

If one goes by the readme and fix the lndrpc manually she will get the following error  on getinfo:

> 2018-5-19 22:12:12 debug: RPC server request: {"method":"getInfo","id":1,"jsonrpc":"2.0"}
E0519 22:12:12.839404574  114166 ssl_transport_security.cc:190] ssl_info_callback: error occured.

> E0519 22:12:12.839457173  114166 ssl_transport_security.cc:979] Handshake failed with fatal error SSL_ERROR_SSL: error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure.
> 2018-5-19 22:12:12 error: RPC server request (1) error undefined
> E0519 22:12:13.841028264  114166 ssl_transport_security.cc:190] ssl_info_callback: error occured.
> 

Best would be if xud will not be sensitive for current wd.

sangaman:
Thanks @offerm, this is something we've known about but haven't fixed yet. We definitely want to be able to run `xud` from any directory. I think now that we're using `dotenv`, probably the best approach is to move the proto file path from the config file to the `.env` file and make it an absolute path.

offerm:
Thanks @sangaman, as a minimum I think the setup instructions should be changed.

sangaman:
Resolved by #85.

PR #85, opened 2018-05-22
Use __dirname to load lndrpc.proto

sangaman:
This solves the bug from #80, allowing `xud` to be run from any working directory without interfering with the default path to `lndrpc.proto`. I don't think there's any need to keep this configurable, as even multiple instances of `xud` should be using the same version of `lnd`.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -1,5 +1,6 @@
 import grpc, { Metadata, ChannelCredentials, StatusObject } from 'grpc';
 import fs from 'fs';
+import path from 'path';
 
 import Logger from '../Logger';
 import BaseClient, { ClientStatus } from '../BaseClient';
@@ -16,7 +17,6 @@ type LndClientConfig = {
   macaroonpath: string;
   host: string;
   port: number;
-  rpcprotopath: string;
 };
 
 /** A class representing a client to interact with lnd. */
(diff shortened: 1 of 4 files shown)


lib/lndclient/LndClient.ts
```diff
@@ -42,6 +42,7 @@ class LndClient extends BaseClient{
     } else {
       const lndCert: Buffer = fs.readFileSync(certpath);
       const credentials: ChannelCredentials = grpc.credentials.createSsl(lndCert);
+      const rpcprotopath = path.join(__dirname, '..', '..', '..', 'lndrpc.proto');
```
moshababo: 1) can be written as `path.join(__dirname, '../../../lndrpc.proto')`. If you knew it and preferred the above, you can leave 
1) any reason for not putting `lndrpc.proto` inside of `lndclient` folder? (considering `lndrpc_pb.d.ts` is already there)

sangaman: Yeah I knew it could be either way, seems like both ways are equally good and effective so I will just leave as is.

As for putting `lndrpc.proto` inside of `lndclient`, I had thought of it but then decided against it. It's not part of our code and it doesn't get compiled, so it seemed like it would be out place in `lib` (`lndrpc_pb.d.ts` does get compiled and is part of our source code). Also, I thought it might cause a problem because the file will actually run within the `dist` folder and I wasn't sure if `.proto` files would get moved over from `lib` to `dist`, maybe it would but I didn't bother testing it.

Issue #74, opened 2018-05-15
DB connection error inspection


moshababo:
At the moment we inspect any `Sequelize.ConnectionError` as if it indicates that the database doesn't exists, although that error type is thrown on other scenarios as well.

A better inspection to the thrown error object should be made.

PR #88, opened 2018-05-25
TS database related migration

moshababo:
TS migration:
* Models initialization
* DB class
* Repositories
* `sequelize` usage
* DB row types

Also solves #74


lib/db/DB.ts
```diff
@@ -91,41 +68,63 @@ class DB {
...
   }
 
-  async dbCreate() {
-    try {
-      this.sequelize = new Sequelize('', this.sequelize.options.username, this.sequelize.options.password, this.sequelize.options);
-      await this.sequelize.authenticate();
-      await this.sequelize.query(`CREATE DATABASE ${this.sequelize.options.database} CHARACTER SET utf8 COLLATE utf8_general_ci;`);
-      await this.sequelize.query(`USE ${this.sequelize.options.database}`);
-      this.close();
+  private createSequelizeInstance = (config: DBConfig|any): Sequelize.Sequelize => {
```
sangaman: `DBConfig|any` I believe is the same as just `any`. I think it's better to switch this to `DBConfig` and then make `database` nullable with `?` in the `DBConfig` declaration.

lib/orderbook/OrderBookRepository.ts
```diff
@@ -1,25 +1,24 @@
...
 };
 
 class OrderbookRepository {
   logger: Logger = Logger.global;
-  models: any;
 
-  constructor(db) {
+  constructor(private db: DB) {
     this.logger = Logger.global;
-    this.models = db.getModels();
```
sangaman: This class  seems to only use `db.models` but never `db` itself. Maybe go back to just having `models` as a class property and drop `db`? We can also change all references to this constructor and have it pass `db.models` instead of `db`. It's also slightly confusing to have `db` as an import as well as `this.db` as a property.

lib/types/db.ts
```diff
@@ -0,0 +1,53 @@
+import Sequelize, { DataTypeAbstract, DefineAttributeColumnOptions } from 'sequelize';
+import { db } from './index';
+
+export type SequelizeAttributes<T extends { [key: string]: any }> = {
+  [P in keyof T]: string | DataTypeAbstract | DefineAttributeColumnOptions
+};
+
+export type CurrencyFactory = {
```
sangaman: I always think of a "factory" in code as something that creates other objects. Here it seems to be used to describe a base set of properties that are used to create the complete version of an object, for example here the `CurrencyInstance`. I think maybe replacing "Factory" with "Request" (or a similar term) for the naming conventions in this file would make more sense - as in a "request" to create an order must have the following properties whereas an "instance" of an order must additionally have some other properties. Same would go for `Peer`, `Order`, and `Pair`.

lib/types/db.ts
```diff
@@ -0,0 +1,53 @@
...
+  id: string;
+};
+
+export type CurrencyAttributes = CurrencyFactory;
+
+export type CurrencyInstance = CurrencyAttributes & Sequelize.Instance<CurrencyAttributes>;
```
sangaman: I'm a little unclear as to how `Sequelize.Instance<CurrencyAttributes>` is being used here, but the result is that all the `Instance` types in this class are treated as `any`. I can assign a number value to `CurrencyInstance`, for example, and my compiler accepts it. So something is clearly off here.

lib/types/db.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type SequelizeAttributes<T extends { [key: string]: any }> = {
+  [P in keyof T]: string | DataTypeAbstract | DefineAttributeColumnOptions
+};
+
+export type CurrencyFactory = {
+  id: string;
+};
+
+export type CurrencyAttributes = CurrencyFactory;
```
sangaman: How should `Instance` differ from `Attributes`? In my mind it would make sense for `Attributes` to be the metadata that `xud` adds to particular objects, and that `Instance` to be a combination of the properties from the `Attributes` type and `Factory` (or "`Request`") type, but maybe I'm not following completely. It might also be helpful to throw in a few comments here since this was not easy for me to grasp right away.

lib/types/db.ts
moshababo: Factory - creating a new record
Attributes - the record
Instance - the record + operational properties

The usage of these “Factory” types is indeed for functions which creates objects (in the context of a database). Their usage in this file as a base set for “Attribute” and “Instance” is not very important - I did it only to avoid repeating property names. What’s important is that in the project level, they have a distinct role, and that's to the define the type for creating new records.

“Factory” might not be a perfect name, but I think that “Request” makes less sense. If you have other names in mind, tell me.

Sequelize actually requires the “Attribute” type when you want to create a new record. The only reason I used “Factory” in addition to that is because “Attribute” cannot support nullables, since it is being used also to define the table columns. 
So I couldn’t find a better way of supporting a more narrow object for creation, besides forking `@types/sequelize`, or adding this “Factory” type.

lib/types/db.ts
moshababo: I answered in other comments, but here's the differences:

Factory - what you need to provide to create a new record
Attributes - the full definition of the record
Instance - Attributes + db operations properties (you can look at `@types/sequelize`)

lib/db/DB.ts
```diff
@@ -14,68 +16,43 @@ type DBConfig = {
...
-  initSequelize() {
-    this.sequelize = new Sequelize({
-      ...this.config,
-      operatorsAliases: false,
-      dialectOptions: {
-        multipleStatements: true,
-      },
-    });
-    this.importModels();
+  static isDbDoesNotExistsError(err: Error): boolean {
```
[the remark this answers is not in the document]
moshababo: * hiding implementation details (more readable code)
* being able to test it (it has high chances of breaking at some point)

I'll fix the spelling mistake

lib/db/DB.ts
moshababo: I didn't want to do it because it can imply that `database` is an optional parameter for `xud`, wheres its an optional parameter only for this tiny helper method. The other option is to create another custom type, just for the "no-database" case. It will also decouple sequelize config from DBConfig, which is a good thing.

lib/orderbook/OrderBookRepository.ts
moshababo: I did it because I thought that calling on `models` is confusing, without having the `db.models` context. But I can change it in favor of passing `db.models` to the constructor instead.

moshababo: I did it because I thought that calling on `models` is confusing without having the `db.models` context. Passing `models` to constructor is not the best approach IMO, specially when you need to pass it along to inner layers (xud->orderbook->orderbookRepository). I can change it back to hold `models` as a class property, and keep passing `db` to constructor. we'll still need to import the `DB` type anyway.

lib/db/DB.ts
sangaman: I like your approach, thanks.

lib/orderbook/OrderBookRepository.ts
sangaman: Cool, that approach sounds good to me.

moshababo: Fixed it, but had to keep `db` as a class member on `P2PRepository` because it has a raw query method (we can wrap it, but it seems to me like an overkill for now).

lib/types/db.ts
sangaman: Thanks for the explanation, makes sense now. I think it'd be helpful to summarize these in some tsdoc comments.

As for naming, I agree "Request" doesn't make that much sense, but "Factory" still throws me off. I looked for a while and didn't find any word that really perfectly fits. Still, some options to consider: "Base," "Core," or "Source." I think "Core" is my favorite, but pick any of these and we'll go with it. Or if you don't like any of them, just stick with "Factory" and make sure to comment the type to clarify what it's for.

lib/types/db.ts
sangaman: Gotcha. I realized my problem was that I didn't run `npm install` after checking out this branch and therefore didn't have the `Sequelize` types available, which is why I was getting `any` here. So you can ignore this.

lib/types/db.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type CurrencyInstance = CurrencyAttributes & Sequelize.Instance<CurrencyAttributes>;
+
+export type PeerFactory = {
+  address: string;
+  port: string;
```
sangaman: One more thing I found here on a second pass, everywhere else that we have a `port` property it's a `number`. We should change this to use `number` to stay consistent.

lib/types/db.ts
moshababo: "Factory" still makes the most sense to me, being used in methods which create new records, so i'll stick to it for now.
I'll add some explanation in the beginning of this file.

lib/types/db.ts
moshababo: Thanks, i'll fix it

Issue #76, opened 2018-05-16
gulp inter-process communication problems


moshababo:
After discovering too many problems with using `gulp` through our tests, due to inter-process communication, I think we should get rid of this approach and create a base layer of task functions that can be consumed directly. `gulp` should still wrap it in order to create an easy tasks CLI. 

TODO:
* change our tasks to a plain JS functions, instead of `gulp` tasks
* update our integration tests to consume them directly, and not through `gulp-runner`
* update our `gulpfile` to expose them through the gulp CLI (as now)

PR #90, opened 2018-05-27
remove gulp tasks from tasks definition

moshababo:
addresses #76 

I left `gulp` for having an easy to use CLI, which also applies the `ts-node/register` hook.


tasks/db/restart.ts
```diff
@@ -1,6 +1,7 @@
...
+import dropTables from './dropTables';
+import init from './init';
 
-gulp.task('db.restart', (done) => {
-  runSequence('db.dropTables', 'db.init', done);
-});
+export default async (testDb?: boolean) => {
+  await dropTables(testDb);
+  await init(testDb);
+}
```
sangaman: Missing new line at end of file. Maybe we should make the `lint` script also lint the `tasks` folder?

tasks/index.ts
```diff
@@ -0,0 +1,5 @@
+import * as db from './db';
+
+export default {
+  db
+};
```
sangaman: New line EOF.

tsconfig.json
```diff
@@ -1,12 +1,14 @@
...
+    "test/*",
+    "tasks/*",
+    "gulpfile.ts"
   ],
   "compilerOptions": {
     "target": "ES2016",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
     "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
     "allowJs": true,                          /* Allow javascript files to be compiled. */
-    "outDir": "./dist",                       /* Redirect output structure to the directory. */
+    "outDir": "./dist/lib",                   /* Redirect output structure to the directory. */
```
sangaman: This seems to be causing me issues. When I compile and run, I get `Error: Cannot find module 'C:\Users\Daniel\Documents\GitHub\xud\dist\lib\Xud.js'`. The resulting directory structure is `dist/lib/lib/...` which seems wrong. If I revert this line, I get `dist/lib/...` which works. I think this may also affect loading of the `lnrpc.proto` file if it is changed. Can we simply leave this line as is?

moshababo: I guess it's a cross os compatibility issue, because before I added this change the compilation output structure was flat, meaning `Xud.js` location was `/dist/Xud.js` instead of `/dist/lib/Xud.js`.
This was introduced after I excluded `/tasks/*` and `gulpfile.ts` from compilation, which ended in `lib` being the only directory being compiled. 

So in order to remove this line we should either find a solution to enforce the `outDir` keep it's nested structure, or to just keep some unnecessary file/s to be compiled along with `lib`.

sangaman: I'm getting a `typedoc` folder within `dist` along with `lib`. Odd, I'm going to test this on linux as well and do some research.

moshababo:
I managed to get my tslint works properly only after I removed the `--project` argument from our npm `lint` script.
It caused a problem of excluding from tslint what we exclude from the compilation output (tasks, tests).

tsconfig.json
sangaman: I didn't get this issue with linux. I did a lot of digging and the only way I could make this issue go away on Windows was by changing `allowJs` to false. Which is weird, because there are no `.js` files in my `lib` folder. I'll open an issue with the typescript project. But for the time being, I don't think we want to be compiling any `.js` files anyway, so maybe just leave this line as is and change `allowJs` to false so it doesn't cause problems on Windows.

sangaman: Actually, another way to fix this was to change the `excludes` option to `includes` and only specifying the lib folder. So maybe we want to use the approach of including only the folders we want, rather than excluding folders we don't want. Either way, I'm fine with changing `allowJs` to false since I don't think we'll be using that.

sangaman:
The `--project` argument is used for one of the tslint rules and throws a warning if it's not specified. What do you think about just compiling the `tests` and `tasks` folders? We might have to include `bin` as well if we use that approach and still want any linting there, but disabling js compilation should mean it doesn't actually end up in our `dist` folder.

I checked pretty extensively if it was possible to lint folders excluded in `tsconfig.json` but it seems that's not currently possible.

tsconfig.json
```diff
@@ -1,12 +1,14 @@
 {
   "exclude": [
-    "test/*"
+    "test/*",
+    "tasks/*",
```
sangaman: The two lines above can simply be `test` and `tasks`. No need to have the `/*` at the end for everything in the folder to be included.

moshababo:
As to the combination of the two problems discussed (tslint + outDir), I ended up with removing `exclude` from `tsconfig`. Let me know if you have a better suggestion.

sangaman:
I think that's good enough for this PR. I have a few changes in mind but it's not really related to gulp tasks, so I'll propose them in a separate PR.

PR #92, opened 2018-05-27
Require member accessibility modifiers

sangaman:
This adds a new tslint rule to require accessibilty modifiers for all class members, and fixes all places in the code currently in violation of this rule. I erred on the side of making things `private` when I was unsure, if members need to be used in other classes in the future we can just change them to public.

### DIFF
--- a/lib/BaseClient.ts
+++ b/lib/BaseClient.ts
@@ -6,19 +6,19 @@ enum ClientStatus {
 }
 
 abstract class BaseClient {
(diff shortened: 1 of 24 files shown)


sangaman:
@moshababo Where I've been leaning regarding testing private static methods is that if a class starts getting more than just a few of them, move them into a separate utils class. Otherwise, allow them to be public for testing purposes.

PR #93, opened 2018-05-28
Add raiden wallet & channel methods

sangaman:
This adds several methods that are useful for managing payment channels and the raiden ETH account, as well as some minor refactoring to better use typescript.

### DIFF
--- a/lib/raidenclient/RaidenClient.ts
+++ b/lib/raidenclient/RaidenClient.ts
@@ -1,15 +1,14 @@
 import http from 'http';
-import assert from 'assert';
 
 import BaseClient, { ClientStatus } from'../BaseClient';
-import errors from './errors';
+import * as errors from './errors';
 
 /**
  * A utility function to parse the payload from an http response.
  */
-async function parseResponseBody(res: http.IncomingMessage): Promise<any> {
+async function parseResponseBody<T>(res: http.IncomingMessage): Promise<T> {
   res.setEncoding('utf8');
-  return new Promise<object>((resolve, reject) => {
+  return new Promise<T>((resolve, reject) => {
     let body: string = '';
     res.on('data', (chunk) => {
       body += chunk;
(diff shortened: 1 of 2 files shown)


lib/raidenclient/RaidenClient.ts
```diff
@@ -103,51 +133,86 @@ class RaidenClient extends BaseClient {
...
 
-    const endpoint: string = `channels/${channel_address}`;
-    const res: any = await this.sendRequest(endpoint, 'GET', undefined);
+  /**
+   * Get info about all non-settled channels.
+   */
+  public getChannels = async (): Promise<[Channel]> => {
+    const endpoint = 'channels';
+    const res = await this.sendRequest(endpoint, 'GET');
+    assert(res.statusCode === 200, `${res.statusCode}: ${res.statusMessage}`);
```
moshababo: I think we should throw an operational error (defined in the `errors` files), and keep the `assert` statements for scenarios that are unexpected, and ideally should never happen (unlike operational errors, which are expected to happen).

And if we're not interested in failing the entire process when that happens, we should remember to always wrap these RaidenClient calls in `try/catch` clause.

lib/raidenclient/RaidenClient.ts
```diff
@@ -4,6 +4,25 @@ import assert from 'assert';
 import BaseClient, { ClientStatus } from'../BaseClient';
 import errors from './errors';
 
+/**
+ * A utility function to parse the payload from an http response.
+ */
+async function parseResponseBody(res: http.IncomingMessage): Promise<any> {
```
moshababo: it's possible to add a generic type for the body parsing and apply it to the return type (instead of `any`)

sangaman:
Thanks for the feedback @moshababo. I added some new error types and used those, the error handling is pretty generic for now and it may eventually make sense to have more specific errors that pertain only to certain calls, but I think this is a decent start. I'll give you a chance to review the new changes if you want, then this can be merged.

PR #95, opened 2018-05-29
Refactor utils with method exports

sangaman:
Some minor refactoring/restructuring of the utils file to better use our typescript patterns.

### DIFF
--- a/lib/utils/utils.ts
+++ b/lib/utils/utils.ts
@@ -1,25 +1,21 @@
-const utils: any = {};
-
 /**
  * Check whether a variable is a non-array object
- * @returns {boolean}
  */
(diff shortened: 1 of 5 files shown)


lib/utils/utils.ts
```diff
@@ -65,4 +61,9 @@ utils.groupBy = (arr: object[], keyGetter: (item: object) => string | number): o
   return ret;
 };
 
-export default utils;
+export {
+  getTsString,
```
moshababo: if you don't want to export them united as the default export, then it's better to do `export const ...` on each one of them.

sangaman: Done.

Issue #99, opened 2018-05-31
Logger contexts


moshababo:
New `Logger` contexts are to be define, in addition to GLOBAL: DB, RPC, P2P, ORDERBOOK, LND, RAIDEN.
They should all have a static instance defined on `Logger` class, that should be used by the appropriate components, instead of `Logger.global`. 
They should all map to the same output file for now, same as GLOBAL (`xud.log`), but should have a unique output format : `[CONTEXT] ...`

michael1011:
I think this issue can be closed?

moshababo:
Yes.

PR #100, opened 2018-05-31
Added contexts for logger

michael1011:
This pr adds contexts to the logger as in #99.

But I am not sure about the fact that every context has its own static instance in the `Logger` class. What about a constructor that takes only the context as argument? `logDir` and `level` are the same anyway.

### DIFF
--- a/lib/Logger.ts
+++ b/lib/Logger.ts
@@ -13,7 +13,13 @@ enum Level {
 }
 
 enum Context {
-  GLOBAL,
+  GLOBAL = 'GLOBAL',
+  DB = 'DB',
+  RPC = 'RPC',
+  P2P = 'P2P',
+  ORDERBOOK = 'ORDERBOOK',
+  LND = 'LND',
+  RAIDEN = 'RAIDEN',
 }
 
 const contextFileMap = {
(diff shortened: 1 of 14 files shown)


lib/Logger.ts
```diff
@@ -13,7 +13,13 @@ enum Level {
...
+  GLOBAL = 'GLOBAL',
+  DB = 'DB',
+  RPC = 'RPC',
+  P2P = 'P2P',
+  ORDERBOOK = 'ORDERBOOK',
+  LND = 'LND',
+  RAIDEN = 'RAIDEN',
 }
 
 const contextFileMap = {
```
sangaman: I think you want to update `contextFileMap` too to have all the new contexts, even though for now they will all point to a single log file.

michael1011:
Isn't it possible to use an `enum` as parameter? Something like:
```
constructor(context: Context) {
```

sangaman:
Yeah I figure an enum would work. I guess one other advantage with static loggers is that you'll never use more than 1 instance of `Logger` per context. Is there a downside to how it's currently set up? Personally I don't have much preference either way.

michael1011:
I just think that
```

  public static global = new Logger({ context: Context.GLOBAL, logDir: undefined, level: undefined });
  public static db = new Logger({ context: Context.DB, logDir: undefined, level: undefined });
  public static rpc = new Logger({ context: Context.RPC, logDir: undefined, level: undefined });
  public static p2p = new Logger({ context: Context.P2P, logDir: undefined, level: undefined });
  public static orderbook = new Logger({ context: Context.ORDERBOOK, logDir: undefined, level: undefined });
  public static lnd = new Logger({ context: Context.LND, logDir: undefined, level: undefined });
  public static raiden = new Logger({ context: Context.RAIDEN, logDir: undefined, level: undefined });
```
isn't the most elegant way solving this.

moshababo:
@michael1011 Thanks!

I added a small fix for removing unnecessary props from the instantiation.

PR #101, opened 2018-06-01
p2p hosts

moshababo:
Another incomplete PR, so it won't get too big.

Main remarks:
* I changed `peers` table to `hosts`. The main reason behind it is to have a clear distinction between a node that you are currently connected to, and an information about a node which can potentially lead to connection.
* I added `bannedHosts` table, for the default blacklist mode. I used there only `address` and not `port` - first because you won't always have that information (in inbound connections the port is meaningless), but mainly because it needs to apply to all ports.
* `Host` class currently just wraps the db row instance, but later it will enable operations on a single host (which will be tied to a peer). Banning does't occur on a single host, because it need to apply to  `address` and not `port`, so that's why it is managed separately, by `HostList`.
* `Host` instance is added to `peer` only after the handshake. Basically there are 3 ways to create a peer: from the hosts list, by RPC call, or by an inbound connection. Only for the first case the Host entity exists beforehand, while in the other two it need to be created only after a successful handshake. So in order to avoid complexity, i'm adding the `Host` instance to `peer` only after the handshake, although In the first case I could have done in the instantiation stage.


lib/p2p/HostList.ts
```diff
@@ -0,0 +1,62 @@
...
+      const address = new SocketAddress(host.address, host.port);
+      this.hosts[address.toString()] = new Host(host);
+    });
+
+    this.bannedHosts = new Set<string>(bannedHosts.map(bannedHost => bannedHost.address));
+
+    this.loaded = true;
+  }
+
+  private create = async (socketAddress: SocketAddress): Promise<Host> => {
```
sangaman: Maybe call this `createHost`? It wasn't obvious to me from the name what was being created.

moshababo:
Because we're identifying a host by the `address` + `port` pair, and because we need to be able to ban an `address` which we don't associate with a port and so with a specific host entity, doing what you suggested is less straight-forward, and seems to me like a hack in the database scheme.

sangaman:
Alright, thanks for the explanation.

Issue #98, opened 2018-05-31
Orderbook persistence


moshababo:
`OrderBook.AddOwnOrder` is matching new incoming orders on `MatchingEngine`, which holds a cache of the database orders in priority queues. When a match occurs, orders are removed from the cache, so we need to remove them from the database too, according to `MatchingResult`. Because that partial matching is supported, we should only update the order quantity, so that an order which has been entirely removed should have a quantity set to 0. In addition, we need to make sure we don't query 0 quantity orders (in the `init` stage). It should be considered as a soft delete.

PR #102, opened 2018-06-02
Orderbook persistence

michael1011:
As in #98.
Orders with 0 quantity don't get queried because of the where clauses in `OrderBookRepository.ts`. `quantity: { [Op.gt]: 0 }` and `{ quantity: { [Op.lt]: 0 }` make sure that the quantity is either greater or less than 0.


lib/orderbook/OrderBook.ts
```diff
@@ -80,7 +80,18 @@ class OrderBook {
     const { matches, remainingOrder } = matchingResult;
 
     if (matches.length > 0) {
-      matches.forEach(this.handleMatch);
+      for (const [index, match] of matches.entries()) {
+        const { maker, taker } = match;
+        this.handleMatch({ maker, taker });
+
+        // Maker is fully matched if it is not the last
+        if ((matches.length - index) > 1) {
```
moshababo: although that's correct, it's better to not make assumptions about `MatchingEngine` here, and to treat all matches in the same way.

lib/orderbook/OrderBook.ts
```diff
@@ -80,7 +80,18 @@ class OrderBook {
...
+          await this.repository.updateOrder(maker, 0);
+        } else {
+          const initialQuantity = await this.repository.getOrderQuantity(maker);
+          await this.repository.updateOrder(maker, initialQuantity - maker.quantity);
```
moshababo: These two operations can be reduced to one, atomic. `sequelize` even expose an interface for it, so you won't have to do SQL for that. Here's an example: https://github.com/sequelize/sequelize/issues/7268

lib/orderbook/OrderBookRepository.ts
```diff
@@ -88,6 +88,19 @@ class OrderbookRepository {
     return this.models.Order.bulkCreate(<db.OrderAttributes[]>orders);
   }
 
+  public updateOrder = async (orders: orders.StampedOrder, quantityLeft: number) => {
```
moshababo: `orders` should be singular (`order`).

moshababo: to make this method even more generic, take as an argument only the order id.

test/integration/OrderBook.spec.ts
```diff
@@ -41,6 +44,16 @@ describe('OrderBook', () => {
...
 
+  it('should match new ownOrder and update matches', async () => {
+    const order: orders.OwnOrder = { pairId: 'BTC/LTC', quantity: 5, price: 100 };
+    const matches = await orderBook.addOwnOrder(order);
+    expect(matches.remainingOrder).to.be.null;
+    const firstMakerLeft = await orderBookRepository.getOrderQuantity(matches.matches[0].maker);
+    expect(firstMakerLeft).to.be.equal(0);
+    const secondMakerLeft = await orderBookRepository.getOrderQuantity(matches.matches[1].maker);
+    expect(secondMakerLeft).to.be.equal(-5.5);
+  });
```
moshababo: if we want to start making this test scenario based on a sequential flow, where each step depends on the previous one's state, we should should start with an empty database (or at least with an empty `orders` table). At the moment, arbitrary dummy values from `tasks.db.init` affect this test results.

michael1011: I used the dummy values from `tasks.db.init` for this test. Should I use the values from the tests before instead?

lib/orderbook/OrderBookRepository.ts
```diff
@@ -88,17 +90,8 @@ class OrderbookRepository {
...
-  }
-
-  public getOrderQuantity = async (order: orders.StampedOrder): Promise<number> => {
-    const result = await this.models.Order.find({
-      where: { id: order.id },
-      raw: true,
-      attributes: ['quantity'],
-    }) as db.OrderInstance;
-    return Number(result.quantity);
+  public updateOrderQuantity = async (orderId: string, usedQuantity: number) => {
```
moshababo: perhaps `quantityDelta` instead of `usedQuantity`. will make it more generic.

lib/orderbook/OrderBookRepository.ts
```diff
@@ -88,17 +90,8 @@ class OrderbookRepository {
...
+    await this.models.Order.update({ quantity: this.sequelize.literal(`quantity - ${usedQuantity}`) }, { where: { id: orderId } });
```
moshababo: i'm still getting a type warning here, despite the change you made on `types/db.ts`. see if there's a solution for that.

test/integration/OrderBook.spec.ts
moshababo: Yes, and please remove `tasks.db.restart` from the beginning of the tests (it's old stuff). 
Instead, you can call `db.dropTables` directly, and init some values yourself, by using the repositories objects, same as in the `init` tasks. You'll need to create some currencies and pairs, but no orders. 
You can also add `db.truncate` in the end of the tests (before `db.close`).

lib/orderbook/OrderBook.ts
```diff
@@ -83,14 +83,7 @@ class OrderBook {
       for (const [index, match] of matches.entries()) {
```
moshababo: * You don't use `index` anymore, so you can remove it, together with `entries()`.
* You can aggregate all the `updateOrderQuantity` async calls into one `Promise.all`, because they are not dependent on each other. It might perform faster.

test/integration/OrderBook.spec.ts
michael1011: Do I really need to drop all tables if I truncate them at the end of the test? And can I delete all `tasks` as they have no use anymore?
I will enable `internalmatching` for the `orderBook` in the test to use the orders created in the two tests before.

And sorry for being so sloppy and careless in this pr :disappointed:

lib/orderbook/OrderBookRepository.ts
michael1011: I don't get a warning at all. Do you get it when compiling or linting?

moshababo:
Thanks.
I made some changes myself, because there were all small ones.

* doing `db.truncate` in the beginning of the tests, instead in the end. each tests file should take care of himself ;)
* made the test db init in steps, due to FK constrains. at the previous approach, it would always fail if the database didn't exist (due to FK issues)
* moved the `Sequelize.literal` to the order's Attributes type instead of the Factory type, because it makes more sense (you don't create an order with a literal type)
* merged master, fixed an existing bug (peerId -> hostId)

PR #103, opened 2018-06-03
Migrate JSON-RPC to gRPC

sangaman:
This is the completion of the gRPC migration from #71. I merged all of @itsbalamurali's commits from the previous PR to maintain authorship and commit history, then rebased onto master.

I moved all our `.proto` files into a `proto` folder. There were a couple of reasons I thought this was best, `xudrpc.proto` has a dependency on `lndrpc.proto` and there may be more later. Most importantly, generating type declarations from proto files (as I did in #59) requires the `http.proto` and `descriptor.proto` files, so they're needed outside of just the web proxy component.

I also realized in testing that I needed to `bind` the `lightning` methods if I wanted to use them via the `unaryCall` method as implemented in #94, so I fixed that.

Aside from that there are no major changes from the approach in the first PR, mostly just bringing things in line with changes from the past two weeks, coding style, and correcting some asynchronous logic. I've run some basic manual tests and confirmed I can call the lnd grpc via the xud grpc, so the basics work.

There are a handful more things that ought to be done related to this migration but which I didn't want to all pack into this PR since it's big enough:

- Generate types for `xudrpc.proto` and apply them where appropriate in `GrpcService.ts` and `GrpcServer.ts`
- Make gRPC and web proxy listening host configurable
- Generate and use tls certificates for the xud gRPC
- Expose RPC API layer to be consumed directly by the server for high level testing
- Refactor `XUClient.ts`

When  merging this let's use a regular merge commit instead of a squash/merge commit to preserve the joint authorship. I tried to keep a clean/logical/compartmentalized commit history for this branch


moshababo:
Also, I didn't saw the `proto` folder you mentioned.

sangaman:
Ahhh I just realized I had rebased to introduce a few changes - including the `proto` folder, but forget to push out the rebased commits. Just did that.

moshababo:
I don't see any new commits. Please check that you pushed it properly.

sangaman:
The changes are there, but instead of adding new commits I had just amended the existing ones to maintain a clean commit history.

sangaman:
I had actually overlooked your general review comments. As for the defined request/response types, I actually already have been working on generating those from the `.proto` file and using them in a separate branch that builds off of this one. As for the "Services" class being separate from gRPC semantics, I agree but I don't think it needs to be in an additional class. Rather, how about renaming `GrpcServices` to `Services` and putting it in its own folder, this would also be a step towards exposing the API layer directly for testing, I figure.

I can add a stub streaming grpc call as part of this PR if you think it's , I've already looked into what's needed for that and it's not too hard. I also think it's fine to implement streaming methods in a separate PR tho.

moshababo:
I see. well, the `webproxy` still throws the exception. It doesn't occur on your env?

moshababo:
The main thing is that we want to have a separate, stand-alone TS class to define the `xud` service. 
This makes the gRPC service class pretty degenerate, but we still need to wrap the service class somehow.

itsbalamurali:
@sangaman having quite a few type errors on the library. Please refer to the screenshot below.

screen shot 2018-06-06 at 1 36 14 pm

I'd say create a new PR for this and port the module into typescript and include it in the project. It's nothing but an ExpressJS Middleware.

@moshababo  what do you think?

package.json
```diff
@@ -32,8 +32,8 @@
       "win32": "node bin\\xucli shutdown"
     },
     "compile": {
-      "linux": "rm -rf ./dist && tsc",
-      "darwin": "rm -rf ./dist && tsc",
+      "linux": "rm -rf ./dist && tsc && cross-os copy:protobufs",
```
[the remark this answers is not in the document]
sangaman: Indeed. I actually already fixed this in a significant set of changes I just pushed to github.

sangaman:
OK, I did a lot more work/testing/refactoring on this. Please take a look over the latest changes because they touch quite a bit.

Regarding the web proxy, @itsbalamurali I get the same error in your screenshot and I decided not to waste any more time trying to fix it as part of this PR. I left `grpc-dynamic-gateway` and the related code in there in case we find a way to get it to work, but I'm more than OK with getting rid of that module and using a different approach. I do think it would be cool to establish a truly RESTful layer and take advantage of the `option` definitions in the proto file, though.

One thing that I think might be worthwhile and would ultimately lead to more robust code is using statically generated grpc clients (for both lnd and xud) instead of the dynamic clients we use now. We'd need to generate `grpc-out` using the `grpc-node-plugin` for `protoc`, and then import the generated file directly as the `Service` instead of creating it at runtime using `grpc.load()`. I'll take another look at how practical that is after this PR is merged.

Updated todo list (I'll create issues for these after this is merged): 

- ~~Generate types for `xudrpc.proto` and apply them where appropriate in `GrpcService.ts` and `GrpcServer.ts`~~ DONE, included in this PR. The generated types however are really more relevant for the client side of things, not for the server.
- Make gRPC and web proxy listening host configurable
- Generate and use tls certificates for the xud gRPC
- ~~Expose RPC API layer to be consumed directly by the server for high level testing~~ DONE, included in this PR
- Refactor `XUClient.ts` - EDIT: This should at a minimum use the generated `xudrpc` types, but it might make sense to just use a static client from generated files as discussed above.
- NEW: Improve gRPC error handling to use [gRPC status codes](https://grpc.io/grpc/node/grpc.html#.status). Right now essentially every error is returning 2 `UNKNOWN`.
- NEW: Investigate and fix the web proxy layer, perhaps by creating our own module in typescript  that's a port of `grpc-dynamic-gateway` and using that. It should also be possible to generate the REST web layer statically, that might be the easier approach but I'll do some more research there.
- NEW: Create tests that test the API layer directly via `Xud.service`.

In addition to the above we can resume adding new rpc methods and/or modifying existing ones, I'll also create issues for calls that are a priority.

lib/service/Service.ts
```diff
@@ -0,0 +1,86 @@
...
+import Pool from '../p2p/Pool';
+import OrderBook from '../orderbook/OrderBook';
+import { orders, matchingEngine } from '../types';
+import LndClient from '../lndclient/LndClient';
+import RaidenClient, { TokenSwapPayload } from '../raidenclient/RaidenClient';
+import grpc from 'grpc';
+import { PairInstance } from '../types/db';
+import { GetInfoResponse } from '../proto/lndrpc_pb';
+import { Orders } from '../orderbook/OrderBookRepository';
+import { OwnOrder } from '../types/orders';
```
moshababo: few unused imports over here

lib/service/Service.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export type XudInfo = {
+  lnd: GetInfoResponse,
+};
+
+/** Class containing the available RPC methods for XUD */
+class GrpcService {
```
moshababo: this class should have nothing to do with gRPC.. so perhaps `Service` instead of `GrpcService`

lib/service/Service.ts
```diff
@@ -0,0 +1,86 @@
...
+  private orderBook: OrderBook;
+  private lndClient: LndClient;
+  private raidenClient: RaidenClient;
+  private pool: Pool;
+  private shutdown: Function;
+  private logger: Logger;
+
+  /** Create an instance of available RPC methods and bind all exposed functions. */
+  constructor(components: GrpcComponents) {
```
moshababo: same as above. in addition, `GrpcComponents` should be removed from `GrpcServer.ts`

lib/service/Service.ts
sangaman: Thanks. I'm trying to get `no-unused-variable` into our base style guide to catch things like this.

lib/service/Service.ts
sangaman: Ah yes this was just an oversight, good catch.

lib/service/Service.ts
sangaman: Yeah this makes sense, maybe rename this type `ServiceComponents`?

sangaman:
Thanks a lot for testing/reviewing @moshababo. The cli works perfectly for me on Windows, so I'm guessing it's likely an environment issue. I'll try to test this myself on linux and see if I can't reproduce and fix what you're seeing.

sangaman:
@moshababo I was able to execute gRPC calls through the cli on linux, so I'll need your help at some point to debug the issue you're experiencing. One thing I did find on my own is that the `npm run proto` call didn't work on linux, so I added `cross-os` scripts for that. I also addressed your latest review comments.

lib/service/Service.ts
moshababo: Maybe just define it as an anonymous type, since it's not really going to be used anywhere else.

moshababo:
Sorry for the false alarm - the problem was on my env, and not due to macOS in general. After investigating the problem I found that the `node-gyp` build had some silent errors, so the `grpc` dependency didn't compile well. After fixing the problem and reinstalling, it worked great. 

But funny enough - after testing it again with your recent commits, i'm getting some other issues on some of the CLI calls (on the server side).
from what I checked: `getpairs` works fine, but `getinfo`, `getorders` & `connect` throw exceptions. 
I'm guessing the you could see these problems on your env too. If you don't, let me know and i'll give you more details.

sangaman:
Thanks for catching these. `getinfo` has always been working for me, I'd suspect if you're getting issues it might be because you're running an older version of `lnd`.

I was able to fix `connect` and `getorders`, I just had a type mismatch on those., The response `connect`  gives right now is pretty basic but we can flesh it out later if needed.

A separate issue I encountered on my own in doing some more testing is that `placeOrder` returns a `StampedOrder`, and gRPC does not play nice with `Date` objects like the `createdAt` property. The object can't be serialized, and instead must be converted to an epoch number or ISO string. I considered creating separate types for grpc only that don't use `Date` and converting the `Date` properties when gRPC calls occur, but that seemed hacky. Instead I'm thinking it's probably best to just switch from using `Date` for any types/objects that may need to be used in rpc responses or requests to either `number` (using `getTime()`) or `string` (using `getISOString()`). `lnd` uses `int64` in their proto file for all datetimes so we could just go with that, or use string for better readability. What do you think?

We should be able to override the default SQL type for `createdAt`, an example is discussed here: https://stackoverflow.com/questions/29652538/sequelize-js-timestamp-not-datetime

I'll hold off on pushing my fixes until we agree on an approach for dates over gRPC.

moshababo:
I don't think it's reasonable to start changing the `Date` fields on types/objects which will be needed for gRPC responses. First because you want to have it as `Date` on JS, and second because there's no way to maintain that approach (in case you don't want to apply this for all types). Same goes for the database - i'm not sure about the internal implementation of `datetime` vs `timestamp`, but what's going on there should not concern us in regards to gRPC serialization problem.

The usual practice is that `Date` is serialized to epoch time (`number` \ `int64`) when leaving the JS runtime. If the gRPC Nodejs server doesn't take care of that well, I suggest that we'll do the serialization/deserialization ourselves, on the gRPC layer.

sangaman:
OK that makes sense. I think I came up with a decent/simple solution in the latest commit, so all dates will be treated as epoch time on the gRPC layer. Here's sample output.

```shell
node bin/xucli placeorder BTCLTC 1 1
{
  "matches": [],
  "remainingOrder": {
    "price": 1,
    "quantity": 1,
    "pairId": "BTCLTC",
    "peerId": 0,
    "id": "7cb35c30-6a99-11e8-bdd0-e7c4f0e14f22",
    "createdAt": "1528406808947"
  }
}
```

Thanks and let me know if you have any more feedback.

moshababo:
Looks great. We can introduce a more robust approach later, where the serialize/deserialize will be defined on the request/response objects themselves.

I wasn't sure how you want to merge this branch history, so i'll let you do it yourself. 

Please make sure to open issues for all the open topics you had in mind, so we won't forget anything.

sangaman:
Thanks @moshababo for all the feedback and @itsbalamurali for getting these changes started! I'll open all related issues later today, plus I have one or two quick things I'll open PRs for myself.

PR #104, opened 2018-06-03
Refactor enums to use typescript

sangaman:
Just some typescript housecleaning. I think this is one of the last remaining parts of the code that hadn't been migrated over to using typescript conventions.


moshababo:
I think it should be moved to `types` folder, no?

sangaman:
Yes that makes sense actually. What do you think about making `errorCodesPrefix` an enum as well and scrapping the `constants` folder for now? @moshababo

moshababo:
I'm not sure it should be made a enum. But we should change the way it's currently being used by the errors files (more OOP, now with TS).

sangaman:
OK, yeah I suppose an enum wouldn't be appropriate for how we're using `errorCodesPrefix`, so I'll just leave that out of this PR and we can take care of that separately. I moved `enums` to the `types` folder.

Issue #105, opened 2018-06-05
Market Orders


kilrau:

moshababo:
This can be implemented now. More info can be found at #109

kilrau:
Needs to wait for https://github.com/ExchangeUnion/Docs/issues/26 to be resolved.

PR #109, opened 2018-06-06
Market Order task - consolidation

moshababo:
A small PR for just defining the order book high-level API.

* making `addPeerOrder` a private method, because it won't be consumed directly by exchanged, but rather by the P2P layer. 
* defining two public methods, to be directly consumed by exchanged: `addFillOrder` & `addMarketOrder`. both are implicitly referring to an `OwnOrder`.

The idea of how to implement a Market Order: 
* Buy orders `price` will be set to `Number.MAX_VALUE`.
* Sell order `price` will set to 0.
* After matching, the remaining order won't become a maker order: it won't be added to the queues (for internal matching), and won't be broadcasted to peers. 

@kilrau 

#105

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -69,7 +69,7 @@ class OrderBook {
     return this.repository.getPeerOrders(pairId, maxResults);
   }
 
-  public addFillOrder = async (order: orders.OwnOrder): Promise<matchingEngine.MatchingResult>  => {
+  public addLimitOrder = async (order: orders.OwnOrder): Promise<matchingEngine.MatchingResult>  => {
     return this.addOwnOrder(order);
   }
 
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -69,7 +69,15 @@ class OrderBook {
     return this.repository.getPeerOrders(pairId, maxResults);
   }
 
-  public addOwnOrder = async (order: orders.OwnOrder): Promise<matchingEngine.MatchingResult>  => {
+  public addFillOrder = async (order: orders.OwnOrder): Promise<matchingEngine.MatchingResult>  => {
+    return this.addOwnOrder(order);
+  }
+
+  public addMarketOrder = async (_order: orders.OwnOrder) => {
+    // TODO: implement
+  }
+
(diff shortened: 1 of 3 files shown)


kilrau:
```
Buy orders price will be set to Number.MAX_VALUE.
Sell order price will set to 0.
```
Didn't we implement it as +-? A market order can be both, buy and sell obviously so I don't understand why the sell order price would always be 0. There should be just one order price and it should be 0. Please correct me if I'm wrong.

`After matching, the remaining order won't become a maker order: it won't be added to the queues (for internal matching), and won't be broadcasted to peers.`
Correct. There is an edge case where all order books (`ownOrder` as well as `peerOrders`) are empty. I think the default case should be the market order sitting in the queue waiting for any new incoming orders and just match.

And finally I worked through some more order book & API stuff and also discussed it with a DAE engineer. Please give me feedback @moshababo @sangaman @itsbalamurali @offerm : https://github.com/ExchangeUnion/Docs/issues/26

moshababo:
We implemented the +- for the quantity, not the price. 
Making the price set to `MAX_VALUE` or `0` is just a trick so the price matching condition (`buyOrder.price >= sellOrder.price`) will always be satisfied, without introducing a new primitive to the matching engine.

About the remaining order (same as unmatched order, just in partial matching context):
> Correct. There is an edge case where all order books (ownOrder as well as peerOrders) are empty. I think the default case should be the market order sitting in the queue waiting for any new incoming orders and just match.

So you're saying that you **do** want the market order to become a maker order for internal matching, but not a maker order in the network, for other peers?

sangaman:
I think it could get a bit messy if we have partially unfilled market orders enter the order queue instead of getting discarded. What price would they have in the queue? If it stays at zero, for instance, what happens when another market order comes in? I feel like it would create unexpected edge cases and possibly unwanted behavior.

kilrau:
>So you're saying that you do want the market order to become a maker order for internal matching, but not a maker order in the network, for other peers?

No. Thought about it again and agree with @sangaman . Let's go with discarding when there is nothing / amount not enough in the order book. @moshababo

kilrau:
>Making the price set to MAX_VALUE or 0 is just a trick so the price matching condition (buyOrder.price >= sellOrder.price) will always be satisfied, without introducing a new primitive to the matching engine.

Checked `getMatchingQuantity` -> makes sense. Good from my side.

PR #110, opened 2018-06-07
Set default 10 maxResults for cli getorders

sangaman:
`getOrders` calls from the command line without specifying `max_results` now default to 10 rows max. Specifying a `max_results` value of 0 will return an unlimited number of orders.


sangaman:
Yes you're right, those two methods should behave similarly. In fact I think `getPeerOrders` is the method that gets used by the `getOrders` rpc call. I refactored those two methods to use as much of the same code as possible, and also made sure all functions in the class are using arrow functions.

PR #117, opened 2018-06-08
Fixed travis build

michael1011:
This PR will Travis make install `rsync` before compiling the Docker image and cause the build to succeed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,6 @@ script:
   - cd docker/xud && docker build -t exchangeunion/xud .
 
 before_install:
+  - sudo apt-get install -y rsync
   - mysql -e 'CREATE USER 'xud'@'localhost'';
   - mysql -e 'GRANT ALL PRIVILEGES ON `xud\_%`.* TO `xud`@`%`;'


sangaman:
Thanks, I was wondering if this was a temporary issue with travis as the grpc2 branch was building fine even after I added the `rsync`command, but this is probably good to have to be sure.

michael1011:
The build still fails for some reason. What about using `cp` instead of `rsync` on Linux and Mac?

sangaman:
Hmm I'll look into it later, it looks like it might an issue with the docker build actually. I'm not sure if it's possible to do that same command with `cp`, in other words copying over the directory structure and only files matching the `.js` extension, but I'll look into that too.

michael1011:
I should have installed `rsync` in the Docker container and not on the Travis VM :man_facepalming: 

Adding `RUN apt update && apt install -y rsync` to the Dockerfile solves this issue

sangaman:
Good call, want to just open a separate PR reverting the line from this one and adding the one above to the dockerfile? Your PR's travis builds should happen now.

Issue #113, opened 2018-06-08
Make gRPC and web proxy listening host configurable


sangaman:
Right now, the gRPC service is hardcoded to listen only on `localhost`. We should make this configurable, leaving `localhost` as the default, to allow for rpc calls to potentially be made from a remote machine.

sangaman:
Resolved with #125.

PR #125, opened 2018-06-09
added gRPC host to config

michael1011:
As in #113. And I renamed the config section `rpc` to `grpc`.

### DIFF
--- a/lib/grpc/GrpcServer.ts
+++ b/lib/grpc/GrpcServer.ts
@@ -36,13 +36,14 @@ class GrpcServer {
    * Start the server and begin listening on the provided port
    * @param port
    */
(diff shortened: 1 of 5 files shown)


lib/grpc/GrpcServer.ts
```diff
@@ -36,13 +36,14 @@ class GrpcServer {
    * Start the server and begin listening on the provided port
    * @param port
    */
-  public listen = (port: number) => {
+  public listen = (host: string, port: number) => {
```
sangaman: Having `port` as the first parameter before `host` seems to be the convention both elsewhere in the code and in places like node's `net` module.

Issue #115, opened 2018-06-08
Refactor XUClient 


sangaman:
`XUClient.ts` currently has quite a bit of code duplication and use of `any` types. It should be refactored to provide better type safety and code reuse. It can use the generated types and methods in `xudrpc_pb.d.ts` and `xudrpc_pb.js`. Another possible approach would be to generate a static gRPC client, as described here: https://github.com/grpc/grpc/tree/master/examples/node/static_codegen.

itsbalamurali:
@sangaman I am closing this as the PR Fix has been already merged.

PR #129, opened 2018-06-11
Generate Xud client from proto

sangaman:
This PR aims to satisfy #115. It replaces the `XudClient.ts` file we were using for `xucli` calls with a client generated from our `xudrpc.proto` file, and refactors the cli commands to use it. It also adds a `.gitattributes` file to mark files and folders that were generated or copied from other projects, and implements the `shutdown` call that was missing from the initial gRPC migration. Finally I added a basic README for how to generate these protobuf files.


sangaman:
Good catch, it was a simple bug where I passed a second new `ConnectRequest` to the `connect` method rather than the one that had already been created and assigned `host` and `port` values.

PR #130, opened 2018-06-11
Pool silent conn error

### LINKED ISSUE #126: Establishing connection to offline peer results in error
After connecting to a peer, restarting `xud` while the peer is offline results in an error during the attempt to reestablish a connection during startup. We should handle and silently ignore peer connection errors (or maybe log with `verbose`) during startup, since it is expected that we won't necessarily be able to reconnect with all known peers.

```shell
(node:18872) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: connect ECONNREFUSED 127.0.0.1:9995
(node:18872) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
2018-6-9 11:52:10 [P2P] debug: Socket Error (localhost:9995): "Socket hangup"
2018-6-9 11:52:10 [P2P] error: peer error undefined
```

moshababo:
Fixes #126 

Plus:
* crashing the process on `unhandledRejection`. there's no reason to swallow them. 
* creating a separate `bootstrap` function, to be accessible from different entry points (tests, etc.).

### DIFF
--- a/lib/bootstrap.ts
+++ b/lib/bootstrap.ts
@@ -4,7 +4,7 @@ export default () => {
   /** Loads environment variables from the file .env */
   dotenv.config();
 
-  /** rethrow unhandled promise rejection, to crash process of an unwrapped instance */
+  /** Rethrow unhandled promise rejection, to crash process of an unwrapped instance */
   process.on('unhandledRejection', (err) => {
     console.log(err);
     throw err;
--- a/lib/bootstrap.ts
+++ b/lib/bootstrap.ts
@@ -5,8 +5,8 @@ export default () => {
   dotenv.config();
 
   /** rethrow unhandled promise rejection, to crash process of an unwrapped instance */
-  process.on('unhandledRejection', err => {
+  process.on('unhandledRejection', (err) => {
     console.log(err);
     throw err;
-  })
-}
+  });
+};
--- a/lib/bootstrap.ts
+++ b/lib/bootstrap.ts
@@ -1,0 +1,12 @@
+import dotenv from 'dotenv';
+
+export default () => {
+  /** Loads environment variables from the file .env */
+  dotenv.config();
+
+  /** rethrow unhandled promise rejection, to crash process of an unwrapped instance */
+  process.on('unhandledRejection', err => {
+    console.log(err);
+    throw err;
+  })
+}
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -61,7 +61,7 @@ class Peer extends EventEmitter {
     return peer;
   }
 
-  public getStatus(): string {
+  public getStatus = (): string => {
     if (this.connected) {
       return `Connected to peer (${this.id})`;
     } else {
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -43,14 +43,6 @@ class Peer extends EventEmitter {
     return this.socketAddress.toString();
   }
 
-  get statusString(): string {
-    if (this.connected) {
-      return `Connected to peer (${this.id})`;
-    } else {
-      return 'Not connected';
-    }
-  }
-
   constructor() {
     super();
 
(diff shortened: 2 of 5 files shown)


lib/p2p/Pool.ts
```diff
@@ -65,10 +65,18 @@ class Pool extends EventEmitter {
...
-    await this.connectPeer(peer);
+    await this.tryConnectPeer(peer);
     return peer;
   }
 
+  private tryConnectPeer = async (peer: Peer): Promise<void> => {
+    try {
+      return await this.connectPeer(peer);
+    } catch (err) {
+      this.logger.warn(`connectPeer failed: ${err}`);
```
michael1011: Maybe the error should be returned to make finding the reason why the connection failed easier.

lib/bootstrap.ts
```diff
@@ -0,0 +1,12 @@
+import dotenv from 'dotenv';
+
+export default () => {
+  /** Loads environment variables from the file .env */
+  dotenv.config();
+
+  /** rethrow unhandled promise rejection, to crash process of an unwrapped instance */
```
michael1011: Shouldn't the TypeDoc comments start with uppercase letters? Do we have some kind of code style guidelines for this?

lib/p2p/Pool.ts
moshababo: I don't think it should return the error, because this method is just about making `connectPeer` errors silent. We should just return what `connectPeer` returns. 
But you made me notice that it returns nothing, so i'll fix that.

`Peer` just throws an error on `ECONNREFUSED` or timeout. If the reason will be important for us, except for logging, we could use the thrown object.

lib/bootstrap.ts
moshababo: Fixed.

I don't think we have guidelines for this.

lib/p2p/Pool.ts
michael1011: I thought that if you try to connect to a peer with `xucli` and it fails you are pretty clueless about the reason why it failed. Especially if you don't know that the error is in the logs.

lib/bootstrap.ts
sangaman: I think we should just be following TypeDoc conventions. We do have some basic contribution guidelines, I can link to some TypeDoc guides from there.

lib/p2p/Peer.ts
```diff
@@ -69,6 +61,14 @@ class Peer extends EventEmitter {
     return peer;
   }
 
+  public getStatus(): string {
```
sangaman: Is there a reason why you switched this from a getter method? Mainly just curious. I do think we should make this an arrow function tho to match the rest of the class.

lib/p2p/Pool.ts
```diff
@@ -65,10 +65,18 @@ class Pool extends EventEmitter {
...
-    await this.connectPeer(peer);
+    await this.tryConnectPeer(peer);
     return peer;
   }
 
+  private tryConnectPeer = async (peer: Peer): Promise<void> => {
+    try {
+      await this.connectPeer(peer);
+    } catch (err) {
+      this.logger.warn(`connectPeer failed: ${err}`);
```
sangaman: I think `warn` might be too "urgent" of a logger level here. To me at least, a warning is cause for concern, whereas failing to connect on startup is is something that will happen with some regularity. I was thinking `verbose` might be the most appropriate, but maybe `info` is fine if we want to log failed connections on startup as a matter of routine.

moshababo: I think that because `connect` returns the connection status, there’s no need to be dependent on the logging there, and changing it to `info` should be fine.

lib/p2p/Peer.ts
moshababo: I switched it away from a getter syntax just as a style preference. I find this syntax more suitable for something like a pseudo property, or just a wrapper that you want it to look like a property. But `status` looked to me like something that might eventually need to retrieve/calculate, so I didn't find that syntax appropriate.

PR #131, opened 2018-06-11
Improved getInfo

### LINKED ISSUE #120: Improve getInfo API method
We have a `getInfo` method that is currently very basic - it simply calls the underlying lnd `getInfo` method and passes this along. We should expand this call to provide more meaningful data about the state of an `xud` instance. Some ideas:

- Number of peers connected
- Number of trading pairs/currencies supported
- Number of peer orders and own orders.
- Info about the Raiden node.

michael1011:
Closes #120 
I had to [edit `lib/proto/lndrpc_pb.d.ts` manually](https://github.com/ExchangeUnion/xud/compare/master...api_getInfo#diff-9b4392c5fa7ac711bb6dfa7e7fdf8252R892) to get `lndClient.listChannels()` working. I don't know why that didn't work in the first place. Even after updating `lndrpc.proto` to the latest version from the LND master branch.

### DIFF
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -1,4 +1,5 @@
 import fs from 'fs';
+import path from 'path';
 import Logger from './Logger';
 import Config from './Config';
 import DB from './db/DB';
@@ -77,6 +78,12 @@ class Xud {
     } catch (err) {
       this.logger.error(err);
     }
+  }
+
+  public static getVersion = (): string => {
+    const packageBuffer = fs.readFileSync(path.join(__dirname, '..', 'package.json'));
+    const packageJson = JSON.parse(packageBuffer.toString());
+    return packageJson.version;
   }
 
   /**
--- a/lib/p2p/PeerList.ts
+++ b/lib/p2p/PeerList.ts
@@ -4,7 +4,7 @@ import SocketAddress from './SocketAddress';
 class PeerList {
   private peers: { [ socketAddress: string ]: Peer } = {};
 
-  public add(peer: Peer): boolean {
+  public add = (peer: Peer): boolean => {
     if (this.has(peer.socketAddress)) {
       return false;
     } else {
@@ -13,17 +13,22 @@ class PeerList {
     }
   }
 
-  public remove(peer:Peer): void {
+  public remove = (peer:Peer): void => {
     delete this.peers[peer.socketAddress.toString()];
   }
 
-  public has(socketAddress: SocketAddress): boolean {
+  public has = (socketAddress: SocketAddress): boolean => {
     return !!this.peers[socketAddress.toString()];
   }
 
-  public forEach(cb: (peer: Peer) => void) {
+  public length = () => {
+    return Object.keys(this.peers).length;
+  }
+
+  public forEach = (cb: (peer: Peer) => void) => {
     Object.keys(this.peers).map(key => cb(this.peers[key]));
   }
+
 }
 
 export default PeerList;
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -19,8 +19,9 @@ type PoolConfig = {
 
 /** A pool of peers for handling all network activity */
 class Pool extends EventEmitter {
+  public  peers: PeerList = new PeerList();
+
   private hosts: HostList;
-  private peers: PeerList = new PeerList();
   private server: Server = net.createServer();
   private logger: Logger = Logger.p2p;
   private connected: boolean = false;
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -4,8 +4,8 @@ import Pool from '../p2p/Pool';
 import OrderBook from '../orderbook/OrderBook';
 import LndClient from '../lndclient/LndClient';
 import RaidenClient, { TokenSwapPayload } from '../raidenclient/RaidenClient';
-import { GetInfoResponse } from '../proto/lndrpc_pb';
 import { OwnOrder } from '../types/orders';
+import Xud from '../Xud';
 
 /**
  * The components required by the API service layer.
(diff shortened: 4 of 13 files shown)


lib/p2p/PeerList.ts
```diff
@@ -13,17 +13,22 @@ class PeerList {
...
     delete this.peers[peer.socketAddress.toString()];
   }
 
-  public has(socketAddress: SocketAddress): boolean {
+  public has = (socketAddress: SocketAddress): boolean => {
     return !!this.peers[socketAddress.toString()];
   }
 
-  public forEach(cb: (peer: Peer) => void) {
+  public length = () => {
```
sangaman: Might be nice to make this a getter function. That way we can use `peers.length` without the () which I think is more consistent.

EDIT: Actually, better yet see below about moving this logic to the `Pool` class.

lib/p2p/PeerList.ts
```diff
@@ -13,17 +13,22 @@ class PeerList {
...
+    return Object.keys(this.peers).length;
+  }
+
+  public forEach = (cb: (peer: Peer) => void) => {
     Object.keys(this.peers).map(key => cb(this.peers[key]));
   }
+
```
sangaman: Drop this blank line.

lib/p2p/Pool.ts
```diff
@@ -19,8 +19,9 @@ type PoolConfig = {
 
 /** A pool of peers for handling all network activity */
 class Pool extends EventEmitter {
+  public  peers: PeerList = new PeerList();
```
sangaman: It seems like the only thing we're using `peers` for outside this class is to get the number of keys aka length. In that case, I think it's better to leave this private and add a public method here to return this info. Also, if we're only concerned with the number of *connected* peers, which for getInfo purposes I think makes sense, we should count only the peers that are currently connected.

proto/xudrpc.proto
```diff
@@ -66,7 +66,38 @@ service XUDService {
 
 message GetInfoRequest {}
 message GetInfoResponse {
-  lnrpc.GetInfoResponse lnd = 1 [json_name = "lnd"];
+    int32 peers = 1 [json_name = "peers"];
+    int32 pairs = 2 [json_name = "pairs"];
```
sangaman: Let's go with `num_peers` and `num_pairs` or similar to clarify that these are counts.

lib/service/Service.ts
```diff
@@ -40,11 +40,75 @@ class Service {
...
+    info.peers = peers.length();
+    info.pairs = pairs.length;
+
+    let peerOrdersCount: number = 0;
+    let ownOrdersCount: number = 0;
+    for (const key in pairs) {
+      const pair = pairs[key];
+
+      const orders = await this.orderBook.getOrders(pair.id, 0);
+      const ownOrders = await this.orderBook.getOwnOrders(pair.id, 0);
```
sangaman: This seems like a computationally expensive loop. Maybe not that big of a deal once we have proper caching and performance optimizations - but I'm wondering if it would be better to explicitly track the number of all orders for informational purposes like this. Then we don't need any loop or queries here. Thoughts? @moshababo

lib/service/Service.ts
```diff
@@ -40,11 +40,75 @@ class Service {
...
+    info.orders = {
+      peer: peerOrdersCount,
+      own: ownOrdersCount,
+    };
+
+    try {
+      const lnd = await this.lndClient.getInfo();
+      const channels = await this.lndClient.listChannels();
+
+      info.lnd = {
```
sangaman: I think it would be useful to have the `chains` property here from the lnd getinfo response, identity_pubkey maybe as well but I'm more on the fence about that one.

Edit: nevermind on the identity_pubkey, we get that info from the `uris` and I don't think we need it twice.

lib/service/Service.ts
```diff
@@ -40,11 +40,75 @@ class Service {
...
+    }
+
+    info.orders = {
+      peer: peerOrdersCount,
+      own: ownOrdersCount,
+    };
+
+    try {
+      const lnd = await this.lndClient.getInfo();
+      const channels = await this.lndClient.listChannels();
```
sangaman: I think it's sufficient to just call `getInfo` and use the number of active and pending (but not inactive channels) from there. But if we do want to stick to two separate calls to lnd, we should make a promises array for these two calls and then do `await Promises.all()` so that the two calls run in parallel, you can see examples of this elsewhere in the code if you search for "promises.all".

lib/service/Service.ts
```diff
@@ -40,11 +40,75 @@ class Service {
...
+    } catch (err) {
+      this.logger.error(`LND error: ${err}`);
+      info.lnd = {
+        error: String(err),
+      };
+    }
+
+    try {
+      const address = await this.raidenClient.getAddress();
+      const channels = await this.raidenClient.getChannels();
```
sangaman: Same here for using `Promises.all()`.

lib/Xud.ts
```diff
@@ -79,6 +80,12 @@ class Xud {
     }
   }
 
+  public static getVersion = (): string => {
```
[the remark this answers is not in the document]
michael1011: Importing just the version isn't possible but the whole file worked fine.

lib/p2p/Pool.ts
michael1011: Isn't every peer in `peers` connected? I thought so because in `broadcastOrder` the order gets send to every peer in `peers` and sending a package to peers that aren't connected would make no sense.

michael1011:
`// Manually changed 'channelsList' to 'channels'`
The manual edit I made doesn't mater anymore if `this.lndClient.listChannels()` isn't called (I used it for the inactive channels which will be removed).

sangaman:
Gotcha, in that case you don't need to wait on anything else. I'm actually not sure it would have made a difference actually, I've noticed that grpc simply appends "List" to every field name that is `repeated`. Pretty annoying, but I did some digging and that's just how it is with no way to just keep field names as they're specified.

One last thing I thought about, you can remove the `import lndrpc.proto` from `xudrpc.proto` since it's not being used any longer.

michael1011:
What to do about this appending `List` issue? `chainsList` and `urisList` are `undefined`. And just `chains` or `uris` returns an error when compiling.

lib/service/Service.ts
```diff
@@ -34,17 +37,89 @@ class Service {
...
+      peerOrdersCount += orders.buyOrders.length + orders.sellOrders.length;
+      ownOrdersCount += ownOrders.buyOrders.length + ownOrders.sellOrders.length;
+    }
+
+    info.orders = {
+      peer: peerOrdersCount,
+      own: ownOrdersCount,
+    };
+
+    if (!this.config.lnd.disable) {
```
sangaman: I think it's better to do `this.lndClient.isDisabled()` here because it's possible that lnd was disabled for some other reason besides configuration - maybe it didn't load properly . Same for Raiden below.

michael1011: Are you sure about that? If `lnd` is enabled in the config you surely want that it is always connected to `xud`. I think `getInfo` should not only show some basic numbers but also the status of your node. And if the connection between `lnd` and `xud` gets interrupted for some reason there should be a simple way for the user to find out apart from the logs.

michael1011:
@sangaman Although I fixed the issue with the `List` suffix with a pretty hacky workaround I would prefer doing this properly when the `LndClient` is refactored. Even if you are not sure that it would make a difference it still worth waiting.

sangaman:
Sorry for the delay. The `chainsList` and `urisList` thing really baffled me, and I couldn't find any good way to get around it. I just opened #135, which I think will help. With those changes, when you call `LndClient.getInfo()` you'll get back a `GetInfoResponse` object, where you'll be able to call `getChainsList()` and `getUrisList()` which will return the data you want.

Another thing I noticed when testing this branch, it seems that the `resolveJsonModule` typescript option actually puts `package.json` in the `dist` folder which then throws off the whole directory structure by putting a `lib` folder inside `dist`. I also couldn't find an easy way around this besides setting `resolveJsonModule` to false, which then means you can't import package.json the way I suggested. Maybe see if you can figure something out there, and if you can't then just load read the package.json file similar to how you were doing before my requested change. The only thing I'd say is that it should be read only once and then the `version` value should be stored in memory, such that each `getInfo` call doesn't require the file to be read from again.

sangaman:
OK, I think we can pick back up with this PR if you get a chance @michael1011. I think instead of the `requireJsonModule` option which seems to mess up the compilation, we can just `require()` package.json. Then you can rebase on the latest changes to LndClient.ts, and you should be able to do read from the `chainsList` and `urisList` keys of the response from `LndClient.getInfo()`.

michael1011:
Thanks! I will take care of it asap.

michael1011:
I probably should have squashed the commits into one before pushing them but at least it is working fine now.

lib/p2p/Pool.ts
```diff
@@ -32,6 +31,10 @@ class Pool extends EventEmitter {
     this.bindServer(this.server);
   }
 
+  get peerCount(): number {
+    return Object.keys(this.peers).length;
```
sangaman: This returns a fixed value of `5`, in other words the number of members in `PeerList`. Since `peers` (on PeerList) is private, we'll need to add some public logic in that class to return the number of peers. I also think it's worthwhile adding in some check to make sure we only count connected peers.

tslint.json
```diff
@@ -21,7 +21,8 @@
   },
   "linterOptions": {
     "exclude": [
-        "lib/lndclient/lndrpc_pb.d.ts"
+        "lib/lndclient/lndrpc_pb.d.ts",
+        "package.json"
```
sangaman: I don't think we need to exclude package.json here any longer since we're not using the `resolveJsonModule` compiler option, right?

michael1011:
Thanks! I think I got it right this time :smile:

tslint.json
michael1011: Yes you are right.

PR #135, opened 2018-06-18
Refactor lndclient

sangaman:
This is similar to the changes in #129 except applied to the LND client. Now all LND calls use the generated request/response types, and the main purpose of LNDClient.ts is to establish the grpc connection and to wrap calls with promises.

I did the bare minimum to make the `getinfo` cli call working, as it is being reworked in #131.

### DIFF
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -43,6 +43,8 @@ class Service {
    */
   public getInfo = async () => {
     const lnd = await this.lndClient.getInfo();
+    delete lnd.chainsList;
+    delete lnd.urisList;
     return { lnd };
   }
 
--- a/README.md
+++ b/README.md
@@ -48,7 +48,6 @@ port = 8885
 
 [lnd]
 disable = false
-datadir = "~/.lnd"
 host = "localhost"
 
 [raiden]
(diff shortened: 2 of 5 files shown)


lib/service/Service.ts
```diff
@@ -42,7 +43,10 @@ class Service {
    * Placeholder for a method to return general information about an Exchange Union node.
    */
   public getInfo = async () => {
-    const lnd = await this.lndClient.getInfo();
+    const getInfoResponse = await this.lndClient.getInfo();
+    const lnd = getInfoResponse.toObject();
```
moshababo: Any reason for exposing gRPC semantics/types outside of `lndClient`? It should return plain objects, without needing to call `toObject()`.

sangaman: Yeah this is part of the "bare minimum" I was referring to above. @michael1011 ran into issues in #131 with using the `toObject` result because of some discrepancies due to how grpc annoyingly appends `List` to the name of array fields. That's also why I had to `delete` the lists objects in the following lines. In some quick tests, grabbing data from the `GetInfoResponse` object seems to work fine. I'd otherwise probably prefer also confining all grpc types to the lndClient class, but evidently it's not that simple.

moshababo: Didn't had a chance to play with it, but I think we better keep `lndClient` encapsulating the solutions for these issues, even if you'll need to define these per call. It shouldn't reach `Service`.

sangaman: OK, done. I removed all references to the grpc/lndrpc types outside of `LndClient.ts` and things still seem to work. The one exception is the `AddressType` enum, I couldn't think of a way to separate that short of creating an identical enum in our enums class.

PR #136, opened 2018-06-18
p2p request-response

moshababo:
Mostly adding  P2P request-response support. In addition:

* new packets: `Hello`, `Ping`, `Pong`
* sending `Ping` on intervals
* define response packet type and response timeout on each packet
* the ability to `wait` for a response, and easily consume it as a `Promise`

The mechanism might change after we'll add a packet header, and do the p2p protocol handshake.

The packets folder files and classes are not well structured, that will be taken care of later.

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -220,7 +220,7 @@ class Peer extends EventEmitter {
     assert(!this.destroyed);
     assert(!this.stallTimer);
 
-    this.stallTimer = setInterval(this.maybeTimeout, Peer.STALL_INTERVAL);
+    this.stallTimer = setInterval(this.checkTimeout, Peer.STALL_INTERVAL);
   }
 
   private initHello = async () => {
@@ -252,7 +252,7 @@ class Peer extends EventEmitter {
   /**
    * Potentially timeout peer if it hasn't responded.
    */
-  private maybeTimeout = () => {
+  private checkTimeout = () => {
     const now = ms();
 
     for (const [packetType, entry] of this.responseMap) {
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -267,7 +267,7 @@ class Peer extends EventEmitter {
   /**
    * Wait for a packet to be received from peer.
    */
-  private addReponseTimeout = (packetType: PacketType, timeout: number): ResponseEntry|null => {
+  private addReponseTimeout = (packetType: PacketType, timeout: number): ResponseEntry | null => {
     if (this.destroyed) {
       return null;
     }
@@ -292,7 +292,7 @@ class Peer extends EventEmitter {
   /**
    * Fulfill awaiting requests response.
    */
-  private fulfillReponse = (packetType: PacketType): ResponseEntry|null => {
+  private fulfillReponse = (packetType: PacketType): ResponseEntry | null => {
     const entry = this.responseMap.get(packetType);
 
     if (!entry) {
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -148,7 +148,7 @@ class Peer extends EventEmitter {
     this.sendRaw(packet.type, packet.toRaw());
 
     if (packet.responseType && packet.responseTimeout) {
-      this.addReponseTimeout(packet.responseType, packet.responseTimeout);
+      this.addResponseTimeout(packet.responseType, packet.responseTimeout);
     }
   }
 
@@ -267,7 +267,7 @@ class Peer extends EventEmitter {
   /**
    * Wait for a packet to be received from peer.
    */
-  private addReponseTimeout = (packetType: PacketType, timeout: number): ResponseEntry | null => {
+  private addResponseTimeout = (packetType: PacketType, timeout: number): ResponseEntry | null => {
     if (this.destroyed) {
       return null;
     }
@@ -292,7 +292,7 @@ class Peer extends EventEmitter {
   /**
    * Fulfill awaiting requests response.
    */
-  private fulfillReponse = (packetType: PacketType): ResponseEntry | null => {
+  private fulfillResponse = (packetType: PacketType): ResponseEntry | null => {
     const entry = this.responseMap.get(packetType);
 
     if (!entry) {
@@ -396,7 +396,7 @@ class Peer extends EventEmitter {
     }
 
     if (status) {
-      const entry = this.fulfillReponse(packet.type);
+      const entry = this.fulfillResponse(packet.type);
       if (entry) {
         entry.resolve(packet);
       }
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -51,7 +51,7 @@ class Peer extends EventEmitter {
   private stallTimer: NodeJS.Timer | null = null;
   private pingTimer: NodeJS.Timer | null = null;
   private lastPingId: string | null = null;
-  private responseMap: Map<PacketType, ResponseEntry> = new Map();
+  private responseMap: Map<PacketType, PendingResponseEntry> = new Map();
   private connectTime: number = 0;
   private banScore: number = 0;
   private lastRecv: number = 0;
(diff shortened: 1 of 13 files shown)


lib/p2p/Peer.ts
```diff
@@ -255,6 +422,102 @@ class Peer extends EventEmitter {
...
+  public reject = (err: any) => {
+    for (const job of this.jobs) {
+      job.reject(err);
+    }
+
+    this.jobs.length = 0;
+  }
+}
+
+class Job {
```
sangaman: How will this be used differently from a regular promise?

lib/p2p/Peer.ts
```diff
@@ -169,6 +218,98 @@ class Peer extends EventEmitter {
...
+        return;
+      }
+    }
+  }
+
+  /**
+   * Wait for a packet to be received from peer
+   */
+
+  private addReponseTimeout = (packetType: PacketType, timeout: number): ResponseEntry|null => {
```
sangaman: Typo - Change "Reponse" to "Response"

lib/p2p/Peer.ts
```diff
@@ -169,6 +218,98 @@ class Peer extends EventEmitter {
...
+    }
+
+    return entry;
+  }
+
+  /**
+   * Fulfill awaiting requests response
+   */
+
+  private fulfillReponse = (packetType: PacketType): ResponseEntry|null => {
```
sangaman: Typo - Change "Reponse" to "Response"

lib/p2p/Peer.ts
```diff
@@ -169,6 +218,98 @@ class Peer extends EventEmitter {
...
+      const entry = this.getOrAddResponseEntry(packetType);
+      entry.addJob(resolve, reject);
+    });
+  }
+
+  /**
+   * Potentially timeout peer if it hasn't responded
+   */
+
+  private maybeTimeout = () => {
```
sangaman: I think a method name like `checkTimeout` or `checkForTimeout` would be more descriptive.

lib/p2p/Peer.ts
```diff
@@ -255,6 +422,102 @@ class Peer extends EventEmitter {
...
+      this.logger.debug(`Peer sent an invalid pong id (${this.id})`);
+      return false;
+    }
+
+    this.lastPingId = null;
+    return true;
+  }
+}
+
+class ResponseEntry {
```
sangaman: The name for this class threw me off a bit. As I understand it this class represents a wait for a response from a peer. In that case I think a name like `PendingResponse` might be better.

lib/p2p/packets/HelloPacket.ts
```diff
@@ -0,0 +1,30 @@
...
+
+type HelloPacketBody = {
+  version: string;
+  nodeKey: string;
+  listenPort: number;
+  pairs: string[];
+};
+
+class HelloPacket extends Packet {
+  public type: PacketType = PacketType.HELLO;
```
sangaman: For consistency I think it looks better to make the type inferred here, aka drop the `: PacketType` to match with `responseType` below.

lib/p2p/packets/HelloPacket.ts
```diff
@@ -0,0 +1,30 @@
...
+  public responseType = PacketType.HELLO;
+  public responseTimeout = 10000;
```
sangaman: All these properties seem like they should be static, since they shouldn't be subject to change and they are inherent to the class. Same goes for other classes that extend `Packet`. I think the best way to do this would be to make these getter functions that return a fixed packet type or number. The getter methods can be defined on the base `Packet` class and then overridden by the child classes.

lib/p2p/packets/PingPacket.ts
```diff
@@ -0,0 +1,27 @@
+import Packet from './Packet';
+import PacketType from './PacketType';
+
+type PingPacketBody = {
+  id: string;
+};
+
+class PingPacker extends Packet {
```
sangaman: Typo, change `PingPacker` to `PingPacket`.

lib/p2p/packets/PongPacket.ts
```diff
@@ -0,0 +1,27 @@
+import Packet from './Packet';
+import PacketType from './PacketType';
+
+type PongPacketBody = {
+  id: string;
+};
+
+class PingPacker extends Packet {
```
sangaman: Typo: I'm guessing this should be `PongPacket` not `PingPacker`.

lib/p2p/Peer.ts
```diff
@@ -255,6 +422,102 @@ class Peer extends EventEmitter {
 
     this.emit('error', { msg, err });
   }
+
+  private sendHello = (): HelloPacket => {
+    // TODO: use real values
+    const packet = new HelloPacket({
+      version: '123',
```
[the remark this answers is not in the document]
moshababo: Sure, will do (in a later PR).

lib/p2p/Peer.ts
```diff
@@ -94,6 +119,26 @@ class Peer extends EventEmitter {
...
+    }
+
+    if (this.stallTimer) {
+      clearInterval(this.stallTimer);
+      delete this.stallTimer;
+    }
+
+    if (this.connectTimeout) {
+      clearTimeout(this.connectTimeout);
+      delete this.connectTimeout;
```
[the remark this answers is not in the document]
moshababo: `clearTimeout` will remove the timeout from the scheduler, but the id will remain on the variable. I think it's good to clean it, to indicate that there's no active scheduler. otherwise you cannot know that.
I'm actually gonna change this to `null` instead of `undefined`. I think it's more appropriate when to trying represent "empty" state.

lib/p2p/Peer.ts
moshababo: It's actually equivalent to a promise executor, not to a promise (which you can call `then` on). It's being used just to resolve promises, not to define/hold promises. We might need to add that later, but I didn't find a use-case for that yet.

lib/p2p/Peer.ts
moshababo: That's what happens when you'r rushing to open a PR ;)

lib/p2p/Peer.ts
moshababo: Fixed.

lib/p2p/Peer.ts
moshababo: Yes, it wasn't a brilliant name.
But i'm not sure about `PendingReponse`, because it doesn't represent a response. 

I changed it to `PendingResponseEntry` for now. Let me know if you have a better name.

lib/p2p/packets/HelloPacket.ts
moshababo: Because we've set on explicit properties typing, I added types instead of dropping then (temp solution, this area will probably be changed in the next PR).

lib/p2p/packets/HelloPacket.ts
moshababo: You're right. That's the bad class structure I was talking about. I originally did it this way to avoid static properties inheritance issues. I'll do some experiments with it, in addition to what you suggested, and fix this design in the next PR.

lib/p2p/Peer.ts
```diff
@@ -255,6 +416,102 @@ class Peer extends EventEmitter {
...
+  }
+
+  private handlePing = (packet: PingPacket): void  => {
+    const { id } = packet.body;
+
+    if (!id) {
+      return;
+    }
+
+    this.sendPong(id + 1);
```
michael1011: Why `id + 1` and not just `id`? `id` shouldn't be altered as it is compared to the original value (`this.lastPingId`) or am I missing something?

moshababo: Oops, I did it to make the pong fail, when I was testing it, and forgot to remove it.
Thanks for catching this ;)

lib/p2p/Peer.ts
sangaman: Ok, what do you think about calling this `Resolver` or something along those lines?

lib/p2p/Peer.ts
sangaman: Is my understanding above correct tho? I am also struggling to come up with a succint name, feels like English is missing a word. Some other ideas, none of which are perfect, `ResponseTimer`, `WaitForResponse` (wait in the noun sense, `AwaitingResponse`.

sangaman:
Looks good, I just made 2 more comments about some potential renaming of the `Job` and `PendingResponseEntry` classes. You can change if you think it makes sense, either way it should be ready to merge.

Issue #123, opened 2018-06-08
Add streaming gRPC call to listen to orderbook events


sangaman:
This issue is about notifying clients when order book events occur such as an `ownOrder` being filled.

Events:
1. ownOrder added success/error (placeOrder)
2. ownOrder removed success/error (cancelOrder)
3. ownOrder removed (due to internal match)
4. ownOrder removed (due to a successful swap that was instantiated remotely)
5. peerOrder added
6. peerOrder removed (due to non-internal match and successful swap)
7. peerOrder removed (due to peer order invalidation)
8. peerOrder removed (due to swap failure)

`placeOrder` response covers `(1)`, `(3)`, and `(6)` (handled by https://github.com/ExchangeUnion/xud/issues/479)
`placeOrderSync` sync response covers `(4)`, `(8)` (handled by https://github.com/ExchangeUnion/xud/issues/479)
`cancelOrder` response covers `(2)`
`subscribeOrders` covers `(4)`, `(5)`, `(7)`, and `(8)`

`placeOrder`, is asynchronous and just returns an immediate success/error of *placing* an order. Swap results come in via `subscribeOrders`. There is a second `placeOrderSync` call which returns fail/success of a swap where we inform the client synchronously if the swap actually was successful or failed.

Note:
We already have `(1)` , `(2)`, and `(3)` taken care of.

moshababo:
We should clearly define the API after finalizing our descussions [here](https://github.com/ExchangeUnion/Docs/issues/26).

In the meanwhile we should create a gRPC streaming call example. I'll open a separate issue for that.

itsbalamurali:
This also requires event emitter implementation on `OrderBook` and `MatchingEngine` beforehand and also predefining the events and their types.

sangaman:
@moshababo I'd say we can make #128 be a very basic version of the orderbook subscription that this issue is for. Maybe start with a `SubscribeOrders` call that triggers an event whenever a `peerOrder` is received. I imagine we'll definitely want something like that and we can always tweak it later if necessary, plus it would be laying the groundwork for this issue. I'd be interested to get started on that soon.

sangaman:
Although @itsbalamurali is right that we need to implement event emitters on the orderbook which has not been done yet, and perhaps that should be done separately from the streaming call example. Another possibility is a `SubscribePeers` call which triggers events when a new peer connects, since we already have some events set up on the p2p side of things.

moshababo:
[...]

PR #144, opened 2018-06-26
gRPC streaming call example

### LINKED ISSUE #128: gRPC streaming call example
Add a gRPC streaming call example, including the listener side: `XUClient` + the CLI layer. It should ideally hang the process and log the streaming to terminal.

michael1011:
Closes #128 

I added the gRPC call `StreamingExample` which will return the current time every second and requires only one call of the client. In gRPC this is called [server-side streaming](https://grpc.io/docs/guides/concepts.html#server-streaming-rpc). It is a dummy call because in #123 @moshababo said that it should be just a POC.

### DIFF
--- a/lib/cli/commands/streamingExample.ts
+++ b/lib/cli/commands/streamingExample.ts
@@ -1,0 +1,15 @@
+import { loadXudClient } from '../command';
+import { Arguments } from 'yargs';
+import { StreamingExampleRequest, StreamingExampleResponse } from '../../proto/xudrpc_pb';
+
+export const command = 'stream';
+
+export const describe = 'example for server-side streaming';
+
+export const handler = (argv: Arguments) => {
+  const request = new StreamingExampleRequest();
+  const call = loadXudClient(argv).streamingExample(request);
+  call.on('data', (message: StreamingExampleResponse) => {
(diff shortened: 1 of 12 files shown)


michael1011:
Yeah you are right all of us should use the same version of `protoc` (btw I was using `3.0.0` before).

What about adding the version `3.4.0` to the [README in the proto folder](https://github.com/ExchangeUnion/xud/blob/master/proto/README.md)? And should we update to the latest version `3.6.0`?

sangaman:
I do think it's a good idea to specify a protoc version in the README. As for latest version vs current version, `lnd` is using 3.4.0 (which is why I had it installed) so that's one reason to stick with that. But that's not to say we couldn't start using the latest, let me know what you think is better. I do think once we settle on a version though we should stick with that until there's an impactful change with protoc.

michael1011:
Lets settle on 3.4.0. 

Does Go need to be installed to regenerate the Javascript protobuf? I tried without and it worked just fine.

sangaman:
Evidently it works fine without go, the changes now all look good and it runs as before. I'll add the version to the readme separately. We can build off this sample call when we implement our first real streaming grpc call. Thanks!

Issue #122, opened 2018-06-08
Add disconnect API method


sangaman:
Right now we can issue `xud` a command to connect to a peer, but we can't manually disconnect yet. We should add a `disconnect` API method via gRPC with `xucli` integration to allow for disconnecting from specific peers.

itsbalamurali:
Closing as implemented and merged #205

Issue #121, opened 2018-06-08
Add removeOrder API method


sangaman:
We should add methods to allow for removing of orders from the XUD orderbook. This is necessary in the case of a user canceling their order or an order being filled internally. This should have a call through the gRPC layer as well as `xucli` integration.

moshababo:
It can be implemented on the `OrderBook` class as `removeOwnOrder` & `removePeerOrder`. The later will be invoked when a peer invalidates his order. 

The order should be removed both from the priority queue and from the database. We currently maintain them separately, so it's pretty messy. 
In addition it also need to be invalidated over the P2P network, so a new packet need to be added: `ORDER_INVALIDATION`.

kilrau:
Does it make sense to add `removePeerOrder` to the gRPC layer since `XChange` will never call this ? @moshababo 

> In addition it also need to trigger invalidation over the P2P network, so a new packet need to be added: ORDER_INVALIDATION.

Yes!

moshababo:
> Does it make sense to add removePeerOrder to the gRPC layer since XChange will never call this ? @moshababo

I don't think so. The Own/Peer semantics shouldn't even be used on these calls. 
Each `xud` client can cancel his own orders  (by invoking `cancelOrder`). 
As for other peers orders - if he don't want them to enter his node, he can block the peer. But there's no reason to actively remove a peer order - that's something that only the P2P layer should do, when `cancelOrder` was invoked by a remote node.

sangaman:
From an API perspective, I prefer the `removeOrder` naming because it is generic. The exchange should call it both when an order is cancelled and when an order is filled internally. Correct me if I'm wrong but from xud's perspective I believe the two are equivalent.

kilrau:
agree with Daniel's `removeOrder` naming, because more generic @itsbalamurali @moshababo

moshababo:
I discussed this with @sangaman, and we agreed that `cancelOrder` is more appropriate.
The reason: `removeOrder` makes sense only in the context of a local order book API (and that's how it should be implemented there), but not for the high-level API, which also need to trigger invoices cancellation, and invalidation over the P2P network.

In regards to what @sangaman wrote above - it doesn't really matter if the exchange is cancelling it 
due to internal cancellation or filling. From `xud` perspective, it's all the same.

kilrau:
kk, let's keep it final from now on: `cancelOrder`

itsbalamurali:
Fixed in #206

Issue #141, opened 2018-06-24
Don't store orders on disk, order bootstrap mechanism


kilrau:
- E.g. public getOrder(timestamp) to expose to other xuds, so on bootup xud can query other xuds for their orders
- default behavior after p2p handshake: call `getOrders`(timestamp)

kilrau:
Related: https://github.com/ExchangeUnion/xud/issues/149

PR #167, opened 2018-06-27
Disconnect Peers & Order cancellation

itsbalamurali:
This PR solves this following issues:

- #121  add removeOrder method implemented as cancel order from Exchange/gRPC API perspective.
- #122  add disconnect method to disconnect from already connected `xud`peers/nodes.


lib/p2p/Pool.ts
```diff
@@ -68,6 +68,20 @@ class Pool extends EventEmitter {
...
+      this.peers.remove(peer);
+      peer.destroy();
+      this.hosts.removeHost(socketAddress);
+    } else {
+      const err = errors.NOT_CONNECTED(address.toString());
+      this.logger.info(err.message);
+      throw err;
+    }
+  }
+
```
moshababo: * if the peer exists, you need to destroy him and to make sure he’s removed from the peers list (you might even just want to delegate the destroying responsibility to the peers list). At the moment you’re checking if he exists, and if so, you’re creating a new peer, and then remove the old peer according to the new peer `socketAddress`. Even if it can somehow work, it need to be fixed.
* you don’t need to remove the host, which is saved for future connection. it might be a good idea to have an option for it, but it shouldn’t be the only supported behavior.
* the disconnection implementation shouldn't be under `removeOutbound`. Although when initiating a new connection it will always be an outbound, when disconnecting this differentiation is not necessary. You'll want to support disconnection also if it's an inbound connection, and anyway there shouldn't be any difference in the implementation of the two.

lib/orderbook/OrderBook.ts
```diff
@@ -77,6 +79,14 @@ class OrderBook {
     // TODO: implement
   }
 
+  public removeOwnOrder = async (id: string): Promise<void> => {
+    return this.repository.removeOrder(id);
```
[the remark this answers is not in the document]
itsbalamurali: so we don't delete from the database regardless, right?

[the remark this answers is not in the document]
moshababo: there's no need for that, because we're not going to save them there anymore (#141)

lib/orderbook/MatchingEngine.ts
```diff
@@ -173,6 +174,20 @@ class MatchingEngine {
 
     return matchingResult;
   }
+
+  public removeOwnOrder = (id: string): void => {
```
moshababo: use `orderId` instead of `id`

lib/orderbook/MatchingEngine.ts
```diff
@@ -173,6 +174,20 @@ class MatchingEngine {
...
+    this.priorityQueues.ownBuyOrders.remove(id, this.removeQueuedComparator);
+    this.priorityQueues.ownSellOrders.remove(id, this.removeQueuedComparator);
+  }
+
+  private removeQueuedComparator = (queuedOrder:any, orderID:string): Boolean => {
```
moshababo: can make static, and also use `orderId` instead of `orderID` (just because that's our usual way)

lib/orderbook/OrderBook.ts
```diff
@@ -84,6 +84,14 @@ class OrderBook {
...
+  }
+
+  public removePeerOrder = async (id: string, hostId: number): Promise<void> => {
```
moshababo: call remove on the matching engine.

lib/orderbook/OrderBookRepository.ts
```diff
@@ -106,6 +106,22 @@ class OrderbookRepository {
   public addPairs = (pairs: db.PairFactory[]): Bluebird<db.PairInstance[]> => {
     return this.models.Pair.bulkCreate(<db.PairAttributes[]>pairs);
   }
+
+  public removeOrder = async (orderId: string, hostId?: number): Promise<void> => {
```
moshababo: this can be removed

moshababo:
@itsbalamurali I wasn't sure if you're ready for another review on the "disconnect peer" part. Let me know if you do.

moshababo:
The methods handlers (CancelOrder, Disconnect) were not provided.

moshababo:
The methods handlers (CancelOrder, Disconnect) were not provided, so I couldn't test this PR.

moshababo:
Closing this because it's blocking other stuff. gRPC/CLI layer will be implemented by #193. The actual Disconnect and CancelOrder implementation should be done separately.

itsbalamurali:
@moshababo i am reopening this since we're done merging #193

moshababo:
Please open separate PR's for Disconnect/CancelOrder implementations.

PR #188, opened 2018-07-04
Market orders

### LINKED ISSUE #105: Market Orders

michael1011:
Closes #105 
I also added market orders to the gRPC interface and the cli.

Although I installed the correct version (3.4.0) of `protoc` there are many unnecessary changes to protobuffs. The only difference was that I compiled it on my laptop this time. @sangaman

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -63,10 +63,17 @@ class OrderBook {
   }
 
   /**
-   * Returns lists of buy and sell orders sorted by price.
+   * Returns lists of buy and sell orders of peers sorted by price.
    */
-  public getOrders(pairId: string, maxResults: number): Promise<Orders> {
+  public getPeerOrders(pairId: string, maxResults: number): Promise<Orders> {
     return this.repository.getPeerOrders(pairId, maxResults);
+  }
+
+  /*
+  * Returns lists of the node's own buy and sell orders sorted by price.
+  */
+  public getOwnOrders(pairId: string, maxResults: number): Promise<Orders> {
+    return this.repository.getOwnOrders(pairId, maxResults);
   }
 
   public addLimitOrder = async (order: orders.OwnOrder): Promise<matchingEngine.MatchingResult>  => {
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -5,6 +5,9 @@ import OrderBook from '../orderbook/OrderBook';
 import LndClient from '../lndclient/LndClient';
 import RaidenClient, { TokenSwapPayload } from '../raidenclient/RaidenClient';
 import { OwnOrder } from '../types/orders';
+import Config from '../Config';
+
+const packageJson = require('../../package.json');
 
 /**
  * The components required by the API service layer.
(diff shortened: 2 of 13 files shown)


lib/service/Service.ts
```diff
@@ -141,14 +141,22 @@ class Service {
   }
 
   /**
-   * Add an order to the order book.
+   * Add a limit order to the order book.
    */
-  public placeOrder = async (order: OwnOrder) => {
```
moshababo: I think we could keep the 2 calls (limit/market) unified on `gRPC`/`Service`, and let them split only when `OrderBook` API is being called. same as you did on CLI, based on if price is specified or not.

@sangaman @kilrau

lib/orderbook/OrderBook.ts
```diff
@@ -80,18 +80,24 @@ class OrderBook {
     return this.addOwnOrder(order);
   }
 
-  public addMarketOrder = async (_order: orders.OwnOrder) => {
-    // TODO: implement
+  public addMarketOrder = async (_order: orders.MarketOrder): Promise<matchingEngine.OrderMatch[]> => {
```
moshababo: the `_` in `_order` can be removed. It was used only when the argument wasn't in use, to relax the compiler.

lib/orderbook/OrderBook.ts
```diff
@@ -80,18 +80,24 @@ class OrderBook {
...
+    const price = _order.quantity > 0 ? Number.MAX_VALUE : 0;
+    const order: orders.OwnOrder = { ..._order, price };
+
+    const result = await this.addOwnOrder(order, true);
+    delete result.remainingOrder;
```
moshababo: The remaining order is being discarded (not added to queues, not broadcasted), but I think we should return it to the client anyway.

sangaman:
@michael1011 What differences are there between your laptop and the other machine you used to run the proto script? Diifferent OS? Just wondering so we can try to pin down what's causing the files to be generated differently.

michael1011:
@sangaman I have no idea. Everything is exactly the same as far as I know. Both times I installed everything as you described it in the [README](https://github.com/ExchangeUnion/xud/blob/master/proto/README.md).

michael1011:
I unified the two calls into one but the cli doesn't work because of inconsistency in the database:

`Error: 13 INTERNAL: Illegal value for createdAtvalue element of type int64: object (not convertible to Long)`

If a order is loaded from the database the data is stored like this `2018-07-05 18:05:43` and not in the unix time format. And the `hostId` in the database is `null` but gRPC does not support a `int32` to be `null`.

Should I take of this issues in this PR?

moshababo:
I think we shouldn't spend time on fixing this, since we want to stop storing orders in database anyway. The `hostId` issue seems to throw an exception, so I would suggest to implement the removal of the orders persistence first: https://github.com/ExchangeUnion/xud/issues/141
It's on @sangaman now, but I don't think he started working on it. So you can grab it instead, and we'll merge this PR after it will be done.

michael1011:
I messed up the branch really bad by rebasing to master and not `origin/master` and force pushing changes which resulted in getting this branch deleted. I will open another PR about market orders ASAP. Sorry guys.

moshababo:
No worries.

sangaman:
I've done that myself, it's a good lesson in git.

Issue #180, opened 2018-06-28
Reorganize packets data structures & files


moshababo:
Do some cleaning and optimizations to the `lib/p2p/packets` folder, before implementing more packets. 

@sangaman please wait with your p2p tasks before i'll do that (will probably get to it only next week). if i'm blocking you, let me know.

PR #189, opened 2018-07-05
packets & messaging

moshababo:
Solves #180 

also added `hash` field to the packet header, which is the checksum of the packet body (for later data integrity checks), and also being used for the request/response mechanism, wheres each response packet is adding `reqHash` to the header.


lib/p2p/Peer.ts
```diff
@@ -292,16 +301,25 @@ class Peer extends EventEmitter {
   /**
    * Fulfill awaiting requests response.
    */
-  private fulfillResponse = (packetType: PacketType): PendingResponseEntry | null => {
-    const entry = this.responseMap.get(packetType);
+  private fulfillResponse = (key: string, packet: Packet): boolean => {
+    if (!key) {
```
sangaman: Are you only checking for empty strings here or also for null values? I think if we have a check like this, we can let `key` be `string | null` and not put the `!` operator when passing parameters to this function.

lib/p2p/Peer.ts
```diff
@@ -242,10 +247,14 @@ class Peer extends EventEmitter {
   /**
    * Wait for a packet to be received from peer. Executed on timeout or once packet is received.
    */
-  private wait = (packetType: PacketType) => {
+  private wait = (key: string, timeout?: number) => {
```
sangaman: I'm thinking perhaps `key` go back to `packetType` here. If the `wait` method is always being used to wait for a specific type of packet, as opposed to a response for a packet we've sent, then I think that makes more sense. Also, I'm wondering will we have a use case for `wait` besides the Hello packet? If so, I think I'd prefer explicitly defining `responseMap` as a map between hashes and PendingResponseEntries, and having a separate flag or something to indicate whether we are waiting for a Hello packet.

lib/p2p/Peer.ts
```diff
@@ -267,23 +276,23 @@ class Peer extends EventEmitter {
   /**
    * Wait for a packet to be received from peer.
    */
-  private addResponseTimeout = (packetType: PacketType, timeout: number): PendingResponseEntry | null => {
+  private addResponseTimeout = (key: string, timeout: number): PendingResponseEntry | null => {
```
sangaman: If the key passed to this method is always expected to be a hash, I think I'd prefer naming it `hash` within this method rather than `key`.

lib/p2p/packets/Packet.ts
```diff
@@ -1,11 +1,36 @@
...
   public abstract body: any;
-  public abstract responseType?: PacketType;
-  public abstract responseTimeout?: number;
-  public abstract toRaw(): string;
+  public abstract messageType: MessageType;
+
+  constructor (body: any, header?: PacketHeader) {
+    this.header = header || {
+      hash: SHA256(body).toString(CryptoJS.enc.Hex),
+      ...header!,
```
sangaman: What is the purpose of the `...header!` line if header is null?

lib/p2p/packets/Packet.ts
```diff
@@ -1,11 +1,36 @@
 import PacketType from './PacketType';
+import CryptoJS from 'crypto-js';
+import SHA256 from 'crypto-js/sha256';
+
+type PacketHeader = {
+  hash?: string;
```
sangaman: Should we call this `resHash` or something? These fields are a bit confusing to me, especially when it's not clear whether a packet is outgoing/ingoing, so making this as clear as possible how they are being used (maybe with some tsdoc) would be helpful.

sangaman: On a separate note, should this really be nullable? I figure we'd always want to and be able to determine the hash of a packet.

lib/p2p/packets/types/HelloPacket.ts
```diff
@@ -0,0 +1,25 @@
...
+type HelloPacketBody = {
+  version: string;
+  nodeKey: string;
+  listenPort: number;
+  pairs: string[];
+};
+
+class HelloPacket extends Packet {
+  public type: PacketType = PacketType.HELLO;
+  public messageType: MessageType = MessageType.UNILATERAL;
```
sangaman: Why the change from a packet with a response to a unilateral one?

lib/p2p/packets/Packet.ts
```diff
@@ -1,11 +1,36 @@
...
 abstract class Packet {
+  public header: PacketHeader;
   public abstract type: PacketType;
   public abstract body: any;
-  public abstract responseType?: PacketType;
-  public abstract responseTimeout?: number;
-  public abstract toRaw(): string;
+  public abstract messageType: MessageType;
+
+  constructor (body: any, header?: PacketHeader) {
```
sangaman: Looks like the `body` parameter isn't being used. We should do `this.body = body` in this constructor right?

lib/p2p/packets/packetUtils.ts
```diff
@@ -1,20 +1,17 @@
 import PacketType from './PacketType';
 import Packet from './Packet';
-import OrderPacket from './OrderPacket';
-import HelloPacket from './HelloPacket';
-import PingPacket from './PingPacket';
-import PongPacket from './PongPacket';
+import * as packetTypes from './types';
 
-const fromRaw = (type:string, body:string): Packet|void => {
+const fromRaw = (type:string, packet:string): Packet|void => {
```
sangaman: There should be spaces between the colon and the type, like `type: string`, for consistency.

lib/p2p/packets/types/HelloPacket.ts
moshababo: The previous design I did was buggy.  If the node will receive `Hello` before it sent an `Hello` , and the generic req/res mechanism will be applied, then a new `Hello` will never be received. This scenario can be easily verified by playing with the intervals (it didn't happen just because we're still testing multiple nodes on one machine only). And because anyway `Hello` should be sent voluntarily, without any prior request, I wasn’t a good fit for this pattern anyway.

lib/p2p/packets/Packet.ts
moshababo: It's being done by the sub classes (implicitly, by defining `public body` in the constructor). 
The only reason for having the base class receive the body, is to calculate the hash from it, and to avoid doing so in all the sub classes. 
If you have a better design for this, feel free to fix.

lib/p2p/Peer.ts
moshababo: Good idea. Fixed, using `undefined` instead.

lib/p2p/Peer.ts
moshababo: Fixed.

lib/p2p/packets/packetUtils.ts
moshababo: Fixed.

lib/p2p/packets/Packet.ts
moshababo: Good catch - it was a bug, after some refactoring which I did. 

I changed it to: `hash: (header && header.hash) || SHA256...`
The reason for this clumsy statement is to support these functionalities:
* it's an incoming packet deserialization, so the full header is provided
* it's a new packet being constructed. partial header might be given (`reqHash` field), or not

I can try to split the two modes somehow, if the code is too confusing.

moshababo:
Thanks, good review, I applied some changes.

lib/p2p/Peer.ts
sangaman: OK, then I think it's good for now but maybe let's revisit this in the future once the p2p layer is fleshed out.

lib/p2p/packets/types/HelloPacket.ts
sangaman: Cool, just so I'm clear - upon connection both nodes will send Hellos and once both Hellos have been received the connection can start being used.

lib/p2p/packets/Packet.ts
sangaman: Ok so `hash` is the hash of the body of the packet, and reqHash is the hash of the packet which we are responding to, or null if we're not responding to a packet. That's fine, I just think comments would be good.

lib/p2p/packets/types/HelloPacket.ts
moshababo: Yep. But we might add more async steps to the handshake later.

lib/p2p/packets/Packet.ts
sangaman: I think it might make sense to make `hash` non-nullable. We could have two constructors for `Packet`, one that accepts a header which we use when deserializing a packet that we've received. The other only takes the body and an optional `reqHash` parameter, and calculates the hash of the new packet so we can set it's value right when we create the packet. Let me know what you think.

lib/p2p/packets/Packet.ts
```diff
@@ -1,11 +1,36 @@
...
+    this.header = {
+      hash: (header && header.hash) || SHA256(JSON.stringify(body)).toString(CryptoJS.enc.Hex),
```
sangaman: Another minor thing, maybe we should use Base64 encoding instead of hex? Fewer characters, and therefore bytes, that need to be transferred and stored in memory that way. Human readability is not a real concern for packet headers as far as I can tell.

lib/p2p/packets/Packet.ts
sangaman: Ah I see now. I think I prefer having it done in the base constructor, that way there's no possibility that a class inheriting `Packet` can forget to put the `public` operator in the constructor and fail to set the body.

lib/p2p/packets/Packet.ts
moshababo: Makes sense to try to separate, specially since TS supports constructor overloading (based on number of arguments). Will do it tomorrow.

lib/p2p/packets/Packet.ts
sangaman: OK right, then it does make sense to declare it on each sub class as you're currently doing. I think then we can just drop `body` as a parameter from the base class constructor, right?

lib/p2p/packets/Packet.ts
moshababo: Fixed. We can also use a different hash function.

sangaman: True... from some quick research it looks like MD5 is considerably faster, and since we'll be calculating these hashes very frequently it's probably worth switching. It has some known collision vulnerabilities but since a collision here would not be catastrophic, I think it's good enough for our purposes.

lib/p2p/packets/Packet.ts
moshababo: No because it's still being used for calculating the header's `hash`.

lib/p2p/packets/Packet.ts
moshababo: Yes, that what I was thinking, just started with SHA256 as a default.

moshababo:
TODO: 
* switch SHA256 to MD5
* better separation for when constructing packets from a serialized format or from scratch

kilrau:
> switch SHA256 to MD5
Why? @moshababo

moshababo:
https://github.com/ExchangeUnion/xud/pull/189#discussion-diff-201452411R24

moshababo:
@sangaman 
1) I changed the hash function
2) I refactored the packets classes once again, to solve the problems we discussed
* not using static methods as factories anymore (easy inheritance)
* `fromRaw`/`toRaw` are managed solely by the base class (for our current JSON-based serialization)
* each sub-class implement `init`, and any additional factory method it wants
* instantiation starts always with an empty object, and must get initialized

I'll open a separate PR for the comments, after this will be merged.

lib/p2p/packets/Packet.ts
```diff
@@ -1,11 +1,43 @@
...
 abstract class Packet {
+  public header!: PacketHeader;
   public abstract type: PacketType;
+  public abstract messageType: MessageType;
   public abstract body: any;
-  public abstract responseType?: PacketType;
-  public abstract responseTimeout?: number;
-  public abstract toRaw(): string;
+
+  public fromRaw(raw: string): Packet {
```
sangaman: `fromRaw` as a non-static method doesn't seem right. What about making this a static method and use generics to define what class gets returned?

lib/p2p/packets/types/HelloPacket.ts
```diff
@@ -0,0 +1,23 @@
...
+  public body!: HelloPacketBody;
+
+  public init(args: HelloPacketBody): HelloPacket {
```
sangaman: This init pattern seems unnecessary and somewhat counterintuitive. Why not just make this a constructor method that accepts a `HelloPacketBody` param. I don't see the difference between `new HelloPacket(body)` and `new HelloPacket().init(body)`.

lib/p2p/packets/Packet.ts
moshababo: Generics won’t work here, because you can’t instantiate a new object by using `T` (unlike in C#). So you will must pass the constructor to the method, which seems hacky to me. 
For the non-static way it won't work either, because a sub-class is not comparable with the base-class.

It works fine as a static method if you implement it on each sub-class, but I wanted to avoid it, because as for now the implementation is exactly the same, and plus, it gives a nice separation of serialization/deserialization from the sub-classes.

lib/p2p/packets/types/HelloPacket.ts
moshababo: It's mainly to avoid constructor overloading (which isn't so straight-forward in TS as I thought), and to be aligned with the non-static approach for the `toRaw/fromRaw`, which its reasons I specified in the other comment. 
Declaring factory methods on an instance is indeed quite counterintuitive, but isn't so uncommon, due to the flexibility it gives.

moshababo:
@sangaman I didn't end up with writing any changes, feel free to push your own commit here if you have a better design in mind.

lib/p2p/packets/Packet.ts
sangaman: Ahh I wasn't aware that this wasn't possible but it seems you are right.

sangaman:
Thanks, I will take one more close look at this tonight and then probably merge. If I can come up with a better design I'll probably do it in a separate PR - I could also perhaps handle the commenting in there as well.

sangaman:
Merging and will propose my changes separately since it's kind of outgrown the scope of this PR. Thanks.

Issue #146, opened 2018-06-27
Orders/Swaps High-level API


moshababo:
Add support on gRPC proto/service to the following high-level API methods, so they could be implemented on the `Service` class later on.

2 modes:
1. (Prio0) `disableMatching` = `false`:
- placeOrder (returns `matchingResult`) (matching result= successfullyMatched[orderID,quantity], remaining[orderID,quantity]
- cancelOrder (returns success/error)
- subscribePeerOrders [streaming]
- subscribeSwaps [streaming]

2. (Prio2) `disableMatching` = `true`:
- placeOrder (returns success/error)
- cancelOrder (returns success/error)
- subscribePeerOrders [streaming]
- subscribeSwaps [streaming]
- executeSwap

PR #193, opened 2018-07-05
Add New Swap & Order Subscription RPC Methods 

itsbalamurali:
⚠️  This is a WIP PR implementing #146.


moshababo:
Same as in #167, methods handlers are not provided.

michael1011:
@moshababo I also finished #194 because I was already at it.

We need an additional issue for subscribing to swaps because the [API method of Raiden](https://raiden-network.readthedocs.io/en/v0.3.0/rest_api.html#querying-channel-events) isn't implemented in our client yet and we don't have LND swaps in XUD right now.

`placeOrder` is already implemented and I think `cancelOrder` is part of #167 isn't it?

lib/cli/commands/subscribePeerOrders.ts
```diff
@@ -0,0 +1,15 @@
+import { loadXudClient } from '../command';
+import { Arguments } from 'yargs';
+import { SubscribePeerOrdersRequest, SubscribePeerOrdersResponse } from '../../proto/xudrpc_pb';
+
+export const command = 'subscribeorders';
```
moshababo: better to use the full name `subscribepeerorders`, to make it more explicit.

lib/service/Service.ts
```diff
@@ -158,11 +162,27 @@ class Service {
...
+   */
+  public executeSwap = async ({ target_address, payload, identifier }: { target_address: string, payload: TokenSwapPayload, identifier: string }) => {
     return this.raidenClient.tokenSwap(target_address, payload, identifier);
   }
+
+  /*
+   * Subscribe to incoming peer orders.
+   */
+  public subscribePeerOrders = async () => {
+    this.pool.on('packet.order', (order: orders.PeerOrder) => this.emit('peerOrder', order));
```
moshababo: I think it makes more sense to implement the subscribes via function-level callbacks, and not via class-level event emitter. So just provide a callback to `subscribePeerOrders`, and call it instead of `this.emit(...`

lib/grpc/GrpcService.ts
```diff
@@ -89,15 +78,30 @@ class GrpcService {
...
 
   public shutdown: grpc.handleUnaryCall<{}, {}> = async (call, callback) => {
     this.unaryCall(call.request, callback, this.service.shutdown);
   }
+
+  /*
+   * See [[Service.subscribePeerOrders]]
+   */
+  public subscribePeerOrders: grpc.handleServerStreamingCall<{}, {}> = async (call) => {
+    this.service.on('peerOrder', (order: orders.PeerOrder) => call.write({ order }));
```
moshababo: I couldn't test it, but it seems that it will never trigger, because you didn't call `service.subscribePeerOrders`, so it will that it will subscribe to `pool`. 

but anyway I think we should use callbacks instead, as I wrote in a different comment.

lib/service/Service.ts
moshababo: Better to trigger/listen on `orderBook` and directly on `pool`. we might not expose merely the packet.

Issue #170, opened 2018-06-27
Handle gRPC Server Startup Errors


itsbalamurali:
Currently, we are not handling gRPC startup errors, which we should in scenarios like the port specified is already in use the gRPC binding is throwing errors quitely which needs to be caught.

itsbalamurali:
Closing on #195

PR #195, opened 2018-07-06
Catch Bind port number to make sure grpc started successfully

itsbalamurali:
Implements #170

### DIFF
--- a/lib/grpc/GrpcServer.ts
+++ b/lib/grpc/GrpcServer.ts
@@ -39,8 +39,8 @@ class GrpcServer {
     try {
       const bindCode = this.server.bind(`${host}:${port}`, grpc.ServerCredentials.createInsecure());
       if (bindCode !== port) {
-        this.logger.error(`gRPC Failed to listen on port: ${port}`);
-        throw `gRPC couldn't bind on port: ${port}`;
+        this.logger.error(`Failed to listen on port: ${port}`);
+        throw new Error(`gRPC couldn't bind on port: ${port}`);
       }
       this.server.start();
       this.logger.info(`gRPC server listening on ${host}:${port}`);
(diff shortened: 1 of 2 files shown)


lib/grpc/GrpcServer.ts
```diff
@@ -36,9 +36,12 @@ class GrpcServer {
    */
   public listen = (port: number, host: string) => {
     assert(Number.isInteger(port) && port > 1023 && port < 65536, 'port must be an integer between 1024 and 65535');
-
     try {
-      this.server.bind(`${host}:${port}`, grpc.ServerCredentials.createInsecure());
+      const bindCode = this.server.bind(`${host}:${port}`, grpc.ServerCredentials.createInsecure());
+      if (bindCode !== port) {
+        this.logger.error(`gRPC Failed to listen on port: ${port}`);
+        throw `gRPC couldn't bind on port: ${port}`;
```
moshababo: Small remarks:
* better not to throw a string.
* you can avoid the local scope logging, since you're not adding any further information
* you can drop the `gRPC` prefix in the local scope logging, since you're already in that context.

lib/grpc/GrpcServer.ts
```diff
@@ -36,9 +36,12 @@ class GrpcServer {
    */
   public listen = (port: number, host: string) => {
     assert(Number.isInteger(port) && port > 1023 && port < 65536, 'port must be an integer between 1024 and 65535');
-
     try {
-      this.server.bind(`${host}:${port}`, grpc.ServerCredentials.createInsecure());
+      const bindCode = this.server.bind(`${host}:${port}`, grpc.ServerCredentials.createInsecure());
+      if (bindCode !== port) {
+        this.logger.error(`Failed to listen on port: ${port}`);
+        throw new Error(`gRPC couldn't bind on port: ${port}`);
```
moshababo: Let's use pre-defined errors, defined per-domain, same as in here, so we can easily track them and give them codes in a later stage.

moshababo: Let's use pre-defined errors, defined per-domain, same as in [here](https://github.com/ExchangeUnion/xud/blob/master/lib/p2p/errors.ts), so we can easily track them and give them codes in a later stage.

sangaman: Agreed.

PR #196, opened 2018-07-06
Merge MatchingEngine Queues

itsbalamurali:
Implements #166

PR #199, opened 2018-07-10
Order bootstrapping

### LINKED ISSUE #141: Don't store orders on disk, order bootstrap mechanism

### LINKED ISSUE #149: p2p packets: get_orders, orders
* `get_orders`: on joining the network, query peers for the list of their existing valid orders. Default behavior is to query all orders, but a base timestamp optional condition should be supported.
* `orders`: `get_orders` response. if timestamp is applied, it should be considered as local.

EDIT: `get_orders` shouldn't only be called when joining the network, but rather on every time that the node is connecting to a new peer. The opposite applies as well: when disconnecting from a peer, remove his orders (#168).

EDIT 2: I think we should drop the timestamp support, because it will never be reliable, and it's actually not needed. If a node will query a peer for its orders, it should assume that it will send all of them. if it will query it incorrectly (without clearing the peer's orders, or after handshaking was completed, when new orders have already arrived), it will simply get duplicated orders, and it won't be the peer's fault.
Since handshaking is blocking the broadcasting of new orders, the edge case would be that between the `orders` response and the handshake finalization, some new orders won't be broadcasted. There are a few solutions for this, but I don't think we even need to address this issue. 
@kilrau @sangaman

michael1011:
Closes #141 
Closes #149 

This PR removes the orders table from the database and stores orders temporarily in the OrderBook. 

And when connecting to another `xud` node both sides will send their ` ownOrders` after the handshake finished.


lib/orderbook/OrderBook.ts
```diff
@@ -29,102 +39,153 @@ class OrderBook {
...
+
+      return result;
+    } else {
+      return orders;
+    }
   }
 
-  public addLimitOrder = async (order: orders.OwnOrder): Promise<matchingEngine.MatchingResult>  => {
+  // TODO: sort when adding new order or getting existing ones?
+  private sortOrdersByPrice = (orders: Orders) => {
```
moshababo: I don't think it make sense, performance-wise, to sort the orders on each get request. That's why we use the priority queues in the matching engine, so the list will be sorted each time a new order is added (slow writes, fast reads). 
It's not clear if we even need this result sorted. For the `GETORDERS` response, the peer will need to re-sort anyway, since he can't rely on it to already be sorted. And for the `getInfo` response, i'm not sure we need it sorted, and even if so, we can support that separately.
At the moment I guess you can just use the priority queue from MatchingEngine, if you want it sorted. But since we're going to merge the own/peer orders queues (#196) it won't be usable anymore soon, because you wouldn't be able to get the sorted `ownOrders` without the `peerOrders`. So I think that in the meanwhile you can just return it unsorted.

lib/orderbook/OrderBook.ts
```diff
@@ -1,26 +1,36 @@
...
+  public matchingEngines: { [ pairId: string ]: MatchingEngine } = {};
+
   private logger: Logger = Logger.orderbook;
   private repository: OrderBookRepository;
   private matchesProcessor: MatchesProcessor = new MatchesProcessor();
-  public pairs: db.PairInstance[] = [];
-  public matchingEngines: {[ pairId: string ]: MatchingEngine} = {};
+
+  private ownOrders: { [ pairId: string ]: Orders } = {};
+  private peerOrders: { [ pairId: string ]: Orders } = {};
```
moshababo: We don't have to do it now,  but these lists will need to support updates/deletes by the keys: `orderId`, `hostId`. 
So we better remember to optimize it to avoid having to do a "full table scan" (at least for `orderId`) on each operation.

lib/p2p/Pool.ts
```diff
@@ -119,18 +121,58 @@ class Pool extends EventEmitter {
...
+        quantity: order.quantity,
+        invoice: invoice as string,
+      };
+      outgoingOrders.push(cast);
     }
   }
 
   private handleOpen = async (peer: Peer, handshakeState: HandshakeState): Promise<void> => {
     this.setPeerHost(peer, handshakeState.listenPort);
+    peer.sendPacket(new GetOrdersPacket());
```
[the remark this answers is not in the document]
michael1011: I would handle duplicates gracefully. There is no harm except a little bit of computational and networking overhead.

moshababo:
@michael1011 please re-assign me as a reviewer when you want me to go over the changes.

lib/orderbook/OrderBook.ts
```diff
@@ -1,130 +1,205 @@
...
 
   constructor(private config: OrderBookConfig, db: DB, private pool?: Pool, private lndClient?: LndClient) {
     this.repository = new OrderBookRepository(db);
     if (pool) {
       pool.on('packet.order', this.addPeerOrder);
+      pool.on('packet.getOrders', async (peer: Peer) => {
```
moshababo: better to keep the constructor clean. wrap it all in a class-level function.

lib/orderbook/OrderBook.ts
```diff
@@ -1,130 +1,205 @@
...
   */
-  public getOwnOrders(pairId: string, maxResults: number): Promise<Orders> {
-    return this.repository.getOwnOrders(pairId, maxResults);
+  public getOwnOrders = (pairId: string, maxResults: number): Orders => {
+    return this.getOrders(maxResults, this.ownOrders[pairId]);
+  }
+
+  private getOrders = (maxResults: number, orders: Orders) => {
+    if (maxResults > 0) {
+      const result = this.initOrders();
```
moshababo: any reason for still doing a push into a result object, and not merely returning an object with the slice result?

lib/orderbook/OrderBook.ts
```diff
@@ -1,130 +1,205 @@
...
     return this.addOwnOrder(order);
   }
 
-  public addMarketOrder = async (_order: orders.OwnOrder) => {
+  public addMarketOrder = (_order: orders.OwnOrder) => {
     // TODO: implement
   }
 
-  private addOwnOrder = async (order: orders.OwnOrder): Promise<matchingEngine.MatchingResult>  => {
+  public castToOutgoingOrder = async (ownOrders: StampedOrder[], outgoingOrders: OutgoingOrder[]) => {
```
moshababo: This is too similar to `broadcastOrder`, although was implemented differently. I'd rather have the logic of adding an invoice and casting to an `OutgoingOrder` implemented only once.

lib/orderbook/OrderBook.ts
```diff
@@ -1,130 +1,205 @@
...
+  }
+
+  private updateOrderQuantity = (type: { [pairId: string]: Orders }, order: StampedOrder, decreasedQuantity: number) => {
+    const array = this.getOrderArray(type, order);
+    array.forEach((item, index) => {
+      if (item.id === order.id) {
+        item.quantity = item.quantity - decreasedQuantity;
+        if (item.quantity === 0) {
+          array.splice(index, 1);
+        }
```
moshababo: Although we might change the current design of keeping both `ownOrders` and `peerOrders` separated from `MatchingEngine`, I think we should already optimize it to avoid full-scan when querying on `orderId`. So instead using an array, just use a plain object, and use `orderId` for the keys.

lib/orderbook/OrderBook.ts
```diff
@@ -1,130 +1,205 @@
...
+
+        for (const { id } of pairs) {
+          const orders = await this.getOwnOrders(id, 0);
+          await Promise.all([
+            this.castToOutgoingOrder(orders.buyOrders, outgoingOrders),
+            this.castToOutgoingOrder(orders.sellOrders, outgoingOrders),
+          ]);
+        }
+
+        const packet = new OrdersPacket({ orders: outgoingOrders });
```
moshababo: `OrderBook` shouldn't construct a new packet. You already implement `sendOrders` under `Peer`. You should just pass it `outgoingOrders`.

lib/orderbook/OrderBook.ts
moshababo: I'm less in favor of the current method design of pushing the result into an a referenced array. Rather just return the results.

lib/orderbook/OrderBook.ts
moshababo: I was referring to doing something like this:
```
return {
  buyOrders: orders.buyOrders.slice(…
  sellOrders: orders.sellOrders.slice(…
}
```

Nothing major, just thought it can make the code even more simple.

lib/orderbook/OrderBook.ts
```diff
@@ -1,151 +1,220 @@
...
-  public matchingEngines: {[ pairId: string ]: MatchingEngine} = {};
+
+  private ownOrders: { [ pairId: string ]: Orders } = {};
+  private peerOrders: { [ pairId: string ]: Orders } = {};
 
   constructor(private config: OrderBookConfig, db: DB, private pool?: Pool, private lndClient?: LndClient) {
     this.repository = new OrderBookRepository(db);
     if (pool) {
-      pool.on('packet.order', this.addPeerOrder);
+      this.registerListeners(pool);
```
moshababo: i'd rather use the class-level method only for event callback, and have them specified in the constructor, not wrapped inside a different function. (Same as it was with `pool.on('packet.order', this.addPeerOrder)`)

lib/orderbook/OrderBook.ts
```diff
@@ -1,151 +1,220 @@
...
   }
 
-  public addLimitOrder = async (order: orders.OwnOrder): Promise<matchingEngine.MatchingResult>  => {
+  private getOrders = (maxResults: number, orders: Orders): Orders => {
+    if (maxResults > 0) {
+      const result = this.initOrders();
+
+      const buyOrders = Object.entries(result.buyOrders);
+      const sellOrders = Object.entries(result.sellOrders);
+      for (let index; index < maxResults; index = index + 1) {
```
moshababo: variable is not initialized

lib/orderbook/OrderBook.ts
```diff
@@ -1,151 +1,220 @@
...
     const invoice = await this.getInvoice(order);
-    if (!invoice) {
-      return;
-    }
 
-    const { createdAt, ...outgoingOrder } = { ...order, invoice };
-    this.pool.broadcastOrder(outgoingOrder);
+    if (!invoice) return;
+
+    const { ...outgoingOrder } = { ...order, invoice };
```
moshababo: This can be written as `const outgoingOrder = { ...order, invoice }`;

But any reason for not dropping the `createdAt`, as it previously was? (since it's not defined in `OutgoingOrder` type)

lib/p2p/Pool.ts
```diff
@@ -6,8 +6,8 @@ import HostList from './HostList';
 import SocketAddress from './SocketAddress';
 import PeerList from './PeerList';
 import P2PRepository from './P2PRepository';
-import { Packet, PacketType, OrderPacket } from './packets';
-import { orders } from '../types';
+import { Packet, PacketType, OrderPacket, OrdersPacket, GetOrdersPacket } from './packets';
```
moshababo: unused type

michael1011:
Your parser works fine and I tested this a bit locally and it seemed to work after I fixed a few more minor details.

lib/orderbook/OrderBook.ts
michael1011: I though that it gets dropped automatically because it is not defined in the type?

moshababo: It doesn't do it, so you can easily end-up with having a run-time object which has more properties then what was defined in TS.

lib/orderbook/OrderBook.ts
```diff
@@ -1,151 +1,212 @@
...
+    if (this.pool) {
+      const outgoingOrder = await this.castToOutgoingOrder(order);
+
+      if (outgoingOrder) {
+        this.pool.broadcastOrder(outgoingOrder);
+      }
     }
+  }
+
+  private castToOutgoingOrder = async (order: orders.StampedOwnOrder): Promise<orders.OutgoingOrder | void> => {
```
moshababo: I think that the name `castToOutgoingOrder` doesn't fit here, because you're not just casting, but also creating an invoice. So perhaps `createOutgoingOrder` is better (and also, maybe change `getInvoice` to `createInvoice`).

moshababo:
@michael1011 for automatic testing, p2p tests will be required, but we haven't progressed much with it (#132), so we can't do it in this PR. Please open a separate issue for that.

In regards to merging this, let's wait for a secondary review, in case we missed something. @sangaman @itsbalamurali

lib/orderbook/OrderBookRepository.ts
```diff
@@ -1,104 +1,18 @@
...
-import { OrderAttributes } from '../types/db';
-
-type Orders = {
-  buyOrders: db.OrderInstance[];
-  sellOrders: db.OrderInstance[];
-};
+import { db } from '../types';
+import DB, { Models } from '../db/DB';
 
 class OrderbookRepository {
```
sangaman: Now that this class isn't actually fetching orders from a db repository any longer, it seems like it's misnamed. In fact, I think we can probably get rid of this class altogether. I figure we can extract the remaining logic in this class (which is 3 lines, really) and move it directly to `Orderbook.ts`.

lib/p2p/Pool.ts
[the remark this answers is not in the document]
sangaman: I think duplicates is better too, we just need to make sure we don't penalize the peer for sending dupes. Maybe no penalizing for duplicates until the `GETORDERS` sequence has completed?

lib/orderbook/OrderBookRepository.ts
moshababo: i'd suggest to keep all `sequelize` object wrapped separately, and keep this class for now (might get more traction later). I think it's reasonable that currency/pairs management is under order book domain. If not, we can open a generic repository. I think it's good to keep it separate from `DB` or service classes.

proto/xudrpc.proto
```diff
@@ -154,10 +154,14 @@ message Order {
...
+  // The id of the host that created this order
+  string host_id = 5 [json_name = "host_id"];
   // A UUID for this order
-  string id = 5 [json_name = "id"];
+  string id = 6 [json_name = "id"];
   // The epoch time when this order was created
-  int64 created_at = 6 [json_name = "created_at"];
+  int64 created_at = 7 [json_name = "created_at"];
+  // Lightning invoice
+  string invoice = 8 [json_name = "invoce"];
```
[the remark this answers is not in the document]
michael1011: This change isn't about functionality but just a workaround to make the gRPC show orders of the type `PeerOrder`. I ran into issues when manually testing the new P2P packets. Lets leave `invoice` for now and change this when the other approach is ready.

lib/p2p/Parser.ts
```diff
@@ -14,18 +14,24 @@ enum ParserErrorType {
...
     const dataStr = data.toString();
-    const type = dataStr.split(' ', 1)[0];
-    const body = dataStr.substring(type.length + 1);
-    try {
-      const packet = packetUtils.fromRaw(type, body);
-      if (packet) {
-        this.emit('packet', packet);
-      } else {
-        this.emit('error', new ParserError(ParserErrorType.UNKNOWN_PACKET_TYPE, type));
+    const packets = dataStr.split('\r\n');
```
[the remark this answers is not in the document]
michael1011: @moshababo came up with this solution when I had this issues:

```
2018-7-12 10:11:59 [P2P] debug: Connected to peer (localhost:8895)
2018-7-12 10:11:59 [P2P] debug: Received data (localhost:8895): HELLO {"version":"123","nodeKey":"123","listenPort":20000,"pairs":["BTC/LTC"]}
GETORDERS null
ORDERS {"orders":[]}
2018-7-12 10:11:59 [P2P] warn: Unparsable peer message: HELLO {"version":"123","nodeKey":"123","listenPort":20000,"pairs":["BTC/LTC"]}
GETORDERS null
ORDERS {"orders":[]}
: SyntaxError: Unexpected token G in JSON at position 74
Executing (default): INSERT INTO `hosts` (`id`,`address`,`port`,`createdAt`,`updatedAt`) VALUES (DEFAULT,'localhost',8895,'2018-07-12 08:11:59','2018-07-12 08:11:59');
2018-7-12 10:12:09 [P2P] debug: Socket Error (localhost:8895): "Peer is stalling (HELLO)"
2018-7-12 10:12:09 [P2P] error: peer error undefined
```

lib/p2p/Pool.ts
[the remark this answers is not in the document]
moshababo: Possible, let's just see it doesn't turn out too hacky to implement.

lib/p2p/Parser.ts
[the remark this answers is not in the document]
sangaman: Gotcha, that makes sense. I figure why not continue using the linebreaks for splitting? We can discuss more later.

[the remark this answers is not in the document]
moshababo: I’m not sure which layer added the `\r\n` - Nodejs Core or the TCP/IP stack, an if it's compatible across platforms. I'd rather define something that we control on our own. In addition, I assume that we need to prepare for long messages being split. 
These are my 2 main concerns at the moment, but we might encounter some more.

michael1011:
I am still not comfortable with rebasing. Could someone check real quick that I did it correctly? @sangaman @moshababo

moshababo:
Looks fine to me, but I usually use merging, so let's wait for @sangaman to check it

moshababo:
@sangaman can we merge this?

sangaman:
My apologies, I actually had written up that review last night and just checked on this PR again to see if it had been merged yet and realized I didn't hit "submit."

sangaman:
Also generally speaking I figure merging master into a feature branch works just as well as rebasing if it's easier for you. Only advantage of rebasing is preserving a specific commit history, but it doesn't matter if it gets squashed before being merged back into master.

moshababo:
@sangaman In regards to the orders sorting, I commented about it here: https://github.com/ExchangeUnion/xud/pull/199#discussion_r201421399

The best solution would be to manage the orders solely on `MatchingEngine`, and to remove the duplicate work from `OrderBook` that was done in this PR. After #196 will be merged, we can support querying only `ownOrders` from `MatchingEngine` priority queues  either by  filtering `peerOrders`, or by managing a separate priority queue  (a bit similar to what was done now in `OrderBook`).

PR #201, opened 2018-07-17
Refactor packets

sangaman:
This is a followup to #189. I think I found a reasonable approach to the packet classes using generics. The body property itself is generic, and all classes extended `Packet` share a single constructor which itself is overloaded in that it accepts either a packet body type and an optional request hash (for creating outgoing packets) or a string (for deserializing incoming packets).

This refactors the Packet class and its derivations in the following ways:

- Adding a generic type parameter to Packet which is used for a generic constructor and to define the `body` type.
- Generic constructor that accepts either a packet body type or a string. If passed a string, it attempts to deserialize it into a packet.
- Renaming `MessageType` to `PacketDirection` to differentiate itself from `PacketType`
- Change the `type` and `messageType`/`direction` parameters to getter functions that return fixed values, since we don't want these values to be overwritten.

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -93,7 +93,7 @@ class Pool extends EventEmitter {
   }
 
   public broadcastOrder = (order: orders.OutgoingOrder) => {
-    const orderPacket = new OrderPacket().fromOutgoingOrder(order);
+    const orderPacket = new OrderPacket(order);
     this.peers.forEach(peer => peer.sendPacket(orderPacket));
   }
 
--- a/lib/p2p/packets/Packet.ts
+++ b/lib/p2p/packets/Packet.ts
@@ -3,41 +3,56 @@ import CryptoJS from 'crypto-js';
 import MD5 from 'crypto-js/md5';
 
 type PacketHeader = {
+  /** The Base64 encoded MD5 hash of the body of this packet. This is used for error checking and as a key to identify packets.*/
   hash: string;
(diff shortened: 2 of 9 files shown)


lib/p2p/Pool.ts
```diff
@@ -93,7 +93,7 @@ class Pool extends EventEmitter {
   }
 
   public broadcastOrder = (order: orders.OutgoingOrder) => {
-    const orderPacket = new OrderPacket().fromOutgoingOrder(order);
```
moshababo: if you don't want to use this factory method, remove it from the class as well (unless you wanted to keep it as an example).

lib/p2p/packets/Packet.ts
```diff
@@ -3,41 +3,61 @@ import CryptoJS from 'crypto-js';
...
+  /** A packet that is sent in response to an incoming packet. */
   RESPONSE,
 }
 
-abstract class Packet {
-  public header!: PacketHeader;
-  public abstract type: PacketType;
-  public abstract messageType: MessageType;
-  public abstract body: any;
+function createHeader(body: any, reqHash?: string): PacketHeader {
```
michael1011: We were defining functions like this: `const createHeader = (body: any, reqHash?: string): PacketHeader => {` and not with the `function` keyword everywhere else.

This has nothing to do with this specific PR but we are mixing the way we define functions quite a bit.

lib/p2p/Pool.ts
sangaman: Thanks, it was just an oversight.

PR #203, opened 2018-07-20
Verify connection to lnd fixes #52

### LINKED ISSUE #52: Verify connection to lnd
There's currently a TODO in the code to verify communication with `lnd` via grpc. This can be done with a simple `getinfo` call to lnd or establishing a stream to listen to invoice payments, for example.

rsercano:
fixes #52

Hello, tried my way to resolve this issue, I'm ready to apply whatever change you need after the review. Also fixed a simple command on docker-compose.yml (it was `npm run run`)


lib/lndclient/LndClient.ts
```diff
@@ -195,12 +212,16 @@ class LndClient extends BaseClient {
       })
       .on('end', () => {
         this.logger.info('invoice ended');
+        this.setStatus(ClientStatus.DISABLED);
+        this.connect(this);
```
michael1011: When `connect` is called here loading the certificate and macaroon again isn't sensible. Initializing `lndClient` should be moved back to the constructor.

rsercano:
@michael1011 added requested changes

docker/docker-compose.yml
```diff
@@ -18,7 +18,7 @@ services:
       - btc
       - lnd
       - db
-    entrypoint: "npm run run"
+    entrypoint: "npm run start"
```
sangaman: Good catch, but this can actually just be `npm start` for the sake of brevity.

lib/lndclient/LndClient.ts
```diff
@@ -26,14 +26,17 @@ interface GrpcResponse {
 class LndClient extends BaseClient {
   private lightning!: LightningClient;
   private meta!: grpc.Metadata;
+  private config!: LndClientConfig;
```
sangaman: Rather than keep the entire config object as a property on the class, I think I'd prefer just keeping `host` and `port`. Even better maybe just have a property called `uri` which is the `${host}:${port}` concatenation and set it in the constructor.

lib/lndclient/LndClient.ts
```diff
@@ -72,6 +73,25 @@ class LndClient extends BaseClient {
     });
   }
 
+  private connect = (lndCLient: LndClient, subscribeInvoices: boolean = false) => {
```
sangaman: You don't need the `lndClient` parameter here, you can just use `this`.

lib/lndclient/LndClient.ts
```diff
@@ -72,6 +73,25 @@ class LndClient extends BaseClient {
...
+    const { host, port } = lndCLient.config;
+
+    lndCLient.lightning = new LightningClient(`${host}:${port}`, lndCLient.credentials);
+
+    lndCLient.getInfo().then((getInfoResponse) => {
```
sangaman: We are trying to use the async/await pattern everywhere in the codebase, so make this an `async` function and do `const getInfoResponse = await this.getInfo();` inside a try/catch statement.

lib/lndclient/LndClient.ts
```diff
@@ -72,6 +73,25 @@ class LndClient extends BaseClient {
...
+      if (getInfoResponse && getInfoResponse.numActiveChannels >= 1) {
```
sangaman: Why do we need there to be at least 1 active channel to consider the connection verified? We can be connected to lnd even without any active channels.

lib/lndclient/LndClient.ts
```diff
@@ -72,6 +73,25 @@ class LndClient extends BaseClient {
...
+        lndCLient.setStatus(ClientStatus.CONNECTION_VERIFIED);
+        if (subscribeInvoices) {
```
sangaman: In my experience, it's necessary to resubscribe to invoices whenever a connection to lnd is reestablished, so we might want to remove this if statement and just call `subscribeInvoices` every time we connect.

lib/lndclient/LndClient.ts
```diff
@@ -72,6 +73,25 @@ class LndClient extends BaseClient {
...
+          lndCLient.subscribeInvoices();
+        }
+      }
+    }).catch((err: any) => {
+      lndCLient.logger.error(`could not fetch info from lnd host, error: ${err}, retrying in 5000 ms`);
+      lndCLient.setStatus(ClientStatus.DISABLED);
+      setTimeout(lndCLient.connect.bind(null, lndCLient), 5000);
```
sangaman: I'm not 100% sure, but I don't believe we need the `bind` statement here because we're using arrow functions.

lib/lndclient/LndClient.ts
```diff
@@ -72,6 +73,25 @@ class LndClient extends BaseClient {
...
+    lndCLient.getInfo().then((getInfoResponse) => {
+      if (getInfoResponse && getInfoResponse.numActiveChannels >= 1) {
+        lndCLient.setStatus(ClientStatus.CONNECTION_VERIFIED);
+        if (subscribeInvoices) {
+          lndCLient.subscribeInvoices();
+        }
+      }
+    }).catch((err: any) => {
+      lndCLient.logger.error(`could not fetch info from lnd host, error: ${err}, retrying in 5000 ms`);
+      lndCLient.setStatus(ClientStatus.DISABLED);
```
sangaman: I don't think we need to set the status to disabled here, because if we are attempting to connect that would mean that the status should already be set to DISABLED.

lib/lndclient/LndClient.ts
rsercano: after arrow func, that's true :)

lib/lndclient/LndClient.ts
rsercano: now just checking if there's a response from getInfo to determine the successful connection

lib/lndclient/LndClient.ts
rsercano: That was my first approach, fixing it again.

lib/lndclient/LndClient.ts
rsercano: already removed, because of arrow function as you mentioned we don't need lndClient anymore.

rsercano:
@sangaman just done :)

lib/lndclient/LndClient.ts
```diff
@@ -44,15 +46,14 @@ class LndClient extends BaseClient {
...
-      const credentials = grpc.credentials.createSsl(lndCert);
-      this.lightning = new LightningClient(`${host}:${port}`, credentials);
+      this.credentials = grpc.credentials.createSsl(lndCert);
 
       const adminMacaroon = fs.readFileSync(macaroonpath);
       this.meta = new grpc.Metadata();
       this.meta.add('macaroon', adminMacaroon.toString('hex'));
-
-      this.setStatus(ClientStatus.CONNECTION_VERIFIED); // TODO: verify connection
+      this.connect();
```
sangaman: I realized upon a second pass that we have an `async` call here within a constructor which is necessarily synchronous. We want to make sure we are connected before we continue with our startup process. Can you make the `connect` method public and move it from the constructor to `Xud.ts` right after we create the client by calling `await this.lndClient.connect();`?

lib/lndclient/LndClient.ts
```diff
@@ -72,6 +73,21 @@ class LndClient extends BaseClient {
     });
   }
 
+  private connect = async () => {
```
sangaman: Change to public.

lib/lndclient/LndClient.ts
michael1011: @sangaman Are you sure? This methods gets also called from the constructor which means always being subscribed to the invoices. `subscribeInvoices` should be private then.

rsercano:
@michael1011 just made subscribeInvoices private, would you please check ? 

There were no tests failing.

lib/lndclient/LndClient.ts
```diff
@@ -72,6 +72,22 @@ class LndClient extends BaseClient {
...
+    this.lightning = new LightningClient(this.uri, this.credentials);
+
+    try {
+      const getInfoResponse = await this.getInfo();
+      if (getInfoResponse) {
+        this.setStatus(ClientStatus.CONNECTION_VERIFIED);
+        this.subscribeInvoices();
+      }
+    } catch (err) {
+      this.logger.error(`could not fetch info from lnd host, error: ${err}, retrying in 5000 ms`);
```
michael1011: You should use `JSON.stringify(err)` else `could not fetch info from lnd host, error: [object Object], retrying in 5000 ms` will get logged.

rsercano:
@michael1011 that's my bad didn't recieve the github notification about your review somehow, but just fixed the it tests.

lib/lndclient/LndClient.ts
rsercano: fixed the it test as well :)

rsercano:
ping @michael1011  :)

rsercano:
if everything is okay, would you please merge ? :) @michael1011 @sangaman

lib/lndclient/LndClient.ts
```diff
@@ -72,6 +72,26 @@ class LndClient extends BaseClient {
     });
   }
 
+  public connect = async () => {
+    if (this.isDisabled()) {
```
sangaman: Setting status to disabled again here is unnecessary, you can just make this `if (!this.isDisabled())` and go right into the connect logic without needing an else statement.

Issue #185, opened 2018-07-01
Peers/Host data structures


kilrau:
Redesign Peers/Host data structures to support future features:
- pubkey based on blocking #162 
- reputation
- get hosts packet response

kilrau:
discussion with @sangaman

moshababo:
Here's a rough design, subject to change. I'm posting it early so you could ask questions / comment about it before i'll revise it. We can edit this message for future updates.

**db tables**

* **hosts**: id (pk), score (def = 0, sum snapshot of reputation history), latestIpAddress, latestNodePubKey (nullable), latestListenPort (nullable)

* **hostIds**: hostId (fk), ipAddress (nullable), nodePubKey (nullable), createdAt

(each row assign either `ipAddress` **or** `nodePubKey` to a `hostId`. The purpose of this table is to associate multiple `ipAddress` and `nodePubKey` with a `hostId`. Reason: a recognizable host entity can have multiple ip addresses and multiple public keys. A new host entity will be instantiated based on `ipAddress`)

* **hostReputation**: id (pk), hostId (fk), event (enum), createdAt

(not specifying score of the event, for easy upgrades)

**app (pool)**

`peer` will be created together with a `host` object, based on `ipAddress`, which is known in the instantiation stage. `host` object will be the gateway to all updates by `peer`.

**create a new host (ipAddress, listenPort)**

1) insert into `hosts` (ipAddress, latestListenPort)
2) insert into `hostIds` (hostId, ipAddress)

**get host (ipAddress)**

select from `hostIds` (by ipAddress)

**xud postinstall**

create new hosts from a list of <ipAddress, listenPort>

 **p2p bootstrap:**

select top X from `hosts`, where `listenPort` is not null and `score` > Y, order by `score`

**new outbound**  

check if ipAddress is already connected (peerList), or if it's the node own ipAddress + listenPort. if not, get or create host, and instantiate `peer` with it. 

**new inbound** 

get or create host, and instantiate `peer` with it. if host score is below threshold, disconnect

**`get_hosts` request iteration**

for each host, new outbound connection

**`get_hosts` packet handling**

return select top X from `hosts`, where `listenPort` is not null and `score` > Y, order by `score`

**after outbound connection failure**

add negative reputation to host (new entry on `hostReputation` table, score is update on `hosts` table)

**after outbound connection success**

add positive reputation to host 

**after handshake**

[...]

PR #204, opened 2018-07-22
Hosts and Get_Hosts Packets

### LINKED ISSUE #147: p2p packets: get_hosts, hosts
* `get_peers`: asking for information about known active peers to help with finding potential nodes in the network.
* `peers`: `get_peers` response. provide information on active peers.

sangaman:
Resolves #147. I split this into two commits, the first is an attempt at some general cleanup, refactoring, and code documentation around the p2p module. The second implements the new packets. Currently the `GetHostsPacket` is empty, but we can modify this in the future to specify some criteria for which hosts should be returned. When we successfully establish an outbound connection to a peer, we send a `GetHostsPacket` which kicks off recursion of connecting to new hosts as long as we keep finding out about new ones.

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -34,13 +34,17 @@ class Pool extends EventEmitter {
     return this.peers.length;
   }
 
-  public connect = async (): Promise<void> => {
+  /**
+   * Initialize the Pool by connecting to known hosts and listening to incoming peer connections, if configured to do so.
+   */
+  public init = async (): Promise<void> => {
     if (this.connected) {
       return;
     }
 
-    for (const host of await this.hosts.get()) {
-      this.addOutbound(host.socketAddress.address, host.socketAddress.port);
+    await this.hosts.load();
+    for (const host of this.hosts.toArray()) {
+      this.addOutbound(host.socketAddress);
     }
 
     if (this.config.listen) {
@@ -65,10 +69,13 @@ class Pool extends EventEmitter {
     this.connected = false;
   }
 
-  public addOutbound = async (address: string, port: number): Promise<Peer> => {
-    const socketAddress = new SocketAddress(address, port);
+  /**
+   * Attempt to add an outbound peer by connecting to a given SocketAddress.
+   * Throws an error if a connection already exists for the provided address.
+   */
+  public addOutbound = async (socketAddress: SocketAddress): Promise<Peer> => {
     if (this.peers.has(socketAddress)) {
-      const err = errors.ADDRESS_ALREADY_CONNECTED(address.toString());
+      const err = errors.ADDRESS_ALREADY_CONNECTED(socketAddress.address);
       this.logger.info(err.message);
       throw err;
     }
@@ -126,12 +133,12 @@ class Pool extends EventEmitter {
         this.emit('packet.order', order);
         break;
       }
-      case PacketType.GETORDERS: {
+      case PacketType.GET_ORDERS: {
         this.emit('packet.getOrders', peer);
         break;
       }
       case PacketType.ORDERS: {
-        const { orders } = packet.body;
+        const { orders } = (packet as OrdersPacket).body;
         orders.forEach((order) => {
           this.emit('packet.order', { ...order, hostId: peer.id });
         });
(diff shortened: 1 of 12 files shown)


lib/p2p/Pool.ts
```diff
@@ -65,16 +69,21 @@ class Pool extends EventEmitter {
...
     if (this.peers.has(socketAddress)) {
-      const err = errors.ADDRESS_ALREADY_CONNECTED(address.toString());
+      const err = errors.ADDRESS_ALREADY_CONNECTED(socketAddress.address);
       this.logger.info(err.message);
       throw err;
     }
 
     const peer = Peer.fromOutbound(socketAddress);
     await this.tryConnectPeer(peer);
+    peer.sendPacket(new GetHostsPacket(null));
```
moshababo: This should be placed in `handleOpen`, with all the other post-handshake operations. 
At the moment it also doesn't apply for inbound connections. Was that on purpose?

lib/p2p/Pool.ts
```diff
@@ -121,22 +130,32 @@ class Pool extends EventEmitter {
...
         break;
-      }
+      case PacketType.GET_HOSTS:
+        peer.sendHosts(this.hosts.toArray());
+        break;
+      case PacketType.HOSTS:
+        const { hosts } = (packet as HostsPacket).body;
+        hosts.forEach(async (host) => {
+          try {
+            this.addOutbound(host.socketAddress);
```
moshababo: `await` is missing (if you want to catch the error)

lib/p2p/Pool.ts
moshababo: Our current design hashes the packet by the body. It means that the body should be unique in some way. that's why I added `ts` field to ping/pong packets body - not a perfect solution, but better then leaving it empty.

sangaman: Yes that was intentional, although I could make it symmetrical. In that case though, should we just have this be part of the handshake process?

sangaman: Yeah I considered this but wasn't sure what to go with. There's a similar issue with the get orders packet and this will be the case with any p2p request that doesn't have anything meaningful to put in the body. I figure for pings the time is useful to determine latency.

Maybe there's some refactoring that can be done here, treating the empty request packets (get orders, get hosts) similarly to the way we treat the Hello packet currently. If that sounds reasonable to you, I'll see what I can come up with.

sangaman: I think this would be related to #208 as well.

lib/p2p/Pool.ts
```diff
@@ -121,22 +130,32 @@ class Pool extends EventEmitter {
...
-        const { orders } = packet.body;
+      case PacketType.ORDERS:
+        const { orders } = (packet as OrdersPacket).body;
         orders.forEach((order) => {
           this.emit('packet.order', { ...order, hostId: peer.id });
         });
         break;
-      }
+      case PacketType.GET_HOSTS:
+        peer.sendHosts(this.hosts.toArray());
```
moshababo: same issue as in https://github.com/ExchangeUnion/xud/issues/208 applies here as well. Need to follow the ping/pong example.

lib/p2p/Pool.ts
moshababo: I don’t think it’s related to how we handle `Hello` packet, because here we have a typical request/response scenarios. 

I think that a timestamp field should be sufficient, because the hash need to be unique only from the sender perspective (he's the one who's waiting for the response). This could be done on the packet type class, so when instantiating a new packet instance, you don't need to provide anything, but the constructor add it.

moshababo: Handshake process is for critical stuff, to block other activities (same as if the peer is not connected yet). So this should be post-handshake.

moshababo: Other option: adding a uuid.
I used timestamp in case it will be useful for other things as well.

sangaman: OK, that makes sense to me just wanted to make sure. Another idea, maybe for a future PR, is that we might want a configuration option to determine whether or not we automatically connect to new hosts that we learn about through getHosts. One might want manual control over which peers they connect to, although it's harmless to at least learn about new peers.

moshababo: Possible, but remember that the new design (#185) would require to save hosts to list before attempting to connect to them.

@kilrau

lib/p2p/packets/types/GetHostsPacket.ts
```diff
@@ -0,0 +1,18 @@
+import Packet, { PacketDirection } from '../Packet';
+import PacketType from '../PacketType';
+
+type GetHostsPacketBody = {
+  ts: number;
+};
+
+class GetHostsPacket extends Packet<GetHostsPacketBody> {
+  public get type() {
+    return PacketType.PING;
```
moshababo: change packet type

lib/p2p/packets/types/HostsPacket.ts
```diff
@@ -0,0 +1,19 @@
...
+import PacketType from '../PacketType';
+import Host from '../../Host';
+
+type HostsPacketBody = {
+  hosts: Host[],
+};
+
+class HostsPacket extends Packet<HostsPacketBody> {
+  public get type() {
+    return PacketType.PONG;
```
moshababo: change packet type

PR #205, opened 2018-07-22
Disconnect Peers

itsbalamurali:
#122 Implements disconnect Peers via RPC

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -152,7 +152,7 @@ class Pool extends EventEmitter {
 
   private handleOpen = async (peer: Peer, handshakeState: HandshakeState): Promise<void> => {
     this.setPeerHost(peer, handshakeState.listenPort);
-   // peer.sendPacket(new GetOrdersPacket({}));
+    peer.sendPacket(new GetOrdersPacket({}));
   }
 
   private setPeerHost = async (peer: Peer, listenPort?: number): Promise<void> => {
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -92,6 +92,20 @@ class Pool extends EventEmitter {
     this.peers.add(peer);
   }
 
+  public disconnectPeer = async (address: string, port: number): Promise<void> => {
+    const socketAddress = new SocketAddress(address, port);
+    if (this.peers.has(socketAddress)) {
+      const peer = this.peers[socketAddress.toString()];
+      this.peers.remove(peer);
+      this.emit('peer.destroy', peer);
+      peer.destroy();
+    } else {
+      const err = errors.NOT_CONNECTED(address.toString());
+      this.logger.info(err.message);
+      throw err;
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


lib/p2p/Pool.ts
```diff
@@ -92,6 +92,20 @@ class Pool extends EventEmitter {
...
 
+  public disconnectPeer = async (address: string, port: number): Promise<void> => {
+    const socketAddress = new SocketAddress(address, port);
+    if (this.peers.has(socketAddress)) {
+      const peer = this.peers[socketAddress.toString()];
+      this.peers.remove(peer);
+      this.emit('peer.destroy', peer);
+      peer.destroy();
+    } else {
+      const err = errors.NOT_CONNECTED(address.toString());
```
moshababo: use `socketAddress` and not solely `address`

lib/p2p/Pool.ts
```diff
@@ -92,6 +92,20 @@ class Pool extends EventEmitter {
     this.peers.add(peer);
   }
 
+  public disconnectPeer = async (address: string, port: number): Promise<void> => {
+    const socketAddress = new SocketAddress(address, port);
+    if (this.peers.has(socketAddress)) {
+      const peer = this.peers[socketAddress.toString()];
```
moshababo: This line assignment doesn't return a peer. 

You need to implement `get` on `PeerList`, and after doing so you can avoid using `has`.

lib/p2p/Pool.ts
```diff
@@ -92,6 +92,20 @@ class Pool extends EventEmitter {
...
+      this.peers.remove(peer);
+      this.emit('peer.destroy', peer);
```
moshababo: what's the purpose of this event? it's not in usage.

lib/service/Service.ts
```diff
@@ -171,8 +171,9 @@ class Service extends EventEmitter {
   /*
    * Disconnect from a connected peer XU node on a given host and port.
    */
-  public disconnect = async (_request: { host: string, port: number}) => {
-    return 'Not implemented';
+  public disconnect = async ({ host, port }: { host: string, port: number}) => {
+    const peer = await this.pool.disconnectPeer(host, port);
```
moshababo: `peer` is not in usage

lib/p2p/Pool.ts
sangaman: My preferred approach would be to make `remove` accept a `SocketAddress` and return the `Peer` that it removed, or null if no peer with that address exists.

lib/service/Service.ts
sangaman: `disconnectPeer` resolves to void anyhow, which is good.

lib/p2p/Pool.ts
moshababo: I used solely `get` because after you do `peer.destroy`, it will trigger an event that will remove it from `peerList` anyway. 

But I implemented `remove` as you suggested, so we'll have it.

PR #206, opened 2018-07-22
Order Cancellation

itsbalamurali:
Implements #121 own order cancellation.

### DIFF
--- a/lib/cli/commands/cancelorder.ts
+++ b/lib/cli/commands/cancelorder.ts
@@ -2,22 +2,22 @@ import { callback, loadXudClient } from '../command';
 import { Arguments } from 'yargs';
 import { CancelOrderRequest } from '../../proto/xudrpc_pb';
 
-export const command = 'cancelorder <id> <pair_id>';
+export const command = 'cancelorder <order_id> <pair_id>';
 
 export const describe = 'cancel an order';
 
 export const builder = {
-  id: {
+  order_id: {
     type: 'string',
   },
-  pairId: {
+  pair_id: {
     type: 'string',
   },
 };
 
 export const handler = (argv: Arguments) => {
   const request = new CancelOrderRequest();
-  request.setId(argv.id);
-  request.setPairId(argv.pairId);
+  request.setOrderId(argv.order_id);
+  request.setPairId(argv.pair_id);
   loadXudClient(argv).cancelOrder(request, callback);
 };
--- a/lib/cli/commands/cancelorder.ts
+++ b/lib/cli/commands/cancelorder.ts
@@ -2,7 +2,7 @@ import { callback, loadXudClient } from '../command';
 import { Arguments } from 'yargs';
 import { CancelOrderRequest } from '../../proto/xudrpc_pb';
 
-export const command = 'cancelorder <id>';
+export const command = 'cancelorder <id> <pair_id>';
 
 export const describe = 'cancel an order';
 
@@ -10,10 +10,14 @@ export const builder = {
   id: {
     type: 'string',
   },
+  pairId: {
+    type: 'string',
+  },
 };
 
 export const handler = (argv: Arguments) => {
   const request = new CancelOrderRequest();
-  request.setId(argv.type);
+  request.setId(argv.id);
+  request.setPairId(argv.pairId);
   loadXudClient(argv).cancelOrder(request, callback);
 };
--- a/lib/orderbook/MatchingEngine.ts
+++ b/lib/orderbook/MatchingEngine.ts
@@ -7,6 +7,7 @@ import Logger from '../Logger';
 
 type PriorityQueue = {
   add: Function;
+  remove: Function;
   heapify: Function;
   peek: Function;
   poll: Function;
@@ -148,6 +149,18 @@ class MatchingEngine {
     }
 
     return matchingResult;
+  }
+
+  public removeOwnOrder = (orderId: string): void => {
+    this.priorityQueues.buyOrders.remove(orderId, MatchingEngine.removeQueuedComparator);
+    this.priorityQueues.sellOrders.remove(orderId, MatchingEngine.removeQueuedComparator);
+  }
+
+  private static removeQueuedComparator = (queuedOrder:any, orderID:string): Boolean => {
+    if (queuedOrder.id === orderID) {
+      return true;
+    }
+    return false;
   }
 }
 
--- a/lib/orderbook/MatchingEngine.ts
+++ b/lib/orderbook/MatchingEngine.ts
@@ -1,13 +1,14 @@
 import assert from 'assert';
 import FastPriorityQueue from 'fastpriorityqueue';
 
-import { orders, matchingEngine, db } from '../types';
+import { orders, matchingEngine } from '../types';
 import { OrderingDirection } from '../types/enums';
 import Logger from '../Logger';
 
 type PriorityQueue = {
   add: Function;
-  remove: Function;
+  removeOne: Function;
+  removeMany: Function;
   heapify: Function;
   peek: Function;
   poll: Function;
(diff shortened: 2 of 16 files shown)


moshababo:
1) After updating `fastpriorityqueue` version, my `package-lock.json` changed completely. which `npm` version are you using @sangaman @michael1011 ? I'm on 6.2.0
2) `fastpriorityqueue` seems to expect a weird comparator for the `remove` function. I implemented it so that the tests would pass, but it looks to me like a bug (FastPriorityQueue.js line 151). I'm pausing here so we'll decide whether to continue with it as is, report an issue, or fork the lib.

itsbalamurali:
@moshababo https://github.com/ExchangeUnion/xud/pull/192/commits/1581462d5987c3f74793e78132213262e2544898
my npm version: 6.1.0
i have written comparator for peer order removal pr

itsbalamurali:
i have updated the package.json but didn't push it

sangaman:
I'm on NPM 6.2 as well, I've been trying to keep it up to date and I figure we might as well while we're still pre-release.

Line 151 FastPriorityQueue looks like it will remove the item if the comparator returns false both when you compare A to B as well as B to A. That makes sense I think, we just need a comparator that will return false if the items are equal.

We might want to fork it for #192 to add a function to remove multiple items in a single iteration through the heap. Right now I think we'd need to call `remove()` for every item that needs to be removed. We can open a PR to the main repo afterwards.

kilrau:
If forking and ironing out the bug is the only way for now, let's do it. Would also open an issue if it's a confirmed bug.

The `remove` function seems only to be in 0.5.0 with this commit from April: https://github.com/lemire/FastPriorityQueue.js/commit/30ab7bd7fd91b305b6698d766e3e702da7b4b3b9#diff-2382578b0b3e69816d0d529481efdb14 

Could be it's gone unnoticed so far.

moshababo:
The comparator is to express priority, same like the default comparator. It means that the best way to use it is just to provide the item we want to remove, without any new comparator.
But since we only have the item id, it doesn't work well.

I forked the lib under ExchangeUnion, and made the refactoring (simple callback instead of a comparator, removed the double checking, removed the input val to rely on closure instead).
It seems to be working fine now (the test is passing). I don't think we need to report this, because it's not a bug, just a design which is less suitable for us. We can open a PR after we'll implement the batch remove. 

@sangaman so i'll leave the changes in `package-lock.json` as is?

moshababo:
I was about to write the missing code (`// TODO: remove from this.ownOrders/this.peerOrders as well`) to remove order from the list which is managed by `OrderBook` as well. But I figured we that we better finalize on our approach with that (following out discussion on gitter), instead of writing code that will probably change tomorrow.
I'm going to open a separate issue for that.

Our tests currently don't catch this issue. So I think we can merge it to master, with that being missing, and fix this issue as soon as possible.

lib/cli/commands/cancelorder.ts
```diff
@@ -2,18 +2,22 @@ import { callback, loadXudClient } from '../command';
 import { Arguments } from 'yargs';
 import { CancelOrderRequest } from '../../proto/xudrpc_pb';
 
-export const command = 'cancelorder <id>';
+export const command = 'cancelorder <id> <pair_id>';
 
 export const describe = 'cancel an order';
 
 export const builder = {
   id: {
```
sangaman: It might be helpful to call this `order_id` to further differentiate from `pair_id`.

lib/orderbook/OrderBook.ts
```diff
@@ -100,6 +100,16 @@ class OrderBook extends EventEmitter {
...
   }
 
+  public removeOrder = (orderId: string, pairId: string): boolean => {
+    const matchingEngine = this.matchingEngines[pairId];
+    if (!matchingEngine) {
+      throw errors.INVALID_PAIR_ID(pairId);
+    }
+
+    return matchingEngine.removeOrder(orderId);
+    // TODO: remove from this.ownOrders/this.peerOrders as well
```
sangaman: Why not implement this as part of this PR? Should be easy enough, to do something like `ownOrders[orderId] = undefined` right?

proto/xudrpc.proto
```diff
@@ -212,6 +212,8 @@ message PlaceOrderResponse {
 
 message CancelOrderRequest {
   string id = 1 [json_name = "order_id"];
+  string pair_id = 2 [json_name = "pair_id"];
+
```
sangaman: Drop this extra blank line.

lib/orderbook/OrderBook.ts
sangaman: Oh I didn't see your explanation @moshababo in your comment above. I still think it makes sense to put it in here because it can be done in one line (correct me if I'm wrong) but I'll leave it up to you.

moshababo: Yes it make sense, I implemented `removeOrder` function on `OrderBook`. The naming is a bit confusing, but we'll live with that until we do the refactoring.

moshababo:
I made the necessary changes. 

I also regenerated the protos, but i'm on protoc 3.6.0. Do you want to upgrade or should I downgrade?

sangaman:
I think best to downgrade protoc on your end for the time being so don't need to update the instructions to generate protos every time a new version comes out. I've been figuring maybe once version 4.x has been out for a while we can upgrade to that. If that's inconvenient for you I can always run it on my machine and check in only the generated files, then do a squash merge.

lib/orderbook/MatchingEngine.ts
```diff
@@ -149,6 +151,15 @@ class MatchingEngine {
 
     return matchingResult;
   }
+
+  public removeOwnOrder = (orderId: string): orders.StampedOwnOrder | null => {
```
sangaman: Maybe we should just call this `removeOrder` and have it accept `StampedOrder` so we can reuse this same method down the road when removing orders due to peer or swap events. The logic shouldn't change whether we're removing a peer order or own order since they're all in the same queue.

moshababo:
Yes, so please regenerate it on your machine, push it and do squash merge. I'll downgrade soon (just tried to, but i'm having some issue).

lib/orderbook/MatchingEngine.ts
```diff
@@ -149,6 +151,15 @@ class MatchingEngine {
...
+    return this.priorityQueues.buyOrders.removeOne(order => order.id === orderId) ||
```
sangaman: For performance reasons I'd guess it's faster to check whether the order is a buy or sell (check if quantity is positive) and then do removeOne on either `buyOrders` or `sellOrders`, but not both. Currently it would be possible for `removeOne` to execute twice if it's not a buy order, which means we iterate through all orders once more than necessary.

lib/orderbook/MatchingEngine.ts
moshababo: That's how it previously was, but I changed it because I figured out that soon we're gonna have to deal here with the separated lists (#215) as well, so not everything will be unified. 

In regards to having it accepting `StampedOrder`, see my other comment.

lib/orderbook/MatchingEngine.ts
moshababo: This information (`quantity`) isn't available in the current design.
It can be added as a parameter to the gRPC/CLI call, but I don't think it's a good idea. 
It's also possible to do a fast lookup on the separated lists on `OrderBook`, but this area is subject to change, so I don't want to introduce more stuff there. And in it's current design, the lists are not unified, so I need to check on both buy/sell lists to find the order. But yes, it will still be faster, and then the complete object will be available.

Is that what you meant when you suggested to accept `StampedOrder` as a parameter? that we'll retrieve it from the separated lists by `orderId`, and pass it to this function?

sangaman:
Yup working on it, I'm seeing if I can squash into separate commits to preserve authorship then merge. Or if not I'll just squash the whole thing through github.

PR #209, opened 2018-07-22
Added market orders

### LINKED ISSUE #105: Market Orders

michael1011:
Closes #105 
I had to recreate #188 because I unintentionally deleted the branch and closed the PR. Sorry for the hassle.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -96,18 +96,19 @@ class OrderBook extends EventEmitter {
     return this.addOwnOrder(order);
   }
 
-  public addMarketOrder = (_order: orders.OwnOrder) => {
-    // TODO: implement
+  public addMarketOrder = (order: orders.MarketOrder) => {
+    const price = order.quantity > 0 ? Number.MAX_VALUE : 0;
+    return this.addOwnOrder({ ...order, price }, true);
   }
 
-  private addOwnOrder = (order: orders.OwnOrder): matchingEngine.MatchingResult => {
+  private addOwnOrder = (order: orders.OwnOrder, discardRemaining: boolean = false): matchingEngine.MatchingResult => {
     const matchingEngine = this.matchingEngines[order.pairId];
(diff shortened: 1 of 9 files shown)


lib/orderbook/OrderBook.ts
```diff
@@ -96,18 +96,19 @@ class OrderBook extends EventEmitter {
...
 
-  public addMarketOrder = (_order: orders.OwnOrder) => {
-    // TODO: implement
+  public addMarketOrder = (order: orders.MarketOrder) => {
+    const price = order.quantity > 0 ? Number.MAX_VALUE : 0;
+    return this.addOwnOrder({ ...order, price }, true);
   }
 
-  private addOwnOrder = (order: orders.OwnOrder): matchingEngine.MatchingResult => {
+  private addOwnOrder = (order: orders.OwnOrder, discardRemaining: boolean = false): matchingEngine.MatchingResult => {
```
moshababo: if `discardRemaining` is `true`, don't broadcast or add to the orders list (line 120)

lib/service/Service.ts
```diff
@@ -148,9 +148,13 @@ class Service extends EventEmitter {
    * Add an order to the order book.
    */
   public placeOrder = async (order: OwnOrder) => {
-    assert(order.price > 0, 'price must be greater than 0');
+    assert(order.price >= 0, 'price must be greater than 0');
```
sangaman: Change the message here to match the updated logic, maybe `price cannot be negative`.

proto/xudrpc.proto
```diff
@@ -26,7 +26,9 @@ service Xud {
     };
   }
   
-  /* Add an order to the order book. */
+  /* Add an order to the order book.
+   * If no price is specified a market order will get added
```
sangaman: Throw a period at the end here to be consistent with the other comments. Also I think it might make sense to say "If price is zero or unspecified a market order will get added" since explicitly setting a zero value for price will have the same effect as not setting it at all.

Issue #157, opened 2018-06-27
Verify data integrity 


moshababo:
Add checksum/hash field to the packet header.
On receiving, verify the packet data integrity.

moshababo:
1st is done.
All the rest can be done separately.

kilrau:
Updated scope of this issue:
- On receiving, verify the packet data integrity via the checksum/hash field.
- Receive data structure of packet is what we expect
- rate limit for certain packet types like `GET_ORDERS` per `nodePubKey`

kilrau:
Do we still need this when we do (packet serialization)[https://github.com/ExchangeUnion/xud/issues/159#issuecomment-435583439]? @moshababo @sangaman

kilrau:
Moved to next milestone, not too critical

moshababo:
@kilrau we won't need to do most of this if we'll change to protobuf serialization (#159). but it's not a trivial change.

kilrau:
https://github.com/ExchangeUnion/xud/issues/159 is a requirement

moshababo:
Lets limit the scope of this issue solely to:

* On receiving, verify the packet data integrity via the checksum/hash field.

In regards to the others:

* Receive data structure of packet is what we expect - 

#159 will make this unnecessary. 

* rate limit for certain packet types like GET_ORDERS per nodePubKey - 

I don't think it's related to here. I've opened #708.

moshababo:
Waiting for #159.

PR #210, opened 2018-07-23
Allow empty packets, packet ids, type in header

### LINKED ISSUE #208: ORDERS packet doesn't contain reqHash
ORDERS packet currently isn't associated with a previous request, so it triggers an unsolicited response packet error.

This should be handled same as ping/pong: `handlePing` gets the packet `hash`, and add it to `pong` as the `reqHash`.

sangaman:
Closes #208.

This commit creates a `GENERIC_REQUEST` packet direction which does not use a body and instead specifies a response packet type. Sending a generic request packet solicits any packet of the corresponding response packet type. This also refactors the `GET_HOSTS` and `GET_ORDERS` packets as generic requests.

An approach like this was discussed in #204. I'm open to feedback but I thought that it would be good to have a working example for further discussion. I've tested successfully between two separate instances running on my machine.


sangaman:
> I think i'd rather keep the linkage.

You mean the linkage to an identifier?

> Also, requests might have optional parameters, but we're defining PacketDirection in a static manner (it's not being serialized). This might create an issue.

I don't think I follow. If all peers are following the same protocol with the same `PacketDirection`s and `PacketType`s, shouldn't it be fine?

> I think that every packet should have a unique identifier (from the sender perspective). Relying merely on hashing won't work, because the packet might lack uniqueness (not only because the body is empty). We can introduce uniqueness to the body, but I think that supporting packets without a body is a good idea. If we'll add the uniqueness to the header, the hashing would not be straight-forward.
> A solution can be to separate between the id and the checksum hash. We can add a uuid as an id field to the header, and if the packet has a body, add hash as well. the response linkage would be by the id.

An identifier might be useful I'm thinking as well - and you're right about the hash not always being sufficient (particularly for bodyless packets). But I think a uuid might be overkill actually. How about tracking a packet count for each peer, using that as the id for the next packet and incrementing each time we send a packet? Nodes responding to a request packet will include its id in the response packet. The Hello packet can always be id 0 and we always expect a packet of id 0. That guarantees packet uniqueness for a particular socket and we don't care about global uniqueness.  A running count should need fewer cycles and less data transmission than a uuid.

I think that might make sense as part of this PR actually, as I'd have to rework some of the code.

As a minor bonus, this might also be a useful statistic to have about a socket connection, total packets sent.

> Another change I wanted to do is to include the packet type inside of the header. currently it's being serialized separately. It would be better if each serialized packet would be one complete JSON, so that the header would need to be deserialized first in order to determine the packet type.

I like this idea. I can implement this in a separate PR unless you want to work on it. I can also open an issue for it if you'd like.

moshababo:
> You mean the linkage to an identifier?

Yes, that the response could be linked to a specific request, not to a packet type. This feature could be useful only if we want new request packets from the same type to invalidate the previous ones, but I don't think this should be the default behavior.

> I don't think I follow. If all peers are following the same protocol with the same PacketDirections and PacketTypes, shouldn't it be fine?

Yes. What I meant and wasn't so clear, is that a packet might have optional parameters, and so an optional body, but it's still being defined as `GENERIC_REQUEST` in the protocol level. That's not creating an issue right now, because it's being used only for the request/response mechanism on the sender side, but it's less of an optimal design IMO. 

> An identifier might be useful I'm thinking as well - and you're right about the hash not always being sufficient (particularly for bodyless packets). But I think a uuid might be overkill actually. How about tracking a packet count for each peer, using that as the id for the next packet and incrementing each time we send a packet? Nodes responding to a request packet will include its id in the response packet. The Hello packet can always be id 0 and we always expect a packet of id 0. That guarantees packet uniqueness for a particular socket and we don't care about global uniqueness. A running count should need fewer cycles and less data transmission than a uuid.
I think that might make sense as part of this PR actually, as I'd have to rework some of the code.
As a minor bonus, this might also be a useful statistic to have about a socket connection, total packets sent.

Sounds like a good idea, if indeed we won't end up needing something which is globally unique, like if we'll need to support a relay network. But for the current requirements what you suggested is suitable.
This should replace the adding of `GENERIC_REQUEST` type. In regards to `responseType` - if we'll link only by the id, we don't need it. But we can leave it there for better code readability, and if we'll want to verify that the linked response is from a specific type. 

> I like this idea. I can implement this in a separate PR unless you want to work on it. I can also open an issue for it if you'd like.

We can leave this one for a later PR, feel free to open an issue and work on it if you're getting to it. And if you can already start doing so, it can be done in this PR as well.

sangaman:
Thanks for the explanation. I just went ahead and took care of everything in this PR. Here's the summary I put in the commit message:

> This commit accomplishes three changes to the p2p protocol and packet structure.
> 
> - The body for packets is now optional, and packets without bodies will not have hashes in their headers.
> - An id is added to the packet header to uniquely identify a packet for a particular socket. Packets that are specific to a particular peer use an id starting from 0 and increment upwards for each packet transmitted. Packets that are broadcast to multiple peers use an id starting from -1 and increment downwards. The hash is now intedned only for error checking and not identification.
> - The packet type is now part of the header and is not transmitted separately, so packets now consist entirely of a single stringified JSON object.

I dropped the `responseType` parameter because I felt it wasn't necessary, especially if we continue with a good request/response naming pattern like GET_HOSTS/HOSTS. Using it to verify that the type of a response matches what we expect didn't seem sufficient either, since a malicious peer could always fake the `type` value in the header and have the body be a total mismatch. That's something we might want to address down the road though.

I also had to tweak my id approach somewhat. I realized that for broadcasting a single packet across multiple sockets, we'd need to call the constructor once for each socket with a different id to ensure uniqueness for that socket. That seemed awkward and inefficient, so the solution I came up with is to track packet count in two places. On `Peer` for packets that are specific to that peer, this starts from 0 and increments up. On `Pool` for packets that get broadcast to multiple peers, this starts from -1 and increments down.  This way something like an `Order` packet can only be created once, and then broadcast to all peers.

This set up seemed to work well in some basic local testing, here's what the output looks like:

```shell
2018-7-26 11:27:24 [P2P] debug: Connected to peer (127.0.0.1:55809)
2018-7-26 11:27:24 [P2P] debug: Received data (127.0.0.1:55809): {"header":{"id":0,"hash":"CX8h7aZdVS4/daHcHzyvkw==","type":"HELLO"},"body":{"version":"123","nodePubKey":"123","listenPort":20000,"pairs":["BTC/LTC"]}}
{"header":{"id":1,"type":"GET_ORDERS"}}
{"header":{"id":2,"type":"GET_HOSTS"}}

2018-7-26 11:27:24 [P2P] debug: Received data (127.0.0.1:55809): {"header":{"reqId":2,"id":3,"hash":"11FxOYiYfpMxmANj4kGJzg==","type":"HOSTS"},"body":[]}

2018-7-26 11:27:24 [P2P] debug: Received data (127.0.0.1:55809): {"header":{"reqId":1,"id":4,"hash":"11FxOYiYfpMxmANj4kGJzg==","type":"ORDERS"},"body":[]}

2018-7-26 11:27:54 [P2P] debug: Received data (127.0.0.1:55809): {"header":{"id":5,"hash":"7umJHKv8n1Vl6LksBRfCFg==","type":"PING"},"body":{"ts":1532618874079}}

2018-7-26 11:27:54 [P2P] debug: Received data (127.0.0.1:55809): {"header":{"reqId":6,"id":6,"hash":"Re5tpSgZLtyJwaQxJ96D1g==","type":"PONG"},"body":{"ts":1532618874086}}
```

moshababo:
Awesome.

We might end up not being able to use a single packet instance for multiple peers (multiple invoices, encryption against the peer pubKey, truly globally unique packet id requirement, etc.). That's not the case now, so there's no reason not to use the "broadcasting" approach.

But I think that the `negativePacketCount` / `packetCount` is a bit weird though, although it works, and solves the problem of `id` uniqueness, while not requiring a uuid. 

Other options:
* simply use a uuid, and count outgoing packets to each peer, if we want that statistics.
* when broadcasting, have each peer receive a packet instance, and create a new instance out of it, which will differ only by `id` (so the "clone" will be shallow).

lib/p2p/Peer.ts
```diff
@@ -451,34 +475,24 @@ class Peer extends EventEmitter {
     return packet;
   }
 
-  private handleHello = (packet: HelloPacket): void => {
+  private handleHello = (packet: packets.HelloPacket): void => {
     this.handshakeState = { ...this.handshakeState, ...packet.body };
-
-    if (this.responseMap.has(packet.type)) {
-      this.fulfillResponseEntry(packet.type, packet);
-    }
```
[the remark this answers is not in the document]
sangaman: I don't think this reintroduces the bug, although I need to update comments and code to make it more clear. Line 254 effectively waits for a packet with id 0, which is the hello packet, so whenever it receives that packet it will be recognized.

sangaman:
@moshababo I was trying to make the second option work at first, but it felt a little off trying to make `id` an optional parameter that's not set on the constructor. I just revisited it though and I think I should be able to make it work.

A third option I'd consider is using uuid for the "broadcast" or multicast packets and the regular incrementing id for everything else. Thoughts or any preference?

lib/p2p/Peer.ts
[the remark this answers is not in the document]
moshababo: I pushed a fix to an old bug I apparently left, which now makes the `Hello` bug become apparent.

HelloPacket direction is `UNILATERAL`, so when handling the packet, you don't try to fulfill the response entry. You also don't attach `reqId` to it. that's why `wait` doesn't resolve.

moshababo:
For now I would prefer the simplest option, to use uuid for all packets. We shouldn't over optimize at this stage.

sangaman:
I went ahead with the uuid for broadcast packets but I didn't want throw away the code I already had for sequential id for packets specific to a peer. I also added the types package for `uuid`.

sangaman:
One thing I wanted to ask, is there a reason `handshakeState` starts off as an empty object rather than undefined? I'm thinking it would make more sense to start off as undefined and make the `HandshakeState` type have non-nullable fields. But it can be a separate PR if that makes sense to you.

lib/p2p/packets/packetUtils.ts
```diff
@@ -1,30 +1,52 @@
...
+        case PacketType.GET_ORDERS:
+          return new packetTypes.GetOrdersPacket(packet);
+        case PacketType.ORDERS:
+          return new packetTypes.OrdersPacket(packet);
+        case PacketType.GET_HOSTS:
+          return new packetTypes.GetHostsPacket(packet);
+        case PacketType.HOSTS:
+          return new packetTypes.HostsPacket(packet);
+        default:
+          throw new ParserError(ParserErrorType.UNKNOWN_PACKET_TYPE, packet.header.type!);
```
moshababo: this will be catch and by the outer `try`/`catch`, so the error type will be misleading.

lib/p2p/packets/packetUtils.ts
```diff
@@ -1,30 +1,52 @@
...
+      }
+    } else {
+      throw new ParserError(ParserErrorType.INVALID_MESSAGE, `${raw}`);
```
moshababo: same as above.

lib/p2p/packets/packetUtils.ts
```diff
@@ -1,30 +1,52 @@
...
+  UNPARSABLE_MESSAGE,
+  INVALID_MESSAGE,
+  UNKNOWN_PACKET_TYPE,
+}
+
+const fromRaw = (raw: string): Packet => {
+  try {
+    const json = JSON.parse(raw);
+    // check that we have the required fields for an incoming packet
+    if (typeof json.header === 'object' && typeof json.header.type === 'string' && typeof json.header.id === 'number') {
```
moshababo: this fail because `id` now is now a string

moshababo:
I originally made `handshakeState` starts as an empty object and the type to have nullable fields so we can continuously append new states into it. 

At the moment I think we won't want to support overriding this state, because it will lead to some difficulties with other features, which depends on it. So if a peer wants to update this state (not a very common scenario), he will need to establish a new connection. 

I agree that it would be better to start it as `undefined`, regardless to whether we want to support overriding or not. Just make sure to fix `handleHello` function as well.
In regards nullable fields - it also depends if the entire state is from `HelloPacket` or not.

moshababo:
I don't see any changes (although I recreated the local branch).

Another thing: the parsing code I commented on is inside of `packetUtils` now. It should return to `Parser`.

sangaman:
Oops, I'd forgotten to push my changes.

sangaman:
Rebased, take another look when you get a chance @moshababo.

lib/p2p/Peer.ts
```diff
@@ -122,10 +131,12 @@ class Peer extends EventEmitter {
...
 
-    const handshakeState: HandshakeState = {};
-
     // let the pool know that this peer is ready to go
-    this.emit('open', handshakeState);
+    this.emit('open', this.handshakeState!);
+
+    // request peer's orders and known hosts
+    this.sendPacket(new packets.GetOrdersPacket());
+    this.sendPacket(new packets.GetHostsPacket());
```
moshababo: putting this here is a bit misleading, because it's post-handshake. why not in `Pool.handleOpen`? That's where these stuff should be.

lib/p2p/Peer.ts
```diff
@@ -262,10 +272,9 @@ class Peer extends EventEmitter {
     const packet = this.sendHello();
 
     if (!this.handshakeState) {
-      // using the packet type as the response key,
-      // to allow waiting for any incoming HelloPacket
-      await this.wait(packet.type, Peer.RESPONSE_TIMEOUT);
-      assert(this.handshakeState);
+      // wait for an incoming HelloPacket
+      await this.wait(PacketType.HELLO, Peer.RESPONSE_TIMEOUT);
+      assert(isHandshakeState(this.handshakeState));
```
moshababo: I figured you added `isHandshakeState` to validate the runtime object, but with this line I can crash other xud's by sending them `Hello` packet with a partial body. `assert` should be used to enforce integrity on scenarios that should never happen (and if they do, xud should crash).

lib/p2p/Peer.ts
```diff
@@ -3,24 +3,31 @@ import net, { Socket } from 'net';
...
 import Host from './Host';
 import SocketAddress from './SocketAddress';
-import Parser, { ParserError, ParserErrorType } from './Parser';
-import { Packet, PacketDirection, PacketType, HelloPacket, PingPacket, PongPacket, HostsPacket, OrdersPacket } from './packets';
+import Parser from './Parser';
+import * as packets from './packets/types';
 import Logger from '../Logger';
-import { ms } from '../utils/utils';
-import { orders } from '../types';
+import { ms, isEmptyObject } from '../utils/utils';
```
moshababo: unused import

lib/p2p/Peer.ts
```diff
@@ -474,34 +498,34 @@ class Peer extends EventEmitter {
...
+      entry.resolve(packet);
+
+      this.handshakeState = packet.body;
     }
   }
 
-  private sendPing = (): PingPacket => {
-    const packet = new PingPacket({ ts: ms() });
+  private sendPing = (): packets.PingPacket => {
+    const packet = new packets.PingPacket({ ts: ms() });
```
moshababo: the packet body can be removed. no need for it anymore

lib/p2p/Peer.ts
```diff
@@ -474,34 +498,34 @@ class Peer extends EventEmitter {
...
-
-    this.sendPong(hash);
+  private handlePing = (packet: packets.PingPacket): void  => {
+    this.sendPong(packet.header.id);
   }
 
-  private sendPong = (pingHash: string): PongPacket => {
-    const packet = new PongPacket({ ts: ms() }, pingHash);
+  private sendPong = (pingId: string): packets.PongPacket => {
+    const packet = new packets.PongPacket({ ts: ms() }, pingId);
```
moshababo: the packet body can be removed. no need for it anymore

lib/p2p/packets/packetUtils.ts
```diff
@@ -1,32 +1,56 @@
...
+  constructor(public type: ParserErrorType, public payload: string) { }
+}
+
+enum ParserErrorType {
+  UNPARSABLE_MESSAGE,
+  INVALID_MESSAGE,
+  UNKNOWN_PACKET_TYPE,
+}
+
+const fromRaw = (raw: string): Packet => {
```
moshababo: This function need to change, because now we parse the packet in order to determine the type. 
Putting the parsing code here, including the error types, isn't a good solution IMO. It should return to `Parser`.

lib/p2p/Peer.ts
```diff
@@ -474,34 +498,34 @@ class Peer extends EventEmitter {
...
+    const entry = this.responseMap.get(PacketType.HELLO);
 
-    if (this.responseMap.has(packet.type)) {
-      this.fulfillResponseEntry(packet.type, packet);
+    if (!entry) {
+      this.logger.debug(`Peer (${this.id}) sent an unsolicited Hello packet`);
+      // TODO: penalize
+    } else {
+      this.responseMap.delete(PacketType.HELLO);
+      entry.resolve(packet);
```
moshababo: We have a bit of code duplication here, but that's fine if you want to close `fulfillResponseEntry` to the usual use-case.

sangaman: Yeah it was intentional, I think it's different enough and it's clearly a special case.

lib/p2p/Peer.ts
sangaman: Good point, although technically I think a peer could crash xud by sending an empty packet body even with the previous `assert(this.handshakeState)` line, right? My thinking was this line was to verify that we received the fields we expected. I can definitely change it back, but we will probably want some explicit checking on the body of received packets, discarding malformed packets and penalizing the peer. That can be a separate PR though. I can create an issue if it makes sense to you.

lib/p2p/Peer.ts
sangaman: I'll move it back.

sangaman:
Rebased and addressed the latest round of feedback. I wound up removing the `packetUtils` class because after moving the `fromRaw` method and error types to `Parser`, there was nothing left.

moshababo:
In regards to your [comment](https://github.com/ExchangeUnion/xud/pull/210#discussion_r206712371):
That line was actually to verify that `handleHello`, at it's previous implementation, was executed. So it was actually just checking our code. 

At the moment the new version is still vulnerable, because the body can be empty (even by TS definition).  Changing line 508 on `Pool.ts` to the following could fix it:
````
this.handshakeState = { ...packet.body }
````
It still doesn't solve the parsing problem in the greater sense, that is specifically problematic because we're using JSON serialization + JS runtime, which means that anything can go into these objects, unless we validate them. We should customize a solution only if we plan to settle on these two approaches.

Feel free to open a separate issue for that, and merge the PR after adding the small fix I suggested. Thanks!

sangaman:
Done. And I'll open an issue to validate received packets, it's definitely something we'd need in the wild. Merging, thanks!

moshababo:
Just because we'll probably tend to forget that, I think it's better to use casting, or just remove the assertion line (274) instead.

sangaman:
I'll remove the assertion, we'll validate the Hello packet properly for the issue I'm about to create. Thanks.

sangaman:
Actually the issue I was going to create is already pretty much covered by #157, I'll add details there.

PR #212, opened 2018-07-23
Define peer events interface

sangaman:
This defines types of the `.on()`, `.once()`, and `.emit()` `EventEmitter` functions for the `Peer` class. If this looks good, I can use the same approach to our other classes that extend `EventEmitter`.

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -23,6 +23,20 @@ type HandshakeState = {
   pairs?: string[];
 };
 
+declare interface Peer {
+  on(event: 'packet', listener: (packet: Packet) => void): this;
+  on(event: 'error', listener: (err: Error) => void): this;
+  once(event: 'open', listener: (handshakeState: HandshakeState) => void);
+  once(event: 'close', listener: () => void);
+  once(event: 'ban', listener: () => void);
+  emit(event: 'connect');
+  emit(event: 'ban');
+  emit(event: 'open', handshakeState: HandshakeState);
+  emit(event: 'close');
+  emit(event: 'error', err: Error);
+  emit(event: 'packet', packet: Packet);
+}
+
 /** Represents a remote XU peer */
 class Peer extends EventEmitter {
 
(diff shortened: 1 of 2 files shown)


lib/p2p/Peer.ts
```diff
@@ -23,6 +23,23 @@ type HandshakeState = {
...
+  once(event: 'close', listener: () => void);
+  once(event: 'ban', listener: () => void);
+  once(event: string, listener: Function): this;
+  emit(event: 'connect');
+  emit(event: 'ban');
+  emit(event: 'open', handshakeState: HandshakeState);
+  emit(event: 'close');
+  emit(event: 'error', err: Error);
+  emit(event: 'packet', packet: Packet);
+  emit(event: string, ...any);
```
moshababo: what's this one for?

lib/p2p/Pool.ts
```diff
@@ -215,7 +215,7 @@ class Pool extends EventEmitter {
       this.logger.error('peer error', err);
     });
 
-    peer.once('open', (handshakeState: HandshakeState) => {
```
moshababo: why removing the type?

lib/p2p/Peer.ts
sangaman: I debated leaving that in - maybe it should be better if left out. This line (and line 29 and 33) allows for generic events without specific types. Basically this says you can `emit` anything where the first parameter is a string plus as many additional parameters of any type. Dropping this line would require anyone adding a new event type to come here and add new line defining the types it will use - and that might be a good thing after all. Let me know what you think.

lib/p2p/Pool.ts
sangaman: The type is already explicitly defined in Peer.ts:30, so I thought restating it here is unnecessary. The compiler will already know that `handshakeState` is of type `HandshakeState`. The advantage is we don't need to explicitly define the type everywhere we call `peer.once('open'...)`, it will already be defined for us.

lib/p2p/Peer.ts
moshababo: Yes I think it would be preferred, because if it's not enforced, we might forget adding a new event type.

sangaman: Agreed, removing them now.

moshababo: You can merge after settling this (either dropping these lines or not).

Issue #176, opened 2018-06-27
Split order ids


moshababo:
Each order can be split, locally or remotely, so representing an order merely by its original `orderId` isn't very effective.
One approach to tackle this would be to generate a new `orderId` for each order split, and to associate them all with the original `orderId`. 
Another approach would be to keep one single `orderId`, but to use the `orderId` + `quantity` tuple for every order-related action (cancellation, swap update, etc.). So for example, if I want to cancel an order, and to invalidate it over the network, i'll do so by specifying the quantity as well, so that in practice each order will be treated as a max sum of potential quantity, which can only be decreased over time. This approach seems more suitable to me, but we need to think this through together, in case I overlooked something.

kilrau:
I thought about it for a bit and would go for the orderID+quantity tuple. This is how it could look like
``
1. placeOrder(internalOrderID, 1)
uuid() = 2494843749, 1 BTC, maker order
save (internalOrderID, 2494843749)
broadcast(2494843749, 1, BTC, 100, LTC)

2. 4023482337 10 LTC for 0.1 BTC coming in as taker, successfully fills 2494843749 partly with 0.1 BTC:
updateLocally(2494843749, 0.9, BTC) and broadcastUpdate(2494843749, 0.9, BTC)
saveToTradeHistory(4023482374, 2494843749, 0.1, BTC, 10, LTC)
``

kilrau:
On a different topic we came accross this again with @sangaman and it seems we don't have a consensus to keep the same orderID and just decrease the quantity on a partial fill. Please both give opinions with detailed reasoning here so we can close this asap @moshababo @sangaman

sangaman:
Ah yeah sorry I hadn't seen this discussion. In gitter I was against the idea of using the Id+quantity tuple to uniquely identify orders. It would complicate such things, for example our orderbook right now is a map of Id to order. Technically it's possible to use Id+qty as the key, but it would be more complicated. Also I can't imagine we'd want to allow for two orders with the same Id but different quantities in our books.

But I think what @moshababo is saying is that the Id is still the sole identifier, but that order-related actions would always require a quantity along with the ID. So when you cancel an order or fill an order, you must specify the amount being filled or canceled. I think that's a good approach. We keep the same Id until an order is completely filled or canceled, and each time the available quantity for an order is reduced we update our peers.

kilrau:
Ok, good to be implemented then @michael1011 . Updated comment from above:

[...]

Issue #150, opened 2018-06-27
p2p packet: order_invalidation


moshababo:
When `cancelOrder` is called on the high-level API, this packet should be sent to all peers. 
When it's being received, it should invoke `removePeerOrder` in the order book.

moshababo:
Yes.

PR #214, opened 2018-07-24
Split order IDs

### LINKED ISSUE #175: Local/Global orderId
Add `orderId` property to the high-level API (`placeOrder`, `cancelOrder`, etc.).
 Inside of `xud` it should be represented as `clientOrderId`, while attaching a globally unique `orderId` to it. The mapping between them should be managed by `xud`, so that the client won't be aware of that.

`peerOrders` should be exposed to the client with their globally unique `orderId`, and not with their `clientOrderId`. (@kilrau)

michael1011:
Closes #175 

I thought `internalId` (because it is doesn't leave the exchange) is more suitable than `clientOrderId`. Thoughts?

### DIFF
--- a/lib/cli/commands/placeorder.ts
+++ b/lib/cli/commands/placeorder.ts
@@ -2,7 +2,7 @@ import { callback, loadXudClient } from '../command';
 import { Arguments } from 'yargs';
 import { PlaceOrderRequest, Order } from '../../proto/xudrpc_pb';
 
-export const command = 'placeorder <pair_id> <order_id> <price> [quantity]';
+export const command = 'placeorder <pair_id> <order_id> <quantity> [price]';
 
 export const describe = 'place an order';
 
--- a/lib/cli/commands/placeorder.ts
+++ b/lib/cli/commands/placeorder.ts
@@ -2,19 +2,22 @@ import { callback, loadXudClient } from '../command';
 import { Arguments } from 'yargs';
 import { PlaceOrderRequest, Order } from '../../proto/xudrpc_pb';
 
-export const command = 'placeorder <pair_id> <quantity> [price]';
+export const command = 'placeorder <pair_id> <order_id> <price> [quantity]';
 
 export const describe = 'place an order';
 
 export const builder = {
+  pair_id: {
+    type: 'string',
+  },
+  order_id: {
+    type: 'string',
+  },
   price: {
     type: 'number',
   },
   quantity: {
     type: 'number',
-  },
-  pair_id: {
-    type: 'string',
   },
 };
 
@@ -22,9 +25,10 @@ export const handler = (argv: Arguments) => {
   const request = new PlaceOrderRequest();
 
   const order = new Order();
-  order.setPrice(argv.price);
-  order.setQuantity(argv.quantity);
   order.setPairId(argv.pair_id);
+  order.setLocalId(argv.order_id);
+  order.setQuantity(argv.quantity);
+  order.setPrice(argv.price);
   request.setOrder(order);
 
   loadXudClient(argv).placeOrder(request, callback);
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -28,6 +28,8 @@ class OrderBook extends EventEmitter {
   private ownOrders: { [ pairId: string ]: Orders } = {};
   private peerOrders: { [ pairId: string ]: Orders } = {};
 
+  private localIds: { [ localId: string ]: string } = {};
+
   constructor(models: Models, private pool?: Pool, private lndClient?: LndClient) {
     super();
 
@@ -65,17 +67,17 @@ class OrderBook extends EventEmitter {
   }
 
   /**
-   * Returns lists of buy and sell orders of peers sorted by price.
+   * Returns lists of buy and sell orders of peers
    */
-  public getPeerOrders = (pairId: string, maxResults: number): { [type: string]: orders.StampedOrder[] } => {
-    return this.getOrders(maxResults, this.peerOrders[pairId]);
+  public getPeerOrders = (pairId: string, maxResults: number): { [type: string]: orders.StampedPeerOrder[] } => {
+    return this.getOrders(maxResults, this.peerOrders[pairId]) as { [type: string]: orders.StampedPeerOrder[] };
   }
 
   /*
-  * Returns lists of the node's own buy and sell orders sorted by price.
+  * Returns lists of the node's own buy and sell orders
   */
-  public getOwnOrders = (pairId: string, maxResults: number): { [type: string]: orders.StampedOrder[] } => {
-    return this.getOrders(maxResults, this.ownOrders[pairId]);
+  public getOwnOrders = (pairId: string, maxResults: number): { [type: string]: orders.StampedOwnOrder[] } => {
+    return this.getOrders(maxResults, this.ownOrders[pairId]) as { [type: string]: orders.StampedOwnOrder[] };
   }
 
   private getOrders = (maxResults: number, orders: Orders): { [ type: string ]: orders.StampedOrder[]} => {
@@ -96,19 +98,26 @@ class OrderBook extends EventEmitter {
     return this.addOwnOrder(order);
   }
 
-  public addMarketOrder = (order: orders.MarketOrder) => {
+  public addMarketOrder = (order: orders.OwnMarketOrder): matchingEngine.MatchingResult => {
     const price = order.quantity > 0 ? Number.MAX_VALUE : 0;
     return this.addOwnOrder({ ...order, price }, true);
   }
 
-  public removeOwnOrder = (orderId: string, pairId: string): boolean => {
+  public removeOwnOrder = (pairId: string, orderId: string): boolean => {
     const matchingEngine = this.matchingEngines[pairId];
     if (!matchingEngine) {
       throw errors.INVALID_PAIR_ID(pairId);
     }
 
-    if (matchingEngine.removeOwnOrder(orderId)) {
-      return this.removeOrder(this.ownOrders, orderId, pairId);
+    let id = orderId;
+
+    if (this.localIds[orderId]) {
+      id = this.localIds[orderId];
+      delete this.localIds[orderId];
+    }
+
+    if (matchingEngine.removeOwnOrder(id)) {
+      return this.removeOrder(this.ownOrders, id, pairId);
     } else {
       return false;
     }
@@ -134,7 +143,6 @@ class OrderBook extends EventEmitter {
       this.broadcastOrder(remainingOrder);
       this.addOrder(this.ownOrders, remainingOrder);
       this.logger.debug(`order added: ${JSON.stringify(remainingOrder)}`);
-
     }
 
     return matchingResult;
@@ -155,19 +163,26 @@ class OrderBook extends EventEmitter {
   }
 
   private updateOrderQuantity = (type: { [pairId: string]: Orders }, order: orders.StampedOrder, decreasedQuantity: number) => {
-    const object = this.getOrderMap(type, order);
-    object[order.id].quantity = object[order.id].quantity - decreasedQuantity;
-    if (object[order.id].quantity === 0) {
-      delete object[order.id];
+    const orderMap = this.getOrderMap(type, order);
+    const id = this.localIds[order.id] ? this.localIds[order.id] : order.id;
+
+    orderMap[id].quantity = orderMap[id].quantity - decreasedQuantity;
+    if (orderMap[id].quantity === 0) {
+      delete orderMap[id];
     }
   }
 
   private addOrder = (type: { [pairId: string]: Orders }, order: orders.StampedOrder) => {
+    if (order['localId']) {
+      this.localIds[order['localId']] = order.id;
+    }
+
     this.getOrderMap(type, order)[order.id] = order;
   }
 
   private removeOrder = (type: { [pairId: string]: Orders }, orderId: string, pairId: string): boolean => {
     const orders = type[pairId];
+
     if (orders.buyOrders[orderId]) {
       delete orders.buyOrders[orderId];
       return true;
@@ -175,6 +190,7 @@ class OrderBook extends EventEmitter {
       delete orders.sellOrders[orderId];
       return true;
     }
+
     return false;
   }
 
@@ -217,7 +233,7 @@ class OrderBook extends EventEmitter {
 
     if (!invoice) return;
 
-    const { createdAt, ...outgoingOrder } = { ...order, invoice };
+    const { createdAt, localId, ...outgoingOrder } = { ...order, invoice };
     return outgoingOrder;
   }
 
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -13,8 +13,8 @@ import Peer from '../p2p/Peer';
 import { Models } from '../db/DB';
 
 type Orders = {
-  buyOrders: { [ id: string ]: orders.StampedOrder };
-  sellOrders: { [ id: string ]: orders.StampedOrder };
+  buyOrders: Map<String, orders.StampedOrder>;
+  sellOrders: Map<String, orders.StampedOrder>;
 };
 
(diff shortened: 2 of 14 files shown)


lib/cli/commands/placeorder.ts
```diff
@@ -2,29 +2,33 @@ import { callback, loadXudClient } from '../command';
 import { Arguments } from 'yargs';
 import { PlaceOrderRequest, Order } from '../../proto/xudrpc_pb';
 
-export const command = 'placeorder <pair_id> <price> <quantity>';
+export const command = 'placeorder <pair_id> <id> <price> <quantity>';
```
sangaman: Maybe let's change `id` to `order_id` to make it more clear here.

lib/orderbook/OrderBook.ts
```diff
@@ -117,10 +117,9 @@ class OrderBook extends EventEmitter {
       });
     }
     if (remainingOrder) {
-      this.broadcastOrder(remainingOrder);
+      this.broadcastOrder(remainingOrder as orders.StampedOwnOrder);
```
sangaman: Instead of casting `remainingOrder` here, I think it makes more sense to change the `MatchingResult` type to use `StampedOwnOrder`. We can also change the `match` function to only accept own orders, since we shouldn't be matching peer orders.

lib/orderbook/OrderBook.ts
```diff
@@ -141,15 +140,22 @@ class OrderBook extends EventEmitter {
...
     }
   }
 
   private addOrder = (type: { [pairId: string]: Orders }, order: orders.StampedOrder) => {
-    this.getOrderMap(type, order)[order.id] = order;
+    const identifier = this.getIdentifier(order);
+    this.getOrderMap(type, order)[order[identifier]] = order;
+  }
+
+  private getIdentifier = (order: orders.StampedOrder): string => {
```
sangaman: I'm unsure about this `getIdentifier` method. I think it's simpler to only use `id` as the key for referencing orders. Mainly because peers who try to fill our orders will not know the internal id, only the global id, and will reference orders using the global id, meaning we need to track some mapping of internal id to global id anyhow. So my preferred approach would be to maintain this mapping and do a lookup of id whenever we are given an id from the API layer. 

If we do decide to keep this, it should probably be `static` or moved out of the `Orderbook` class, maybe even in a utils class because we might need to use it elsewhere down the road.

lib/orderbook/OrderBook.ts
```diff
@@ -141,15 +140,22 @@ class OrderBook extends EventEmitter {
...
+    return ('internalId' in order) ? 'internalId' : 'id';
```
sangaman: Also, if we do keep this method I think it should return the identifier itself, not the name of the identifier.

lib/orderbook/OrderBook.ts
michael1011: I completely forgot about that. Having a map with the `localId` as key and `id` as value makes most sense you are right.

sangaman:
Well actually I get the exact the same generated files as you do when I checkout this branch and do `npm run proto`. Maybe they were mistakenly changed in another PR  using a different version of protoc without me noticing. So let's just stick with the changes as is.

lib/orderbook/OrderBook.ts
```diff
@@ -28,6 +28,8 @@ class OrderBook extends EventEmitter {
   private ownOrders: { [ pairId: string ]: Orders } = {};
   private peerOrders: { [ pairId: string ]: Orders } = {};
 
+  private localIds: { [ localId: string ]: string } = {};
```
sangaman: I'd rather call this `localIdMap` and maybe add a comment block explaining that this maps local orders Ids to global order Ids.

Also, I think this is virtually identical functionally speaking but I think it's a little more appropriate to define this as a `Map<string, string>`. That's the pattern we use for `responseMap` in the Peer class.

lib/orderbook/OrderBook.ts
```diff
@@ -96,19 +98,26 @@ class OrderBook extends EventEmitter {
...
     const matchingEngine = this.matchingEngines[pairId];
     if (!matchingEngine) {
       throw errors.INVALID_PAIR_ID(pairId);
     }
 
-    if (matchingEngine.removeOwnOrder(orderId)) {
-      return this.removeOrder(this.ownOrders, orderId, pairId);
+    let id = orderId;
+
+    if (this.localIds[orderId]) {
```
sangaman: Rather than doing this local id lookup every time we remove an order, I think it's better to create a wrapper method called `removeOwnOrderByLocalId` which takes `pairId` and `localId`, performs this lookup, and then calls and returns the result of `removeOwnOrder`. If it doesn't find a matching value for `localId` it returns false.

proto/xudrpc.proto
```diff
@@ -218,7 +220,7 @@ message CancelOrderRequest {
 }
 
 message CancelOrderResponse {
-  string result = 1 [json_name = "result"];
+  bool result = 1 [json_name = "result"];
```
sangaman: Let's rename this to `canceled` so that it's more clear what the response means - true means an order was canceled and false means it wasn't. We can put a comment in that says that too.

lib/cli/commands/placeorder.ts
```diff
@@ -2,29 +2,33 @@ import { callback, loadXudClient } from '../command';
 import { Arguments } from 'yargs';
 import { PlaceOrderRequest, Order } from '../../proto/xudrpc_pb';
 
-export const command = 'placeorder <pair_id> <quantity> [price]';
+export const command = 'placeorder <pair_id> <order_id> <price> [quantity]';
```
sangaman: Was it intentional to make `price` required and `quantity` optional? I think we should keep this the other way around, `<order_id> <quantity> [price]`.

proto/xudrpc.proto
```diff
@@ -191,10 +191,12 @@ message Order {
   string host_id = 5 [json_name = "host_id"];
   // A UUID for this order
   string id = 6 [json_name = "id"];
+  // The local ID for this oder
+  string local_id = 7 [json_name = "local_id"];
```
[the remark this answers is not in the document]
michael1011: `GetOrders` and `SubscribePeerOrders` show `PeerOrder` with the field `id`. So we can't remove `id` from this message.

michael1011:
What about adding an additional debug message for removing orders? Right now we have just one when an order gets added.

proto/xudrpc.proto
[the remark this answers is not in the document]
moshababo: @michael1011 is right in that we're we're dropping `localId` when broadcasting an order, so a `peerOrder` doesn't have that field. 
@sangaman is right in that `id` is irrelevant for an `ownOrder`, because it suppose to be invisible to the client.
 
At the moment `Order` message definition contains all possible fields. Many calls using it, but require only some of the fields. I think we should define all the types more precisely. 
If i'm creating a new `PlaceOrderRequest` object, I might end up thinking that I need to set `HostId` \ `Invoice` \ `CreatedAt`,  which in practice would be ignored. 
If you agree on this, it can be addressed in a separate PR.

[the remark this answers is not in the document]
sangaman: Yes, `PlaceOrderRequest` ought to be fixed to accept the same parameters that are possible on the xucli placeorder call. Doesn't have to be part of this PR but that's something that could be improved.

I originally thought we can drop id even though `GetOrders` and `SubscribePeerOrders` use peer orders because I suspected there's no value in returning the xud global order id to the client. But in thinking about this some more, we will need it for the use case where the exchange does all the matching - they need to be able to specify which peer orders to take by the XUD id.

However, I'm also thinking now we might have a case where an exchange is letting XUD do all the matching and doesn't want to have to create and track their own Ids. Which makes me think maybe `localId` should be optional when placing an order. Maybe even tracking localIds at all can be something that's configurable? Also something that can be done in a separate PR, but I wanted to see what you thought.

lib/orderbook/OrderBook.ts
```diff
@@ -154,31 +169,38 @@ class OrderBook extends EventEmitter {
...
+  private addOrder = (type: Map<String, Orders>, order: orders.StampedOrder) => {
+    if (order['localId']) {
+      this.localIdMap[order['localId']] = order.id;
+    }
+
     this.getOrderMap(type, order)[order.id] = order;
   }
 
-  private removeOrder = (type: { [pairId: string]: Orders }, orderId: string, pairId: string): boolean => {
+  private removeOrder = (type: Map<String, Orders>, orderId: string, pairId: string): boolean => {
```
moshababo: Any reason for not deleting the order id from `localIdMap`?

michael1011:
I meant when an orders gets removed. We already have a message when one gets added. @sangaman

sangaman:
Oops sorry that's what I meant as well. Debug message for both adding and removing orders.

lib/orderbook/OrderBook.ts
```diff
@@ -25,8 +25,14 @@ class OrderBook extends EventEmitter {
...
 
-  private ownOrders: { [ pairId: string ]: Orders } = {};
-  private peerOrders: { [ pairId: string ]: Orders } = {};
+  private ownOrders: Map<String, Orders> = new Map<String, Orders>();
+  private peerOrders: Map<String, Orders> = new Map<String, Orders>();
+
+  /**
+   * This map has the localId of and order as key and the global id as value
+   * which makes possible to cancel orders by their local id without iterating
+   */
```
moshababo: Maybe let's write a more general description, something like: A map between an order localId and id (global).

lib/orderbook/OrderBook.ts
```diff
@@ -25,10 +25,14 @@ class OrderBook extends EventEmitter {
...
 
-  private ownOrders: { [ pairId: string ]: Orders } = {};
-  private peerOrders: { [ pairId: string ]: Orders } = {};
+  private ownOrders: Map<String, Orders> = new Map<String, Orders>();
+  private peerOrders: Map<String, Orders> = new Map<String, Orders>();
 
-  private localIds: { [ localId: string ]: string } = {};
+  /**
+   * This map has the localId of and order as key and the global id as value
+   * which makes possible to cancel orders by their local id without iterating
```
sangaman: Minor typos:

"localId of **and** order"
"which makes **it** possible"

And I think I'd also change "to cancel orders" to "to reference orders" since we might use this for more than just canceling down the road, and I'd drop "without iterating" at the end and add a period.

lib/service/Service.ts
```diff
@@ -162,7 +162,8 @@ class Service extends EventEmitter {
    * Cancel placed order from the orderbook.
    */
   public cancelOrder = async ({ pairId, orderId }: { pairId: string, orderId: string }) => {
-    return this.orderBook.removeOwnOrder(pairId, orderId);
+    return {
+      canceled: this.orderBook.removeOwnOrderByLocalId(pairId, orderId) };
```
sangaman: Could you put this return statement in one line?

proto/xudrpc.proto
```diff
@@ -220,7 +220,8 @@ message CancelOrderRequest {
 }
 
 message CancelOrderResponse {
-  bool result = 1 [json_name = "result"];
+  // True means that the order was canceled and false means that it wasn't
```
sangaman: So I know this comment is essentially verbatim what I said in my last comment here, but I wasn't thinking about the actual wording at the time and when I read it here it doesn't mesh with the way other comments are worded. I'd prefer "Indicates whether an order was successfully canceled" if that's alright.

lib/orderbook/OrderBook.ts
```diff
@@ -96,12 +102,22 @@ class OrderBook extends EventEmitter {
     return this.addOwnOrder(order);
   }
 
-  public addMarketOrder = (order: orders.MarketOrder) => {
+  public addMarketOrder = (order: orders.OwnMarketOrder): matchingEngine.MatchingResult => {
     const price = order.quantity > 0 ? Number.MAX_VALUE : 0;
     return this.addOwnOrder({ ...order, price }, true);
```
moshababo: I think we should verify `localId` uniqueness in the beginning of `addOwnOrder` method. Client shouldn't place a new order with an id which already exists. At the moment the later `localId` will override the previous one (`addOrder`) for the remaining order. swaps updates would also be problematic. And at the moment it creates a bug that I can't cancel an order if I re-used it's `localId`.

proto/xudrpc.proto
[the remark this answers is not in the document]
michael1011: I agree that there are cases in which tracking `localId` is not necessary. 

Can I take care of that and separating the gRPC messages? Because when this PR gets merged all of my issues for the prealpha are done.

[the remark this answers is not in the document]
moshababo: @kilrau what do you say about `localId` being optional?

If so, yes, it can and better be done in this PR. gRPC messages separation is not important for pre-alpha IMO.

test/unit/MatchingEngine.spec.ts
```diff
@@ -17,6 +17,11 @@ const createOrder = (price: number, quantity: number, date = ms()): orders.Stamp
   invoice: '',
 });
 
+const createOwnOrder = (price: number, quantity: number, data = ms()): orders.StampedOwnOrder => {
+  const { hostId, invoice, ...ownOrder } = { ...createOrder(price, quantity, data), localId: uuidv1() };
```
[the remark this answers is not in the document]
sangaman: Separate public methods for peer and own orders would be good I think but they should try to reuse logic in a private method if possible.

lib/orderbook/OrderBook.ts
```diff
@@ -65,17 +71,17 @@ class OrderBook extends EventEmitter {
...
-  * Returns lists of the node's own buy and sell orders sorted by price.
+  * Returns lists of the node's own buy and sell orders
   */
-  public getOwnOrders = (pairId: string, maxResults: number): { [type: string]: orders.StampedOrder[] } => {
-    return this.getOrders(maxResults, this.ownOrders[pairId]);
+  public getOwnOrders = (pairId: string, maxResults: number): { [type: string]: orders.StampedOwnOrder[] } => {
+    return this.getOrders(maxResults, this.ownOrders[pairId]) as { [type: string]: orders.StampedOwnOrder[] };
   }
 
   private getOrders = (maxResults: number, orders: Orders): { [ type: string ]: orders.StampedOrder[]} => {
```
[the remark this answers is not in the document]
sangaman: It might be best just to use the `Map<string, StampedOwnOrder[]>` type since we've started using `Map` in more places in this PR.

lib/orderbook/OrderBook.ts
michael1011: No I completely forgot about that. Thanks!

michael1011:
@moshababo You are talking about P2P packets to update the quantity of an order?

moshababo:
Yes, the p2p packet is one part (#150). 
To solve #176, this packet should be sent when an order is being split on a local xud (due to a partial match), or by client (`cancalOrder`). The `OrderBook` remove functions should support that. 

It doesn't need to be addressed in this PR.

michael1011:
@moshababo Yeah I will address #176 in another PR.

lib/orderbook/OrderBook.ts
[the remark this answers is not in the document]
moshababo: What bothered me about it is that it's hard to quickly understand what `type` means, because we have a few splitting dimensions for orders. 
But it's fine to leave it as is, because we're going to re-arrange this area anyway for #215.

lib/orderbook/OrderBook.ts
```diff
@@ -98,25 +105,41 @@ class OrderBook extends EventEmitter {
...
   }
 
-  public removeOwnOrder = (orderId: string, pairId: string): boolean => {
+  public removeOwnOrderByLocalId = (pairId: string, localId: string): boolean => {
+    const id = this.localIdMap[localId];
+
+    if (id === undefined) {
+      return false;
+    } else {
+      delete this.localIdMap[localId];
```
moshababo: It would be better to have the API balanced, so if we're adding to `localIdMap` at `addOrder`, we should remove it from there at `removeOrder`, and not 2 levels above that. But it's fine to leave it like this for now, because of the upcoming #215.

PR #218, opened 2018-07-27
gRPC error handling and status codes #116

### LINKED ISSUE #116: Improve gRPC error handling and status code
Currently, gRPC error handling is naive and every error is returning 2 UNKNOWN or 13 INTERNAL. We should intelligently use [gRPC status codes](https://grpc.io/grpc/node/grpc.html#.status) based on the type of errors being handled.

rsercano:
fixes #116


rsercano:
Hello @sangaman are you planning to add this to 1.0.0 alpha release ?

sangaman:
@rsercano Yes.

lib/service/Service.ts
```diff
@@ -53,78 +54,20 @@ class Service extends EventEmitter {
...
-
-      peerOrdersCount += Object.keys(peerOrders.buyOrders).length + Object.keys(peerOrders.sellOrders).length;
-      ownOrdersCount += Object.keys(ownOrders.buyOrders).length + Object.keys(ownOrders.sellOrders).length;
-    }
-
-    info.orders = {
-      peer: peerOrdersCount,
-      own: ownOrdersCount,
-    };
+    await this.constructOders(info);
```
sangaman: typo in "Oders" - although I don't think we need to break this into a separate method in any case.

lib/service/Service.ts
```diff
@@ -133,72 +76,269 @@ class Service extends EventEmitter {
...
+        message: 'could not connect to raiden',
+      };
+      this.logger.error(`raiden error: ${serviceError}`);
+      return {
+        error: serviceError,
+      };
+    }
+  }
+
+  private connectToLnd = async() => {
```
sangaman: This does not connect to lnd, it's jsut calling getinfo, and I don't think we really need a separate method for this logic.

lib/service/Service.ts
```diff
@@ -133,72 +76,269 @@ class Service extends EventEmitter {
...
+   * @throws ServiceError with status.FAILED_PRECONDITION code
    */
   public getPairs = () => {
-    return this.orderBook.getPairs();
+    try {
+      return this.orderBook.getPairs();
+    } catch (err) {
+      const serviceError: ServiceError = {
+        code: status.FAILED_PRECONDITION,
+        metadata: err,
```
sangaman: `metadata` is actually supposed to be grpc.metadata object, so I'm not sure this works. Same for other cases where `metadata` is assigned a value. https://grpc.io/grpc/node/grpc.Metadata.html

lib/service/Service.ts
```diff
@@ -133,72 +76,269 @@ class Service extends EventEmitter {
...
+        name: 'internal',
+        message: 'could not get orders',
+      };
+      throw serviceError;
+    }
   }
 
   /**
    * Add an order to the order book.
+   * @throws ServiceError with status.INTERNAL code
```
sangaman: `@throws` is not a keyword for typedoc, same for other lines where @throws is used: http://typedoc.org/guides/doccomments/

I might just drop these @throws comment lines entirely, if there's something important or unique to note maybe put it in the method description.

lib/service/Service.ts
```diff
@@ -133,72 +76,269 @@ class Service extends EventEmitter {
...
+        code: status.FAILED_PRECONDITION,
+        metadata: err,
+        name: 'order constructon',
+        message: 'could not construct orders',
+      };
+      throw serviceError;
+    }
+  }
+
+  private connectToRaiden = async() => {
```
sangaman: This is also not connecting to raiden, just getting info from the API, and like the other newly added private methods I think this can be dropped.

sangaman:
You can look at #234 for an example of a particular call for how I'm thinking this should be best implemeented.

lib/service/Service.ts
rsercano: @sangaman actually this's just for better handling of exceptions

lib/service/Service.ts
rsercano: @sangaman same as "Orders", getInfo method is huge, I guess we need to seperate it to handle exceptions in a better way ?

lib/service/Service.ts
rsercano: will check that ! :)

lib/service/Service.ts
rsercano: sure, will fix

lib/service/Service.ts
rsercano: same here :)

rsercano:
Hello @sangaman for me;  

> I think the heavy lifting for the issue this PR is trying to close is in going up through the code to the methods called by Service.ts

yes, that's the correct way, but that would take loads of time to understand underlying code base and that doesn't fit my man-hour rate unfortunately, if that's what you want I guess I would give up on this. 

On the other hand I checked #234 and [here](https://github.com/ExchangeUnion/xud/pull/234/files#diff-388131b223817eb339ddab0b29f399b0) seems okay if you know what kind of errors would come from external service which requires a deep knowledge as I mentioned previously

Therefore, should I handle every exception on underlying contexts which are called by `Service.ts` if that's the case I would give up on this probably, would you enlighten me ?

rsercano:
Let me try another one, in a few hours :) @sangaman I'll try as you've done

sangaman:
Now that #234 is finalized, you can see the approach for this sort of error handling. It's all handled within GrpcService.ts, and it's about detecting certain error types and using the appropriate status code based on the error. Up to you on whether or not it's a good use of your time, but this isn't a high priority issue so you can leave it open if you think you'll get around to it. Otherwise you can close and we'll deal with this another time. Thanks.

rsercano:
Thanks @sangaman I'll check and give it an another shot I guess, I know it's hard to tell project architecthure and code design in detail from remote, so thx for explaining again ^^

sangaman:
Closing in favor of #302.

rsercano:
yeah sorry, couldn't find spare time to check this..

sangaman:
No worries, thank you!

PR #219, opened 2018-07-27
Docker ltcd

itsbalamurali:
Includes LTCD


docker/docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
     build:
       context: xud/
     volumes:
-      - ../:/opt/xud
+      #- ../:/opt/xud
```
itsbalamurali: @kilrau here is the line.

Issue #215, opened 2018-07-24
Orders data structures


moshababo:
To support internal matching, `ownOrders` and `peerOrders` are merged in `MatchingEngine` priority queues, but we still need to support an access to the orders separately. For `ownOrders` it's for the `GET_ORDERS` packet response, and for `peerOrders` it's for `subscribePeerOrders` gRPC call.

First we need to decide on whether or not we need them sorted (by price), when being served separately. I don't see a requirement for that right now, so if someone does, say so. 

In addition, these lists are currently being managed inside of `OrderBook`, which is separated from the main work in `MatchingEngine`. I think it will be better to have everything being managed under one place, so that would be in inside of `MatchingEngine`. 

In addition, @sangaman suggested to make the priority queue a secondary index, so that the orders would be saved elsewhere, to allow easy updates (via `orderId`, the pk). To do so, the queue will still require to have the order's `price` & `createdAt`, because that's what we're sorting on. So I don't see a benefit from splitting the order object. We can have the same object sitting both on the priority queue and on a separate list, and use the later for fast pk lookups, when doing an read/update. add/delete will require operations on both.

sangaman:
> We can have the same object sitting both on the priority queue and on a separate list ({ [orderId: string]: Order }), and use the later for fast pk lookups, when doing read/update. add/delete will require operations on both.

This seems like the best approach.

sangaman:
Also if we merge the `OrderBook` and `MatchingEngine` classes together I feel like we should come up with a more descriptive name, particularly since we'll still need the lists even if we're not in matching mode. I'm not sure what the best name is right now though, and it might depend on the specifics of how we implement things.

moshababo:
I think we should keep both `OrderBook` & `MatchingEngine`. The later will contain all logic for orders management, per a specific pair. It should be easily tested with unit tests. Yes, It won't be focused solely on matching, but matching involves with orders management anyway. 
`OrderBook` is the higher layer, which interacts with the p2p layer (listening, broadcasting), emitting events, creating invoices, etc. It shouldn't contain orders management logic.

moshababo:
We still haven't thought through the implementation details of that mode (what will still need to be managed inside `xud`). We might be able to cut off `MatchingEngine` entirely. But even if we don't, it's not a big deal.

kilrau:
[...]

PR #222, opened 2018-07-29
order_invalidation packet

moshababo:
Implements #150.

This doesn't implement #176, which should send the packet in a case of a split or a full match.

In addition, this packet need to be sent in case of a remote match / swap.


lib/service/Service.ts
```diff
@@ -162,7 +162,8 @@ class Service extends EventEmitter {
    * Cancel placed order from the orderbook.
    */
   public cancelOrder = async ({ orderId, pairId }: { orderId: string, pairId: string }) => {
-    return this.orderBook.removeOwnOrder(orderId, pairId);
+    this.pool.broadcastOrderInvalidation(orderId, pairId);
+    return this.orderBook.removeOwnOrder(orderId, pairId).toString();
```
michael1011: Why are you using `toString()` and not `JSON.stringify()`?

lib/p2p/Pool.ts
```diff
@@ -114,6 +114,15 @@ class Pool extends EventEmitter {
   public broadcastOrder = (order: OutgoingOrder) => {
     const orderPacket = new OrderPacket(order);
     this.peers.forEach(peer => peer.sendPacket(orderPacket));
+
+    // TODO: send only to peers which accepts the pairId
```
michael1011: One of the first things we should address after the prealpha is adding the supported pairs to the `Hello` packet and implementing it because we have quite a bunch of `TODO` comments about that.

lib/orderbook/MatchingEngine.ts
```diff
@@ -152,9 +152,12 @@ class MatchingEngine {
     return matchingResult;
   }
 
-  public removeOwnOrder = (orderId: string): orders.StampedOwnOrder | null => {
-    return this.priorityQueues.buyOrders.removeOne(order => order.id === orderId) ||
-      this.priorityQueues.sellOrders.removeOne(order => order.id === orderId);
+  public removeOwnOrder = (orderId: string): orders.StampedOrder | null => {
```
michael1011: Why `removeOwnOrder` and `removePeerOrder`? We should either unify them or change the return types to `orders.StampedOwnOrder` and `orders.StampedPeerOrder` because they are doing the same thing.

lib/p2p/packets/types/OrderInvalidationPacket.ts
```diff
@@ -0,0 +1,20 @@
+import Packet, { PacketDirection } from '../Packet';
+import PacketType from '../PacketType';
+
+type OrderInvalidationPacketBody = {
+  orderId: string;
+  pairId: string;
+  // quantity: number;  TODO: implement quantity invalidation
```
michael1011: Which quantity will be sent in the `OrderInvalidationPacketBody`? How much was filled or how much is remaining?

lib/service/Service.ts
moshababo: Considering it's just a boolean, I thought it doesn't make much of a difference

lib/orderbook/MatchingEngine.ts
moshababo: I agree that this design is weird. I did it as a preparation to #215, which would require to have them separated. It would require us to redo things anyway.

lib/p2p/packets/types/OrderInvalidationPacket.ts
moshababo: How much was filled, because that's what you are invalidating. Unless we'll find it safer/easier to do otherwise.

lib/orderbook/MatchingEngine.ts
sangaman: Why not cast the order we return as a `StampedOwnOrder`?

lib/orderbook/MatchingEngine.ts
```diff
@@ -152,9 +152,12 @@ class MatchingEngine {
...
+    return this.removeOrder(orderId);
+  }
+
+  public removePeerOrder = (orderId: string): orders.StampedOrder | null => {
```
sangaman: We can cast the return result as `StampedPeerOrder`.

moshababo:
Thanks for the feedback. I'm merging

PR #225, opened 2018-07-30
Event types

sangaman:
This adds event types like with #212 to our other `EventEmitter` classes. I also fixed a bug where the peer id was accidentally getting assigned to the `hostId` value on peer orders.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -19,6 +19,11 @@ type Orders = {
   sellOrders: Map<String, orders.StampedOrder>;
 };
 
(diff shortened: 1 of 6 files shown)


sangaman:
I rebased and added types for the new order invalidation event. I also went ahead and created a new type called `OrderIdentifer` that's used as the body for the `OrderInvalidation` packet. I wanted a type to use on the events, and my thinking is that we're going to have this orderId/pairId/quantity come up often as we migrate order-related actions to expecting quantity to specify what portion of an order is being acted upon. Take a quick look if you can to make sure it looks good to you.

PR #228, opened 2018-07-30
Priority queue types & removePeerOrders fix

sangaman:
I created a typed interface for our priority queues in the MatchingEngine.ts class, and in the process detected and fixed a bug with the way peer orders were previously being removed.

### DIFF
--- a/lib/orderbook/MatchingEngine.ts
+++ b/lib/orderbook/MatchingEngine.ts
@@ -16,6 +16,7 @@ type PriorityQueue = {
   isEmpty: Function;
   has: Function;
 };
+import { StampedOrder, StampedOwnOrder, StampedPeerOrder } from '../types/orders';
 
 type PriorityQueues = {
   buyOrders: PriorityQueue;
@@ -160,11 +161,15 @@ class MatchingEngine {
     return this.removeOrder(orderId) as orders.StampedPeerOrder;
   }
 
-  public removePeerOrders = (predicate: Function): orders.StampedPeerOrder[] => {
+  public removePeerOrders = (hostId: number): StampedPeerOrder[] => {
+    const callback = (order: StampedOrder) => {
+      return (order as StampedPeerOrder).hostId === hostId;
+    };
+
     return [
-      ...this.priorityQueues.buyOrders.removeMany(predicate),
-      ...this.priorityQueues.sellOrders.removeMany(predicate),
-    ];
+      ...this.priorityQueues.buyOrders.removeMany(callback),
+      ...this.priorityQueues.sellOrders.removeMany(callback),
+    ] as StampedPeerOrder[];
   }
 
   public isEmpty = (): boolean => {
--- a/lib/orderbook/MatchingEngine.ts
+++ b/lib/orderbook/MatchingEngine.ts
@@ -1,31 +1,35 @@
 import assert from 'assert';
 import FastPriorityQueue from 'fastpriorityqueue';
 
-import { orders, matchingEngine } from '../types';
+import { matchingEngine } from '../types';
 import { OrderingDirection } from '../types/enums';
 import Logger from '../Logger';
-
-type PriorityQueue = {
-  add: Function;
-  removeOne: Function;
-  removeMany: Function;
-  heapify: Function;
-  peek: Function;
-  poll: Function;
-  trim: Function;
-  isEmpty: Function;
-  has: Function;
-};
 import { StampedOrder, StampedOwnOrder, StampedPeerOrder } from '../types/orders';
 
+interface PriorityQueue<T> {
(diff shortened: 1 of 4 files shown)


lib/orderbook/MatchingEngine.ts
```diff
@@ -152,16 +157,20 @@ class MatchingEngine {
...
-      this.priorityQueues.sellOrders.removeOne(order => order.id === orderId);
+  public removeOwnOrder = (orderId: string): StampedOwnOrder | undefined => {
+    return (this.priorityQueues.buyOrders.removeOne(order => order.id === orderId) ||
+      this.priorityQueues.sellOrders.removeOne(order => order.id === orderId)) as StampedOwnOrder | undefined;
   }
 
-  public removePeerOrders = (predicate: Function): orders.StampedPeerOrder[] => {
+  public removePeerOrders = (hostId: number): StampedPeerOrder[] => {
+    const callback = (a: StampedOrder) => {
+      return (a as StampedPeerOrder).hostId === hostId;
```
moshababo: changing `a` to `order` will make more sense

lib/orderbook/MatchingEngine.ts
```diff
@@ -152,16 +157,20 @@ class MatchingEngine {
...
-  public removeOwnOrder = (orderId: string): orders.StampedOwnOrder | null => {
-    return this.priorityQueues.buyOrders.removeOne(order => order.id === orderId) ||
-      this.priorityQueues.sellOrders.removeOne(order => order.id === orderId);
+  public removeOwnOrder = (orderId: string): StampedOwnOrder | undefined => {
+    return (this.priorityQueues.buyOrders.removeOne(order => order.id === orderId) ||
+      this.priorityQueues.sellOrders.removeOne(order => order.id === orderId)) as StampedOwnOrder | undefined;
   }
 
-  public removePeerOrders = (predicate: Function): orders.StampedPeerOrder[] => {
+  public removePeerOrders = (hostId: number): StampedPeerOrder[] => {
```
moshababo: If later we'll need to remove multiple peer orders by a different param, we'll need another function for that. I used a predicate function instead of a regular param to make the interface more open.

lib/orderbook/OrderBook.ts
```diff
@@ -181,7 +181,7 @@ class OrderBook extends EventEmitter {
   private removePeerOrders = async (peer: Peer): Promise<void> => {
     if (peer.hostId) {
       this.pairs.forEach((pair) => {
-        this.matchingEngines[pair.id].removePeerOrders(order => order === peer.hostId!);
+        this.matchingEngines[pair.id].removePeerOrders(peer.hostId!);
```
moshababo: good catch

test/unit/MatchingEngine.spec.ts
```diff
@@ -214,14 +216,14 @@ describe('MatchingEngine.removePeerOrders', () => {
     const secondHostId = 2;
 
     expect(engine.isEmpty()).to.be.true;
-    expect(engine.removePeerOrders(() => true)).to.be.empty;
+    expect(engine.removePeerOrders(0)).to.be.empty;
```
moshababo: this was meant to test removal of all orders. If you want to use `hostId` instead of a predicate function, you can remove this line.

lib/orderbook/MatchingEngine.ts
sangaman: I see, I guess I'm thinking it's better to have the predicate logic in the method itself. It certainly seems better for testing, that's how the previous bug slipped through tests. We can create more methods down the road. Another option might be to define the predicates in one place, so that the tests use exactly the same function as the main code.

test/unit/MatchingEngine.spec.ts
sangaman: I can bring this back with another predicate defined outside of the tests, or I can create a `clearPeerOrders` method. But I also think it's OK to leave this test out since it's more a test of FastPriorityQueue than our code, at least until we start using a "clear" function on orders.

lib/orderbook/MatchingEngine.ts
moshababo: The bugs was slipped because we didn't do yet integration for that section from `OrderBook`, only to the underlying `MatchingEngine`.
Defining the predicates in a separate place won't be a good idea IMO.

We can keep it like this if you prefer, and reassess the design later.

test/unit/MatchingEngine.spec.ts
moshababo: Yes, if you remove the predicate function support, just remove this line.

sangaman:
Rebased and updated package-lock.json to use the latest version of our FastPriorityQueue fork. Merging, thanks.

PR #234, opened 2018-07-31
placeOrder gRPC error handling

### LINKED ISSUE #229: [Crash] placeOrder with Invalid pairID
Current Behavior: placeOrder with Invalid pairID causes XUD to crash, due to unhandled gRPC error
How it should be: gracefully handle all gRPC errors

```
2018-7-30 13:37:14 [RPC] error: invalid pairId: BTC/LTC
TypeError: Incorrectly typed arguments to startBatch
    at handleError (/home/kilrau/xud/node_modules/grpc/src/server.js:71:8)
    at sendUnaryData (/home/kilrau/xud/node_modules/grpc/src/server.js:597:9)
    at GrpcService.<anonymous> (/home/kilrau/xud/dist/grpc/GrpcService.js:40:17)
    at Generator.throw (<anonymous>)
    at rejected (/home/kilrau/xud/dist/grpc/GrpcService.js:5:65)
    at <anonymous>
    at runMicrotasksCallback (internal/process/next_tick.js:121:5)
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)
/home/kilrau/xud/dist/bootstrap.js:13
        throw err;
        ^

TypeError: Incorrectly typed arguments to startBatch
    at handleError (/home/kilrau/xud/node_modules/grpc/src/server.js:71:8)
    at sendUnaryData (/home/kilrau/xud/node_modules/grpc/src/server.js:597:9)
    at GrpcService.<anonymous> (/home/kilrau/xud/dist/grpc/GrpcService.js:40:17)
    at Generator.throw (<anonymous>)
    at rejected (/home/kilrau/xud/dist/grpc/GrpcService.js:5:65)
    at <anonymous>
    at runMicrotasksCallback (internal/process/next_tick.js:121:5)
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! xud@1.0.0-alpha.1 start: `node dist/Xud.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the xud@1.0.0-alpha.1 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/kilrau/.npm/_logs/2018-07-30T20_37_14_899Z-debug.log
```

[2018-07-30T20_37_14_899Z-debug.log](https://github.com/ExchangeUnion/xud/files/2242807/2018-07-30T20_37_14_899Z-debug.log)

sangaman:
Fixes #229. Partially addresses #116. This gracefully handles invalid `placeOrder` requests such that they don't crash the main xud process. I changed `xucli` to not print the stack trace, since I don't think it's helpful for developers/testers and it's even less helpful for end users. I also tried to standardize the `errors` classes to not use the `any` type, and I added a `.` separator in the `code` after the prefix.

### DIFF
--- a/lib/grpc/GrpcService.ts
+++ b/lib/grpc/GrpcService.ts
@@ -1,4 +1,4 @@
-import grpc from 'grpc';
+import grpc, { status, ServiceError } from 'grpc';
 import Logger from '../Logger';
 import Service from '../service/Service';
 import { isObject } from '../utils/utils';
@@ -8,6 +8,8 @@ import { GetInfoResponse } from '../proto/lndrpc_pb';
 import { Orders } from 'lib/orderbook/OrderBook';
 import { MatchingResult } from '../types/matchingEngine';
 import { OwnOrder } from '../types/orders';
+import { default as orderErrors, errorCodes as orderErrorCodes } from '../orderbook/errors';
+import { default as serviceErrors, errorCodes as serviceErrorCodes } from '../service/errors';
 
 function serializeDateProperties(response) {
   Object.keys(response).forEach((key) => {
@@ -36,8 +38,20 @@ class GrpcService {
       const response = serializeDateProperties(rawResponse);
       callback(null, response);
     } catch (err) {
+      // if we recognize this error, return a proper gRPC ServiceError with a descriptive and appropriate code
+      let grpcError: ServiceError | undefined;
+      switch (err.code) {
+        case serviceErrorCodes.INVALID_ARGUMENT:
+          grpcError = { ...err, code: status.INVALID_ARGUMENT };
+          break;
+        case orderErrorCodes.INVALID_PAIR_ID:
+          grpcError = { ...err, code: status.NOT_FOUND };
+          break;
+      }
       this.logger.error(err);
-      callback(err, null);
+
+      // return grpcError if we've created one, otherwise pass along the caught error as UNKNOWN
+      callback(grpcError ? grpcError : err, null);
     }
   }
 
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -1,6 +1,6 @@
 import Logger from '../Logger';
 import Pool from '../p2p/Pool';
-import OrderBook from '../orderbook/OrderBook';
+import OrderBook, { OrderArrays } from '../orderbook/OrderBook';
 import LndClient from '../lndclient/LndClient';
 import RaidenClient, { TokenSwapPayload } from '../raidenclient/RaidenClient';
 import { OwnOrder } from '../types/orders';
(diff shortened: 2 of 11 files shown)


kilrau:
Does this also fix `Cannot read property 'buyOrders' of undefined` mentioned in #229 ?

sangaman:
@kilrau Nope, but I can reproduce it so I should be able to fix it in a separate commit as part of this PR. I'll ping you when it's ready.

sangaman:
@kilrau The `getorders` without specifying a `pairId` issue should be fixed with the latest commit.

lib/service/Service.ts
```diff
@@ -139,22 +138,52 @@ class Service extends EventEmitter {
...
-    assert(order.price >= 0, 'price cannot be negative');
-    assert(order.quantity !== 0, 'quantity must not equal 0');
-    if (order.price === 0) {
-      return this.orderBook.addMarketOrder(order);
-    } else {
-      return this.orderBook.addLimitOrder(order);
+    if (order.price < 0) {
+      throw {
+        message: 'price cannot be negative',
+        code: status.INVALID_ARGUMENT,
```
moshababo: Sorry for being the separation of concerns police, but `Service` shouldn't know/use anything from `gRPC`.

sangaman: That's fine, I just wasn't seeing an easy way to do this without `Service` at least knowing the error codes. Do you have an alternative to suggest? I guess I could put all error handling in `GrpcService.unaryCall` and then create error types for these invalid argument errors - or maybe go back to AssertionErrors and always treat those as `INVALID_ARGUMENT`.

moshababo: simplest solution would be to create a list of errors for `Service` which would be the equivalent to these of gRPC you wanted to use. `GrpcService` will catch them, and link them to the gRPC errors. This way we can continue to expand how we want to expose internal errors to the gRPC client. At the moment we have no support for that. 

It might be what you suggested, but I wasn't sure.

sangaman:
Rebased and restructured to move all grpc related error handling to GrpcService.ts. I went with using `AssertionError` to indicate an invalid argument. That makes sense to me, because we should be asserting up front parameter values to detect if they are values that are never valid, such as a negative price, and this seems like the simplest way to do it.

Here's some sample error output from the cli:

```shell
xucli placeorder KDL/CNB 1 1 1
Error: 5 NOT_FOUND: invalid pairId: KDL/CNB

xucli placeorder KDL/CNB 1 1 -1
Error: 3 INVALID_ARGUMENT: price cannot be negative

xucli placeorder KDL/CNB 1 0 1
Error: 3 INVALID_ARGUMENT: quantity must not equal 0
```

lib/service/Service.ts
```diff
@@ -139,23 +138,48 @@ class Service extends EventEmitter {
...
+      };
+    }
+    if (order.quantity === 0) {
+      throw {
+        message: 'quantity must not equal 0',
+        code: status.INVALID_ARGUMENT,
+      };
+    } */
```
moshababo: this can be removed

lib/grpc/GrpcService.ts
```diff
@@ -36,8 +38,21 @@ class GrpcService {
       const response = serializeDateProperties(rawResponse);
       callback(null, response);
     } catch (err) {
+      // if we recognize this error, return a proper gRPC ServiceError with a descriptive and appropriate code
+      let grpcError: ServiceError | undefined;
+      if (err instanceof AssertionError) {
+        grpcError = { name: err.name, message: err.message, code: status.INVALID_ARGUMENT };
```
moshababo: binding `AssertionError` to `INVALID_ARGUMENT` will probably lead to some misleading errors, because we're not limiting the `assert` usage only for arguments validations. So I think we should create a custom, cross-domain error type for that. You can open a separate issue if you don't want to deal with this right now.

lib/service/Service.ts
sangaman: Ah yes, just forgot to. Thanks.

sangaman:
I created an errors class for `service` with an `INVALID_ARGUMENT` error, and a simple method that throws that error along with a message if an expected condition is false. See if that looks good to you @moshababo.

I also modified the p2p errors class to match the others; I'd overlooked it before.

moshababo:
Looks good, thanks. 
We'll probably need that not only for the service layer, so I think we should eventually create a list of cross-domain errors. We can wait with that until it will be actually required.

Issue #106, opened 2018-06-05
Raiden ERC20 swap integration


kilrau:
- Raiden ERC20 swap
- Raiden/Lightning swap

sangaman:
This issue is on hold until Raiden reintroduces swap functionality, which was removed after v0.3.

kilrau:
Waiting for mainnet release. Asked Lefteris to keep us in the loop on the HTLC design, so we can help to make them compatible with ln.

kilrau:
duplicate, closing

PR #243, opened 2018-08-01
Set handshake data and raiden address on startup

sangaman:
This commit discovers our raiden address on startup and defines handshake data to be sent in the hello packet to newly connected peers.

This functionality is a prerequisite for #106 and #236.


lib/Xud.ts
```diff
@@ -53,13 +53,28 @@ class Xud {
...
+
+      const handshakeData = {
+        pairs,
+        version: '1.0',
+        nodePubKey: this.nodeKey.nodePubKey,
+        listenPort: this.config.p2p.listen ? this.config.p2p.port : undefined,
+        raidenAddress: this.raidenClient.address,
+      };
+
+      this.pool.init(handshakeData);
```
michael1011: The `handshakeData` is used just once in this line. In my mind we don't need to define a new variable. So that the object gets created and the function called in one statement.

lib/raidenclient/RaidenClient.ts
```diff
@@ -72,6 +72,7 @@ type Channel = OpenChannelPayload & {
 class RaidenClient extends BaseClient {
   private port: number;
   private host: string;
+  public address?: string;
```
michael1011: I always moved the declaration of public variables above the private ones. Is that something we should care about?

sangaman: Yeah I think that's a good idea, I'd be in favor of following that convention. There's a tslint rule for it too that we can add: https://palantir.github.io/tslint/rules/member-ordering/

sangaman:
Thanks Michael, I made those changes.

PR #245, opened 2018-08-01
Market order docs

### LINKED ISSUE #232: Add marketOrder to XUCLI
Couldn't test this, if it's already working by setting price 0, pls add this into the TypeDoc as comment

michael1011:
Closes #232 

I added documentation about creating market orders with `xucli`.

### DIFF
--- a/lib/cli/commands/placeorder.ts
+++ b/lib/cli/commands/placeorder.ts
@@ -4,7 +4,7 @@ import { PlaceOrderRequest, Order } from '../../proto/xudrpc_pb';
 
 export const command = 'placeorder <pair_id> <order_id> <quantity> [price]';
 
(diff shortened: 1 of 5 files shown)


lib/cli/commands/placeorder.ts
```diff
@@ -4,7 +4,7 @@ import { PlaceOrderRequest, Order } from '../../proto/xudrpc_pb';
 
 export const command = 'placeorder <pair_id> <order_id> <quantity> [price]';
 
-export const describe = 'place an order';
+export const describe = 'place an order, if the price is 0 or unspecified -> market order';
```
sangaman: How about "place an order, if price is 0 or unspecified a market order is placed"

Issue #239, opened 2018-07-31
How to daemonize XUD with systemd


kilrau:

michael1011:
@kilrau you said that you will add links to additional Markdown files to the readme. Because they don't exist yet I will just post the systemd config here:

```
[Unit]
Description=Exchange Union Daemon
After=lnd.service

[Service]
User=<user which should run LightningTip>
Group=<group of the user above>
ExecStart=<path to the executable>

[Install]
WantedBy=multi-user.target
```

kilrau:
Danke dir, https://github.com/ExchangeUnion/xud/pull/248

PR #244, opened 2018-08-01
P2P Seeding

itsbalamurali:
This pr implements:
- seeding from official bootstrap XU Nodes (#233) .
  - xud1.test.exchangeunion.com
  - xud2.test.exchangeunion.com
  - xud3.test.exchangeunion.com

PR #248, opened 2018-08-01
Readme v2

kilrau:
- distinction between dev setup (manual) and exchanges/users (docker)
- added raiden, docker, systemd instructions
- minor fixes
- closes https://github.com/ExchangeUnion/xud/issues/239


kilrau:
Closes https://github.com/ExchangeUnion/xud/issues/239

instructions-for-devs.md
```diff
@@ -0,0 +1,174 @@
...
+
+## 2. Prepare your environment 
+
+To initialize the database with some testing data, run the following command from `~/xud`:
+
+```bash
+npm run db:init
+```
+
+Setup raiden according to [this guide](https://github.com/ExchangeUnion/xud/blob/raiden_swap/lib/raidenclient/README.md).
```
michael1011: We have to update this link when the Raiden PR gets merged.

instructions-for-devs.md
```diff
@@ -0,0 +1,174 @@
...
+```bash
+~/xud/bin $ ./xud
+2018-6-3 15:14:27 [GLOBAL] info: config loaded
+2018-6-3 15:14:27 [DB] info: connected to database. host:localhost port:3306 database:xud
+2018-6-3 15:14:28 [P2P] info: pool server listening on 0.0.0.0:8885
+2018-6-3 15:14:28 [RPC] info: GRPC server listening on port 8886
+2018-6-3 15:14:28 [RPC] info: gRPC Web API proxy listening on port 8080
+```
+
+Optional command line arguments to override defaults and settings in the [configuration](#configuration) file for a specific `xud` instance
```
michael1011: Should be `#5-configuration-optional` instead of `#configuration`

instructions-for-exchanges.md
```diff
@@ -0,0 +1,54 @@
...
+
+We use `docker-compose` to package `xud`, `mysql`, `lnd`, `btcd`, `ltcd`, `raiden` & `geth` together to make deploying these services as easy as typing a few commands. All configuration between `xud`, `mysql` and other containers are handled automatically by their `docker-compose` config file.
+
+The goal of this is `docker-compose` being able to lauch `xud` along with its dependencies.
+
+## 1. Prerequisites
+* docker (e.g. apt install docker-compose)
+* docker-compose (e.g. apt install docker-compose)
+
+To install these on e.g. Ubuntu, run:
```
michael1011: I would prefer if you link to the [Docker documentation](https://docs.docker.com/install/linux/docker-ce/ubuntu/) instead of writing the commands itself. Especially because the Docker packages in the Ubuntu repositories are fairly outdated and we should use the latest version.

instructions-for-exchanges.md
```diff
@@ -0,0 +1,54 @@
...
+```bash
+apt install docker
+apt install docker-compose
+```
+
+## 2. Install
+```bash
+git clone https://github.com/ExchangeUnion/xud -b 1.0.0-prealpha
```
michael1011: Will we create a branch, tag or both for the prealpha?

I would personally prefer a tag so that we could do the cloning like this:
```bash
git clone https://github.com/ExchangeUnion/xud.git
cd xud
git checkout tags/v1.0.0-prealpha
```

instructions-for-exchanges.md
itsbalamurali: Agree with @michael1011

instructions-for-exchanges.md
sangaman: I'd prefer a tag as well.

instructions-for-devs.md
```diff
@@ -0,0 +1,174 @@
...
+
+Developers:
+```bash
+git clone https://github.com/ExchangeUnion/xud
+cd xud
+npm install
+```
+For everyone else, we recommend cloning the [latest release](https://github.com/ExchangeUnion/xud/releases):
+```bash
+git clone https://github.com/ExchangeUnion/xud -b 1.0.0-prealpha
```
sangaman: Here too I don't think we should specify a particular release or tag, although linking to our release page is good. I think the best approach is probably to publish our releases to npm, and then people can just do `npm install xud` to get the latest release. Until we have a proper release published to npm we can just leave this extra set of commands out I'd say.

README.md
```diff
@@ -10,132 +10,23 @@ This repo contains the early stages of the Exchange Union Daemon ("xud") which e
...
-To initialize the database with default testing data, run the following command:
-
-```bash
-npm run db:init
-```
-
-## Documentation
-
-[Read the TypeDoc here](https://exchangeunion.github.io/xud-typedoc/).
+## Questions?
```
sangaman: Maybe call this something like "Development & Testing Discussion"? Also, add a blank line after the heading.

instructions-for-exchanges.md
sangaman: Is the idea that exchanges should only be installing via docker? I wouldn't expect that to always be the case.

instructions-for-devs.md
kilrau: We will.

instructions-for-devs.md
kilrau: Good catch!

instructions-for-exchanges.md
kilrau: Fixed.

@sangaman Both are not mutually exclusive, that's why I added the (manual) (docker) hints. It's just much faster to make a decision which installation method to use when adding a role.

instructions-for-exchanges.md
kilrau: Agreed and fixed. Pls check.

kilrau:
Pls check again @michael1011 @sangaman

instructions-for-devs.md
```diff
@@ -0,0 +1,199 @@
...
+
+## 2. Prepare your environment 
+
+To initialize the database with some testing data, run the following command from `~/xud`:
+
+```bash
+npm run db:init
+```
+
+Setup raiden according to [this guide](https://github.com/ExchangeUnion/xud/blob/raiden_swap/lib/raidenclient/README.md).
```
michael1011: If the prealpha doesn't have the Raiden swap there is no point for a user to setup Raiden.

instructions-for-devs.md
```diff
@@ -0,0 +1,199 @@
...
+
+```
+
+Examples:
+```bash
+./xucli connect xud1.test.exchangeunion.com 8885
+# Manually connect to another xud instance
+
+./xucli placeorder BTC/LTC 1337 1 99
+#Places a new limit order BUYING 1 BTC for 99 LTC, assigning the orderId 1337
```
michael1011: I would prefer if there is a space between the comment itself and the `#`

instructions-for-devs.md
```diff
@@ -0,0 +1,199 @@
...
+```
+ With args:
+ ```
+nodemon --watch dist -e js bin/xud --lnd.disable=true
+```
+### Daemonize `xud`
+If you want to daemonize `xud`, so you don't have to keep a terminal open and xud starts on bootup, you can do this for example on Ubuntu by adding the following to `systemd`:
+
+```bash
+[Unit]
```
michael1011: This service file doesn't work properly. My bad. @itsbalamurali did come up with a better solution:
```
[Unit]
Description= XUD - ExchangeUnion Daemon
ConditionPathExists=/opt/xud/dist/
After=network.target

[Service]
User=xud
Group=xud
WorkingDirectory=/opt/xud/
ExecStart=/usr/bin/node /opt/xud/dist/Xud.js
Type=simple

[Install]
WantedBy=multi-user.target
```

instructions-for-devs.md
kilrau: Fixed.

instructions-for-devs.md
kilrau: No need for nice words, I just overlooked it ;) - fixed.

instructions-for-devs.md
kilrau: Swaped it. Thx for clearing that up!

instructions-for-devs.md
```diff
@@ -0,0 +1,203 @@
...
+
+
+`xud` uses [MySQL](https://www.mysql.com/) or [MariaDB](https://mariadb.org/). You will have to install one of those and create a user *xud* and a database *xud* and grant all permissions for the new database to the *xud* user.
+
+## 2. Prepare your environment 
+
+To initialize the database with some testing data, run the following command from `~/xud`:
+
+```bash
+npm run db:init
```
michael1011: In #252 it was discussed that this command is just for the test database although @itsbalamurali changed it in #244 I am confused now.

instructions-for-devs.md
```diff
@@ -0,0 +1,203 @@
...
+git checkout tags/1.0.0-prealpha -b 1.0.0-prealpha
+npm install
+```
+
+
+`xud` uses [MySQL](https://www.mysql.com/) or [MariaDB](https://mariadb.org/). You will have to install one of those and create a user *xud* and a database *xud* and grant all permissions for the new database to the *xud* user.
+
+## 2. Prepare your environment 
+
+To initialize the database with some testing data, run the following command from `~/xud`:
```
michael1011: Lets not assume that the users will always clone xud into their home directory.

instructions-for-devs.md
kilrau: Changes from #244 will be reverted, I adjusted the readme to make it clear.

instructions-for-devs.md
kilrau: Removed the ~, even though I think it's ok for instructional purposes. We can assume ppl who don't install in ~ , know what ~ is :)

kilrau:
Please delete the https://exchangeunion.github.io/xud-typedoc/ landing page when merged @sangaman

instructions-for-devs.md
```diff
@@ -0,0 +1,206 @@
...
+  --db.port           Port for SQL database                             [number]
+  --db.username       User for SQL database                             [string]
+  --lnd.certpath      Path to the SSL certificate for lnd               [string]
+  --lnd.disable       Disable lnd integration                          [boolean]
+  --lnd.host          Host of the lnd gRPC interface                    [string]
+  --lnd.macaroonpath  Path of the admin macaroon for lnd                [string]
+  --lnd.port          Port of the lnd gRPC interface                    [number]
+  --p2p.listen        Listen for incoming peers                        [boolean]
+  --p2p.port, -p      Port to listen for incoming peers                 [number]
+  --raiden.disable    Disable raiden integration                       [boolean]
```
offerm: suggest changing the default to false --raiden.disable

instructions-for-devs.md
```diff
@@ -0,0 +1,206 @@
...
+
+./xucli placeorder BTC/LTC 1339 1
+# Places a new market order BUYING 1 BTC for the best LTC market price, assigning the orderId 1339
+
+```
+
+
+## Configuration (optional)
+
+This *optional* configuration file uses [TOML](https://github.com/toml-lang/toml) and by default should be saved at  `~/.xud/xud.conf` on Linux or `AppData\Local\Xud\xud.conf` on Windows (run `xud` at least once for this folder to be created). Default settings which can be overridden are shown below.
```
offerm: macOs - should be saved in ~/.xud/xud.conf (this is not the right place for mac users but it is working).

instructions-for-devs.md
```diff
@@ -0,0 +1,206 @@
...
+port = 3306
+host = "localhost"
+
+[p2p]
+listen = true
+port = 8885
+#make sure this port is reachable from the internet
+
+[lnd]
+disable = false
```
offerm: suggest changing "disable = " from false to true

instructions-for-devs.md
```diff
@@ -0,0 +1,206 @@
...
+host = "localhost"
+
+[raiden]
+disable = false
```
offerm: suggest changing "disable = " from false to true

instructions-for-devs.md
```diff
@@ -0,0 +1,206 @@
...
+host = "localhost"
+port = 5001
+```
+
+An empty *xud* database with schema gets automatically created when launching `xud`. To initialize an additional *xud_test* database with some pre-filled testing data, run the following command from the `~/xud` folder:
```
offerm: Do you create a database when launching xud? 
Are you sure the use has permission for that? see comment above.

instructions-for-devs.md
michael1011: Wouldn't be the correct folder on macOS be `~/Library/Application Support`?

instructions-for-devs.md
```diff
@@ -0,0 +1,206 @@
...
+```bash
+git clone https://github.com/ExchangeUnion/xud 
+cd xud
+git checkout tags/1.0.0-prealpha.1 -b 1.0.0-prealpha.1
+npm install
+```
+
+
+`xud` uses [MySQL](https://www.mysql.com/) or [MariaDB](https://mariadb.org/). You will have to install one of those and create a user *xud* and a database *xud* and grant all permissions for the new database to the *xud* user.
+
```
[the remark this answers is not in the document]
kilrau: Fixed

instructions-for-devs.md
```diff
@@ -0,0 +1,206 @@
...
+Use `./xucli --help` to get up-to-date, optional command line arguments to override defaults and settings in the [configuration](#configuration-optional) file for a specific `xud` instance
+
+```bash
+~/xud/bin $ ./xud --help
+Options:
+  --help              Show help                                        [boolean]
+  --version           Show version number                              [boolean]
+  --xudir, -x         Data directory for xud                            [string]
+  --db.database       SQL database name                                 [string]
+  --db.host           Hostname for SQL database                         [string]
```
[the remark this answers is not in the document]
kilrau: Good idea - We'll do that as part of the CLI --help command and then update the readme. https://github.com/ExchangeUnion/xud/issues/283

instructions-for-devs.md
```diff
@@ -0,0 +1,206 @@
...
+  --db.port           Port for SQL database                             [number]
+  --db.username       User for SQL database                             [string]
+  --lnd.certpath      Path to the SSL certificate for lnd               [string]
+  --lnd.disable       Disable lnd integration                          [boolean]
```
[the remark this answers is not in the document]
kilrau: it is?

instructions-for-devs.md
kilrau: It was. Somebody changed that. Acceptable for now though, we'll switch that to false once the lnd/raiden integration actually works.

instructions-for-devs.md
kilrau: Yes, fixed that as above.

PR #256, opened 2018-08-02
Replace hostId with peerId on orders

sangaman:
This replaces the `hostId` identifier on orders with `peerId`. Since orders are no longer being persisted between sessions, and since all orders for a peer are removed once that peer is disconnected, we no longer need to associate orders with a persistent identifier like `hostId`. This will make it easier to associate matched, swapped, or invalidated orders with an actual live peer.

This functionality is a prerequisite for #106 and #236.

### DIFF
--- a/lib/orderbook/MatchingEngine.ts
+++ b/lib/orderbook/MatchingEngine.ts
@@ -165,9 +165,9 @@ class MatchingEngine {
     return this.removeOrder(orderId) as StampedPeerOrder | undefined;
   }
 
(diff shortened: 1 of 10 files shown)


kilrau:
Will solve https://github.com/ExchangeUnion/xud/issues/273

sangaman:
Per @michael1011 's comment, I think it would be nice to have #256 merged before this so we don't need to make changes just to immediately throw them out after #256, so if you have a chance to review that soonish @moshababo that would be great.

PR #264, opened 2018-08-03
ListPeers gRPC call

sangaman:
With all the p2p testing and development going on, I figured this call would be handy. ListPeers returns key information about all connected peers. I've done some basic testing locally, here's what the output looks like:

```
xucli listpeers
{
  "peersList": [
    {
      "socketAddress": "localhost:9885",
      "nodePubKey": "123",
      "direction": 1,
      "supportedPairsList": [
        "BTC/LTC"
      ],
      "xudVersion": "123",
      "timeConnected": 433
    }
  ]
}
```


lib/p2p/Peer.ts
```diff
@@ -97,6 +107,17 @@ class Peer extends EventEmitter {
...
   }
 
+  public get info(): PeerInfo {
+    return {
+      socketAddress: this.id,
+      nodePubKey: this.handshakeState ? this.handshakeState.nodePubKey : undefined,
+      direction: this.direction,
+      supportedPairs: this.handshakeState ? this.handshakeState.pairs : undefined,
+      xudVersion: this.handshakeState ? this.handshakeState.version : undefined,
+      timeConnected: (this.connectTime - Date.now()) / 1000,
```
michael1011: Should be something like: `Math.round((Date.now() - this.connectTime) / 1000)` so that the value is positive and an integer.

lib/p2p/Pool.ts
```diff
@@ -41,7 +42,7 @@ class Pool extends EventEmitter {
     this.hosts = new HostList(new P2PRepository(db));
   }
 
-  get peerCount(): number {
+  public get peerCount(): number {
```
offerm: Is there any locking mechanism for the data structure to avoid parallel update/deletes?

offerm:
I suggest to change:
peerList -> peers
socketAddress -> address
direction -> inbound (with value true/false)
supportedPairsList -> pairList


There are many code changes which are not related to this PR.

lib/p2p/Pool.ts
sangaman: It's a single threaded process, so there won't be race conditions as far as I know.

sangaman:
Thanks @michael1011 and @offerm. I rebased and addressed your suggestions. I also went ahead and renamed `time_connected` to `seconds_connected` to make the output more clear.

Offer, there's no property called `peerList` to change, unless you mean the PeerList class, but I like leaving that as is. Definitely share other suggested changes unrelated to this PR when you get a chance.

sangaman:
@offerm I rebased and fixed a bug, you should be able to pull this down and test - it's working well for me:

```bash
C:\Users\Daniel\Documents\GitHub\xud>node bin/xucli listpeers
{
  "peersList": [
    {
      "address": "192.168.1.1:49195",
      "nodePubKey": "123",
      "inbound": true,
      "pairsList": [
        "BTC/LTC"
      ],
      "xudVersion": "123",
      "secondsConnected": 13
    },
    {
      "address": "75.75.58.83:8885",
      "nodePubKey": "123",
      "inbound": false,
      "pairsList": [
        "BTC/LTC"
      ],
      "xudVersion": "123",
      "secondsConnected": 13
    }
  ]
}
```

offerm:
Looks good now.

One change to consider (as part of this PR or as a follow-up PR): The address value that the output shows for inbound connection does not bring any value. for example 
```
    {
      "address": "127.0.0.1:65398",
      "nodePubKey": "0245780f9b2e6b7047cbe4d61c2b6d7434cc63c943dcd62c7ac604f936c76ed1c4",
      "inbound": true,
      "pairsList": [],
      "xudVersion": "1.0",
      "secondsConnected": 418
    },
```

The port 65398 is not the listen port but a dynamic port allocated for an accepted connection.  Instead, consider showing the listen port (if exists), or beak the address into  host and listen port fields (response object  only).


```
{
  "peersList": [
    {
      "address": "127.0.0.1:65398",
      "nodePubKey": "0245780f9b2e6b7047cbe4d61c2b6d7434cc63c943dcd62c7ac604f936c76ed1c4",
      "inbound": true,
      "pairsList": [],
      "xudVersion": "1.0",
      "secondsConnected": 418
    },
    {
      "address": "localhost:8895",
      "nodePubKey": "0245780f9b2e6b7047cbe4d61c2b6d7434cc63c943dcd62c7ac604f936c76ed1c4",
      "inbound": false,
      "pairsList": [],
      "xudVersion": "1.0",
      "secondsConnected": 418
    },
    {
      "address": "xud1.test.exchangeunion.com:8885",
      "nodePubKey": "123",
      "inbound": false,
      "pairsList": [
        "BTC/LTC"
      ],
      "xudVersion": "123",
      "secondsConnected": 165
    },
    {
      "address": "127.0.0.1:49161",
      "nodePubKey": "03ba461704d788c51778a5de0a15f290d18c8bc5b97e1244b04bfb1c2224135976",
      "inbound": true,
      "pairsList": [],
      "xudVersion": "1.0",
      "secondsConnected": 28
    }
  ]
}
```

sangaman:
You are right @offerm, but I think that should be taken care of when we are properly advertising connection info in the handshake.

Issue #242, opened 2018-07-31
Don't allow a peer to invalidate a peerOrder which he doesn't own


moshababo:
And penalize him if he tries to 😎

moshababo:
I can't recall we settled on any signing protocol yet. So the current situation is that a peer can invalidate other peers orders. The simplest solution would be to verify that the invalidation came form the same socket that the order came from (since we don't persist them). Or instead just use `hostId`, which is already being added to the order, and controlled by the local `xud`. 

If you're referring to signing the p2p packet, on the node pubKey, then it can still attempt to invalidate a different peer order. Are you talking about signing on the order id? or something else?

michael1011:
The peer should also be penalized if it tries to invalidate the id of an order that was never broadcasted.

kilrau:
I still don't get this. As I spoke with Daniel we already have a sign & verify function prepared and will use it in various places, e.g. signing the handshake `nodePubKey` p2p packet (https://github.com/ExchangeUnion/xud/issues/341).

So I guess my question is:
Why don't we make anyone wanting to invalidate an order *sign* a this invalidation packet and anyone getting this invalidation packet *verify* that the order was signed by the `nodePubKey`s private key of the order.

sangaman:
Even without signing, we should be able to tell because we'll know the peer's pubkey and can see that it differs from the pubkey he is trying to invalidate. We should probably discuss some more about which packets actually need to be signed, it's possible that it's enough to only sign the handshake. If we encrypt p2p traffic, that will take care of this also.

kilrau:
You are right, since we already verified the pubkey via signature on handshake, we can just make sure that the invalidation packet comes from the same socket. Since we have direct socket connections with everyone, this works 👍 

Encrypting will definitely solve that too, but is *optional* (whitelist mode).

sangaman:
Encrypting traffic might be worthwhile even without whitelist mode, but it's a separate discussion. It would prevent MITM attacks and snooping.

kilrau:
agree, I already put that discussion to a later stage: https://github.com/ExchangeUnion/xud/issues/158

kilrau:
So if all (@moshababo @offerm @michael1011 @itsbalamurali ) agree, this task now is about:
check if invalidation packet is coming from the same socket connection than the order.

https://github.com/ExchangeUnion/xud/issues/341 is a prerequisite

michael1011:
Yeah that sounds reasonable.

[...]

PR #265, opened 2018-08-03
Added order cancellations to subscribePeerOrders

### LINKED ISSUE #194: Implement subscribePeerOrders
Pass to subscribed client all incoming peer order, and all order cancellations (updates).
In addition, all existing peer orders need to be published as well (the client need to be fully synced to `xud`). We need to find a good technical solution on how to sync between them, so we won't get any duplicates or missing orders. 

We can start by only implementing the publishing of new incoming orders, as it's the most important and simple one (for prealpha release).

michael1011:
Closes #194 for the prealpha

I added the logic in the orderbook to the constructor for now but will move it into an own function when I take care of the handling of partially filled orders.

This PR also fixes some minor bugs but for which I didn't want to open a separate PR.


lib/orderbook/OrderBook.ts
```diff
@@ -121,14 +127,17 @@ class OrderBook extends EventEmitter {
     return this.addOwnOrder({ ...order, price }, true);
   }
 
-  public removeOwnOrderByLocalId = (pairId: string, localId: string): boolean => {
+  public removeOwnOrderByLocalId = (pairId: string, localId: string): { removed: boolean, globalId: string } => {
     const id = this.localIdMap[localId];
 
     if (id === undefined) {
-      return false;
+      return { removed: false, globalId: '' };
```
itsbalamurali: globalId can be undefined rather than a empty string.

lib/p2p/Pool.ts
```diff
@@ -128,8 +128,8 @@ class Pool extends EventEmitter {
     // TODO: send only to peers which accepts the pairId
   }
 
-  public broadcastOrderInvalidation = (orderId: string, pairId: string) => {
-    const orderInvalidationPacket = new OrderInvalidationPacket({ orderId, pairId });
+  public broadcastOrderInvalidation = (id: string, pairId: string) => {
```
itsbalamurali: any reason behind renaming `orderId` to `id`? `orderId` was much clearer and easily understood at first glance 😄

lib/service/Service.ts
```diff
@@ -181,8 +181,12 @@ class Service extends EventEmitter {
    * Cancel placed order from the orderbook.
    */
   public cancelOrder = async ({ orderId, pairId }: { orderId: string, pairId: string }) => {
```
itsbalamurali: Please check the return types and values.

proto/xudrpc.proto
```diff
@@ -188,7 +188,7 @@ message Order {
   // The id of the peer that created this order
   int32 peer_id = 4 [json_name = "peer_id"];
```
itsbalamurali: i guess we have no more `host_id` its just `peer_id` alone as per @sangaman's recent PR #256

michael1011: #256 wasn't merged yet. The one who gets merged after the other will have to resolve the conflict.

lib/service/Service.ts
michael1011: What do you mean? I am returning the boolean  `canceled` which is the exact same value as before.

itsbalamurali: @michael1011 never mind i overlooked 👍

lib/p2p/Pool.ts
michael1011: I thought that is kind of redundant because there is already 'order' in the name of the function and it makes the callback of `subscribePeerOrders` in `Service.ts` more elegant because I can use `{ cancelled: true, ...order }` instead of assigning the value of `orderId` to `id` and removing the `orderId` property.

I revert the change if you want to though? @itsbalamurali

proto/xudrpc.proto
```diff
@@ -77,7 +77,7 @@ service Xud {
     };
   }
   
-  /* Subscribe to incoming peer orders. */
+  /* Subscribe to incoming peer orders and cancellations. */
```
sangaman: I also think I prefer "Subscribe to peer order events" here. I think eventually it would be good to add more detailed comments to our proto file to clearly define how each call works, and there we can elaborate on exactly which events are covered like new incoming orders, orders being canceled, orders being filled between other peers, and how orders can be only partially filled or invalidated. But for a one-liner that's my preference.

proto/xudrpc.proto
sangaman: You can leave the field unchanged though right? Or is there a reason this has to be changed? I do think you're right that it *should* be an int but it's gonna get dropped anyway.

proto/xudrpc.proto
```diff
@@ -197,6 +197,8 @@ message Order {
   int64 created_at = 8 [json_name = "created_at"];
   // Lightning invoice
   string invoice = 9 [json_name = "invoice"];
+  // Indicates whether an order was cancelled 
+  bool cancelled = 10 [json_name = "cancelled"];
```
sangaman: The American in me thought this was spelled "canceled" and upon googling it turns out that both are correct and "cancelled" is the British spelling while "canceled" is the American spelling. I think we should change this though because we're already using "canceled" elsewhere in the code (see `CancelOrderResponse`) but not "cancelled" and we should stick to one spelling convention. "cancelled" is also used a few other places in this PR so just do a find all.

michael1011:
@sangaman I did *some* testing locally. It is fine if this PR gets moved to the milestone `1.0.0-prealpha.2` and gets merged sometime next week.

@kilrau should this PR test too because it will solve #273.

sangaman:
I think this just needs to have the conflicts resolved and then can be merged, thanks.

lib/orderbook/OrderBook.ts
sangaman: I think we should change this to `undefined` as well.

lib/orderbook/OrderBook.ts
```diff
@@ -51,10 +53,14 @@ class OrderBook extends EventEmitter {
     this.repository = new OrderBookRepository(models);
     if (pool) {
       pool.on('packet.order', this.addPeerOrder);
-      pool.on('packet.orderInvalidation', body => this.removePeerOrder(body.orderId, body.pairId)); // TODO: implement quantity invalidation
+      pool.on('packet.orderInvalidation', (body: orders.OrderIdentifier) => {
+        // TODO: implement quantity invalidation
+        if (this.removePeerOrder(body.id, body.pairId)) {
```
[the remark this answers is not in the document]
sangaman: Could you put `// TODO: handle case of bad invalidation packet that doesn't remove an order, penalize peer` here?

lib/service/Service.ts
```diff
@@ -209,10 +213,11 @@ class Service extends EventEmitter {
   }
 
   /*
-   * Subscribe to incoming peer orders.
+   * Subscribe to incoming peer orders and cancellations.
```
sangaman: For consistency's sake let's have this say `Subscribe to peer order events.` as well.

lib/types/orders.ts
```diff
@@ -36,7 +36,7 @@ export type OutgoingOrder = MarketOrder & {
 };
 
 export type OrderIdentifier = {
-  orderId: string;
+  id: string;
```
sangaman: I'd prefer sticking with `orderId` here as well just for extra clarity.

michael1011:
@sangaman I wrote about penalizing peers for invalidating orders that don't exist #242 because penalizing for invalidating an order that isn't its is closely related.

sangaman:
OK that works for me.

lib/orderbook/OrderBook.ts
```diff
@@ -51,10 +53,14 @@ class OrderBook extends EventEmitter {
     this.repository = new OrderBookRepository(models);
     if (pool) {
       pool.on('packet.order', this.addPeerOrder);
-      pool.on('packet.orderInvalidation', body => this.removePeerOrder(body.orderId, body.pairId)); // TODO: implement quantity invalidation
+      pool.on('packet.orderInvalidation', (body: orders.OrderIdentifier) => {
+        // TODO: implement quantity invalidation
+        if (this.removePeerOrder(body.orderId, body.pairId)) {
+          this.emit('peerOrder.invalidation', body);
```
moshababo: It would be better to emit from `removePeerOrder`. 
Mainly because that's how you did it on the batch removal (`removePeerOrders`), but also to disallow removing without emitting.

lib/service/Service.ts
```diff
@@ -209,10 +213,16 @@ class Service extends EventEmitter {
...
 
   /*
-   * Subscribe to incoming peer orders.
+   * Subscribe to peer order events.
    */
   public subscribePeerOrders = async (callback: Function) => {
-    this.orderBook.on('peerOrder', order => callback(order));
+    this.orderBook.on('peerOrder.incoming', order => callback(order));
+    this.orderBook.on('peerOrder.invalidation', order => callback({
+      canceled: true,
```
moshababo: I think this interface design is a bit problematic.
What about having an enum, or boolean, to represent the event type, and then the payload? 
`{ type/isCancellation, payload }`

We can discuss that separately, for a different PR. Don't let this delay you from merging, if it's blocking #266

michael1011: I was thinking that we will rework the gRPC messages in the alpha stage anyway and just adding the `canceled` boolean was the simplest solution so that it can get merged for prealpha.

But I agree on the new design you described. Could you write it down in an issue so that we don't forget about it?

lib/orderbook/OrderBook.ts
michael1011: I will fix that in #266 because I had to resolve conflicts in this line anyway.

lib/service/Service.ts
moshababo: Yep, sure.

kilrau: Just did that: https://github.com/ExchangeUnion/xud/issues/298

PR #266, opened 2018-08-03
Order invalidation implementation

### LINKED ISSUE #176: Split order ids

michael1011:
Closes #176 

The `OrderBook` sends the `OrderInvalidationPacket` after a match. After the Raiden swap integration the packet should be sent after a successful swap. And the `OrderBook` is able to handle `OrderInvalidationPacket` with a quantity (= partially filled orders).

And I would prefer that #265 gets merged first.

### DIFF
--- a/lib/orderbook/MatchesProcessor.ts
+++ b/lib/orderbook/MatchesProcessor.ts
@@ -31,7 +31,6 @@ class MatchesProcessor {
 
   public executeSwap(_args) {
     // TODO: execute the swap procedure
-    // TODO: invalidate the maker if it is of the type ownOrder
   }
 }
 
--- a/lib/orderbook/MatchesProcessor.ts
+++ b/lib/orderbook/MatchesProcessor.ts
@@ -31,6 +31,7 @@ class MatchesProcessor {
 
   public executeSwap(_args) {
     // TODO: execute the swap procedure
+    // TODO: invalidate the maker if it is of the type ownOrder
   }
 }
 
--- a/lib/orderbook/MatchingEngine.ts
+++ b/lib/orderbook/MatchingEngine.ts
@@ -161,8 +161,18 @@ class MatchingEngine {
     return this.removeOrder(orderId) as StampedOwnOrder | undefined;
   }
 
-  public removePeerOrder = (orderId: string): StampedPeerOrder | undefined => {
-    return this.removeOrder(orderId) as StampedPeerOrder | undefined;
+  public removePeerOrder = (orderId: string, decreasedQuantity?: number): StampedPeerOrder | undefined => {
+    const order = this.removeOrder(orderId) as StampedPeerOrder | undefined;
+
+    if (order && decreasedQuantity) {
+      order.quantity = order.quantity - decreasedQuantity;
+      this.addPeerOrder(order);
+
+      // Return how much was removed
+      return { ...order, quantity: decreasedQuantity };
+    }
+
+    return order;
   }
 
   public removePeerOrders = (peerId: string): StampedPeerOrder[] => {
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -11,6 +11,7 @@ import Logger from '../Logger';
 import LndClient from '../lndclient/LndClient';
 import { ms } from '../utils/utils';
 import { Models } from '../db/DB';
+import { OrderIdentifier } from '../types/orders';
 
 type OrdersMap = Map<String, Orders>;
 
@@ -37,7 +38,7 @@ class OrderBook extends EventEmitter {
 
   private logger: Logger = Logger.orderbook;
   private repository: OrderBookRepository;
-  private matchesProcessor: MatchesProcessor = new MatchesProcessor();
+  private matchesProcessor = new MatchesProcessor();
 
   private ownOrders: OrdersMap = new Map<String, Orders>();
   private peerOrders: OrdersMap = new Map<String, Orders>();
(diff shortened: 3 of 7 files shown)


lib/orderbook/OrderBook.ts
```diff
@@ -296,6 +315,16 @@ class OrderBook extends EventEmitter {
...
   private handleMatch = ({ maker, taker }): void => {
     this.logger.debug(`order match: ${JSON.stringify({ maker, taker })}`);
+    // TODO: send the packet after a successful the swap
+    if (this.pool) {
+      if (!this.isOwnOrder(maker)) {
+        this.pool!.broadcastOrderInvalidation({
+          orderId: maker.id,
+          pairId: maker.pairId,
+          quantity: maker.quantity,
+        });
```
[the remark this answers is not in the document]
michael1011: Makes sense when keeping #242 in mind.

lib/orderbook/OrderBook.ts
```diff
@@ -132,6 +131,26 @@ class OrderBook extends EventEmitter {
...
+          this.updateOrderQuantity(this.peerOrders, order, body.quantity);
+        } else {
+          this.logger.verbose(`Invalid order id: ${body.orderId}`);
+        }
+      } else {
+        this.logger.warn(`Invalid pairId: ${body.pairId}`);
+      }
+    } else {
+      this.removePeerOrder(body.orderId, body.pairId);
+    }
```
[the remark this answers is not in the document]
moshababo: Actually that's not the case - the orders maps and `MatchingEngine` don't hold the same instance, if the order was previously split. So I think this should all be implemented under `removePeerOrder` (which now should support `quantity` too), and propagate down to `MatchingEngine` as well.

lib/orderbook/MatchingEngine.ts
```diff
@@ -161,8 +161,18 @@ class MatchingEngine {
...
 
-  public removePeerOrder = (orderId: string): StampedPeerOrder | undefined => {
-    return this.removeOrder(orderId) as StampedPeerOrder | undefined;
+  public removePeerOrder = (orderId: string, decreasedQuantity: number | undefined = undefined): StampedPeerOrder | undefined => {
+    const order = this.removeOrder(orderId) as StampedPeerOrder | undefined;
+
+    if (order && decreasedQuantity) {
+      order.quantity = order.quantity - decreasedQuantity;
+      this.addPeerOrder(order);
+
```
moshababo: instead of removing & adding, we can possibly just do an update.
`quantity` doesn't affect the order priority, so it shouldn't create an issue if we'll just update it. 
I was looking at the priority queue `forEach` method, but it seems to copy the entire queue before being executed, which might make it worse then removing & adding.
We might be able to do an item lookup similarly to how we currently remove items, just that we'll expose it instead of removing it. 
We can take care of this later on, but tell me if you want to open an issue for this. (@sangaman)

lib/orderbook/OrderBook.ts
```diff
@@ -210,10 +217,14 @@ class OrderBook extends EventEmitter {
...
 
-  private updateOrderQuantity = (order: orders.StampedOrder, decreasedQuantity: number) => {
+  private updateOrderQuantity = (order: orders.StampedOrder, decreasedQuantity: number): boolean => {
     const isOwnOrder = orders.isOwnOrder(order);
     const orderMap = this.getOrderMap(isOwnOrder ? this.ownOrders : this.peerOrders, order);
 
+    if (!orderMap[order.id]) {
+      return false;
+    }
+
```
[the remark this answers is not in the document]
michael1011: If we just get the `id` of the order as argument we would have to make two additional lookups:
- checking if the order is either a buy or sell one. Right now we determine it by the quantity
- check if the order is an `ownOrder` by looking up the `id` parameter in `localIdMap`

So I would prefer if we keep and argument of the type `orders.StampedOrder`

lib/orderbook/MatchesProcessor.ts
```diff
@@ -31,6 +31,7 @@ class MatchesProcessor {
 
   public executeSwap(_args) {
     // TODO: execute the swap procedure
+    // TODO: invalidate the maker if it is of the type ownOrder
```
moshababo: This function will probably be used only to instantiate swaps, so the maker order can't be an `ownOrder`, and no invalidation is required. In this case, I think we should remove this comment.

lib/orderbook/OrderBook.ts
```diff
@@ -156,18 +152,35 @@ class OrderBook extends EventEmitter {
...
     }
 
-    if (matchingEngine.removePeerOrder(orderId)) {
-      return this.removeOrder(this.peerOrders, orderId, pairId);
-    } else {
-      return false;
+    const order = ordersMap[orderId];
+
+    if (order) {
+      if (matchingEngine.removePeerOrder(orderId, decreasedQuantity)) {
```
moshababo: Sorry if my previous comment wasn't clear, but I meant doing `if (order && matchingEngine...)`

lib/orderbook/MatchingEngine.ts
sangaman: I agree actually that we should be updating the existing order. We can just get the order by id and then change the quantity.

Issue #231, opened 2018-07-30
Add "disconnected from peer" console output


kilrau:
peer 1 initiates connection to peer 2.

Current behavior:
on disconnect, peer 2, sees "disconnected from peer"
on disconnect, peer 1, doesn't see anything

How it should be:
on disconnect, both see "Disconnected from <PubKey> <IP> <Port>"

PR #281, opened 2018-08-06
Peer disconnect log message

itsbalamurali:
This PR implements disconnected peer message logging (#231)


michael1011:
Weirdly just one of the Travis tests is failing. Could you trigger the failing one to run again @sangaman

sangaman:
It still fails for the same reason. The pr build passes because it's using the `fastpriorityqueue` version published to npm, whereas the `push` version (the feature branch) looks like it's still using the github version which is causing problems. I figure it should be fine to merge, but it'd be nice to clear up the issue by getting the latest from master if we can.

PR #293, opened 2018-08-07
Improved getOrders

### LINKED ISSUE #258: ./xucli getorders to return ownOrders + peerOrders
Current output only contains peerOrders:
```
{
  "buyOrdersList": [],
  "sellOrdersList": []
}

```
It should include ownOrders:
```
  "buyownOrdersList": [],
  "sellownOrdersList": [],
  "buypeerOrdersList": [],
  "sellpeerOrdersList": []
```

michael1011:
Fixes #258 

This PR:
- adds `ownOrders` to the `getOrders` command
- fixes a bug that caused `getOrders` to show nothing
- makes the argument `pair_id` mandatory because handling `max_results` with multiple pairs adds unnecessary complexity

### DIFF
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -56,6 +56,44 @@ class Service extends EventEmitter {
     this.logger = Logger.rpc;
   }
 
+  /*
+   * Cancel placed order from the orderbook.
+   */
+  public cancelOrder = async ({ orderId, pairId }: { orderId: string, pairId: string }) => {
+    const { removed, globalId } = this.orderBook.removeOwnOrderByLocalId(pairId, orderId);
+
+    if (removed) {
+      this.pool.broadcastOrderInvalidation({
+        pairId,
+        orderId: globalId,
+      });
+    }
+    return { canceled: removed };
+  }
+
+  /**
+   * Connect to an XU node on a given host and port.
+   */
+  public connect = async ({ host, port }: { host: string, port: number }) => {
+    const peer = await this.pool.addOutbound(new SocketAddress(host, port));
+    return peer.getStatus();
+  }
+
+  /*
+   * Disconnect from a connected peer XU node on a given host and port.
+   */
+  public disconnect = async ({ host, port }: { host: string, port: number}) => {
+    await this.pool.closePeer(host, port);
+    return 'success';
+  }
+
+  /**
+   * Execute an atomic swap
+   */
+  public executeSwap = async ({ target_address, payload, identifier }: { target_address: string, payload: TokenSwapPayload, identifier: string }) => {
+    return this.raidenClient.tokenSwap(target_address, payload, identifier);
+  }
+
   /**
    * Get general information about this Exchange Union node.
    */
@@ -112,13 +150,10 @@ class Service extends EventEmitter {
 
     if (!this.config.raiden.disable) {
       try {
-        const [address, channels] = await Promise.all([
-          this.raidenClient.getAddress(),
-          this.raidenClient.getChannels(),
-        ]);
+        const channels = await this.raidenClient.getChannels();
 
         info.raiden = {
-          address,
+          address: this.raidenClient.address!,
           channels: channels.length,
           // Hardcoded for now until they expose it to their API
           version: 'v0.3.0',
@@ -133,14 +168,6 @@ class Service extends EventEmitter {
     }
 
     return info;
-  }
-
-  /**
-   * Get the list of the order book's available pairs.
-   * @returns A list of available trading pairs
-   */
-  public getPairs = () => {
-    return this.orderBook.getPairs();
   }
 
   /**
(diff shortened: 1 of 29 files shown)


lib/service/Service.ts
```diff
@@ -144,26 +144,35 @@ class Service extends EventEmitter {
   }
 
   /**
-   * Get a list of standing peer orders from the order book for a specified trading pair, or for all
-   * trading pairs if no pair is specified.
+   * Get a list of standing peer orders from the order book for a specified trading pair
```
sangaman: Add a period at the end of the comment.

lib/service/Service.ts
```diff
@@ -144,26 +144,35 @@ class Service extends EventEmitter {
...
    */
   public getOrders = ({ pairId, maxResults }: { pairId?: string, maxResults?: number }) => {
```
sangaman: I think here actually `pairId` and `maxResults` will never be undefined, only the default protobuf values of empty string and 0, so we can drop the `?` nullable operators from the method signature.

lib/service/Service.ts
```diff
@@ -144,26 +144,35 @@ class Service extends EventEmitter {
...
-    const ret: OrderArrays = {
-      buyOrders: [],
-      sellOrders: [],
+    checkArgument(pairId !== undefined && pairId !== '', 'pairId has to be specified');
```
sangaman: Likewise I think we can drop the `undefined` check here and just check for empty string. Also let's say `pairId must be specified` here just to match the language of the other `checkArgument` messages.

lib/service/Service.ts
sangaman: Also, it's no longer returning only peer orders so we should remove "peer" here.

lib/service/Service.ts
```diff
@@ -144,26 +144,35 @@ class Service extends EventEmitter {
...
-    } else {
-      this.orderBook.pairs.forEach((pair) => {
-        const orderArrays = this.orderBook.getPeerOrders(pair.id, maxResults ? maxResults : 0);
-        ret.buyOrders.concat(orderArrays.buyOrders);
-        ret.sellOrders.concat(orderArrays.sellOrders);
-      });
-    }
-    return ret;
+
+    const concatOrders = (orders: OrderArrays, peerOrders: boolean) => {
```
sangaman: Since we're no longer concatenating results from multiple pairs, I don't think we need any of this `concat` stuff any more. You can just do `getPeerOrders` and `getOwnOrders` and return them in an object as `peerOrders` and `ownOrders` properties.

proto/xudrpc.proto
```diff
@@ -229,10 +236,10 @@ message GetOrdersRequest {
   uint32 max_results = 2 [json_name = "max_results"];
 }
 message GetOrdersResponse {
-  // A list of buy orders sorted by descending price
-  repeated Order buy_orders = 1 [json_name = "buy_orders"];
-  // A list of sell orders sorted by ascending price
-  repeated Order sell_orders = 2 [json_name = "sell_orders"];
+  // A list of the orders of peers
```
sangaman: I think I prefer just "A list of peer orders" - it's short and simple but I'm not sure what else would be useful to say.

proto/xudrpc.proto
```diff
@@ -229,10 +236,10 @@ message GetOrdersRequest {
...
+  Orders peer_orders = 1 [json_name = "peer_orders"];
+  // A list of your orders
```
sangaman: The use of second person "your" here feels out of place. Maybe "A list of orders placed locally" or "A list of orders belonging to this xud instance."

lib/service/Service.ts
```diff
@@ -144,26 +144,26 @@ class Service extends EventEmitter {
...
-   * Get a list of standing peer orders from the order book for a specified trading pair, or for all
-   * trading pairs if no pair is specified.
+   * Get a list of standing orders from the order book for a specified trading pair.
    */
-  public getOrders = ({ pairId, maxResults }: { pairId?: string, maxResults?: number }) => {
-    const ret: OrderArrays = {
-      buyOrders: [],
-      sellOrders: [],
+  public getOrders = ({ pairId, maxResults }: { pairId: string, maxResults: number }) => {
+    checkArgument(pairId !== '', 'pairId must be specified');
```
sangaman: Maybe add one more `checkArgument` here to make sure maxResults is not negative.

lib/service/Service.ts
```diff
@@ -144,26 +144,26 @@ class Service extends EventEmitter {
...
+
+    const result: { [ type: string ]: OrderArrays } = {
+      peerOrders: {
```
sangaman: You can combine these lines with the ones below rather than initializing with empty arrays:

```javascript
const result = {
  peerOrders: this.orderBook.getPeerOrders(pairId, maxResults),
  ownOrders: this.orderBook.getOwnOrders(pairId, maxResults),
}
```

Issue #276, opened 2018-08-03
Order matching doesn't work


kilrau:
I mentioned this [here](https://github.com/ExchangeUnion/xud/pull/244#issuecomment-410115395) before, matching doesn't work. First order (xud1) goes thru, second order (xud2) throws:

```
kilrau@ubuntu:~/xud/bin$ ./xucli -p 54322 placeorder BTC/LTC 1 1 99
(node:6225) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
Error: 2 UNKNOWN: Cannot read property 'quantity' of undefined
```
```
8/3/2018, 4:43:22 PM [ORDERBOOK] debug: order match: {"maker":{"price":99,"quantity":-1,"pairId":"BTC/LTC","peerId":0,"hostId":4,"id":"ddeafc60-9776-11e8-9485-d51bda3e6ac4","invoice":"dummyInvoice","createdAt":1533339741996},"taker":{"price":99,"quantity":1,"pairId":"BTC/LTC","peerId":0,"hostId":"","id":"02234b50-9777-11e8-a286-430dcc42b522","localId":"1","createdAt":1533339802757,"invoice":""}}
8/3/2018, 4:43:22 PM [RPC] error: TypeError: Cannot read property 'quantity' of undefined
    at OrderBook.updateOrderQuantity (/home/kilrau/xud/dist/orderbook/OrderBook.js:182:62)
    at matches.forEach (/home/kilrau/xud/dist/orderbook/OrderBook.js:151:26)
    at Array.forEach (<anonymous>)
    at OrderBook.addOwnOrder (/home/kilrau/xud/dist/orderbook/OrderBook.js:149:25)
    at OrderBook.addLimitOrder (/home/kilrau/xud/dist/orderbook/OrderBook.js:94:25)
    at Service.<anonymous> (/home/kilrau/xud/dist/service/Service.js:132:53)
    at Generator.next (<anonymous>)
    at /home/kilrau/xud/dist/service/Service.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (/home/kilrau/xud/dist/service/Service.js:3:12)
    at Service.placeOrder (/home/kilrau/xud/dist/service/Service.js:129:38)
    at GrpcService.<anonymous> (/home/kilrau/xud/dist/grpc/GrpcService.js:38:43)
[...]

Issue #253, opened 2018-08-02
[Crash] on connect


kilrau:
latest master, latest p2p-seeding - same behavior:

```
kilrau@ubuntu:~/xud/bin$ ./xud
8/2/2018, 12:18:01 AM [GLOBAL] info: config loaded
Executing (default): SELECT 1+1 AS result
8/2/2018, 12:18:02 AM [DB] info: connected to database. host:localhost port:3306 database:xud
8/2/2018, 12:18:02 AM [DB] verbose: Executing (default): CREATE TABLE IF NOT EXISTS `hosts` (`id` INTEGER auto_increment , `address` VARCHAR(39) NOT NULL, `port` INTEGER NOT NULL, `pubKey` VARCHAR(255) UNIQUE, `createdAt` DATETIME NOT NULL, `updatedAt` DATETIME NOT NULL, PRIMARY KEY (`id`)) ENGINE=InnoDB;
8/2/2018, 12:18:02 AM [DB] verbose: Executing (default): CREATE TABLE IF NOT EXISTS `bannedHosts` (`address` VARCHAR(39) , `createdAt` DATETIME NOT NULL, `updatedAt` DATETIME NOT NULL, PRIMARY KEY (`address`)) ENGINE=InnoDB;
8/2/2018, 12:18:02 AM [DB] verbose: Executing (default): CREATE TABLE IF NOT EXISTS `currencies` (`id` VARCHAR(255) , PRIMARY KEY (`id`)) ENGINE=InnoDB;
8/2/2018, 12:18:02 AM [DB] verbose: Executing (default): SHOW INDEX FROM `hosts`
8/2/2018, 12:18:02 AM [DB] verbose: Executing (default): SHOW INDEX FROM `bannedHosts`
8/2/2018, 12:18:02 AM [DB] verbose: Executing (default): SHOW INDEX FROM `currencies`
[...]

PR #299, opened 2018-08-07
Enforce noImplicitAny rule

sangaman:
This enables the `noImplicitAny` typescript compiler option and makes the necessary changes to comply with the new rule, which include:

- Moving our GrpcExpressMiddleware.ts file to its own `grpc-dynamic-gateway` package (forked from its parent repository) and adding it to the project dependencies. This also required updating `grpc` to the latest version to match what's used in `grpc-dynamic-gateway`.
- Moving the `PriorityQueue` interface to type definitions for our forked `fastpriorityqueue` package.
- Adding various @types packages to define types for packages in use.
- Adding a blank type definition for the `winston` package, which currently does not have types defined for v3.0.0, as a temporary hack. Type definitions should be bundled in the next release of `winston`, at which point this blank definition can be removed.
- Adding an `ORDER_NOT_FOUND` error to handle the cases where we unexpectedly get back `undefined` when looking up an order.
- Fixing a bug in `removePeerOrder `that was incorrectly retrieving an order from `ordersMap`.
- Explicitly defining types anywhere else in our code base where `any` was implied.

I think this will be a valuable rule for preventing bugs now that we are starting a release cycle. Just recently issues #276, #253, and #229 resulted from an `undefined` value on an implicit `any` type, and this should prevent that.


lib/orderbook/OrderBook.ts
```diff
@@ -11,20 +11,19 @@ import Logger from '../Logger';
 import LndClient from '../lndclient/LndClient';
 import { ms } from '../utils/utils';
 import { Models } from '../db/DB';
-import { OrderIdentifier } from '../types/orders';
 
-type OrdersMap = Map<String, Orders>;
+type OrdersMap = Map<string, Orders>;
```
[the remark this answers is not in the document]
sangaman: `String` is an object, `string` is a primitive. I'm generally used to seeing `string` used (that's the type of any string literals) and it's what we're using through most of the code base. They behave a little differently, so I wanted to standardize everything on `string`.

tsconfig.json
```diff
@@ -7,25 +7,20 @@
...
 
     /* Additional Checks */
-      "noUnusedLocals": false,               /* Report errors on unused locals. */
-      "noUnusedParameters": true,            /* Report errors on unused parameters. */
-      "noImplicitReturns": true,             /* Report error when not all code paths in function return a value. */
-      "noFallthroughCasesInSwitch": true,    /* Report errors for fallthrough cases in switch statement. */
+    "noUnusedLocals": false,                  /* Report errors on unused locals. */
+    "noUnusedParameters": true,               /* Report errors on unused parameters. */
+    "noImplicitReturns": true,                /* Report error when not all code paths in function return a value. */
+    "noFallthroughCasesInSwitch": true,       /* Report errors for fallthrough cases in switch statement. */
```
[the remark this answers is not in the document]
sangaman: I don't see any whitespace at the end of the line, or do you mean between the comma and the comment? I put that there so all the comments are vertically aligned.

lib/orderbook/MatchingEngine.ts
```diff
@@ -62,8 +46,8 @@ class MatchingEngine {
     };
   }
 
-  public static getMatchingQuantity = (buyOrder: StampedOrder, sellOrder: StampedOrder): number => {
-    if (buyOrder.price >= sellOrder.price) {
+  public static getMatchingQuantity = (buyOrder?: StampedOrder, sellOrder?: StampedOrder): number => {
```
moshababo: why params are optional?

lib/orderbook/OrderBook.ts
```diff
@@ -154,15 +163,14 @@ class OrderBook extends EventEmitter {
...
-    const ordersMap = this.peerOrders[pairId];
+    const ordersMap = this.peerOrders.get(pairId);
     if (!matchingEngine || !ordersMap) {
       this.logger.warn(`Invalid pairId: ${pairId}`);
       return false;
     }
 
-    const order = ordersMap[orderId];
-
-    if (order && matchingEngine.removePeerOrder(orderId, decreasedQuantity)) {
```
moshababo: was this causing a bug, or you just preferred to rely solely on finding the order in `MatchingEngine`?

lib/orderbook/OrderBook.ts
```diff
@@ -238,23 +246,21 @@ class OrderBook extends EventEmitter {
     });
   }
 
-  private updateOrderQuantity = (order: orders.StampedOrder, decreasedQuantity: number): boolean => {
+  private updateOrderQuantity = (order: orders.StampedOrder, quantityToDecrease: number) => {
```
moshababo: renaming is good, but there are many `decreasedQuantity` usages in other places. should rename them all.

sangaman:
I resolved conflicts and addressed the code redundancy in OrderBook.ts in a separate commit - I also added a few comments and did some variable renaming to try to make it more clear.

Typescript 3.0 is definitely something we should move to eventually. I'd read the release notes and nothing jumped out as me as being immediately useful which is why I'm ok with staying with the current version. I also think after this PR is merged it would make sense to search the codebase for explicit `any` types and try to replace them with a real type where it's practical to do so, maybe along the way we'd come across a variable that would make sense to be `unknown` at which point we should definitely upgrade. It doesn't look like the upgrade would break anything for us.

lib/orderbook/OrderBook.ts
sangaman: Yes it's a bug, I'd mentioned it:

> Fixing a bug in removePeerOrderthat was incorrectly retrieving an order from ordersMap.

`ordersMap` maps to an `Orders` type which has `buyOrders` and `sellOrders` - not order Ids - as keys. However, `MatchingEngine.removePeerOrder` returns the order we're interested in for us so that's what I used.

lib/orderbook/OrderBook.ts
sangaman: Will do, thanks.

lib/orderbook/MatchingEngine.ts
sangaman: I could maybe handle this differently by doing undefined checks before ever calling this method. But basically lines 75 and 76 define a method and line 81 calls it with `oppositeOrder` which can be `undefined` because `peek()` will return `undefined` if the queue is empty.

moshababo: I see, I thought it shouldn't be a problem because we're checking `isEmpty` before, but obviously it doesn't matter for the return type. I'd prefer that this method won't be effected by these issues, but do whatever you think is better.

sangaman: Yeah I see what you're saying, I'll refactor this a bit.

sangaman:
I renamed all instances of `decreasedQuantity` to `quantityToDecrease` to make it more clear, and I refactored `getMatchingQuantity` not to accept undefined parameters. Instead, we do a quick check of the `peek` result before calling `getMatchingQuantity`. That's definitely a more sensible approach than before.

PR #302, opened 2018-08-08
Grpc error handling

### LINKED ISSUE #116: Improve gRPC error handling and status code

### LINKED ISSUE #272: [Crash] Disconnect an already disconnected peer
Same as with the connect to already connected host, disconnect of an already disconnected host leads to a crash of `xud`. As I'm sure we could dig up more of these, will https://github.com/ExchangeUnion/xud/pull/218 fix all of these error handling issues? If yes, let's make it part of `1.0.0-prealpha.2` for next week.
```bash
kilrau@ubuntu:~/xud/bin$ ./xucli -p 1235 disconnect localhost 54321
(node:4746) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
Error: 14 UNAVAILABLE: TCP Read failed

kilrau@ubuntu:~/xud/bin$ ./xucli -p 1235 disconnect localhost 54321
(node:4746) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
Error: 14 UNAVAILABLE: TCP Read failed
```

sangaman:
Fixes #272. Closes #116 - although the issue highlighted there will be a continuous effort but one that is handled about as well as we can for now. This handles all our defined errors within the project and asserts that parameters are valid. I left `getOrders` alone so as not to conflict with #293, as well as the raiden related errors since that's not really in a complete state.

I also added a couple of comments here to our proto file.


lib/service/Service.ts
```diff
@@ -193,6 +196,9 @@ class Service extends EventEmitter {
    * Connect to an XU node on a given host and port.
    */
   public connect = async ({ host, port }: { host: string, port: number }) => {
+    checkArgument(host !== '', 'host must be specified');
+    checkArgument(port > 1023 && port < 65536, 'port must be an integer between 1024 and 65535');
```
michael1011: I think we should wrap `checkArgument` calls we use multiple times into a function or at least define the messages of those calls in one place so that they stay consistent.

lib/service/Service.ts
```diff
@@ -178,6 +178,9 @@ class Service extends EventEmitter {
    * Cancel placed order from the orderbook.
    */
   public cancelOrder = async ({ orderId, pairId }: { orderId: string, pairId: string }) => {
+    checkArgument(orderId !== '', 'orderId must be specified');
+    checkArgument(pairId !== '', 'pairId must be specified');
```
michael1011: What about checking the `pairId` and `orderId` of `placeOrder`?

sangaman: That already has these checks, those were put in previously.

michael1011: It has checks for that in the orderbook not in `Service.ts`. And those checks in the orderbook are just making sure that the values are correct and not if the argument was specified which will result in a different error code.

It fine if we leave it the way it is right now.

sangaman: Oh oops, you're absolutely right, my eyes had tricked me. I'll address that.

sangaman: I added a check for `pairId`, but not for `orderId` since it's actually `localId` that ought to be passed in and I know we discussed generating one if none is provided. So I left it alone for now.

sangaman:
Per @michael1011 's request, I refactored the way the argument checks are happening. Now there are predefined functions with corresponding error messages that check for argument validity and can be reused throughout the class.

Issue #162, opened 2018-06-27
PubKey based blocking


moshababo:
1. Add to handshake: exchange  `nodePubKey` (done)

2. When blocking a peer, we're doing it according to his ip address (without specifying the port, which is irrelevant). This means we can disconnect from him immediately, before handshaking, when he tries to reconnect. A support for blocking according to `nodePubKey` need to be added. This info will be known only after handshaking. The implementation need to take into consideration that one ip address could have multiple public keys, and vise versa.

kilrau:
Yep, definitely needed. Important to prevent Sybil attacks since staked coins are PER PUBKEY.

itsbalamurali:
@moshababoI meant this when I meant there should be a way to disconnect via public key.

moshababo:
Yes, I didn't thought it's important enough to implement it merely for the manual disconnection, but since we'll need it for automatic blocking anyway, we'll better make it work also for the gRPC call as you suggested.
Anyway, to implement this we'll need to do some changes in the way we're managing the peers/hosts list. Please be in touch with me before starting to implement it.

PR #315, opened 2018-08-09
Peer/Host restructuring

### LINKED ISSUE #155: Advertise listenPort
Advertise p2p `listenPort` on handshaking.
When a specific host entry is attached to a peer, update its `listenPort` field, to allow future outbound connections to this peer.

### LINKED ISSUE #177: Prevent a node from being connected to itself
Make sure a node won’t get connected to itself, nor saving itself in the host list. It’s a likely scenario when querying other peers for their list of peers (#147), but it can also happen by mistake, when initiating a connection manually. 
We might just want to check if before attempting to add a new outbound connection, and not in a new inbound connection, because the same node will control these two sides. 

#181 is a prerequisite.

### LINKED ISSUE #269: [Crash] Random 127.0.0.1 added to hosts list, connection leads to crash
**Current behavior:**
Running lastest `xud` `master` for a while adds a couple of 127.0.0.1 with different ports to my host list and we don't know where they are coming from. See below. Also interesting: the one finally connected and sending unparsable data doesn't end up in the `hosts` table:

```bash
8/3/2018, 12:18:57 PM [P2P] debug: Received data (xud1.test.exchangeunion.com:8885): {"header":{"reqId":"11ebab10-9752-11e8-b710-bb411fbe5c72","id":"11f23ac0-9752-11e8-9aed-a3a6a4367476","type":"PONG"}}

8/3/2018, 12:19:02 PM [P2P] debug: Connected to peer (127.0.0.1:40118)
8/3/2018, 12:19:02 PM [P2P] debug: Received data (127.0.0.1:40118): 
8/3/2018, 12:19:02 PM [P2P] warn: Unparsable peer message: : SyntaxError: Unexpected token  in JSON at position 0
8/3/2018, 12:19:02 PM [P2P] info: Socket closed (127.0.0.1:40118)
8/3/2018, 12:19:02 PM [P2P] warn: disconnected peer (127.0.0.1:40118) hostId is missing
Error: Peer was destroyed
    at Peer.close (/home/kilrau/xud/dist/p2p/Peer.js:103:30)
    at Socket.socket.once (/home/kilrau/xud/dist/p2p/Peer.js:294:22)
    at Object.onceWrapper (events.js:273:13)
    at Socket.emit (events.js:182:13)
    at TCP._handle.close (net.js:599:12)
/home/kilrau/xud/dist/bootstrap.js:13
        throw err;
        ^

Error: Peer was destroyed
    at Peer.close (/home/kilrau/xud/dist/p2p/Peer.js:103:30)
    at Socket.socket.once (/home/kilrau/xud/dist/p2p/Peer.js:294:22)
    at Object.onceWrapper (events.js:273:13)
    at Socket.emit (events.js:182:13)
    at TCP._handle.close (net.js:599:12)
kilrau@ubuntu:~/xud/bin$ mysql -u xud -p
Enter password: 
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 216
Server version: 5.7.23-0ubuntu0.18.04.1 (Ubuntu)

Copyright (c) 2000, 2018, Oracle and/or its affiliates. All rights reserved.

Oracle is a registered trademark of Oracle Corporation and/or its
affiliates. Other names may be trademarks of their respective
owners.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

mysql> use xud
Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed
mysql> SELECT * FROM hosts;
+----+-----------------------------+-------+--------+---------------------+---------------------+
| id | address                     | port  | pubKey | createdAt           | updatedAt           |
+----+-----------------------------+-------+--------+---------------------+---------------------+
| 18 | xud1.test.exchangeunion.com |  8885 | NULL   | 2018-08-03 01:27:28 | 2018-08-03 01:27:28 |
| 19 | 0.tcp.ngrok.io              | 19154 | NULL   | 2018-08-03 05:44:44 | 2018-08-03 05:44:44 |
| 20 | 127.0.0.1                   | 37542 | NULL   | 2018-08-03 06:17:46 | 2018-08-03 06:17:46 |
| 21 | 75.75.58.83                 |  8885 | NULL   | 2018-08-03 17:09:15 | 2018-08-03 17:09:15 |
| 22 | 127.0.0.1                   | 39812 | NULL   | 2018-08-03 18:38:50 | 2018-08-03 18:38:50 |
| 23 | 127.0.0.1                   | 39850 | NULL   | 2018-08-03 18:40:15 | 2018-08-03 18:40:15 |
| 24 | 127.0.0.1                   | 39882 | NULL   | 2018-08-03 18:40:15 | 2018-08-03 18:40:15 |
| 25 | 0.tcp.ngrok.io              | 18324 | NULL   | 2018-08-03 18:40:15 | 2018-08-03 18:40:15 |
+----+-----------------------------+-------+--------+---------------------+---------------------+
8 rows in set (0.00 sec)
```
**How it should be:**
there shouldn't be random 127.0.0.1 entries ending up in the hosts table, unparsable data should be handled gracefully. Not connecting to 127.0.0.1 will be dealt with in https://github.com/ExchangeUnion/xud/issues/177

### LINKED ISSUE #271: [Discussion] Socket connection can be opened twice
At least worth discussing: If I have an open connection from my `localhost` to e.g. `xud1.test.exchangeunion.com`, `xud1.test.exchangeunion.com` allows me to open a new connection back to my `localhost`. Ergo, I end up with 2 independent socket connections. Not sure if that's desirable and leads to unwanted behavior.

Also, ping pongs only show on the initiating side of the socket connection, IMO they should show on both ends since this connection can be used by both parties.

### LINKED ISSUE #289: 'socket closed' messages
It looks like there are way too many of them and everything get's like printed 4 or 5 times.

### LINKED ISSUE #305: hosts table (db) should include only valid entries
A valid entry is an entry that can be used by the node and by nodes connected to it to open outbound connections.

record an entry only upon successful connection.

For inbound connection:
a. don't record it in the table at all, or
b. record it in the table only if the connected peer provided it's listen port
(I opt for a. since we can't validate the listen port for an incoming connection.

don't add a record if the host is localhost or 127.0.0.1 or node didn't provide listen port or the peer PubKey is the same as our PubKey (self-connection).

For outbound connection: 
don't add a record if the host is localhost or 127.0.0.1 or the peer PubKey is the same as our PubKey (self-connection).

sangaman:
Fixes #289. Fixes #271. Fixes #269. Closes #155. Implements parts of #185 and #162. Edit: Closes #177. Closes #305.

This is a structural change to the p2p layer with breaking database changes.

- The term to represent a known xud instance that we are not necessarily connected to is changed from "Host" to "Node" and all references have been renamed.
- The unique identifier for a node used throughout the codebase is their `nodePubKey` rather than their socket address.
- External addresses for incoming connections are configurable and are shared during handshake.
- On receiving NODES packets, connections are only attempted to nodes that we were not previously aware of.

I'm opening this PR a bit early for purposes of getting input. Todo before merging:

- ~~Hardcode the nodePubKeys of the three xud seed nodes.~~ Edit: Done.
- ~~Update README.~~ Edit: Done.
- Testing between two running xud instances. Edit: Basic testing done.

```shell
2018-8-10 09:14:22 [P2P] debug: Connected to peer 127.0.0.1:59443
2018-8-10 09:14:22 [P2P] debug: Received data (127.0.0.1:59443): {"header":{"id":"4c0529a0-9c9f-11e8-8b78-dfb49cdc6ab6","hash":"9S+96qeN0grAs834/Y6ESg==","type":"HELLO"},"body":{"pairs":[],"version":"1.0","nodePubKey":"029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345","addresses":[]}}

2018-8-10 09:14:22 [P2P] debug: Received data (029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345): {"header":{"id":"4c07c1b0-9c9f-11e8-8b78-dfb49cdc6ab6","type":"GET_ORDERS"}}
{"header":{"id":"4c07c1b1-9c9f-11e8-8b78-dfb49cdc6ab6","type":"GET_NODES"}}

Executing (default): INSERT INTO `nodes` (`id`,`nodePubKey`,`addressesText`,`createdAt`,`updatedAt`) VALUES (DEFAULT,'029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345','[]','2018-08-10 13:14:22','2018-08-10 13:14:22');
2018-8-10 09:14:22 [P2P] debug: Received data (029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345): {"header":{"reqId":"4c07c1b0-9c9f-11e8-a3cf-41f24daf02b0","id":"4c0cf1d0-9c9f-11e8-8b78-dfb49cdc6ab6","hash":"J/XyycqWKe0kLqZ0sVdlDw==","type":"NODES"},"body":[{"nodePubKey":"1","addresses":[{"host":"xud1.test.exchangeunion.com","port":8885}]},{"nodePubKey":"2","addresses":[{"host":"xud2.test.exchangeunion.com","port":8885}]},{"nodePubKey":"3","addresses":[{"host":"xud3.test.exchangeunion.com","port":8885}]}]}

2018-8-10 09:14:22 [P2P] debug: Received data (029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345): {"header":{"reqId":"4c079aa0-9c9f-11e8-a3cf-41f24daf02b0","id":"4c130c50-9c9f-11e8-8b78-dfb49cdc6ab6","hash":"11FxOYiYfpMxmANj4kGJzg==","type":"ORDERS"},"body":[]}

2018-8-10 09:14:52 [P2P] debug: Received data (029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345): {"header":{"id":"5de8c870-9c9f-11e8-8b78-dfb49cdc6ab6","type":"PING"}}

2018-8-10 09:14:52 [P2P] debug: Received data (029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345): {"header":{"reqId":"5de8a160-9c9f-11e8-a3cf-41f24daf02b0","id":"5de91690-9c9f-11e8-8b78-dfb49cdc6ab6","type":"PONG"}}
```


sangaman:
Travis is getting DB errors that I wasn't getting locally, so I'll have to look into those. It compiles, runs, and passes tests for me locally in its current state although I did need to drop all my xud databases.

sangaman:
I fixed the DB dependency issue that was causing travis to fail by changing the column to a TEXT type named `addressesText` which stores the JSON in string format. The code then has a `VIRTUAL` column named `addresses` that doesn't exist on the table itself, but has a getter and setter that parses and stringifies the `addressesText` value.

Only thing missing is the `nodePubKey` values for the seed nodes, and more testing would be nice as I've only tested between two local nodes on my machine so far.

instructions-for-devs.md
```diff
@@ -142,12 +144,15 @@ host = "localhost"
 
 [p2p]
 listen = true
-port = 8885
-#make sure this port is reachable from the internet
+port = 8885 # Make sure this port is reachable from the internet
```
michael1011: `# Make sure this port is reachable from the internet` Maybe we can phrase this differently and refer to firewalls and does that port have to be reachable when you just want to have outbound connections?

instructions-for-devs.md
```diff
@@ -142,12 +144,15 @@ host = "localhost"
...
+# A string array of reachable socket addresses for this node, port defaults to 8885 if unspecified
```
michael1011: Shouldn't the default port be the config value `p2p.port`?

instructions-for-devs.md
offerm: Another comment about this subject:
The listen port may not be the port which is open on the firewall. for example, xud may listen on 8885 but the firewall may open 8888 and have port forwarding role from 8888 to 8885 on the xud machine. This is a very common situation (just consider 2 servers behind the FW running xud).

As such we must have a listen port and an external port which we publish to the peers.

michael1011: @offerm I think this is already handled in this PR by `p2p.addresses`. There you can specify the port that is sent in the handshake for each address.

lib/p2p/NodeList.ts
```diff
@@ -0,0 +1,96 @@
...
+/** Represents a list of nodes for managing network peers activity */
+class NodeList {
+  private nodes = new Map<string, NodeInstance>();
+
+  constructor(private repository: P2PRepository) {}
+
+  /**
+   * Return a [[NodeInstance]] array for nodes that haven't been banned.
+   */
+  public toArray = (): NodeInstance[] => {
```
moshababo: not in usage

lib/p2p/NodeList.ts
```diff
@@ -0,0 +1,96 @@
...
+      }
+    });
+    nodes.length = unbannedNodeCount;
+    return nodes;
+  }
+
+  /**
+   * Return a [[NodeFactory]] array for nodes that haven't been banned and have known addresses.
+   */
+  public toFactoryArray = (): NodeFactory[] => {
```
moshababo: I think using the `Factory` term here is a bit confusing, because it's from the DB row context. Maybe create a different type, with the same properties.

lib/p2p/Pool.ts
```diff
@@ -30,42 +32,55 @@ interface Pool {
...
 
     this.logger.info('Connecting to known / previously connected peers');
-    await this.hosts.load();
-    for (const host of this.hosts.toArray()) {
-      this.addOutbound(host.socketAddress);
-    }
+    await this.nodes.load();
+    this.nodes.forEach((node) => {
+      this.connectNode(node);
+    });
```
moshababo: connecting to banned nodes as well?

lib/p2p/Pool.ts
```diff
@@ -85,20 +100,32 @@ class Pool extends EventEmitter {
     this.connected = false;
   }
 
+  /** Attempt to create an outbound connection to a node using its known listening addresses. */
+  private connectNode = async ({ addresses, nodePubKey }: NodeFactory) => {
+    for (let n = 0; n < addresses.length; n += 1) {
+      try {
+        await this.addOutbound(addresses[n], nodePubKey);
```
moshababo: why connecting to all the node's addresses? if one succeed, isn't it sufficient? specially if they are all using the same pubkey.

instructions-for-devs.md
sangaman: That would be good, yes.

lib/p2p/Pool.ts
```diff
@@ -85,20 +100,32 @@ class Pool extends EventEmitter {
...
-   * If the connection is successful, send a [[GetHostsPacket]] to discover new XUD hosts to connect to.
+   * Attempt to add an outbound peer by connecting to a given socket address.
+   * Throws an error if a connection to a node with the given nodePubKey exists or
+   * if the connection handshake shows a different nodePubKey than the one provided.
+   * @param nodePubKey the nodePubKey of the node to connect to
    */
-  public addOutbound = async (socketAddress: SocketAddress): Promise<Peer> => {
-    if (this.peers.has(socketAddress)) {
-      const err = errors.ADDRESS_ALREADY_CONNECTED(socketAddress.address);
+  public addOutbound = async (address: Address, nodePubKey: string): Promise<Peer> => {
```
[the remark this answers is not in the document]
sangaman: I don't think it's completely necessary, and actually in my first approach I didn't add the pk to the `connect gRPC method, but it seemed worth it to me. The advantage is we know who we're connecting to, meaning that we can tell before even attempting the connection if we are already connected, and that we know which pk to expect from the handshake. In cases where we already know the pk, either from saved nodes or nodes we hear about from other peers, there's no downside. For the gRPC method it makes the command a bit more cumbersome particularly from the cli, but I still thought it made sense for purposes of knowing right away if we are already connected. This is also the approach `lnd` uses, connecting requires specifying the node pub key up front.

lib/db/models/Node.ts
```diff
@@ -0,0 +1,35 @@
...
+    id: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
+    nodePubKey: { type: DataTypes.STRING, unique: true },
+    addressesText: { type: Sequelize.TEXT },
+    addresses: {
+      type: Sequelize.VIRTUAL,
+      get(this: db.NodeInstance) {
+        return JSON.parse(this.addressesText);
+      },
+      set(this: db.NodeInstance, value: Address[]) {
+        this.setDataValue('addressesText', JSON.stringify(value));
```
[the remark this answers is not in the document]
sangaman: I hadn't thought of that but it does make querying for node via address awkward (I think it could still be possible with `SELECT * WHERE addresses LIKE '%host:'1.2.3.4',port:8885%`, but not the nicest query). 

The advantage as I see it is that `addresses` can be updated with a single, simple UPDATE query on handshake and it makes loading nodes from the db a little simpler because it's just querying rows from a single table. But if querying node via address is something we'll do regularly, it should probably be changed. I can't think of a real need to query node by address off the top of my head, can you?

As an aside, updating the addresses on handshake is actually not something that's implemented in this PR. So I will either take care of that here or in a separate PR after this is merged.

lib/p2p/SocketAddress.ts
```diff
@@ -15,13 +16,26 @@ class SocketAddress {
...
+  }
+
+  /**
+   * Return this SocketAddress as an [[Address]] object.
+   */
+  public toAddress = (): Address => {
+    return {
+      host: this.host,
+      port: this.port,
+    };
```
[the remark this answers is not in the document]
sangaman: It's perhaps a bit confusing, I tried to explain in comments but maybe I need to make it more clear. `SocketAddress` is a class with a handy `toString` method. `Address` is just a type representing the host+port tuple. Maybe we can drop `SocketAddress` though and just use `Address` plus a utils method to convert an Address to the host:port string we use in several places though.

lib/p2p/Peer.ts
```diff
@@ -416,17 +414,23 @@ class Peer extends EventEmitter {
 
     this.socket.once('close', (hadError) => {
       // emitted once the socket is fully closed
-      if (hadError) {
-        this.logger.warn(`Socket closed due to error (${this.id})`);
+      if (this.nodePubKey === undefined) {
```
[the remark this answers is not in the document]
sangaman: Sounds good, but keep the `verbose` log if we're connected but pre-handshake?

lib/p2p/NodeList.ts
sangaman: That's true, so basically create an alias for the type? I'll look into it.

lib/p2p/Pool.ts
```diff
@@ -191,44 +218,44 @@ class Pool extends EventEmitter {
...
-    if (!peer.inbound) {
-      const host = await this.hosts.getOrCreateHost(peer);
-      peer.setHost(host);
-    } else if (listenPort) {
-      const socketAddress = new SocketAddress(peer.socketAddress.address, listenPort);
-      const host = await this.hosts.getOrCreateHost(peer);
-      peer.setHost(host);
+  private handleOpen = async (peer: Peer): Promise<void> => {
+    if (this.nodes.isBanned(peer.nodePubKey!)) {
+      // TODO: Ban IP address for this session if banned peer attempts repeated connections.
```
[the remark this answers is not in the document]
sangaman: I think in both places, or rather right after the handshake. The if statement here is supposed to catch peers who are banned who connect to us anyway. The Peer class doesn't know about which nodes are banned which is why the logic is here. But I think if the handshake is bad or invalid, we can also put logic to ban before emitting the `open` event.

sangaman:
Conflicts resolved and suggested changes addressed, there are still a few open questions to @moshababo. I created a type alias called `NodeConnectionInfo` for `NodeFactory`, let me know if you think it's good to keep it that way.

lib/p2p/Peer.ts
```diff
@@ -122,13 +108,19 @@ class Peer extends EventEmitter {
 
   public getStatus = (): string => {
     if (this.connected) {
-      return `Connected to peer (${this.id})`;
+      return `Connected to peer ${this.socketAddress.toString()}`;
```
michael1011: The public key of the node is used as identifier everywhere else. Even in the log messages. Maybe you can add it here too.

lib/p2p/Peer.ts
```diff
@@ -416,17 +413,26 @@ class Peer extends EventEmitter {
...
-      if (hadError) {
-        this.logger.warn(`Socket closed due to error (${this.id})`);
-      } else {
-        this.logger.info(`Socket closed (${this.id})`);
+      if (this.connected) {
+        // don't log anything on close if we've never connected as the connection failure is logged elsewhere
+        if (this.nodePubKey === undefined) {
+          this.logger.verbose(`Socket closed prior to handshake (${this.socketAddress.toString()})`);
+        } else if (hadError) {
+          this.logger.warn(`Peer ${this.nodePubKey} socket closed due to error (${this.socketAddress.toString()})`);
```
michael1011: Is `this.socketAddress.toString()` returning the error in this case or does it get printed in a different place?

instructions-for-devs.md
```diff
@@ -74,6 +75,30 @@ Options:
...
+  --p2p.addresses     String array of reachable addresses                [array]
+  --p2p.listen        Listen for incoming peers                        [boolean]
+  --p2p.port, -p      Port to listen for incoming peers                 [number]
+  --raiden.disable    Disable raiden integration                       [boolean]
+  --raiden.port       Port for raiden REST service                      [number]
+  --rpc.host          gRPC service host                                 [string]
+  --rpc.port, -r      gRPC service port                                 [number]
+  --webproxy.disable  Disable web proxy server                         [boolean]
+  --webproxy.port     Port for web proxy server                         [number]
+>>>>>>> 789fccd... update readme
```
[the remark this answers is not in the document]
sangaman: Thanks, this was actually auto-merged. I'll fix it.

lib/db/models/Node.ts
```diff
@@ -0,0 +1,35 @@
+import Sequelize from 'sequelize';
+import { db } from '../../types';
+import { Address } from '../../types/p2p';
+
+export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes) => {
+  const attributes: db.SequelizeAttributes<db.NodeAttributes> = {
+    id: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
```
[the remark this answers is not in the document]
sangaman: Eventually we'll have a table to track node reputation events - and perhaps more tables to track successful trade history or other information related to nodes - and being able to use an integer `id` as a foreign key will be better for those tables.

lib/p2p/Peer.ts
sangaman: At this point we might not know the public key of the node (at least if it's an incoming connection), this is being called before the handshake from the `initConnection` method. But you're right that it's a little unusual. Maybe it should say `Connected to peer [nodePubKey]` if it has a valid handshakeState and `Connected pre-handshake to peer [socketAddress]` otherwise?

lib/p2p/Peer.ts
sangaman: Yeah the error would get logged elsewhere, but this message is a little unclear. Maybe it should just not print the socket address if we have a nodePubKey value to print instead.

instructions-for-devs.md
kilrau: ># A string array of reachable socket addresses for this node, port defaults to p2p.port if unspecified
addresses = [ "4.8.15.16:8885" ]

Well done @sangaman , that's exactly how it should be! This will allow even people in heavily censored countries to propagate `p2p.addresses` with port 443

sangaman:
I removed the `SocketAddress` class and replaced it with `addressUtils` helper methods. Let me know what you think.

instructions-for-devs.md
michael1011: @kilrau The port has to be between `1024` and `65535`. But there are still plenty options to choose from. 

And when you are in a heavily censored country changing the port won't help you much. Maybe we can provide a tutorial for setting up XUD with Tor in the future.

lib/p2p/Pool.ts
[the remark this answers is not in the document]
moshababo: I see, so let’s just wait for @kilian to approve this (in case he didn’t already)

lib/p2p/Pool.ts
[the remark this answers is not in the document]
moshababo: We were talking earlier about blocking inbound peers prior to handshake, based on IP address. Just wanted to know if we plan to support that, because if we do, we'll need to query nodes by IP address. If we don't, we're more vulnerable to DDoS.

This doesn't apply for banning within session, if we'll just collect banned IP addresses.
But we might want to apply this to blacklist/whitelist modes, which should persist, and might refer to a node entity - not strictly an IP address.

But anyway, we'll deal with that when we'll get to it.

lib/db/models/Node.ts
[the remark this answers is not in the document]
moshababo: Main scenario for querying by address would be to block an inbound peer, prior to handshake, based on the node entity (see my other comment). We might end up having other usages as well.

Squeezing a many-to-one relationship in a row is a known problem RDBMS, forcing you to normalize. This is why MongoDB became popular, followed by the JSON columns (PostgreSQL and then MySQL).
Concatenating the array is usually a fine for a dumb column, one that you wouldn't do anything with except projecting. Yes, querying would be possible, but pretty awful. 

At the moment, even if we'll want to query by address, we'll do it in-memory (after deserializing), so no DB support is needed. So I guess you can leave this situation as-is for now, and change it if we'll need to.

lib/p2p/Peer.ts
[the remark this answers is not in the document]
moshababo: I think it should be on `info`, same as the other non-error disconnection

lib/p2p/NodeList.ts
moshababo: Yep, something like this.

lib/utils/addressUtils.ts
```diff
@@ -0,0 +1,32 @@
+import { Socket } from 'net';
+import { Address } from '../types/p2p';
+import assert from 'assert';
+
+/** Helper methods for interacting with the [[Address]] type. */
+export const addressUtils = {
```
moshababo: This should be exported as the default.

Besides that, I preferred the class implementation, as it was before, specially because `toString` is called many times. In my earlier comment I was just wandering about the naming, but it wasn't an issue. 

Both solutions are fine.

lib/p2p/Peer.ts
```diff
@@ -416,17 +417,26 @@ class Peer extends EventEmitter {
...
     this.socket.once('close', (hadError) => {
       // emitted once the socket is fully closed
-      if (hadError) {
-        this.logger.warn(`Socket closed due to error (${this.id})`);
-      } else {
-        this.logger.info(`Socket closed (${this.id})`);
+      if (this.connected) {
+        // don't log anything on close if we've never connected as the connection failure is logged elsewhere
+        if (this.nodePubKey === undefined) {
+          this.logger.verbose(`Socket closed prior to handshake (${addressUtils.toString(this.socketAddress)})`);
```
moshababo: can be changed to `info`

lib/p2p/Pool.ts
```diff
@@ -86,19 +104,55 @@ class Pool extends EventEmitter {
...
+   * @param ignoreKnown whether to ignore nodes we are already aware of, defaults to false
+   * @returns a promise that will resolve when all outbound connections resolve
+   */
+  private connectNodes = (nodes: NodeConnectionIterator, ignoreKnown = false) => {
+    const connectionPromises: Promise<void>[] = [];
+    nodes.forEach((node) => {
+      // check that this node is not ourselves, that it has listening addresses,
+      // and that either we haven't heard of it, or we're not ignoring known nodes and it's not banned
+      if (node.nodePubKey !== this.handshakeData.nodePubKey && node.addresses.length > 0 &&
+        (!this.nodes.has(node.nodePubKey) || (!ignoreKnown && !this.nodes.isBanned(node.nodePubKey)))) {
```
moshababo: Suggestion: all conditions here are negative (including `ignoreKnown`). I think that making them positive will make it easier to follow. So basically checking whether to drop the node, instead of whether to connect to it. `ignoreKnown` can be turned to `allowKnown`,

lib/p2p/Pool.ts
```diff
@@ -86,19 +104,55 @@ class Pool extends EventEmitter {
...
+      }
+    });
+    return Promise.all(connectionPromises);
+  }
+
+  /**
+   * Attempt to create an outbound connection to a node using its known listening addresses.
+   */
+  private connectNode = async ({ addresses, nodePubKey }: NodeConnectionInfo) => {
+    for (let n = 0; n < addresses.length; n += 1) {
```
moshababo: `forEach` instead? parallel to `connectNodes`

lib/utils/addressUtils.ts
sangaman: The reason I wanted a type was for the `NodeFactory` type definition plus a handful of other places where we were using host+port tuples. Using `SocketAddress` wouldn't have worked I believe because it was a class with a public `toString` method. I agree the current `toString` usage is a bit awkward, but if it's ok with you then we can leave it.

lib/p2p/Pool.ts
sangaman: The reason I used a regular for loop is to be able to use the `break` command after the first successful connection, which I don't believe works with `.forEach()`.

lib/p2p/Pool.ts
sangaman: What would that look like? I can't use a `continue` statement here unless I change it from a `foreach`. And fwiw if we are checking whether to skip a node, `ignoreKnown` would make more sense I figure.

I know it's a pretty complex `if` statement, that's why I added the comments above to try to make it more clear.

offerm:
@sangaman consider to use the LND format:

03b46799265e53fd8255915f4f62929728f42c6144e89d82626a64ca1f09ea792e@35.236.118.88: 8885

Both on connect and getInfo (so copy/paste is easy).

Can be in a separate PR.

sangaman:
Yes I like the lnd connection string format, but I'm thinking of putting that in a separate PR just because this one is pretty big already and I'm hoping to merge it soon.

lib/p2p/Pool.ts
moshababo: `return` should be the same for that manner.

But you can leave it as is if it doesn't work for you.

sangaman: Oh right `return` would work, I didn't think of that. But I think I'll leave it as is if that's OK. One of the checks does happen to be positive (node.addresses.length > 0).

sangaman:
OK, Resolved conflicts again and addressed some of @moshababo's comments.

@kilrau Thanks for catching that bug with the cli, it's fixed now. I had to regenerate the type declarations for xudrpc and I used the new script I posted in #338. Give it a quick test again if you can.

sangaman:
Had to make some changes to the sanity tests including a new test case, also fixed what I believe is a bug in that both nodes were using the same port. Travis tests pass but looks like the node isn't shutting down properly, so I'll look into that. Still good for testing and I think just needs this sanity test/shutdown thing straightened out.

Sanity tests don't work for me at all when I do them locally, whether on master or this branch. I get this every time.

```bash
       "before all" hook:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

kilrau:
CLI `connect` still has issues @sangaman :disappointed: 
~/xud/bin$ ./xucli connect 03b46799265e53fd8255915f4f62929728f42c6144e89d82626a64ca1f09ea792e 35.236.118.88 8885
{
  "result": "Not connected"
}
2018-8-15 16:56:05 [P2P] warn: error while opening connection to peer 03b46799265e53fd8255915f4f62929728f42c6144e89d82626a64ca1f09ea792e: Connection timed out.

Interestingy also now it accepts to not specify `nodePubKey` OR `port` and still fails as timeout. Did you make one of them optional finally?

sangaman:
Can you provide an example of `xucli connect` not requiring nodePubKey? Port should not be required, but if I try to pass just host I get:

```shell
xucli connect 35.236.118.88
xucli connect <node_pub_key> <host> [port]

connect to an xu node

Options:
  --help          Show help                                            [boolean]
  --version       Show version number                                  [boolean]
  --rpc.port, -p  RPC service port                      [number] [default: 8886]
  --rpc.host, -h  RPC service hostname           [string] [default: "localhost"]
  --node_pub_key  nodePubKey of peer to connect                         [string]
  --host          hostname or IP address                                [string]
  --port          listening port number               [number] [default: "8885"]

Not enough non-option arguments: got 1, need at least 2
```

sangaman:
`banned` in the db is a tinyint(1), it's just what sequelize uses for booleans. It looks like the banned peer you tried connecting to resulted in `Not Connected`, right? I think we can take a closer look at banned peer behavior in a future issue/PR though.

I haven't tested explicitly trying to connect to oneself, in my tests it looks like it connects but fails the handshake. I can put in a check up front, but I'd rather do it in a separate PR to keep this one from getting bigger than it already is. It should be low priority, you have to intentionally be trying to mess things up by commanding xud to connect to yourself. What's prevented here is connecting to yourself when other peers tell you about yourself.

I fixed the builds and added more p2p sanity test cases, but I did detect a weird edge case bug I'll open an issue for later, it existed before this PR. I also fixed a bug with initialization, it was possible for the web proxy initialization to prevent xud from starting if it failed.

lib/p2p/Peer.ts
```diff
@@ -120,24 +106,45 @@ class Peer extends EventEmitter {
...
+   * and emit the `open` event if everything succeeds. Throw an error on unexpected handshake data.
+   * @param handshakeData our handshake data to send to the peer
+   * @param nodePubKey the expected nodePubKey of the node we are opening a connection with
+   */
+  public open = async (handshakeData: HandshakeState, nodePubKey?: string): Promise<void> => {
     assert(!this.opened);
     this.opened = true;
 
+    if (this.closed) {
+      throw 'wtf';
```
michael1011: Good error message :rofl:

sangaman: Haha, I put this in as a sanity test for the sanity tests when I was trying to see why they'd never terminate. I'll probably change this to something a little more formal - probably just an assert like above - but it's really something that should never happen.

kilrau:
Agree, put the upfront check in a separte issue: https://github.com/ExchangeUnion/xud/issues/352

sangaman:
Ok last set of changes I hope, but I made it so that attempting to connect to yourself throws an error upfront *before* the handshake, and added a test case for that. @kilrau doing a few quick tests then this can be merged.

PR #318, opened 2018-08-10
LND support for BTC and LTC (prep for atomic swap)

offerm:
This PR add support for 2 LNDs (one for BTC chain and one for LTC chain).
In additon, it fixes issues related to reconnect to LND after LND restart while XUD is active. Startup swquance is not important any more.
Also, there is different handling now if case LND is not configured (not suppose to be connected ) and LND is disconnected


sangaman:
I've only taken a quick look at the changes so far, but conceptually I think it would be good to be able to configure any number of running lnd instances in a generic way. So like for example, instead of `lndbtc.certpath` as a config option we'd make `lnd.certpath` an array. And in our proto definition, `getInfo` wouldn't have `lndbtc` and `ltclnd` but rather a field of repeated `LndInfo`s.

sangaman:
Yeah that would probably be better for the config section at least, something like this:

```toml
[lnd]
disable = false

[lnd.btc]
port = 9735

[lnd.ltc]
port = 10735
```

offerm:
@sangaman regarding protoc - I have noticed these changes but have no idea why we have these.
I'm using 3.4.0.

> admins-MacBook-Pro-2:xud admin$ protoc --version
> libprotoc 3.4.0
> 

Let me know if you have any idea.

offerm:
@sangaman @michael1011 regarding your comments about supporting multiple LNDs via array or map, here is why I haven't implemented it like this:

a. the user can of multiple LNDs is not clear. LND currently supports only BTC and LTC. Do you see a practical use case for 3 LNDs? If only 2 LNDs are used, the current implementation is preferred (IMHO)

b. when done with array/map we need to carefully plan the config file structure, command line arguments, protobuf, and default values. This is not that trivial and I don't want to spend time discussing it now (my focus is on bringing the swap to XUD)

c. The preferred solution (array/map or something else) is yet unknow and may be impacted for the swap implementation.  I'm not sure yet what kind of searches we would need for example.

As such, my suggestion is to support now 2 LNDs via `lndbtc` and `lndltc` and have a task (with much lower priority) for changing it to array/map.

It is better to make baby steps here.

kilrau:
> As such, my suggestion is to support now 2 LNDs via lndbtc and lndltc and have a task (with much lower priority) for changing it to array/map.

Agree, please open the task for let's say beta milestone or even 1.0 @offerm

offerm:
@sangaman please review. Need to merge this PR to continue working on the swap.

Offer

lib/lndclient/errors.ts
```diff
@@ -2,14 +2,20 @@ import errorCodesPrefix from '../constants/errorCodesPrefix';
 
 const codesPrefix = errorCodesPrefix.LND;
 const errorCodes = {
-  LND_IS_DISABLED: codesPrefix.concat('.1'),
+  LND_IS_DISCONNECTED: codesPrefix.concat('.1'),
+  LND_IS_NOT_CONFIGURED: codesPrefix.concat('.2'),
```
sangaman: Here too I actually prefer "disabled" over "not configured" which I think is more clear.

lib/raidenclient/RaidenClient.ts
```diff
@@ -83,22 +83,24 @@ class RaidenClient extends BaseClient {
...
     this.port = port;
     this.host = host;
+
     if (disable) {
-      this.setStatus(ClientStatus.DISABLED);
+      return;
     }
+    this.setStatus(ClientStatus.DISCONNETED);
+
+    this.init();
```
sangaman: Here too we shouldn't call the async `init` method from with the synchronous constructor.

lib/raidenclient/RaidenClient.ts
```diff
@@ -83,22 +83,24 @@ class RaidenClient extends BaseClient {
...
   }
 
   /**
    * Discover our raiden address.
    */
   public async init() {
-    if (!this.isDisabled()) {
```
sangaman: I think we should leave this disabled check in place, we don't want to allow initialization if Raiden is supposed to be disabled.

lib/raidenclient/RaidenClient.ts
```diff
@@ -110,7 +112,7 @@ class RaidenClient extends BaseClient {
    */
   private sendRequest = (endpoint: string, method: string, payload?: object): Promise<http.IncomingMessage> => {
     return new Promise((resolve, reject) => {
-      if (this.isDisabled()) {
+      if (!this.isConnected()) {
         reject(errors.RAIDEN_IS_DISABLED);
```
sangaman: I think we should have two checks with two separate errors here as well, one if raiden is disabled and one if it's disconnected.

lib/service/Service.ts
```diff
@@ -1,13 +1,14 @@
 import Logger from '../Logger';
 import Pool from '../p2p/Pool';
 import OrderBook, { OrderArrays } from '../orderbook/OrderBook';
-import LndClient from '../lndclient/LndClient';
+import LndClient, { LndClientConfig } from '../lndclient/LndClient';
```
sangaman: I actually think we should be able to drop config information from this class given the distinction between an lnd client being "disabled" and "disconnected", see comments below.

lib/service/Service.ts
```diff
@@ -1,13 +1,14 @@
...
 import RaidenClient, { TokenSwapPayload } from '../raidenclient/RaidenClient';
 import { OwnOrder } from '../types/orders';
 import Config from '../Config';
 import { EventEmitter } from 'events';
 import SocketAddress from '../p2p/SocketAddress';
 import errors from './errors';
+import lnderrors from '../lndclient/errors';
```
sangaman: Rename this `lndErrors` to match naming used elsewhere like the errors imports in GrpcService.ts.

lib/service/Service.ts
```diff
@@ -48,13 +51,54 @@ class Service extends EventEmitter {
...
     this.pool = components.pool;
     this.config = components.config;
 
     this.logger = Logger.rpc;
   }
+  private getLndInfo = async (config: LndClientConfig, lndClient: LndClient)  => {
+
+    let info: any = {};
+
+    if (config.disable) {
```
sangaman: I think it's better to call lndClient.isDisabled() here to determine whether lnd is disabled or simply disconnected. Previously we were using the config because the lnd client status didn't make a distinction between being disabled or merely disconnected, but I think it's good to have that distinction and we should take advantage of it here.

proto/xudrpc.proto
```diff
@@ -139,7 +139,8 @@ message GetInfoResponse {
   int32 num_pairs = 2 [json_name = "num_pairs"];
   string version = 3 [json_name = "version"];
   OrdersCount orders = 4 [json_name = "orders"];
-  LndInfo lnd = 5 [json_name = "lnd"];
+  LndInfo lndbtc = 5 [json_name = "lndbtc"];
+  LndInfo lndltc = 7 [json_name = "lndltc"];
```
sangaman: We're early enough in development that I don't think there are backwards compatibility concerns with proto field numbering, I'd prefer if this was named 6 (and raiden below 7) to keep them ordered.

lib/BaseClient.ts
```diff
@@ -1,13 +1,14 @@
 import Logger from './Logger';
 
 enum ClientStatus {
-  DISABLED,
+  DISCONNETED,
```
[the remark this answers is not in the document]
offerm: Done

lib/BaseClient.ts
```diff
@@ -1,13 +1,14 @@
...
   CONNECTION_VERIFIED,
+  NOT_CONFIGURED,
```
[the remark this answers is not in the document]
offerm: Done

lib/Xud.ts
```diff
@@ -20,7 +20,8 @@ class Xud {
   private logger: Logger = Logger.global;
   private config: Config;
   private db!: DB;
-  private lndClient!: LndClient;
+  private lndBtcClient!: LndClient;
+  private lndLtcClient!: LndClient;
```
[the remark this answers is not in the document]
offerm: done

lib/Xud.ts
```diff
@@ -50,15 +51,16 @@ class Xud {
...
+      this.lndLtcClient = new LndClient(this.config.lndltc);
 
+      // setup raiden client and connect if configured
       this.raidenClient = new RaidenClient(this.config.raiden);
-      await this.raidenClient.init();
 
       this.pool = new Pool(this.config.p2p, this.db);
 
-      this.orderBook = new OrderBook(this.db.models, this.pool, this.lndClient);
+      this.orderBook = new OrderBook(this.db.models, this.pool, this.lndBtcClient);
```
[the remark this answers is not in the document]
offerm: To be handled in the next phase. OrderBook is not using the lnd client anyway and I'm not yet sure it will stay there at all.

lib/Xud.ts
```diff
@@ -50,15 +51,16 @@ class Xud {
...
 
-      this.lndClient = new LndClient(this.config.lnd);
-      await this.lndClient.connect();
+      // setup LND clients and connect if configured
+      this.lndBtcClient = new LndClient(this.config.lndbtc);
+      this.lndLtcClient = new LndClient(this.config.lndltc);
 
+      // setup raiden client and connect if configured
       this.raidenClient = new RaidenClient(this.config.raiden);
-      await this.raidenClient.init();
```
sangaman: Restore this `init` statement, we want it to be called from here and not from the constructor so we can properly wait for it.

lib/Xud.ts
```diff
@@ -50,15 +51,16 @@ class Xud {
       this.db = new DB(this.config.db);
       await this.db.init();
 
-      this.lndClient = new LndClient(this.config.lnd);
-      await this.lndClient.connect();
```
sangaman: Restore this `connect` statement (for both lnds), we want it to be called from here and not from the constructor so we can properly wait for it.

lib/BaseClient.ts
```diff
@@ -17,10 +18,10 @@ abstract class BaseClient {
     this.logger.info(`${this.constructor.name} status: ${ClientStatus[val]}`);
     this.status = val;
   }
-
-  public isDisabled(): boolean {
-    return this.status === ClientStatus.DISABLED;
+  public isConnected(): boolean {
+    return this.status === ClientStatus.CONNECTION_VERIFIED;
   }
+
```
sangaman: We can drop this blank line, but also here I'm thinking we should have `isDisconnected` and `isDisabled` methods as well. Or maybe it'd be cleaner to just have a expose the `status` field via a getter method, up to you.

lib/service/Service.ts
offerm: done

lib/service/Service.ts
offerm: Done

lib/service/Service.ts
offerm: done

proto/xudrpc.proto
offerm: not sure why you think we should keep the sequence here (what value does it give beside being unique?)  but  I did it.

done.

lib/lndclient/LndClient.ts
```diff
@@ -41,30 +41,32 @@ class LndClient extends BaseClient {
...
-      this.setStatus(ClientStatus.DISABLED);
-    } else {
-      this.uri = `${config.host}:${config.port}`;
-      const lndCert = fs.readFileSync(certpath);
-      this.credentials = grpc.credentials.createSsl(lndCert);
-
-      const adminMacaroon = fs.readFileSync(macaroonpath);
-      this.meta = new grpc.Metadata();
-      this.meta.add('macaroon', adminMacaroon.toString('hex'));
+      return;
```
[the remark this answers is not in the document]
offerm: Not sure I agree here. The multiple return style makes the code much more readable. Constructors are not a special case in that sense. 

with current logic, the if-else flow has one thread so It is "manageable" but if you have to add another if to one of the "then" part it will become very complicated and uneasy to ready.

Unless this is critical for you I prefer to leave it like this.

lib/lndclient/LndClient.ts
```diff
@@ -76,18 +78,19 @@ class LndClient extends BaseClient {
   }
 
   public connect = async () => {
-    if (!this.isDisabled()) {
+    if (!this.isConnected()) {
```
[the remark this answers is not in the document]
offerm: This should not happen anyway. I will add some more checks.

done.

lib/lndclient/LndClient.ts
```diff
@@ -177,8 +180,8 @@ class LndClient extends BaseClient {
    * Attempt to close an open channel.
    */
   public closeChannel = (fundingTxId: string, outputIndex: number, force: boolean): void => {
-    if (this.isDisabled()) {
-      throw(errors.LND_IS_DISABLED);
+    if (!this.isConnected()) {
```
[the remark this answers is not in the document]
offerm: Added a check for `disabled` and changed to disconnected.

done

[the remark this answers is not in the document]
offerm: Note that this is not a safe code. LND may become disconnected between  the call to closeChannel (when it was connected) and the check.
better not to check for diconnect at all and make the call.
the call may also create the connection.
If you like to change it it should be in a different task.

lib/lndclient/LndClient.ts
```diff
@@ -206,8 +209,8 @@ class LndClient extends BaseClient {
    * Subscribe to events for when invoices are settled.
    */
   private subscribeInvoices = (): void => {
-    if (this.isDisabled()) {
-      throw(errors.LND_IS_DISABLED);
+    if (!this.isConnected()) {
```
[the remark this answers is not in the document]
offerm: done

same as above

lib/lndclient/errors.ts
offerm: done

lib/raidenclient/RaidenClient.ts
offerm: done

offerm:
@sangaman I did most of the changes you suggested. I'm still checking why protoc is making these changes (i think I'm using a special 3.4.0 version for LND).

Few comments:

1. See my response for the if-else vs multi returns in the constructor. 

2. You wrote if a few comments that we want to wait at XUD startup for LNDs and Raiden to be connected. IMHO this is wrong. We should not assume that LNDs are available when we start XUD and we better complete the startup and be able to do work which is not related to LND. Completing the startup also make the RPC available and users/apps can use `getInfo` to see that LND didn't start properly.

3. regarding starting `LND` in async vs sync mode - please note that even the current mode is not doing it in sync mode. Only the first attempt is done in sync mode, but, if LND is not available, reconnect is attempted in async mode. 

4. If we initiate `connect` from the constructor we can make the `connect` function private (not in the PR) which is better. Also, Xud.js code is much cleaner.

Taking all this into account I don't think we should restore the connect calls into Xud.js.

Also, please keep in mind that I "fixed" LND lifecycle management but haven't done so for Raiden. This should be done in a different PR.

lib/lndclient/LndClient.ts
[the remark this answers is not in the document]
sangaman: I am mainly just not used to seeing `return`s anywhere in constructors. I can't find anywhere that suggests it's a rule but it seems unconventional to me. To me they are equally readable, I'm not sure what exactly what you mean by adding another if, to me the difference is just between having `return` statements in all the `if` checks up front vs wrapping the rest of the constructor in an `else` bracket. It's not a big deal, but maybe someone else can chime in.

sangaman:
1. I responded there as well. Basically if it's normal to have `return` statements in constructors then it's fine as is - I just can't say I've seen that before.

2. I slightly misspoke when I said we'll have to wait for lnd to be connected (rather than wait for a connection attempt) before starting xud, and as you correctly pointed out that's not the current behavior (which is intentional). I do think we do want to at least wait for the initial connection attempt to happen before fully starting the xud server. If the connect attempt fails, we set a timer to make reconnection attempts and move on.

There are other considerations here too. Since the default status is "DISABLED", it's possible that a `getinfo` call right after starting xud will show the status as disabled instead of disconnected since the initial connection attempt will not have failed yet. And calling an async function from within a synchronous function like the constructor could also result in unhandled rejections which would be messy.

3. That's correct, see 2.

4. If we could wait for an async function inside the constructor then I'd agree it would be nice to make the connect function private and put it in the constructor, but we can't. I don't think it's that messy though, and this is also the pattern we use for every other class that has an asynchronous `connect()` or `init()` routine so we'd just be sticking to that.

5. This is true. Maybe we should rename `connect` to something like `verifyConnection` because that's really what's happening. But I still don't think it's a good idea to put it in the constructor, strictly for properly handling async functions.

Raiden can definitely be in a different PR, although it will benefit from the distinction between `DISABLED` and `DISCONNECTED`.

lib/lndclient/LndClient.ts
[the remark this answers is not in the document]
offerm: @sangaman see this discussion on how if-else-if can become messy https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement
mainly the second answer.

Anyway, I will change it to single return and keep current structure (readable). Actually having an enhancment of checking validity of disable certpath and macaroonpath all at once.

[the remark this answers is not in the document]
offerm: done

lib/Xud.ts
offerm: done

lib/raidenclient/RaidenClient.ts
offerm: done

offerm:
@sangaman should be good now. please review.

Only thing left is to find why we differ on the protoc. Let's meet only to debug it,

Offer

lib/Xud.ts
```diff
@@ -50,15 +51,24 @@ class Xud {
...
+      // setup LND clients and connect if configured
+      this.lndbtcClient = new LndClient(this.config.lndbtc);
+      if (!this.lndbtcClient.isDisabled()) {
```
sangaman: We don't need these `isDisabled` checks here because we already check for disabled status within the connect method, which is the better place for this sort of check I think.

lib/raidenclient/RaidenClient.ts
sangaman: I don't see the separate checks here, could you make sure you put those in? Although since this is a Raiden class and this PR is focusing on lnd, if you want to leave this as is that's fine too and we'll get back to Raiden in a future PR.

lib/service/Service.ts
```diff
@@ -48,13 +51,54 @@ class Service extends EventEmitter {
...
+    this.lndBtcClient = components.lndBtcClient;
+    this.lndLtcClient = components.lndLtcClient;
     this.raidenClient = components.raidenClient;
     this.pool = components.pool;
     this.config = components.config;
 
     this.logger = Logger.rpc;
   }
+  private getLndInfo = async (lndClient: LndClient)  => {
+
```
sangaman: Can you drop this blank line, and add one above `private getLndInfo` to space it from the close bracket above it? Just for consistent spacing.

lib/lndclient/LndClient.ts
```diff
@@ -38,24 +38,30 @@ class LndClient extends BaseClient {
...
-    } else if (!fs.existsSync(macaroonpath)) {
+      shouldConnect = false;
+    }
+    if (!fs.existsSync(macaroonpath)) {
       this.logger.error('could not find lnd macaroon, is lnd installed?');
-      this.setStatus(ClientStatus.DISABLED);
-    } else {
+      shouldConnect = false;
+    }
+    if (shouldConnect) {
```
sangaman: What's the appeal of using this pattern of `shouldConnect` instead of a chain of if-elses as before? I suppose this way we can log errors for both `macaroonpath` and `certpath` in one go which is nice, but just want to make sure I understand the changes here. If we do keep this pattern, I think the flag should be renamed to `shouldEnable`.

lib/lndclient/LndClient.ts
```diff
@@ -76,18 +82,27 @@ class LndClient extends BaseClient {
   }
 
   public connect = async () => {
-    if (!this.isDisabled()) {
+    if (this.isDisabled()) {
+      this.logger.error(`can't connect to lnd. lnd is disabled`);
```
sangaman: I think a `warn` statement is a little more appropriate here since we I don't consider this an error exactly.

lib/lndclient/LndClient.ts
```diff
@@ -76,18 +82,27 @@ class LndClient extends BaseClient {
...
+      return;
+    }
+    if (this.isConnected()) {
+      this.logger.error(`trying to connect to already connected lnd`);
```
sangaman: I think this should also be a `warn` and I'd rather it say "can't connect to lnd, lnd is already connected because "trying to connect" makes it sound like we're giong forward with the connection attempt.

lib/raidenclient/RaidenClient.ts
```diff
@@ -83,22 +83,25 @@ class RaidenClient extends BaseClient {
...
   public async init() {
-    if (!this.isDisabled()) {
-      try {
-        this.address = await this.getAddress();
-      } catch (err) {
-        this.logger.error('could not get raiden address', err);
-        this.status = ClientStatus.DISABLED;
-      }
+    if (this.isDisabled()) {
+      this.logger.error(`can't init raidan. raidan is disabled`);
```
sangaman: "raiden" instead of "raidan."

lib/Xud.ts
offerm: We need it here in order to avoid an error message that will be issued in case you try to connect().

lib/service/Service.ts
offerm: done.

lib/lndclient/LndClient.ts
offerm: code is more readable.
multiple checks in one run

renamed.

done

lib/lndclient/LndClient.ts
offerm: See above.
calling connect() for a disabled LND should be considered an error. Hence the message.

lib/lndclient/LndClient.ts
offerm: changed the text.

One should not try to connect to a connected LND. This should be considered an error.

lib/raidenclient/RaidenClient.ts
offerm: done

lib/Xud.ts
sangaman: Ah yeah that's right, we can leave as is.

lib/lndclient/LndClient.ts
sangaman: I've thought about this a bit more and I agree actually, but in that case we should throw a `LND_IS_DISABLED` error to let the caller of `connect` know that the method failed and the system is in an unexpected state - right now that's not happening.

But for the case of already being connected, I don't think we should treat it as an error. The state of the lnd client is already in the desired/expected state, but we're unnecessarily calling the method. I think that's exactly the sort of thing that warn statements usually indicate, unexpected behavior that doesn't compromise core functionality.

Again, and as you pointed out, what this method is really doing is verifying that we *can* connect to LND (and that we're subscribed to invoice updates).

lib/Xud.ts
sangaman: Ah yeah you are right, we can leave this as is.

sangaman:
No problem about Raiden, we'll leave that for another PR. See my additional comment about the error checking/handling the the lnd `connect` method. I think we should throw an error if lnd is disabled but merely warn if lnd is already connected.

lib/lndclient/LndClient.ts
offerm: Well, I disagree. connect() should be called only when we are disconnected, hence it should be an error.

regarding disable throwing an error, I think this is an edge case and the developer will be able to find it if reported as an error or thrown.

feel free to change it after I'm done with the swap.

sangaman: Ok, I will revisit this method another time.

sangaman:
Looks like you figured out the unnecessary changes to the generated proto files - great. Only thing left is to resolve conflicts - I took a look at them and they are very minor. I'm looking forward to the next steps and getting lnd swaps working.

offerm:
@sangaman - this is ready

kilrau:
Awesome!!! @offerm @sangaman

PR #327, opened 2018-08-12
log 'socket closed' msg only if already connected

moshababo:
Small one, addressing #289 

if socket wasn't previously connected, the error msg is a bit verbose, because we're already printing the connection failure reason.

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -416,11 +416,14 @@ class Peer extends EventEmitter {
 
     this.socket.once('close', (hadError) => {
       // emitted once the socket is fully closed
-      if (hadError) {
-        this.logger.warn(`Socket closed due to error (${this.id})`);
-      } else {
-        this.logger.info(`Socket closed (${this.id})`);
+      if (this.connected) {
+        if (hadError) {
+          this.logger.warn(`Socket closed due to error (${this.id})`);
+        } else {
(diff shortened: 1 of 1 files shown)


sangaman:
I handled this in a slightly different way in #315 because I noticed the same thing you did.

```javascript
this.socket.once('close', (hadError) => {
  // emitted once the socket is fully closed
  if (this.nodePubKey === undefined) {
	this.logger.verbose(`Socket closed prior to handshake (${this.socketAddress.toString()})`);
  } else if (hadError) {
	this.logger.warn(`Peer ${this.nodePubKey} socket closed due to error (${this.socketAddress.toString()})`);
  } else {
	this.logger.info(`Peer ${this.nodePubKey} socket closed (${this.socketAddress.toString()})`);
  }
  this.close();
});
``

I left the log message in but made it verbose so it won't show even in debug mode. Also it logs the verbose message instead of the regular message as long as it's pre-handshake, rather than pre-connection. Likewise I believe we already would log the error for why the handshake failed, so it makes sense to me to do it that way. Let me know what you think.

moshababo:
It's good to log that the socket closed prior to handshake. But it should exclude cases where connection wasn't established yet.

Both of these cases will be logged by `Pool.tryOpenPeer`. But in case of pre-handshake & post-connection, logging it from socket close event make sense.

I'll comment about this in your PR.

Issue #285, opened 2018-08-06
fix `npm run db:init`


kilrau:
- Seed nodes & test data into db `xud_test` on `db:init`, no additional commands (this is what this task is about), this works *mostly* on the current master but always ends in some errors. To be cleaned up.
- Seed nodes (only) into `xud` db on first run (done)

kilrau:
when done, bring back https://github.com/ExchangeUnion/xud/pull/286

kilrau:
https://github.com/ExchangeUnion/xud/pull/244#issuecomment-410887432

sangaman:
Closed via #328.

PR #328, opened 2018-08-12
db:init task fix

moshababo:
Solved #285 

If we want to keep the hosts values hardcoded in `DB` class, then this is the solution. It was failing because these values insertion was duplicated.


sangaman:
Ahhh yes that makes sense. Only thing here is that the commit from #327 made it into this PR, we should take it out to keep them separate. Can you do that? I can take care of it too if you'd rather.

moshababo:
Yes that was by mistake, I reverted it, and since we're squashing it won't be noticed, isn't it?

sangaman:
Yeah that works, thanks!

PR #329, opened 2018-08-12
add node_modules caching

itsbalamurali:


sangaman:
That looks like the right way to do it, but oddly one of the travis builds still took over 27 minutes and failed. Could just be more issues on travis' end. I'll restart the builds and investigate a bit.

sangaman:
Well it took a lot of restarted builds but it finally worked. Maybe this won't solve the intermittent issues with Travis (for all I know there it might be entirely out of our hands) but I figure it won't hurt to enable caching anyway.

Issue #284, opened 2018-08-06
Long packets being split resulting in unparseable message


sangaman:
I've observed large packets being split into smaller messages, which makes the packets themselves unparseable. This will require some investigation.

```bash
2018-8-6 18:45:20 [P2P] debug: Received data (xud1.test.exchangeunion.com:8885): {"header":{"reqId":"65cc4710-99ca-11e8-ac3f-f7f569b019d7","id":"67c509d0-99ca-11e8-bc0e-c56cce5370a8","hash":"kv8B5Ns2QhekD9lI5cYjlw==","type":"HOSTS"},"body":[{"address":"49.35.229.203","port":20000},{"address":"216.38.151.30","port":20000},{"address":"0.tcp.ngrok.io","port":15583},{"address":"75.75.58.83","port":20000},{"address":"49.35.38.125","port":20000},{"address":"73.71.151.38","port":20000},{"address":"49.35.29.59","port":58523},{"address":"216.38.151.30","port":49325},{"address":"216.38.151.30","port":49683},{"address":"212.89.173.216","port":42191},{"address":"216.38.151.30","port":50098},{"address":"212.89.173.216","port":43722},{"address":"216.38.151.30","port":50162},{"address":"0.tcp.ngrok.io","port":18324},{"address":"199.203.79.137","port":61565},{"address":"localhost","port":8885},{"address":"127.0.0.1","port":45880},{"address":"199.203.79.137","port":61599},{"address":"199.203.79.137","port":62545},{"address":"49.35.108.155","port":55662},{"address":"199.203.79.137","port":62824},{"address":"199.203.79.137","port":62823},{"address":"199.203.79.137","port":62822},{"address":"199.203.79.137","port":62893},{"address":"199.203.79.137","port":62956},{"address":"199.203.79.137","port":62959},{"address":"199.203.79.137","port":62961},{"address":"199.203.79.137","port":63044},{"address":"199.203.79.137","port":63051},{"address":"199.203.79.137","port":63048},{"address":"49.35.108.155","
[...]

PR #330, opened 2018-08-12
Wire protocol 

moshababo:
Addressing #284, and also changes the way we handle multiple packets stream.

I preferred not to use anymore the `\r\n` sign anymore, because it can potentially be found in a string property as well. Instead, the packet length is specified.

Putting this inside of the packet JSON, in the header, was not possible (packet string must be complete in order to be parsable), so I added it as the first 4 bytes in the stream.

I couldn't reproduce yet the packet splitting - couldn't connect to the remote nodes (my ports are blocked), and locally it never happen. Will do it soon. 
Wanted to open this PR so you could give feedback on the approach.

If the approach is acceptable, I can add some length constrains for safety.


lib/p2p/Parser.ts
```diff
@@ -60,20 +60,46 @@ interface Parser {
 
 /** Protocol packet parser */
 class Parser extends EventEmitter {
+  private pending: Buffer[] = [];
+  private waiting: number = 0;
```
sangaman: These names aren't that clear. I think maybe `pendingBuffers` or `pendingPacket` for the first and `remainingBytes` for the second are a couple of suggestions.

lib/p2p/Parser.ts
```diff
@@ -60,20 +60,46 @@ interface Parser {
...
+      this.parsePacket(this.pending);
+      this.resetCycle();
+      if (size < chunk.length) { // multiple packets
+        this.feed(chunk.slice(size));
       }
-    });
+    }
+  }
+
+  private resetCycle = () => {
```
sangaman: Maybe call this `resetBuffer`? Or maybe better just put these two lines into the `read` function after we parse a packet, unless you think it's going to be called elsewhere later on.

lib/p2p/Parser.ts
```diff
@@ -60,20 +60,46 @@ interface Parser {
...
+  private read = (size: number, chunk: Buffer) => {
+    this.pending.push(chunk.slice(0, size));
+
+    if (size > chunk.length) { // packet isn't complete
+      this.waiting = size - chunk.length;
+    } else { // chunk is finalizing the packet
+      this.parsePacket(this.pending);
+      this.resetCycle();
+      if (size < chunk.length) { // multiple packets
+        this.feed(chunk.slice(size));
```
offerm: are you taking here the first part or the left over?
also, what if the leftover size is <4? you will not be able to read the length.

offerm:
See this.
https://stackoverflow.com/questions/17312242/node-js-net-library-getting-complete-data-from-data-event
I think the first answer shows correctly how this should be done

moshababo:
I can handle the scenario where chunks would split on the 4 bytes. Also, I can add another 1 byte flag, following the 4 bytes, to ensure that i'm not reading just a random 4 bytes and interpreting it as a number (and if I am, the chunk would be dropped).

Or, we can use the delimiter approach. With specifying packet size, I know the size of the packet from the beginning, so I can drop it it if it's too big. With the delimiter approach, I have to reach my maximum before I know about it.
Also, the delimiter is a bit less safe - even if we'll use a unique one, it will be known, and can be injected somehow into one of the strings that a node might use in packets. 

So in the overall, specifying size is more complicated, but a bit more safe. Both can be used.
I'm used to seeing size specifying in p2p protocols. Not using it was an option only because we're using JSON serialization. 

So tell me what you think, and i'll finalize this.

moshababo:
I can handle the scenario where chunks would split on the 4 bytes. Also, I can add another 1 byte flag, following the 4 bytes, to ensure that i'm not reading just a random 4 bytes and interpreting it as a number (and if I am, the chunk would be dropped).

Or, we can use the delimiter approach. With specifying packet size, I know the size of the packet from the beginning, so I can drop it it if it's too big. With the delimiter approach, I have to reach my maximum before I know about it.
Also, the delimiter is a bit less safe - even if we'll use a unique one, it will be known, and can be injected somehow into one of the strings that a node might use in packets.

So in the overall, specifying size is more complicated, but a bit more safe. Both can be used.
I'm used to specifying size in p2p protocols. Not using it was an option only because we're using JSON serialization.

So tell me what you think, and i'll finalize this.

offerm:
as you wrote both approaches should work.

Looks to me that `size` works better when you can read from the connection directly and you can specify the amount you want to read. This way, once you know the size of the message, you can be sure you are not reading anything belonging to the next message. Also, there is normally an `available()` function that you can use to be sure that there are at least 4 bytes ready to be read.

In our case, the `data` is a callback. In this case, I think a separator will work better.

you should do something like this:
```
    conn.on('data', function(data) {
        completeData += data;
        var dataArray = completeData.split('your delimiter');
        if(dataArray.size > 1) { //If our data was split into several pieces, we have a complete chunk saved in the 0th position in the array
            doWorkOnTheFirstHalfOfData(dataArray[0]);
            completeData = dataArray[1];// The second portion of data may yet be incomplete, thise may need to be more complete logic if you can get more than one delimeter at a time...
        }
    });
```
Basically, you append the received data. If there is a separator in the data you handle all chunks to the left of the last separator. Note that the code above is not doing that properly and is not ready for getting multiple separators in one data part.

moshababo:
There's no problem of reading min/max length, because the buffer array length is known. So I didn't understand the argument against it.

offerm:
Try to write the code and you will see.
What if you get a 3 bytes buffer on the first feed() call?

moshababo:
Then I wait for the next chunk, and only then try to parse the number. Before I have 4 bytes, i'm iterating until i'll get them. After I have it, i'm iterating until i'll consume the packet.
It will add some more code, and I didn't write it yet, but I don't see a problem with the approach.

offerm:
As I wrote above

> as you wrote both approaches should work.

Do it as you like. My feeling is that the seperator option will lead a simple code which is way easier to understand.

moshababo:
Code simplicity is important, but the problem we were discussing couldn't even be solved if we'll use a separator.

moshababo:
me and @offerm talked, and I was wrong, the problem can be addressed while using a separator, while having an overall simpler solution. 

@sangaman waiting for your take on this to finalize.

moshababo:
Yes, i'll do the necessary refactoring.

moshababo:
Latest commit:
* changed the parser to use delimiter. it's fairly simple now, only 7 lines
* added some tests
* refactored `Packet` a bit. mostly due to the deep equal checks in the tests

lib/p2p/Parser.ts
```diff
@@ -60,20 +60,30 @@ interface Parser {
...
-        return;
-      }
-      try {
-        const packet = fromRaw(packetStr);
-        this.emit('packet', packet);
-      } catch (err) {
-        this.emit('error', err);
-      }
-    });
+    this.pending += data.toString();
```
offerm: Suggest to protect the max size of this.pending to avoid accumulation of a very large message

lib/p2p/Parser.ts
```diff
@@ -60,20 +60,30 @@ interface Parser {
...
+    const index = this.pending.indexOf(this.delimiter);
+    if (index > -1) {
```
offerm: Suggest to have tests for:
- empty Buffer
- buffer containing one delimiter only (no other data)
- buffer containing 5  delimiters (no other data)

lib/p2p/Parser.ts
sangaman: I was thinking of having this implemented later but maybe not too hard to take care of it here, what's a reasonable number? 4 MB? I can imagine the ORDERS packet getting fairly large.

lib/p2p/Parser.ts
```diff
@@ -60,20 +60,30 @@ interface Parser {
 
 /** Protocol packet parser */
 class Parser extends EventEmitter {
+  private pending = '';
```
sangaman: I raised a similar comment earlier but I'd prefer calling this `buffer` or maybe `pendingPacket`. Actually, since it might contain multiple packets I don't think `pendingPacket` would be too good, but I still like `buffer`.

lib/p2p/packets/Packet.ts
```diff
@@ -40,6 +40,8 @@ abstract class Packet<T = any> implements PacketInterface {
   public abstract get direction(): PacketDirection;
   public body?: T;
   public header: PacketHeader;
+  /** Wire protocol delimiter. */
+  public static PROTOCOL_DELIMITER = 'ↂ₿ↂ';
```
sangaman: Nice! What are those 2 things on the side?

lib/p2p/Parser.ts
```diff
@@ -60,20 +60,30 @@ interface Parser {
...
-      }
-      try {
-        const packet = fromRaw(packetStr);
-        this.emit('packet', packet);
-      } catch (err) {
-        this.emit('error', err);
-      }
-    });
+    this.pending += data.toString();
+    const index = this.pending.indexOf(this.delimiter);
```
sangaman: I have a minor performance concern here where `indexOf` will get slower and slower as `pending` gets larger, but I realize it's not as simple as checking the `data.toString()` for the delimiter because the delimiter could be split between two chunks. I'm pretty sure this is O(n^2) complexity with the size of the packet and something we'll be doing very often so I'm just concerned and maybe we should at least make a note to revisit this.

One solution that just came to mind is tracking the last received chunk and the current chunk, and concatenating those then checking the result for the delimiter. We'd clear the last chunk on creating a packet.

Not necessary for this PR, but something to think about.

lib/p2p/Parser.ts
```diff
@@ -60,20 +60,30 @@ interface Parser {
...
+    if (index > -1) {
+      this.parsePacket(this.pending.slice(0, index));
+      const next = this.pending.slice(index + this.delimiter.length);
+      this.pending = '';
+      this.feed(Buffer.from(next));
```
sangaman: We're converting a buffer to a string and then back to a buffer here, I'm thinking it might be better to make `feed()` accept a string parameter and do `toString()` from where we call the method in Peer.ts. It would also make the tests a little cleaner, since we have `Buffer.from` calls in multiple places that we wouldn't need if `feed` took a string.

test/unit/Parser.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+  testPacket(pingPacket);
+  testPacket(helloPacket);
+
+  testSplitPacket(pingPacket);
+  testSplitPacket(helloPacket);
+
+  testConcatenatedPackets([pingPacket, helloPacket, pingPacket]);
+
+  testConcatenatedAndSplitOnTheDelimiter([pingPacket, helloPacket, pingPacket]);
+
```
sangaman: Can drop this blank line.

lib/p2p/packets/Packet.ts
```diff
@@ -56,15 +58,20 @@ abstract class Packet<T = any> implements PacketInterface {
   constructor(bodyOrPacket?: T | PacketInterface, reqId?: string) {
     if (isPacketInterface(bodyOrPacket)) {
       // we are deserializing a received packet from a raw JSON string
-      this.body = bodyOrPacket.body;
+      delete bodyOrPacket.header.type; // deleting type from header because it's being used only for the wire protocol
```
sangaman: This is smart.

lib/p2p/packets/Packet.ts
```diff
@@ -56,15 +58,20 @@ abstract class Packet<T = any> implements PacketInterface {
...
       this.header = bodyOrPacket.header;
+
+      if (bodyOrPacket.body) {
+        this.body = bodyOrPacket.body;
```
sangaman: What's the advantage of this over `this.body = bodyOrPacket.body`? I think it winds up undefined either way when the body is undefined.

moshababo: if it's undefined, then the property is set to undefined. it's not the same as if the property wouldn't be set at all. So if doing a deep equal check on the objects, it fails. It's possible to just remove the undefined properties before doing the equal check, but I thought that this would be better.

lib/p2p/Parser.ts
moshababo: I left it as is because I thought that maybe at some point we'll want to read the bytes array and not the string. But I can do what you suggester for now.

lib/p2p/packets/Packet.ts
moshababo: Just some random thing I found on the unicode table. I think it's a roman number. Feel free to change the delimiter if you want.

lib/p2p/Parser.ts
moshababo: I didn't want `buffer` to be confused with the `Buffer` type. But I can use it if you prefer, since `feed` will receive a string anyway.

moshababo: I didn't want `buffer` to be confused with the `Buffer` type. But I can use it if you prefer, since `feed` will receive a string now.

lib/p2p/Parser.ts
moshababo: Your solution could improve it, you can open a separate issue for that if you want. Not high priority at the moment.

moshababo:
Latest commits:
* parser now receives a string
* renamed `pending` to `buffer`
* max buffer size support added, default to 4 MB. we can open an issue for making it configurable, if we want
* test code improved, and more test were added

The travis build results are not consistent. It's sometime fails on the max buffer size checks.
JS string supposed to be UTF-16, 2 bytes. But I checked it with `Buffer.byteLength`, and it was only 1 byte. I ended up using it for the actual max length checks. 
If it still doing problems, it means that converting a fixed size buffer to string, and then calculating the string byte length, results in a different size. 
If that's the case, I'll need to find a different solution. I just want to make sure it's not just another travis issue.

lib/p2p/packets/Packet.ts
sangaman: Ah yeah that's what you meant by deep checks, ok cool, thanks.

lib/p2p/packets/Packet.ts
sangaman: Nah it's fine, as long as it works.

sangaman:
The default encoding when you do Buffer.toString() is utf-8. Same for Buffer.from(), it assumes utf-8 unless specified otherwise. In some quick tests I did, the `ↂ₿ↂ` comes out as 9 bytes with utf-8 and `abc` comes out as 3 bytes. It does look like you can't easily calculate the bytes of a string - you have to convert back to buffer first I think. You could maybe keep a running number tally of the number of bytes along with the string buffer perhaps. Ultimately I don't think this needs to hold up this PR unless it's consistently giving us build issues - do the tests fail occasionally for you when running locally? I'll try it myself later.

Also 9 bytes sounds like kind of a lot for a separator, but it's not a big deal.

test/unit/Parser.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    parser.feed(delimiter);
+  });
+
+  it(`should buffer a max buffer length`, (done) => {
+    parser = new Parser(delimiter, 10);
+    wait()
+      .then(() => expect.fail())
+      .catch((err) => {
+        expect(err).to.equal('timeout');
```
sangaman: I'm getting an unhandled rejection error from this line not every time I run the tests locally but most times. If you merge master, I'm using the `chai-as-promised` plugin for some of the sanity tests. Maybe you can use it here as well with the sanity tests as an example.

sangaman:
So I actually did a rebase myself and refactored two of the test cases and now the tests pass every time for me, but I'm still seeing some unhandled rejection warnings that I assume are coming from other test cases. Here's what the refactored test cases look like:

```javascript
  it(`should try parse a standalone delimiter and fail`, () => {
    expect(wait()).to.be.rejectedWith(ParserErrorType.UNPARSEABLE_MESSAGE.toString());

    parser.feed(delimiter);
  });

  it(`should buffer a max buffer length`, () => {
    parser = new Parser(delimiter, 10);
    expect(wait()).to.eventually.be.fulfilled;

    parser.feed(Buffer.allocUnsafe(10).toString());
  });
```

If you want I can push my changes to the branch and you can finish it, let me know.

test/unit/Parser.spec.ts
moshababo: The 'timeout' exception is expected. If you're talking about tests timeout (20 sec), then it doesn't happen for me locally.
I just saw your `chai-as-promised` usage - looks nice.

sangaman:
OK cool, well the tests pass for me consistently now. If we can expose the parser error objects and maybe clean up the unhandled exception warnings in the test output in a future PR, that would be great, but I think this is good for now. Thanks.

PR #333, opened 2018-08-13
Set version on handshake

sangaman:
Since we're having releases and using versioning now - and since there will likely be breaking changes between pre-release versions - I thought it would be helpful to share the actual local version on handshake.


sangaman:
Oh you mean to indicate that it's in a dev branch state between releases?

sangaman:
Ah yeah that makes sense, but maybe we should only make that version change when introducing breaking p2p changes between releases.

Issue #320, opened 2018-08-11
Error: 14 UNAVAILABLE: Connect Failed


zhouzhou84:
### Description：
I follow the tutorial and finish the step 1,2,3, it's alright.
but when I use the command `./xucli connect xud1.test.exchangeunion.com 8885`,
it shows 
```
ubuntu@ip-172-31-15-69:~/xud/bin$ ./xucli connect xud1.test.exchangeunion.com:8885
(node:21935) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
Error: 14 UNAVAILABLE: Connect Failed
```
what's wrong?

### system configuration
```
ubuntu@ip-172-31-15-69:~/xud/bin$ uname -a
Linux ip-172-31-15-69 4.4.0-1061-aws #70-Ubuntu SMP Fri May 25 21:47:34 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

ubuntu@ip-172-31-15-69:~/xud/bin$ node -v
v10.8.0

ubuntu@ip-172-31-15-69:~/xud/bin$ nodejs -v
v4.2.6

ubuntu@ip-172-31-15-69:~/xud/bin$ geth version
Geth
Version: 1.8.13-stable
Git Commit: 225171a4bfcc16bd12a1906b1e0d43d0b18c353b
Architecture: amd64
Protocol Versions: [63 62]
Network Id: 1
Go Version: go1.10
Operating System: linux
GOPATH=/home/ubuntu/gocode
GOROOT=/usr/lib/go-1.10

ubuntu@ip-172-31-15-69:~/xud/bin$ lncli -v
lncli version 0.3

```

sangaman:
Can you confirm that xud is running?

itsbalamurali:
@zhouzhou84 please open two separate terminals and in first one run xud via `./bin/xud` or ` npm start` and after `xud` starts up successfully in second terminal run the connect command via `./bin/xucli connect xud1.test.exchangeunion.com 8885`.

zhouzhou84:
@sangaman  it's my fault, I should start it first, BUT when I run `./xud`, something about DB goes wrong 
```
ubuntu@ip-172-31-15-69:~/xud/bin$ ./xud
8/11/2018, 12:07:36 PM [GLOBAL] info: config loaded
[...]

PR #334, opened 2018-08-13
Update instructions-for-devs.md

kilrau:
"make sure xud is running" as per https://github.com/ExchangeUnion/xud/issues/320#issuecomment-412271220

### DIFF
--- a/instructions-for-devs.md
+++ b/instructions-for-devs.md
@@ -33,7 +33,7 @@ mysql> GRANT ALL ON *.* TO 'xud'@'localhost';
 
 ## 2. Start `xud`
 
-Open a new terminal and launch a `xud` process
+Open a new terminal and launch a `xud` process.
 
 ```bash
 ~/xud/bin $ ./xud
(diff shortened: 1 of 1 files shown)


instructions-for-devs.md
```diff
@@ -78,7 +78,7 @@ Options:
 
 ## 3. Interact with `xud` via Command-Line Interface
 
-Interact with an `xud` process, identified by its `rpc` host and port using `xucli`. For getting up-to-date CLI commands, use `./xucli --help` as shown below:
+Open a new terminal (make sure `xud` is running either in a different terminal or as [service](#daemonize-xud)) and interact with `xud`, identified by its `rpc` host and port using `xucli`. For getting up-to-date CLI commands, use `./xucli --help` as shown below:
```
sangaman: The wording seems a a bit excessive to me. We say to open a new terminal in the previous section already. Having the daemonize instructions are good (although it should say "as **a** service") but I think they should go in **Start xud**. It's also possible to interact with xud on another machine, which this doesn't make clear. 

Maybe we can say "Interact with a *running* xud process" if you feel it needs more clarification, but I think it's OK as is.

kilrau: Well not clear enough obviously, as we could see some user had an issue understanding that `xud` needs to be running in a different terminal/process in order for `xucli` to interact with it. But agree on the excessive wording. Better?

Issue #142, opened 2018-06-24
XUD sends order twice


kilrau:
- with same ID
- with different ID

Receiving XUD just drops it.

moshababo:
If an order with the same Id arrives, the order book should ignore it, and the p2p layer should penalize the peer (an issue for this task will be open soon).

If a duplicated order with a different Id arrives, there's no practical way to protect against it, except with our general anti-spam mechanisms, which will be applied to all orders.

moshababo:
For now we just need to make sure that order book does not add an order with a duplicated order id.

PR #337, opened 2018-08-14
drop duplicated peer orders

moshababo:
Addressing #142

At the moment it's checking that the order is duplicated, and not solely `id`. 
More precisely, that the id+pairId+orderSide tuple is duplicated. That's because we're dividing the orders by pairId, and then by sell/buy. 
Having a support for a duplication only on the order id would be messy right now. If we want that support, it should be a part of #215.

I didn't add tests, because it requires p2p tests, and not integration tests. But I temporarily turned `addPeerOrder` method to public, run some quick tests, and it looks fine.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -232,10 +232,15 @@ class OrderBook extends EventEmitter {
     return matchingResult;
   }
 
+  /**
+   * Add peer order
+   * @returns false if it's a duplicated order or with an invalid pair id, otherwise true
+   */
   private addPeerOrder = (order: orders.PeerOrder): boolean => {
     const matchingEngine = this.matchingEngines.get(order.pairId);
     if (!matchingEngine) {
       this.logger.debug(`incoming peer order invalid pairId: ${order.pairId}`);
+      // TODO: penalize peer
       return false;
     }
 
@@ -289,6 +294,10 @@ class OrderBook extends EventEmitter {
     return true;
   }
 
+  /**
+   * Add an order to an order map
+   * @returns false if an order with the same id already exist, otherwise true
+   */
   private addOrder = (ordersMap: OrdersMap, order: orders.StampedOrder) => {
     if (this.isOwnOrdersMap(ordersMap)) {
       const { localId } = order as orders.StampedOwnOrder;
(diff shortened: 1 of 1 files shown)


lib/orderbook/OrderBook.ts
```diff
@@ -286,7 +295,13 @@ class OrderBook extends EventEmitter {
       this.localIdMap.set(localId, order.id);
     }
 
-    this.getOrderMap(ordersMap, order).set(order.id, order);
+    const orderMap = this.getOrderMap(ordersMap, order);
+    if (orderMap.has(order.id)) {
+      return false;
+    } else {
+      orderMap.set(order.id, order);
+      return true;
```
sangaman: Could you add a comment block to this method explaining that it attempts to add an order if none exists in the ordermap, `@returns true if the order was added, false if the order is a duplicate`.

lib/orderbook/OrderBook.ts
```diff
@@ -232,18 +232,27 @@ class OrderBook extends EventEmitter {
     return matchingResult;
   }
 
-  private addPeerOrder = (order: orders.PeerOrder) => {
+  private addPeerOrder = (order: orders.PeerOrder): boolean => {
     const matchingEngine = this.matchingEngines.get(order.pairId);
     if (!matchingEngine) {
       this.logger.debug(`incoming peer order invalid pairId: ${order.pairId}`);
-      return;
+      return false;
```
sangaman: Maybe add a TODO here as well to penalize the peer?

lib/orderbook/OrderBook.ts
```diff
@@ -232,18 +232,27 @@ class OrderBook extends EventEmitter {
     return matchingResult;
   }
 
-  private addPeerOrder = (order: orders.PeerOrder) => {
+  private addPeerOrder = (order: orders.PeerOrder): boolean => {
```
sangaman: A comment block would be good here as well to explain that it returns true if the order was added, false if it's a duplicate order or for an invalid pair.

lib/orderbook/OrderBook.ts
moshababo: In this case it's specifically a duplicated order id, so I changed your text a bit

PR #343, opened 2018-08-15
 Separated placeorder command

### LINKED ISSUE #249: CLI Market Order / Limit Order clear distinction
1. We would want `placeOrder` to require a '0' for a market order in CLI+gRPC for security reasons. Don't make this param optional. 
2. Change the documentation.

Opened a new issue, so you don't have to change https://github.com/ExchangeUnion/xud/pull/245 but rather open a separate PR for the next milestone.

michael1011:
Closes #249 

This PR separates the CLI command `placeorder` into `limitorder` and `marketorder`.

### DIFF
--- a/lib/cli/commands/orders/marketorder.ts
+++ b/lib/cli/commands/orders/marketorder.ts
@@ -1,0 +1,22 @@
+import { Arguments } from 'yargs';
+import * as limitorder from './limitorder';
+
+export const command = 'marketorder <pair_id> <order_id> <quantity>';
+
+export const describe = 'place a market order';
+
+export const builder = {
+  pair_id: {
+    type: 'string',
+  },
+  order_id: {
+    type: 'string',
+  },
+  quantity: {
+    type: 'number',
+  },
+};
+
+export const handler = (argv: Arguments) => {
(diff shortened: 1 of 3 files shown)


lib/cli/commands/orders/marketorder.ts
```diff
@@ -0,0 +1,22 @@
...
+  order_id: {
+    type: 'string',
+  },
+  quantity: {
+    type: 'number',
+  },
+};
+
+export const handler = (argv: Arguments) => {
+  limitorder.handler(argv);
```
sangaman: Do we need to set a 0 price here like `{ ...argv, price: 0 }` or does it work as is? When we do `setPrice` in the limitorder file we set an undefined value there instead of a number. I'm guessing it's probably fine, I just haven't tested it. If it works for you, then we can keep as is.

michael1011: It is working right now because the gRPC server receives the value 0 for `price` even if it set to `undefined` on the client side.

sangaman: That's what I thought, works for me.

PR #344, opened 2018-08-15
Refactor gRPC

sangaman:
This aims to be at least an intermediate solution to #339.  It's in an unfinished state but I wanted feedback before I went and applied the same pattern to every call. `getInfo` is the main call to look at, you can ignore the others for the purposes of initial review. It works and can be tested. I also refactored `connect` and `cancelOrder`.

This uses our static generated files for creating the xud grpc server - doing so requires that we use the actual generated gRPC classes for the callback. Passing a plain javascript object with the right fields won't work. This restricts the code to only be able to return valid grpc responses.

The current structure is basically that Service.ts deals with our internal types and plain javascript objects but doesn't touch anything grpc-related. It passes values to the GrpcService.ts class which in turn creates the grpc response object and passes it to the callback - or if it encounters an error it passes the error to the callback. The logic to map internal errors to grpc errors is located within the `getServiceResponse` method. The functions in GrpcService.ts aim to follow the general structure shown below:

```javascript
function sayHello(call, callback) {
  var reply = new messages.HelloReply();
  reply.setMessage('Hello ' + call.request.getName());
  callback(null, reply);
}
````

I also refactored the getInfo logic itself a bit and moved some of it to the `LndClient` and `RaidenClient` classes.

If this looks good I'll apply the same approach to all the calls and update the PR.

### DIFF
--- a/lib/grpc/GrpcServer.ts
+++ b/lib/grpc/GrpcServer.ts
@@ -5,18 +5,16 @@ import Logger from '../Logger';
 import GrpcService from './GrpcService';
 import Service from '../service/Service';
 import errors from './errors';
+import { XudService } from '../proto/xudrpc_grpc_pb';
 
 class GrpcServer {
   private server: Server;
 
   constructor(private logger: Logger, service: Service) {
     this.server = new grpc.Server();
-    const xudrpcProtoPath = path.join(__dirname, '..', '..', 'proto', 'xudrpc.proto');
-    const protoDescriptor = grpc.load(xudrpcProtoPath, 'proto', { convertFieldsToCamelCase: true });
-    const { xudrpc }: any = protoDescriptor;
 
     const grpcService = new GrpcService(logger, service);
-    this.server.addService(xudrpc.Xud.service, {
+    this.server.addService(XudService, {
       cancelOrder: grpcService.cancelOrder,
       connect: grpcService.connect,
       disconnect: grpcService.disconnect,
--- a/lib/grpc/GrpcService.ts
+++ b/lib/grpc/GrpcService.ts
@@ -2,144 +2,303 @@
 import grpc, { status } from 'grpc';
 import Logger from '../Logger';
 import Service from '../service/Service';
-import { isObject } from '../utils/utils';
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by sangaman** [comment]
I finished refactoring all the methods per my approach described above. I tested about half the gRPC calls manually and everything looks good.

**Comment by offerm** [comment]
Looks better!

Still more todo (IMHO).

See this code:
```
  public getInfo: grpc.handleUnaryCall<xudrpc.GetInfoRequest, xudrpc.GetInfoResponse> = async (call, callback) => {
    try {
      const serviceResponse = await this.getServiceResponse(this.service.getInfo, call);
      const response = new xudrpc.GetInfoResponse();

```
Why can't it call `this.service.getInfo()` directly? Like this:

```
  public getInfo: grpc.handleUnaryCall<xudrpc.GetInfoRequest, xudrpc.GetInfoResponse> = async (call, callback) => {
    try {
      const serviceResponse = await this.service.getInfo();
      const response = new xudrpc.GetInfoResponse();
```

the `getServiceResponse` is not really needed and just complicates things.

Also, why can't `this.service.getInfo()` returns `GetInfoResponse`? you will be able to delete lines 128:164 and get a really nice and clean code.  
Not to metion that you can delete the `XudInfo` type which is not really used....

And you still keep the separation and be able to test the service methods as today.

sangaman:
I have `getServiceResponse` in there mainly to have all the error handling logic to convert internal error codes to grpc error codes in one place (and a minor bonus is not having to write `.request.toObject()` for every method). But I can change it so I just take that error handling logic and use it in the catch blocks of the methods if you think that's cleaner, so the try portion looks like like `this.service.getInfo()`.

If `service.getInfo()` returned a `GetInfoResponse`, we'd just need to take those lines 128:164 and move them to Service.ts right? They can't go away because they're needed to set the values on the response, afaik. And then our testing of `service.getInfo` would have to work with the grpc objects instead of plain JSON objects, which is what I'd want to avoid unless we decide to make the jump and use grpc objects everywhere they are applicable.

sangaman:
@offerm Take a look now and see what you think. If it's good I'll squash the two most recent commits and we can merge.

PR #350, opened 2018-08-16
Switch to upstream fastpriorityqueue

sangaman:
Unless we have any more plans for changes to the `fastpriorityqueue` package, I figure we might as well switch back to the upstream version rather than our fork.


kilrau:
Good call! Always preferable to have an "official" version. We don't want to maintain any forks ideally.

sangaman:
There's an issue with the way the upstream package was deployed actually, it has `npm` as a dependency for some mysterious reason which is causing travis builds to fail. I'll open an issue there.

PR #351, opened 2018-08-16
Replace init task with seed values

### LINKED ISSUE #331: integration tests failing due to connection to seed nodes
`WebProxy.spec.ts` is failing due to connection to seed nodes, which cause an error on `xud.shutdown`.
We should avoid making `xud` instance always connect to the seed nodes.

sangaman:
Closes #331.

This commit makes xud add seed values for pairs and currencies when a new database is created during startup. These are in addition to the seed nodes. An `initDb` config option can be set to `false` to disable initializing new databases with these seed values. This makes the `db:init` task obsolete.

I didn't add a value to the `[db]` config section because we've had that line up with Sequelize options and `initValues` would have broken that. I tried this approach at first but it was getting messy making changes to DB.ts to accommodate it. So instead I put `initDb` as a top level config option.


kilrau:
works like a charme, but it's still based on `hosts`, so you might have adjust this to `nodes` as per #315 ? @sangaman

sangaman:
Yup, I will refactor once #315 is merged.

sangaman:
@kilrau Rebase done, do you want to try this again one more time now that it's using `nodes`?

PR #355, opened 2018-08-17
Update node addresses on handshake

sangaman:
This makes it so the database is updated with listening addresses from a node after the handshake, even if we already had a record for that node in the db. Currently, once a row is written to the nodes table it won't be updated, but it makes sense that we should update with the freshest info we've received from a peer.

I also added a new test suite for `Pool` with a couple of test cases that test this new functionality.

Lastly I caught a few issues during testing with the way we were initializing and shutting down the Pool, the `listen` and `unlisten` methods were not resolving properly, and fixed those here as well.

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -38,7 +38,7 @@ interface NodeConnectionIterator {
 class Pool extends EventEmitter {
   private nodes: NodeList;
   private peers: PeerList = new PeerList();
-  private server: Server = net.createServer();
+  private server?: Server;
   private connected = false;
   /** The local handshake data to be sent to newly connected peers. */
   private handshakeData!: HandshakeState;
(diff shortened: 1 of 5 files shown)


lib/p2p/Pool.ts
```diff
@@ -333,12 +333,36 @@ class Pool extends EventEmitter {
     this.peers.forEach(peer => peer.close());
   }
 
-  private listen = (port: number): void => {
-    this.server.listen(port, '0.0.0.0');
+  /**
+   * Start listening for incoming p2p connections on the given port.
+   * @return a promise that resolves once the server is listening, or rejects if it fails to listen
+   */
+  private listen = (port: number) => {
```
michael1011: Shouldn't we define the return type even if it is just `Promise<void>`?

sangaman: It's implied in this case, I've been letting return types be implicitly defined for short/simple methods. Since the first line starts with `return new Promise<void>` I think it's clear enough in this case, but let me know if you think differently.

kilrau:
Didn't test it properly, but looks good.

Issue #119, opened 2018-06-08
Test API layer via exposed Xud.service methods


sangaman:
Our API layer is currently exposed via the `service` property on the main `xud` instance. We should create test cases for the exposed methods to test the API integration with `xud`.

kilrau:
This task includes defining details (feedback loop via https://gitter.im/exchangeunion/Lobby)

fundrequestbot:
[![](https://fundrequest.io/requests/43/badge)](https://fundrequest.io/requests/43) [![](https://fundrequest.io/assets/img/powered-by-fundrequest-badge.svg)](https://fundrequest.io)
This issue has been funded using [FundRequest](https://fundrequest.io/requests/43). A developer can claim [the reward](https://fundrequest.io/requests/43) by submitting a pull request referencing this issue. ([How to Close Issues via Pull Requests?](https://help.github.com/articles/closing-issues-using-keywords)) e.g. `fixes #119`

* For more help on how to claim on issue, please visit our [help section](https://help.fundrequest.io).
* Looking for more? Feel free to [browse](https://fundrequest.io/requests) through all funded requests.

kilrau:
Interested in continuing this? @ImmanuelSegol @erkarl

sangaman:
#357 took care of a lot of this, I think going forward we should actually focus on the approach outlined in #588, since a lot of the rpc methods require p2p connections and swap clients. I'd say this issue can probably be closed.

PR #357, opened 2018-08-18
Service tests

sangaman:
This creates a test suite with a sampling of some test cases for our API service methods which are exposed via gRPC, and is a step towards resolving #119.

Running these tests actually revealed a bug with the shutdown call, which I actually accidentally introduced myself in #332. As part of that PR I added an `await` statement to the shutdown procedure to wait on the grpc server (among other things) to close before returning, but the grpc server wouldn't close until the method returned a response, thus they were waiting on each other and would hang. The first commit fixes that and adds a `shutdown` event for the `Xud` class to notify listeners when the server has completely shut down, it's used in the service shutdown test.

### DIFF
--- a/proto/xudrpc.proto
+++ b/proto/xudrpc.proto
@@ -75,7 +75,7 @@ service Xud {
     };
   }
   
-  /* Shutdown the xud daemon. */
+  /* Begin shutting down xud. */
   rpc Shutdown(ShutdownRequest) returns (ShutdownResponse) {
     option (google.api.http) = {
       post: "/v1/shutdown"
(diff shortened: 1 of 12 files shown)


proto/xudrpc.proto
```diff
@@ -75,7 +75,7 @@ service Xud {
     };
   }
   
-  /* Shutdown the xud daemon. */
+  /* Begin shutting down the xud daemon. */
```
michael1011: Isn't XUD = Exchange Union Daemon? What about removing the redundant `daemon` here?

sangaman: Yeah maybe i'll rename "daemon" to "server"? Or just drop daemon and say "Begin shutting down xud"?

michael1011: I would prefer just `Begin shutting down xud` but it is up to you

sangaman: Done, I also changed a comment in the code from "Inititate graceful shutdown of the server" to "Initiate graceful shutdown of xud"

PR #361, opened 2018-08-20
Add tslint circular import rule

sangaman:
This adds a tslint plug-in and rule to check for circular imports, which we have been trying to avoid in our codebase. The rule caught one violation, a minor one with the `NodeConnectionInfo` for `NodeFactory`. I corrected it just by respecifying the type from scratch, which is not too bad because it's a short type definition.


offerm:
Please don't merge it until we clean the ln-swap-berlin branch.

PR #362, opened 2018-08-20
Set lnd pubkeys on handshake

sangaman:
This refactors the lnd `connect` method to store the lnd server's pub key and renames it `verifyConnection`. It then saves the lnd pub key as part of the node's handshake info to be shared with peers.

### DIFF
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -61,14 +61,15 @@ class Xud {
       await this.db.init();
 
       const initPromises: Promise<void>[] = [];
+
       // setup LND clients and connect if configured
       this.lndbtcClient = new LndClient(this.config.lndbtc, loggers.lnd);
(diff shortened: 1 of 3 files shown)


sangaman:
Yes neither am I, I'll definitely refactor it to use a map of currency to LND client soon, although maybe not until after Berlin.

As for the whole connection uri, I just hadn't considered it but it might be a good idea. We only need the pubkey for handling swaps. I guess one consideration is that there might not be a connection uri in lnd. But maybe it's something we could treat as an optional piece of data, and display in the `listPeers` output? Let me know what you think, although if it's not a small change I'd probably just put it in a separate pR.

michael1011:
I agree on adding the uri as additional data. We don't need it right now but it will be useful in the future to connect to the LNDs of the other XUD automatically.

Issue #290, opened 2018-08-07
Solve `npm install xud` error


kilrau:
`npm install xud` and `npm install -g xud` both end up in:

```bash
> grpc@1.13.1 install /home/kilrau/node_modules/grpc
> node-pre-gyp install --fallback-to-build --library=static_library

node-pre-gyp WARN Using needle for node-pre-gyp https download 
[grpc] Success: "/home/kilrau/node_modules/grpc/src/node/extension_binary/node-v64-linux-x64-glibc/grpc_node.node" is installed via remote

> secp256k1@3.5.0 install /home/kilrau/node_modules/secp256k1
> npm run rebuild || echo "Secp256k1 bindings compilation fail. Pure JS implementation will be used."


> secp256k1@3.5.0 rebuild /home/kilrau/node_modules/secp256k1
> node-gyp rebuild

make: Entering directory '/home/kilrau/node_modules/secp256k1/build'
  CXX(target) Release/obj.target/secp256k1/src/addon.o
  CXX(target) Release/obj.target/secp256k1/src/privatekey.o
../src/privatekey.cc: In function ‘Nan::NAN_METHOD_RETURN_TYPE privateKeyNegate(Nan::NAN_METHOD_ARGS_TYPE)’:
../src/privatekey.cc:73:30: warning: ignoring return value of ‘int secp256k1_ec_privkey_negate(const secp256k1_context*, unsigned char*)’, declared with attribute warn_unused_result [-Wunused-result]
   secp256k1_ec_privkey_negate(secp256k1ctx, &private_key[0]);
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  CXX(target) Release/obj.target/secp256k1/src/publickey.o
  CXX(target) Release/obj.target/secp256k1/src/signature.o
  CXX(target) Release/obj.target/secp256k1/src/ecdsa.o
  CXX(target) Release/obj.target/secp256k1/src/ecdh.o
  CC(target) Release/obj.target/secp256k1/src/secp256k1-src/src/secp256k1.o
../src/secp256k1-src/src/secp256k1.c: In function ‘secp256k1_ec_pubkey_parse’:
../src/secp256k1-src/src/secp256k1.c:22:8: warning: nonnull argument ‘pubkey’ compared to NULL [-Wnonnull-compare]
     if (EXPECT(!(cond), 0)) { \
        ^
[...]

PR #367, opened 2018-08-21
Configure package for distribution

sangaman:
Solves #290. This prepares the xud package for distribution and installation via `npm install -g xud` with just the compiled files needed for execution and no source code for development. The `google-protobuf` package is included in the `grpc_tools_node_protoc_ts` dev dependency, and is added as a regular dependency for distribution.

I tested this myself by installing the package globally and confirmed that `xud` and `xucli` were added to my path.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "docs": "typedoc --out typedoc --module commonjs --target es6 lib",
     "lint": "tslint --project tsconfig.json && tslint --config tslint-alt.json 'bin/*' 'test/**/*.ts' 'tasks/**/*.ts'",
     "nodemon:watch": "nodemon --watch dist -e js dist/Xud.js",
-    "postinstall": "npm run compile",
+    "prepublishOnly": "npm run compile",
     "proto": "cross-os proto && cross-os swagger && cross-os protodocs",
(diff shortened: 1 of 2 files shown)


sangaman:
One quick note tho, I had to remove the `postinstall` script because it was compiling and end users won't be able to compile any longer. So for development/testing purposes, it's best to run xud with `npm run dev` instead of `npm start` or to do `npm run compile` before launching via the `xud` command.

package.json
```diff
@@ -91,6 +91,15 @@
     "url": "https://github.com/ExchangeUnion/xud/issues"
   },
   "homepage": "https://github.com/ExchangeUnion/xud#readme",
+  "files": [
```
michael1011: Is this a whitelist for files to be uploaded to NPM or should we add an additional `.npmignore` to make sure that the `lib` folder doesn't get distributed?

sangaman: It's a whitelist, I tried to use an npmignore file but it was not working as expected. While googling to try to fix it I came across this which convinced me to go the whitelist approach instead. https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d

PR #368, opened 2018-08-21
Resolve peer-pool circular import

sangaman:
This is a follow-up to #359 to resolve a circular import, TODOs, and other unresolved issues as well as fix failing tests.

Reslolves a circular import between the Peer and Pool classes by moving swap packet handling logic into Pool. Also makes some minor naming and logging changes regarding the swap packets.

### DIFF
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -84,7 +84,7 @@ class Xud extends EventEmitter {
       if (!this.raidenClient.isDisabled()) {
         initPromises.push(this.raidenClient.init());
       }
-      this.pool = new Pool(this.config.p2p, loggers.p2p, this.db, this.lndbtcClient, this.lndltcClient);
+      this.pool = new Pool(this.config.p2p, loggers.p2p, this.db);
 
       this.orderBook = new OrderBook(this.logger, this.db.models, this.pool, this.lndbtcClient, this.raidenClient);
       initPromises.push(this.orderBook.init());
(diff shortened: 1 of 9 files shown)


moshababo:
Following our discussion yesterday, I think we should do further separation of the swap logic from the p2p layer. 
We can keep all the packet handlers in `Pool` (even if there would be many of them), but remove logic related to swaps. 
I think it would be good to introduce `swaps` domain/folder, and to start putting all the logic there, in various files/classes.
`SwapDeals` can move there (currently in `orderbook` domain), and also it's active instance (currently in `Pool`).
`executeSwap` can be put there as well, while creating operational errors for all the ones which are used there.

This will free up both `Service` and `Pool`, which should me more thin. 

Another thing: In order to follow the entire swap procedure from one entry point, for easier testing and code flow, we can use the same pattern that was used in `initHello` - send a packet, and then `wait` for the response. That pattern can be extended and used instead of stand-alone packet handlers, which create one logical flow, but one that is hard to follow. (It's just an idea - we don't have to do that)

sangaman:
Ok great I was actually thinking the same thing re: a `swaps` folder. For the handlers, I'm thinking of a file with utility functions that do things like create a preimage and hash, update a swap deal, and initiate a swap. Does that sound good to you?

Regarding using the `initHello` style flow, I'll look at that too but I agree it's less necessary. We'd have to have two flows, one for maker and one for taker, also it's possible that we drop the latter packets in the future in which case the "flow" would be handing a single packet anyway.

sangaman:
@moshababo Take a look and tell me what you think.

sangaman:
@offerm If you want me to split the second refactoring commit into a separate commit I can, but please at least review the first commit so we can move forward with this.

moshababo:
@sangaman I started to review your changes before you removed them, looked good in the overall, but I need to give another look.

If we're in a hurry for Berlin we can postpone this refactoring.

PR #375, opened 2018-08-22
Fixed addresses bug

michael1011:
This PR fixes a bug that caused the external IP of the node not getting added to the `addresses` array in `Pool.ts`.


moshababo:
It was because it was previously added to `handshakeData` instead of `this.addresses`, or was it something else?

michael1011:
It was because it was added to `handshakeData.addresses` which was overwritten by `this.addresses` afterwards.

Issue #373, opened 2018-08-21
Unit test Parser problem


offerm:
Getting this when running npm run test: 

> 
>   Parser
>     ✓ should parse PING
>     ✓ should parse HELLO
>     ✓ should parse PING split
>     ✓ should parse HELLO split
>     ✓ should parse PING HELLO PING concatenated
>     ✓ should parse PING HELLO PING concatenated and split on the delimiter
>     ✓ should not try to parse an empty string
>     ✓ should try parse a standalone delimiter and fail
>     ✓ should buffer a max buffer length
> (node:36058) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'indexOf' of undefined
>     at Object.compatibleMessage (/Users/admin/github/xud/node_modules/check-error/index.js:82:29)
>     at getBasePromise.then.reason (/Users/admin/github/xud/node_modules/chai-as-promised/lib/chai-as-promised.js:191:77)
>     at process._tickCallback (internal/process/next_tick.js:68:7)
> (node:36058) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
> (node:36058) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
>     ✓ should not buffer when max buffer size exceeds
> 

While there is such an error, the test shows success.

sangaman:
Yup I made note of this in #330 and @moshababo said he would address in a future PR. We can leave this open to track the issue.

moshababo:
@sangaman From what I recall, that wasn't a problem, because the timeout is expected. 
So I didn't figure yet what was the problem on your env / travis (temporarily).

sangaman:
Oh maybe I misunderstood then. The parser tests have been passing fine in my env , they weren't at one time but it was fixed. It still seems like we should try to handle those promise rejections though? I haven't looked closely into it but the output log seems concerning.

moshababo:
Unhandled promise rejections is the problem i'm talking about, which I couldn't reproduce. We definitely shouldn't have these.

sangaman:
Ohhh I see, thanks for clarifying.

PR #376, opened 2018-08-22
parser unit tests

moshababo:
addressing #373 

The result of this test isn't consistent for some reason. I saw the problem, fixed it, but when I revoked my changes, the problem didn't reproduce.

I expect it to be a problem with `chai-as-promised` comparing thrown exception object.
I made a fix, so please check it on your env.
If that wouldn't work, i'll change the the expectation for `wait()` to simply get rejected, without any expected error.

If that wouldn't work, then the only think I can think of now is that emitting events isn't always synchronous, so line 20 isn't stable.

### DIFF
--- a/test/unit/Parser.spec.ts
+++ b/test/unit/Parser.spec.ts
@@ -8,6 +8,7 @@ chai.use(chaiAsPromised);
 
 describe('Parser', () => {
   const delimiter = Packet.PROTOCOL_DELIMITER;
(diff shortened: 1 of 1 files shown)


moshababo:
Travis is passing

moshababo:
Travis doesn't throw unhandled promise rejections after I applied my 2nd suggestion:  expect `wait()` to get rejected, without any expected error.

PR #377, opened 2018-08-22
Peer disconnect log message format

itsbalamurali:
Fixes log message format as in #231

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -214,7 +214,7 @@ class Pool extends EventEmitter {
     const peer = this.peers.get(nodePubKey);
     if (peer) {
       peer.close();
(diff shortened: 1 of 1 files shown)


lib/p2p/Pool.ts
```diff
@@ -214,7 +214,7 @@ class Pool extends EventEmitter {
     const peer = this.peers.get(nodePubKey);
     if (peer) {
       peer.close();
-      this.logger.info(`Disconnected from ${peer.nodePubKey} @ ${addressUtils.toString(peer.socketAddress)}`);
+      this.logger.info(`Disconnected from <${peer.nodePubKey}>@${addressUtils.toString(peer.socketAddress)}`);
```
sangaman: I don't think we actually want the `<>` surrounding the nodepubkey, but it would be good to remove the spaces on either side of the `@` as you've done.

PR #379, opened 2018-08-22
p2p sanity config fix

moshababo:
The tests could not run on my env because config wasn't loaded, and in my case it's necessary, because i'm using a password for the database.

In addition to that, I changed a bit how we create the config for each xud instance in the test.

### DIFF
--- a/test/p2p/sanity.spec.ts
+++ b/test/p2p/sanity.spec.ts
@@ -7,7 +7,7 @@ import Config from '../../lib/Config';
 
 chai.use(chaiAsPromised);
 
-const createConfig = (instanceId: number, p2pPort: number) => ({
+const createConfig = (instanceId: number, p2pPort: number, config: Config) => ({
   instanceId,
   p2p: {
     listen: true,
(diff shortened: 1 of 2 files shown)


test/p2p/sanity.spec.ts
```diff
@@ -37,14 +37,14 @@ describe('P2P Sanity Tests', () => {
   let nodeTwo: Xud;
 
   before(async () => {
-    nodeOneConfig = createConfig(1, 9001);
-    nodeTwoConfig = createConfig(2, 9002);
+    const config = new Config().load();
```
sangaman: I've been thinking, since the config loading process is now synchronous and expected to stay that way, should we just call `load()` from within the constructor and skip having to make this second call? It feels like it would be a lot cleaner that way.

moshababo: Indeed, but i'm usually not in favor of introducing complicated fetching procedures in the constructor, even though they are synchronous. It might also be good to be able to get the config object, without overriding it with the TOML config.

PR #380, opened 2018-08-22
Refactoring swaps

sangaman:
This implements swaps refactoring discussed in #368, putting swaps into it's module so as not to bloat `Pool` and have it handling with things unrelated to managing a pool of peers. We'll wait to merge until after we have basic swap functionality working, but we can start discussing it now.

Edit: I refactored this and rebased off the latest in master plus #399 (which is needed for successful swaps via order matches). This follows the structure suggested by @moshababo in his first impressions, where `Swaps` contains pool and the lightning clients, and `OrderBook`/`MatchesProcessor` contains swaps. I made `Swaps` an event emitter, for now it just has placeholder events but in a future PR I'll put in hooks in the order book to handle successful swaps (by invalidating the maker order) or failed swaps (by removing the peer order that failed the swap and attempting to match with a different order if we are the taker). This also adds the ability to reject deals if the currency is not supported, such as when lnd is disabled, so that we can still test order matching without swaps.

I've tested this to confirm swaps are still working.


lib/Xud.ts
```diff
@@ -76,7 +77,10 @@ class Xud {
       if (!this.raidenClient.isDisabled()) {
         initPromises.push(this.raidenClient.init());
       }
-      this.pool = new Pool(this.config.p2p, loggers.p2p, this.db, this.lndbtcClient, this.lndltcClient);
+
+      const swaps = new Swaps(loggers.swaps);
+
+      this.pool = new Pool(this.config.p2p, loggers.p2p, this.db, swaps);
 
       this.orderBook = new OrderBook(this.logger, this.db.models, this.pool, this.lndbtcClient, this.raidenClient);
```
moshababo: what about having `swaps` to handle the lnd/raiden clients, and to be used also for creating invoices, expose the swaps events, pubkeys, etc. ?

lib/service/Service.ts
```diff
@@ -195,36 +195,10 @@ class Service extends EventEmitter {
...
-      makerCoin: body.makerCoin,
-      createTime: Date.now(),
-    };
-    this.pool.swapDeals.add(deal);
-    this.logger.debug(' swap deal: ' + JSON.stringify(deal));
-
-    this.logger.debug('sending to peer ' + payload.nodePubKey + ': ' + JSON.stringify(body));
-    const packet = new packets.DealRequest(body);
+    const peer = this.pool.getPeer(payload.nodePubKey);
+    this.swaps.beginSwap(takerCurrency, makerCurrency, takerPubKey, payload.receivingAmount, payload.sendingAmount, peer);
```
moshababo: It might be better for `Swaps` should contain `Pool` (same as `OrderBook`), instead of the opposite. Here it's just about having it query the peer by itself, but it could make other things easier as well (direct packet listening, etc.). It will create a unidirectional flow, instead of `Pool` using `Swaps` as a util.

Also, `Swaps` should probably be inside `OrderBook` (or `MatchesProcessor`, in case it will still be needed), to activate swaps.

lib/Xud.ts
sangaman: Yeah I like that, makes sense.

lib/service/Service.ts
sangaman: Sounds good and makes sense as well.

lib/swaps/Swaps.ts
```diff
@@ -0,0 +1,336 @@
...
+      }
+    }
+    return undefined;
+  }
+
+  public addDeal = (deal: SwapDeal) => {
+    this.deals.push(deal);
+  }
+
+  public getDealByHash = (r_hash: string): SwapDeal | undefined => {
```
michael1011: What about using a map instead of an array to make this method way more efficient?

sangaman: I agree, I just didn't get to it in this PR. I'm thinking going forward there should be a single global deal id that both maker and taker use the way we do order ids (instead of each having their own) and then a mapping of dealId to deal and r_hash to deal.

kilrau:
`Service.executeSwap` looks good to have for Berlin! @sangaman

sangaman:
Yes although I am hoping to demo it via an order match so it's more relevant to what we're doing.

lib/service/Service.ts
```diff
@@ -331,17 +306,18 @@ class Service extends EventEmitter {
...
-      this.logger.error('Something went wrong. Can\'t find deal: ' + hash);
-      return 'Somthing is worng';
+      const msg = `Something went wrong. Can't find deal: ${hash}`;
+      this.logger.error(msg);
+      return msg;
     }
 
-	// If I'm the taker I need to forward the payment to the other chanin
-	// TODO: check that I got the right amount before sending out the agreed amount
+    // If I'm the taker I need to forward the payment to the other chanin
```
michael1011: Typo: `chain`

lib/swaps/Swaps.ts
```diff
@@ -0,0 +1,336 @@
...
+      try {
+        const sendPaymentResponse = await cmdLnd.sendPaymentSync(request);
+        if (sendPaymentResponse.getPaymentError()) {
+          throw sendPaymentResponse.getPaymentError();
+        }
+
+        // swap succeeded!
+        const r_preimage = Buffer.from(sendPaymentResponse.getPaymentPreimage_asB64(), 'base64').toString('hex');
+        this.logger.debug(`swap succeeded! ${r_preimage}`);
+        this.emit('swapCompleted');
```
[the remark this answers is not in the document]
sangaman: Yeah I was between those two, not sure which is better. There's also the `dealRejected` event to consider, however I don't know if we'll have any more "deal______" events but we probably will have more swap-related events. We use camel case for events in a few places, but I do feel that if we wind up with a lot of swap-related events the 'swap.completed' format is nicer.

offerm:
Suggest [again] to defer this until we complete these issues: #431 #430 #421 #413

kilrau:
As agreed, just merge @sangaman

PR #381, opened 2018-08-22
Make lint script autofix errors

sangaman:
This makes running `npm run lint` fix any errors that can be automatically fixed, typically trivial errors involving spacing and straightforward syntax rules.


sangaman:
Ahhhh that's a good point @michael1011, I wasn't aware. We could have a separate script for travis builds that doesn't fix I suppose. Any better ideas?

sangaman:
I'll have to test this behavior myself again, if it needs to be run twice in order to pass then travis shouldn't be an issue, right? I'll do some experimentation.

michael1011:
I just have to run it once to fix all fixable errors and let is pass:

```
> xud@1.0.0-prealpha.1 lint /home/michael/Git/JavaScript/xud
> tslint --fix --project tsconfig.json && tslint --config tslint-alt.json 'bin/*' 'test/**/*.ts' 'tasks/**/*.ts'

Fixed 5 error(s) in /home/michael/Git/JavaScript/xud/lib/utils/utils.ts
```

Yeah if we really want it to fix the errors a separate script for Travis would make sense.

sangaman:
Same behavior here, one run and it fixes errors and doesn't fail. So I added a `lintNoFix` script and updated our travis config to use it. I also fixed a warning for the recently added "no-circular-imports" rule.

sangaman:
@michael1011 Let me know if the current setup looks good to you and we can merge if so.

sangaman:
Ah yes thanks, that was just an oversight, definitely want to fix on both runs. I'll handle that and merge.

PR #388, opened 2018-08-22
Swap orderbook integration

sangaman:
This is the first step towards integration of swaps with the orderbook. The initial DealRequest packet is now triggered by a match with a peer order with the deal specifics from the order.

This currently treats each swap attempt as a success, assuming the "happy path", but going forward it will need to track updates to the swap flow. I think that will depend partly on #380, where the new `Swaps` class can be an event emitter to emit events when a swap executes, fails, times out, etc... and the orderbook can listen to these events and update the orders accordingly.

I took the buffer/queue out of MatchesProcessor because I believe that all matches will be handled sequentially and synchronously. We might not need this file/class going forward since it currently doesn't have any state.

I left in the currency & amount fields in the deal packets to allow for the `executeSwap` rpc call to continue working, but we should remove these once we're past the demo/early testing stage.

I also did a bit of refactoring and commenting to our orders types class, but the types are unchanged aside from the removal of `invoice` which is not relevant to how we're doing swaps currently.

### DIFF
--- a/lib/orderbook/MatchesProcessor.ts
+++ b/lib/orderbook/MatchesProcessor.ts
@@ -1,38 +1,85 @@
 import Logger from '../Logger';
 import { OrderMatch } from '../types/matchingEngine';
-import { isPeerOrder } from '../types/orders';
+import { isPeerOrder, StampedPeerOrder, StampedOwnOrder } from '../types/orders';
 import RaidenClient from '../raidenclient/RaidenClient';
 import Pool from '../p2p/Pool';
+import * as packets from '../p2p/packets';
+import uuidv1 from 'uuid/v1';
+import { SwapDeal } from './SwapDeals';
+import { SwapDealRole } from '../types/enums';
 
 class MatchesProcessor {
-  private buffer: any[];
+  constructor(private logger: Logger, private pool?: Pool, private raidenClient?: RaidenClient) { }
 
-  constructor(private logger: Logger, private pool?: Pool, private raidenClient?: RaidenClient) {
-    this.buffer = [];
-  }
-
-  public add(match: OrderMatch) {
-    this.buffer.push(match);
-
+  public process = (match: OrderMatch) => {
     if (isPeerOrder(match.maker)) {
-      this.notifyPeer(match);
-    } else if (isPeerOrder(match.taker)) {
       this.executeSwap(match);
     } else {
+      // internal match
       this.notifyClient(match);
     }
   }
 
-  public notifyPeer(_args: any) {
-    // TODO: notify the peer to trigger swap execution from his side
-  }
-
-  public notifyClient(_args: any) {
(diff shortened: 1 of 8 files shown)


lib/orderbook/MatchesProcessor.ts
```diff
@@ -1,38 +1,90 @@
...
-  public add(match: OrderMatch) {
-    this.buffer.push(match);
+  constructor(private logger: Logger, private pool?: Pool, private raidenClient?: RaidenClient) { }
 
+  public process = (match: OrderMatch) => {
     if (isPeerOrder(match.maker)) {
-      this.notifyPeer(match);
-    } else if (isPeerOrder(match.taker)) {
       this.executeSwap(match);
+    } else if (isPeerOrder(match.taker)) {
```
moshababo: This condition is no longer relevant - taker is always `ownOrder`.

We have only 2 conditions: either maker is `peerOrder` (non-internal match), or it's not (internal match).

lib/orderbook/MatchesProcessor.ts
```diff
@@ -1,38 +1,90 @@
...
+
+      this.pool.swapDeals.add(deal);
+      this.logger.debug('Initiating swap deal: ' + JSON.stringify(deal));
+
+      // TODO: Subscribe to swap events to handle swap resolution
+
+      const packet = new packets.DealRequest(dealRequestBody);
+
+      peer.sendPacket(packet);
+    }
```
moshababo: Was this implemented separately from `Service.executeSwap` because you wanted to postpone dealing with that and with #380 to a later stage? or was there a different reason?

sangaman: Yes that's the main reason, also they just function a bit different because `Service.executeSwap` doesn't pertain to any particular order.

lib/orderbook/MatchesProcessor.ts
sangaman: Yeah I thought this too but wasn't sure to drop it. Basically this is relevant to the, as far as I understand, still unresolved question of what to do when you receive an order from a peer that *should* match immediately. But I'm pretty sure that shouldn't be considered a "match" in any case, so I'll cut out this code and make sure we have an issue for it to discuss.

sangaman:
Removed the dead code. @moshababo, do are you agree that we might not have a need for MatchesProcessor any longer? I'm not seeing a need for the queue/buffer that was there before.

moshababo:
Removing the queue/buffer is fine, but we might eventually still need `MatchesProcessor` to notify on internal matches, and to listen to to the swap result and finalize it.

sangaman:
Yeah, my thought is just that having a class for `MatchesProcessor` isn't really necessary if it's just a collection of stateless functions - but it's not important now and we can see how things take shape as we add more match handling logic.

PR #389, opened 2018-08-23
Don't wait for outgoing connections on start

### LINKED ISSUE #387: Remove "wait for node connection success" from xud startup process
> Currently we wait for all connections to be reestablished to peers, which is normally quite fast but can take 5-10 seconds if there are timeouts.

This is issue is about removing this wait and dependency on connection success and just letting the connections resolve after the xud server is up and ready.

sangaman:
Resolves #387. This no longer waits for connections to resolve on start before completing the startup process.

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -104,7 +104,11 @@ class Pool extends EventEmitter {
 
     this.logger.info('Connecting to known / previously connected peers');
     await this.nodes.load();
-    await this.connectNodes(this.nodes);
+    this.connectNodes(this.nodes).then(() => {
+      this.logger.info('Completed start-up connections to known peers.');
+    }).catch((reason) => {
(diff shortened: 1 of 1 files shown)


lib/p2p/Pool.ts
```diff
@@ -104,7 +104,10 @@ class Pool extends EventEmitter {
 
     this.logger.info('Connecting to known / previously connected peers');
     await this.nodes.load();
-    await this.connectNodes(this.nodes);
+    const connectPromise = this.connectNodes(this.nodes);
+    connectPromise.then(() => {
+      this.logger.info('Completed start-up connections to known peers.');
+    });
```
moshababo: * no need for `connectPromise`. `then` can be applied to `connectNodes`
* need to add `catch`. it's currently failing tslint

sangaman: Thanks I should've caught this myself, done.

PR #390, opened 2018-08-23
Configurable log level

michael1011:
This PR:
- makes the log level configurable
- adds verbose messages for relevant packets
- changes the log level of a few already existing messages
- changes the default log level to verbose


sangaman:
I think checking the log level in Config.ts and using the default value if it's an invalid one would be a good idea.

I think `verbose` is too low of a level for default, although I just realized I'd misunderstood the levels all this time. I thought `debug` was above `verbose`, where `verbose` was almost like a "trace" level but I guess that's really the "silly" level. I'd say the default should probably be "info" - and the log messages for things like raw incoming p2p messages should be "silly." Maybe the parsed packets themselves can be verbose. The way I'm thinking, under normal operation there will be lots of packets going back and forth and it will quickly fill the logs. Regular operation (which again I'm thinking of as "info" level) should have relatively clean and brief messages for meaningful events like peers connecting/dropping, order events, swap events, etc...

bin/xud
```diff
@@ -14,6 +14,11 @@ const { argv } = require('yargs')
       type: 'string',
       alias: 'x',
     },
+    loglevel: {
```
sangaman: Our convention for other args is snake_case with the underscore so let's stick to that, although I like e `-l` alias.

lib/Config.ts
```diff
@@ -47,6 +49,7 @@ class Config {
 
     // default configuration
     this.initDb = true;
+    this.loglevel = process.env.NODE_ENV === 'production' ? Level.INFO : Level.VERBOSE;
```
sangaman: Oh did you mean verbose for the default non-production logger levels? I'd prefer keeping it to default to DEBUG for testing/development. Eventually, certainly by beta, we should have the default NODE_ENV be production though.

lib/Logger.ts
```diff
@@ -127,6 +128,11 @@ class Logger {
   public debug = (msg: string) => {
     this.log(Level.DEBUG, msg);
   }
+
+  public silly = (msg: string) => {
```
sangaman: So I know the underlying winston level is `SILLY`, but I might prefer calling this method `trace` for our own purposes. `logger.trace()` sounds more sensible to me than `logger.silly()`. Thoughts?

lib/p2p/Pool.ts
```diff
@@ -177,7 +177,6 @@ class Pool extends EventEmitter {
       throw err;
     } else if (this.peers.has(nodePubKey)) {
       const err = errors.NODE_ALREADY_CONNECTED(nodePubKey, address.host);
-      this.logger.warn(err.message);
```
sangaman: Is there a reason you dropped this message? Is something similar being logged elsewhere? Just wondering.

test/integration/OrderBook.spec.ts
```diff
@@ -18,7 +18,7 @@ describe('OrderBook', () => {
   before(async () => {
     const config = new Config();
     await config.load();
-    const loggers = Logger.createLoggers();
+    const loggers = Logger.createLoggers(Level.DEBUG);
```
sangaman: For our automated test output, would it make sense to keep the log output limited? We may only be really interested in errors or maybe warnings. Also, I'm thinking we should have a way to not write anything to files for automated test logging, so maybe we also need a flag on whether to log to file or just to console. That can be a separate PR or can be handled here.

michael1011:
The sequelize logging comes down to setting the `logging` option to `logger.silly`. For example:
```
public getNodes = async (): Promise<db.NodeInstance[]> => {
    return this.models.Node.findAll({
      logging: this.logger.silly,
    });
  }
```
The annoying thing is that we would have to set it for *every* sequelize call. But we don't have that many database queries anyways so I could do that in that PR if you agree? @sangaman

test/integration/OrderBook.spec.ts
michael1011: Do you think of a cli flag or just an argument for `Logger.createLoggers`?

If you mean a cli flag: what about adding a cli flag for the path of the logfile and setting it to `/dev/null` in case you don't want to write to a file?

bin/xud
sangaman: Yeah you're right actually - I was confusing the `xucli` params which use snake_case with the `xud` and config args which use camel case. I'd be fine with switching xucli over to camel case too actually for the sake of consistency? That can definitely be a separate PR tho. I think I went with snake case earlier because the rpc fields in the proto definition use it, but that actually shouldn't matter to us.

test/integration/OrderBook.spec.ts
sangaman: I think we can probably just have an argument in Logger.createLoggers which we'd only set from tests. I do think a configurable logging path would be a good idea too, but `/dev/null` wouldn't work on windows and I'm not sure there's an equivalent. Another option would be to set it to an empty string when we don't want file output, then our code can check for an empty log file path and disable logging to file.

Do you agree with setting the logger level to WARN or ERROR in the tests though?

sangaman:
Actually I think in DB.ts line 61:

```javascript
const options = { logging: this.logger.verbose };
```

I switched this to `const options = { logging: false };` and it disabled sql query logging. I bet if you switch it to const `options = { logging: this.logger.trace };` (after renaming `logger.silly` to `logger.trace`) it will log using the SILLY level and we won't see it during regular debugging.

test/integration/OrderBook.spec.ts
michael1011: Yeah I agree on setting the logging level to `warn` in the tests. Right now they are way too cluttered imo.

I will take care of the logging path in another PR because this one is already way bigger than I planned.

michael1011:
Renaming `silly` to `trace` turned out to be way trickier than I expected. I had to create our own log levels and logic for coloring them. 

And I also set the log level of the database queries. Take a look when you get time @sangaman

sangaman:
Oh, I was thinking just name a method that logs `SILLY` messages as `trace` for when we call it internally. But if you were able to create a new level and it works then that's cool, I'll check it out soon.

michael1011:
I tried changing just that single line but it didn't work so I did those other changes but feel free to correct me if I got it wrong.

sangaman:
Just that one line seems to work for me... I pulled down this branch, undid the changes to DB.ts and the p2p/orderbook files, then changed line 61 to `const options = { logging: this.logger.trace };` and here's what the output looks like on startup, no db logging:

```bash
2018-8-26 16:14:05 [GLOBAL] info: config loaded
2018-8-26 16:14:05 [GLOBAL] info: Local nodePubKey is 029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345
Executing (default): SELECT 1+1 AS result
2018-8-26 16:14:05 [DB] info: connected to database. host:localhost port:3306 database:xud
2018-8-26 16:14:05 [LND] info: LndClient status: DISCONNECTED
2018-8-26 16:14:05 [LND] info: trying to verify connection to lnd with uri: localhost:10002
2018-8-26 16:14:05 [LND] info: LndClient status: DISCONNECTED
2018-8-26 16:14:05 [LND] info: trying to verify connection to lnd with uri: localhost:10001
2018-8-26 16:14:05 [LND] info: LndClient status: CONNECTION_VERIFIED
2018-8-26 16:14:05 [LND] info: LndClient status: CONNECTION_VERIFIED
2018-8-26 16:14:05 [P2P] info: Connecting to known / previously connected peers
Executing (default): SELECT `id`, `nodePubKey`, `addressesText`, `banned`, `createdAt`, `updatedAt` FROM `nodes` AS `Node`;
2018-8-26 16:14:05 [P2P] info: Completed start-up connections to known peers.
2018-8-26 16:14:05 [P2P] info: p2p server listening on 0.0.0.0:8885
2018-8-26 16:14:05 [RPC] info: gRPC server listening on localhost:8886
```

When I set it to `const options = { logging: this.logger.debug };` the db logging lines are back with the debug setting as expected:

```bash
2018-8-26 16:14:29 [GLOBAL] info: config loaded
2018-8-26 16:14:29 [GLOBAL] info: Local nodePubKey is 029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345
Executing (default): SELECT 1+1 AS result
2018-8-26 16:14:29 [DB] info: connected to database. host:localhost port:3306 database:xud
2018-8-26 16:14:29 [DB] debug: Executing (default): CREATE TABLE IF NOT EXISTS `nodes` (`id` INTEGER auto_increment , `nodePubKey` VARCHAR(255) NOT NULL UNIQUE, `addressesText` TEXT NOT NULL, `banned` TINYINT(1), `createdAt` DATETIME NOT NULL, `updatedAt` DATETIME NOT NULL, PRIMARY KEY (`id`)) ENGINE=InnoDB;
2018-8-26 16:14:29 [DB] debug: Executing (default): CREATE TABLE IF NOT EXISTS `currencies` (`id` VARCHAR(255) , `tokenAddress` VARCHAR(255), PRIMARY KEY (`id`)) ENGINE=InnoDB;
2018-8-26 16:14:29 [DB] debug: Executing (default): SHOW INDEX FROM `nodes`
2018-8-26 16:14:29 [DB] debug: Executing (default): SHOW INDEX FROM `currencies`
2018-8-26 16:14:29 [DB] debug: Executing (default): CREATE TABLE IF NOT EXISTS `pairs` (`id` VARCHAR(255) , `baseCurrency` VARCHAR(255) NOT NULL, `quoteCurrency` VARCHAR(255) NOT NULL, `swapProtocol` ENUM('LND', 'RAIDEN') NOT NULL, PRIMARY KEY (`id`), FOREIGN KEY (`baseCurrency`) REFERENCES `currencies` (`id`) ON DELETE NO ACTION ON UPDATE CASCADE, FOREIGN KEY (`quoteCurrency`) REFERENCES `currencies` (`id`) ON DELETE NO ACTION ON UPDATE CASCADE) ENGINE=InnoDB;
2018-8-26 16:14:29 [DB] debug: Executing (default): SHOW INDEX FROM `pairs`
2018-8-26 16:14:29 [LND] info: LndClient status: DISCONNECTED
2018-8-26 16:14:29 [LND] info: trying to verify connection to lnd with uri: localhost:10002
2018-8-26 16:14:29 [LND] info: LndClient status: DISCONNECTED
2018-8-26 16:14:29 [LND] info: trying to verify connection to lnd with uri: localhost:10001
2018-8-26 16:14:29 [LND] info: LndClient status: CONNECTION_VERIFIED
2018-8-26 16:14:29 [LND] info: LndClient status: CONNECTION_VERIFIED
2018-8-26 16:14:29 [P2P] info: Connecting to known / previously connected peers
Executing (default): SELECT `id`, `nodePubKey`, `addressesText`, `banned`, `createdAt`, `updatedAt` FROM `nodes` AS `Node`;
2018-8-26 16:14:29 [P2P] info: Completed start-up connections to known peers.
2018-8-26 16:14:29 [P2P] info: p2p server listening on 0.0.0.0:8885
2018-8-26 16:14:29 [RPC] info: gRPC server listening on localhost:8886
```

Is it not the same for you?

michael1011:
It is the same for me. But keep in mind that you have to pass `options` as parameter to affect the log level of the query. And I think that all should have the same log level.

sangaman:
I see what you mean, with just that line future queries will have the default logger, which I didn't realize at first. I did some digging and if we pass the logger setting to the `Sequelize` constructor, it carries forward to all future queries. Reverting everything in DB.ts and the orderbook/p2p files and then changing `createSequelizeInstance()` to add this `logging` line as shows below seems to work and apply to all queries. I suggest keeping it at trace since we shouldn't really need to see the queries in the logs.

```javascript
  private createSequelizeInstance = (config: SequelizeConfig): Sequelize.Sequelize => {
    return new Sequelize({
      ...config,
      logging: this.logger.trace,
      dialect: 'mysql',
      operatorsAliases: false,
      dialectOptions: {
        multipleStatements: true,
      },
    });
  }
```

michael1011:
One more thing: since we are just logging the levels `warn` and `error` for the tests I would push for the "recommended setup" which comes down to committing the content of your `.vscode` folder. That should make debugging of the tests easier. @sangaman

Issue #311, opened 2018-08-09
Connection retry logic of an momentarily unavailable peer


kilrau:
I don't think we got this yet: https://github.com/ExchangeUnion/xud/issues/185#issuecomment-411595146

Do this now or as part of https://github.com/ExchangeUnion/xud/issues/178 ?

sangaman:
Yeah we had discussed an exponential backoff on connection attempts, and I think that would be a pretty decent and simple way to implement this functionality at least at first. We'll keep trying to connect with longer waits between attempts after each failure.

kilrau:
Proposal:
The minimum amount of time, in seconds, that `xud` will wait before it tries to reconnect to the source or target destination after a failure.

This attribute works with `Minimum Delay`, `Maximum Delay` and `Incremental Delay` (all in seconds) attributes. After the first failure to connect to a destination, `xud` will wait for the number of seconds defined by `Minimum Delay`.

If the second trial also fails, it will increase its waiting time by the number of seconds defined by `Incremental Delay`. The `maximum delay` time is defined by Maximum Delay. Once the waiting time is increased to the maximum value, `xud` will not increase its waiting time anymore.

Once `xud` successfully connects to a peer, its waiting time will be reset to the initial value defined by `Minimum Delay`.

Source: https://docs.oracle.com/cd/E13222_01/wls/docs81/ConsoleHelp/domain_messagingbridge_config_connectionretry.html

sangaman:
I think a week isn't that long if/when we target non-exchanges that don't expect to be online 24/7. It would be a good thing to have configurable I think, but something around there may be fine for now. But mainly I think a 5 second fixed backoff is too little and will result in way too many connection attempts, I think an exponential backoff (5, 10, 20, 40, 80, 160, etc) makes a lot more sense.

Also I don't think we should delete the node from the db per se, maybe we just drop its listening addresses from the db indicating that we no longer believe those are good and we stop trying to connect to them.

kilrau:
adjusted

moshababo:
@kilrau 

> Once the waiting time is increased to the maximum value, xud will not increase its waiting time anymore.

It will not increase, or reconnection attempts should stop?

Another thing: `Maximum Delay` is smaller then `DeleteNodes Delay`, so how would the later ever be reached?
We can instead stop trying to connect to a node if the last successful connection attempt was too long ago.  

At the moment we're not listing a node in db before our first successful connection. 
[...]

Issue #386, opened 2018-08-22
Share peers


kilrau:
- Share currently connected peers only, not entire 'nodes' table
- Connect to newly received peers dynamically at run time (after receiving nodes via `GET_NODES`)

sangaman:
I can take care of the first bullet point. The second bullet point is mistaken, we do connect to newly received nodes when received not only on startup. For the third point, you mean gossiping to all our peers every time we connect to a new peer? I think that would get very noisy. Rather, we've discussed having the ability to query peers to send their known nodes.

sangaman:
I think better than a query interval may be a way to manually trigger it, like through an rpc call. Maybe let's open another issue for that to discuss because I think it's not exactly related to the first point.

kilrau:
EDIT: moved 3rd bullet point into: https://github.com/ExchangeUnion/xud/issues/402

PR #392, opened 2018-08-23
connection retries

moshababo:
Solves #311 

* retries are blocking `initConnection`
* retries are not activated when manually trying to connect to a peer. if it would, the user won't get a response for a very long time upon failure, since the retries are blocking.
* if max retries period exceeds, and peer was previously connected (exists on `nodes` table)
peer supposed to get banned. Banning doesn't work now for pre-handshake peers, but can be fixed. I'm not sure banning is the right solution, because in the future it might also block inbound connections. Deleting or soft deleting the node entry might not be the right thing either. Perhaps penalizing, which we haven't implemented yet. 
Anyway - that will hardly happen, because it will trigger only after 1 week, and only for previously connected peers.
The main thing now is to verify that trying to retry over an entire week for every node we're hearing about is reasonable. I think it's too much for nodes we never successfully connected with.

### DIFF
--- a/lib/grpc/GrpcService.ts
+++ b/lib/grpc/GrpcService.ts
@@ -119,9 +119,9 @@ class GrpcService {
    */
   public disconnect: grpc.handleUnaryCall<xudrpc.DisconnectRequest, xudrpc.DisconnectResponse> = async (call, callback) => {
     try {
-      const disconnectResponse = await this.service.disconnect(call.request.toObject());
+      await this.service.disconnect(call.request.toObject());
       const response = new xudrpc.DisconnectResponse();
-      response.setResult(disconnectResponse);
+      response.setResult('success');
       callback(null, response);
     } catch (err) {
       callback(this.getGrpcError(err), null);
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -1,6 +1,7 @@
 import P2PRepository from './P2PRepository';
 import { NodeInstance, NodeFactory } from '../types/db';
-import { NodeConnectionInfo, Address } from '../types/p2p';
+import { Address } from '../types/p2p';
+import addressUtils from '../utils/addressUtils';
 
 /** Represents a list of nodes for managing network peers activity */
 class NodeList {
@@ -69,9 +70,32 @@ class NodeList {
   public updateAddresses = async (nodePubKey: string, addresses: Address[] = []) => {
     const node = this.nodes.get(nodePubKey);
     if (node) {
-      node.addresses = addresses;
+      // avoid overriding the `lastConnected` field for existing matching addresses unless a new value was set
+      node.addresses = addresses.map((newAddress) => {
+        const oldAddress = node.addresses.find(address => addressUtils.areEqual(address, newAddress));
+        if (oldAddress && !newAddress.lastConnected) {
+          return oldAddress;
+        } else {
+          return newAddress;
+        }
+      });
+
       await this.repository.updateNode(node);
       return true;
+    }
+
+    return false;
+  }
+
+  public removeAddress = async (nodePubKey: string, address: Address) => {
+    const node = this.nodes.get(nodePubKey);
+    if (node) {
+      const index = node.addresses.findIndex(existingAddress => addressUtils.areEqual(address, existingAddress));
+      if (index > -1) {
+        node.addresses = [...node.addresses.slice(0, index), ...node.addresses.slice(index + 1)];
+        await this.repository.updateNode(node);
+        return true;
+      }
     }
 
     return false;
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -39,7 +39,6 @@ interface Peer {
 
 /** Represents a remote XU peer */
 class Peer extends EventEmitter {
-  // TODO: properties documentation
   public socketAddress!: Address;
   public inbound!: boolean;
   public connected = false;
@@ -62,10 +61,16 @@ class Peer extends EventEmitter {
   private static STALL_INTERVAL = 5000;
   /** Interval for pinging peers. */
   private static PING_INTERVAL = 30000;
-  /** Socket connection timeout for outbound peers. */
-  private static CONNECTION_TIMEOUT = 10000;
   /** Response timeout for response packets. */
   private static RESPONSE_TIMEOUT = 10000;
+  /** Socket connection timeout for outbound peers. */
+  private static CONNECTION_TIMEOUT = 10000;
+  /** Connection retries min delay. */
+  private static CONNECTION_RETRIES_MIN_DELAY = 5000;
+  /** Connection retries max delay. */
+  private static CONNECTION_RETRIES_MAX_DELAY = 3600000;
+  /** Connection retries max period. */
+  private static CONNECTION_RETRIES_MAX_PERIOD = 604800000;
 
   public get nodePubKey(): string | undefined {
     return this.handshakeState ? this.handshakeState.nodePubKey : undefined;
@@ -122,14 +127,17 @@ class Peer extends EventEmitter {
    * and emit the `open` event if everything succeeds. Throw an error on unexpected handshake data.
    * @param handshakeData our handshake data to send to the peer
    * @param nodePubKey the expected nodePubKey of the node we are opening a connection with
+   * @param retryConnecting whether to retry to connect upon failure
    */
-  public open = async (handshakeData: HandshakeState, nodePubKey?: string): Promise<void> => {
+  public open = async (handshakeData: HandshakeState, nodePubKey?: string, retryConnecting?: boolean): Promise<void> => {
     assert(!this.opened);
     assert(!this.closed);
+    assert(this.inbound || nodePubKey);
+    assert(!retryConnecting || !this.inbound);
 
     this.opened = true;
 
-    await this.initConnection();
+    await this.initConnection(retryConnecting);
     this.initStall();
     await this.initHello(handshakeData);
 
@@ -234,7 +242,7 @@ class Peer extends EventEmitter {
    * Ensure we are connected (for inbound connections) or listen for the `connect` socket event (for outbound connections)
    * and set the [[connectTime]] timestamp. If an outbound connection attempt errors or times out, throw an error.
    */
-  private initConnection = async () => {
+  private initConnection = async (retry = false) => {
     assert(this.socket);
 
     if (this.connected) {
@@ -244,7 +252,12 @@ class Peer extends EventEmitter {
       return;
     }
 
-    return new Promise<void>((resolve, reject) => {
+    return new Promise((resolve, reject) => {
+      const address: Address = { port: this.socketAddress.port, host: this.socketAddress.host };
+      const startTime = Date.now();
+      let retryDelay = Peer.CONNECTION_RETRIES_MIN_DELAY;
+      let retries = 0;
+
       const cleanup = () => {
         if (this.connectTimeout) {
           clearTimeout(this.connectTimeout);
@@ -254,11 +267,6 @@ class Peer extends EventEmitter {
           this.socket.removeListener('error', onError);
           this.socket.removeListener('connect', onConnect);
         }
-      };
-
-      const onError = (err: Error) => {
-        cleanup();
-        reject(err);
       };
 
       const onConnect = () => {
@@ -271,16 +279,41 @@ class Peer extends EventEmitter {
         resolve();
       };
 
-      const onTimeout = () => {
+      const onError = (err: Error) => {
         cleanup();
-        reject(new Error('Connection timed out.'));
+
+        if (!retry) {
+          this.close();
+          reject(err);
+          return;
+        }
+
+        if (Date.now() - startTime + retryDelay > Peer.CONNECTION_RETRIES_MAX_PERIOD) {
+          this.close();
+          reject(errors.CONNECTING_RETRIES_MAX_PERIOD_EXCEEDED);
+          return;
+        }
+
+        this.logger.debug(
+          `Connection attempt #${retries + 1} to peer (${addressUtils.toString(this.socketAddress)}) ` +
+          `failed: ${err.message}. retrying in ${retryDelay / 1000} sec...`,
+        );
+
+        setTimeout(() => {
+          retryDelay = Math.min(Peer.CONNECTION_RETRIES_MAX_DELAY, retryDelay * 2);
+          retries = retries + 1;
+          this.socket!.connect(address);
+          bind();
+        }, retryDelay);
       };
 
-      this.socket!.once('connect', onConnect);
+      const bind = () => {
+        this.socket!.once('connect', onConnect);
+        this.socket!.once('error', onError);
+        this.connectTimeout = setTimeout(() => onError(new Error('Connection timed out')), Peer.CONNECTION_TIMEOUT);
+      };
 
-      this.socket!.once('error', onError);
-
-      this.connectTimeout = setTimeout(onTimeout, Peer.CONNECTION_TIMEOUT);
+      bind();
     });
   }
 
@@ -434,8 +467,8 @@ class Peer extends EventEmitter {
         } else {
           this.logger.info(`Peer ${this.nodePubKey} socket closed`);
         }
+        this.close();
       }
-      this.close();
     });
 
     this.socket.on('data', (data) => {
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -1,6 +1,6 @@
 import net, { Server, Socket } from 'net';
 import { EventEmitter } from 'events';
-import errors from './errors';
+import errors, { errorCodes } from './errors';
 import Peer, { PeerInfo } from './Peer';
 import NodeList from './NodeList';
 import PeerList from './PeerList';
@@ -115,7 +115,7 @@ class Pool extends EventEmitter {
 
     this.logger.info('Connecting to known / previously connected peers');
     await this.nodes.load();
-    this.connectNodes(this.nodes).then(() => {
+    this.connectNodes(this.nodes, false, true).then(() => {
       this.logger.info('Completed start-up connections to known peers.');
     }).catch((reason) => {
       this.logger.error('Unexpected error connecting to known peers on startup', reason);
@@ -163,16 +163,17 @@ class Pool extends EventEmitter {
    * If the node is banned, already connected, or has no listening addresses, then do nothing.
    * @param nodes a collection of nodes with a `forEach` iterator to attempt to connect to
    * @param ignoreKnown whether to ignore nodes we are already aware of, defaults to false
+   * @param retryConnecting whether to attempt retry connecting, defaults to false
    * @returns a promise that will resolve when all outbound connections resolve
    */
-  private connectNodes = (nodes: NodeConnectionIterator, ignoreKnown = false) => {
+  private connectNodes = (nodes: NodeConnectionIterator, ignoreKnown = false, retryConnecting = false) => {
     const connectionPromises: Promise<void>[] = [];
     nodes.forEach((node) => {
       // check that this node is not ourselves, that it has listening addresses,
       // and that either we haven't heard of it, or we're not ignoring known nodes and it's not banned
       if (node.nodePubKey !== this.handshakeData.nodePubKey && node.addresses.length > 0 &&
         (!this.nodes.has(node.nodePubKey) || (!ignoreKnown && !this.nodes.isBanned(node.nodePubKey)))) {
-        connectionPromises.push(this.connectNode(node));
+        connectionPromises.push(this.tryConnectNode(node, retryConnecting));
       }
     });
     return Promise.all(connectionPromises);
@@ -181,14 +182,27 @@ class Pool extends EventEmitter {
   /**
    * Attempt to create an outbound connection to a node using its known listening addresses.
    */
-  private connectNode = async ({ addresses, nodePubKey }: NodeConnectionInfo) => {
-    for (let n = 0; n < addresses.length; n += 1) {
+  private tryConnectNode = async (node: NodeConnectionInfo, retryConnecting = false) => {
+    const { addresses, nodePubKey } = node;
+
+    // sort by lastConnected desc
+    const sortedAddresses = [...addresses].sort((a, b) => {
+      if (!a.lastConnected) return 1;
+      if (!b.lastConnected) return -1;
+      return b.lastConnected - a.lastConnected;
+    });
+
+    for (const address of sortedAddresses) {
       try {
-        await this.addOutbound(addresses[n], nodePubKey);
-        break; // once we've successfully established an outbound connection, stop attempting new connections
-      } catch (err) {
-        this.logger.info(err);
-      }
+        await this.addOutbound(address, nodePubKey, false);
+        return; // once we've successfully established an outbound connection, stop attempting new connections
+      } catch (err) {}
+    }
+
+    if (retryConnecting && sortedAddresses.length && sortedAddresses[0].lastConnected) {
+      try {
+        await this.addOutbound(sortedAddresses[0], nodePubKey, true);
+      } catch (err) {}
     }
   }
 
@@ -199,7 +213,7 @@ class Pool extends EventEmitter {
    * @param nodePubKey the nodePubKey of the node to connect to
    * @returns the connected peer
    */
-  public addOutbound = async (address: Address, nodePubKey: string): Promise<Peer> => {
+  public addOutbound = async (address: Address, nodePubKey: string, retryConnecting: boolean): Promise<Peer> => {
     if (nodePubKey === this.handshakeData.nodePubKey) {
       const err = errors.ATTEMPTED_CONNECTION_TO_SELF;
       this.logger.warn(err.message);
@@ -210,7 +224,7 @@ class Pool extends EventEmitter {
     }
 
     const peer = Peer.fromOutbound(address, this.logger);
-    await this.tryOpenPeer(peer, nodePubKey);
+    await this.openPeer(peer, nodePubKey, retryConnecting);
     return peer;
   }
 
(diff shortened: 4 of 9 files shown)


sangaman:
> if max retries period exceeds, and peer was previously connected (exists on nodes table)
peer supposed to get banned. Banning doesn't work now for pre-handshake peers, but can be fixed. I'm not sure banning is the right solution, because in the future it might also block inbound connections.

I definitely don't think we want to ban. I like the idea of a soft delete. Maybe we add another boolean column to the nodes table, and don't try connecting to them again unless it's done manually or we hear about them again from a `NODES` packet from another peer. If we're able to successfully reconnect, then we undelete them. Another option for a soft delete is to delete their known listening addresses. I figure this matter can be dealt with in another PR though.

Also, I agree that we don't need to retry connecting to nodes that we've never connected with. I think one try is enough. If it's a manual connection, failing lets the user know to retry when the issue blocking the connection is resolved. If it's automatic, like a node we learned about from a peer, #386 makes it such that we should only be learning about nodes that are online from peers.

I see retrying as valuable when attempting to reconnect to a peer who has temporarily gone down, or whom we've connected to successfully in the past but was offline at the time xud started.

lib/p2p/Peer.ts
```diff
@@ -122,14 +127,15 @@ class Peer extends EventEmitter {
    * and emit the `open` event if everything succeeds. Throw an error on unexpected handshake data.
    * @param handshakeData our handshake data to send to the peer
    * @param nodePubKey the expected nodePubKey of the node we are opening a connection with
+   * @param retryConnecting whether to retry to connect upon failure
    */
-  public open = async (handshakeData: HandshakeState, nodePubKey?: string): Promise<void> => {
+  public open = async (handshakeData: HandshakeState, nodePubKey?: string, retryConnecting?: boolean): Promise<void> => {
```
sangaman: I think it's clear enough to just call this new param `retry` which is also consistent with how this flag is named elsewhere.

Edit: Actually after looking through more changes it looks like `retryConnecting` is actually used in most places. Personally I still prefer `retry` but if you want to keep `retryConnecting` then I'd use that same naming in the `initConnection` method.

lib/p2p/Peer.ts
```diff
@@ -239,6 +245,10 @@ class Peer extends EventEmitter {
     }
 
     return new Promise((resolve, reject) => {
+      const startTime = Date.now();
+      let retryDelay: number | undefined;
```
sangaman: How about we initialize `retryDelay` by setting it to `CONNECTION_RETRIES_MIN_DELAY`, and then move the line that doubles the retry delay to after the `retryDelay > Peer.CONNECTION_RETIES_MAX_PERIOD` check. It feels a bit cleaner that way, and saves a bunch of null checks on `retryDelay`.

lib/p2p/Peer.ts
```diff
@@ -265,16 +270,45 @@ class Peer extends EventEmitter {
...
+          reject(err);
+          return;
+        }
+
+        retryDelay = !retryDelay
+          ? Peer.CONNECTION_RETRIES_MIN_DELAY
+          : Math.min(Peer.CONNECTION_RETIES_MAX_DELAY, retryDelay * 2);
+
+        if (Date.now() - startTime + retryDelay > Peer.CONNECTION_RETIES_MAX_PERIOD) {
+          this.close();
```
sangaman: Here too I don't think we need to `close` because we're not connected in the first place.

lib/p2p/Peer.ts
```diff
@@ -265,16 +270,45 @@ class Peer extends EventEmitter {
...
 
-      this.socket!.once('error', onError);
+        setTimeout(() => {
+          retries = retries + 1;
+          this.socket!.connect({ port: this.socketAddress.port, host: this.socketAddress.host });
+          listen();
+        }, retryDelay);
+      };
+
+      const listen = () => {
```
sangaman: "listen" seems a bit inaccurate here, it makes it sound like we're listening for incoming connections. Maybe "bind"?

lib/p2p/Peer.ts
```diff
@@ -265,16 +270,45 @@ class Peer extends EventEmitter {
...
+          : Math.min(Peer.CONNECTION_RETIES_MAX_DELAY, retryDelay * 2);
+
+        if (Date.now() - startTime + retryDelay > Peer.CONNECTION_RETIES_MAX_PERIOD) {
+          this.close();
+          reject(new Error(`${retries} connection retries failed`));
+          this.emit('ban');
+          return;
+        }
+
+        this.logger.info(
```
sangaman: I think `debug` or perhaps `verbose` would be better here, this feels like we'd have a lot of messages in the logs that aren't particularly useful.

moshababo:
> have you tested this yet?

I tested it manually. I’d like to add some automatic tests as well, after we’ll settle on everything.

> Another thing that occurred to me while I was going through this, how do we handle a case where we have a timer set to retry a connection to a node and in the meantime that node successfully connects to us. Maybe we need to make some check before retrying a connection to ensure we're not already connected?

It’s not handled at the moment. If the inbound node would suddenly become available for us for outbound as well, then a duplicated connection supposed to be detected. If it’s not available for us, it will continue to retry.  I can add some checks to avoid that, hopefully it won’t become to verbose. 

> Also if I follow correctly, Pool will contain Peers that we're not currently connected to, but instead are only pending reconnection attempts. I think that may affect some other logic that treats all the peers in Pool as actively connected. Which makes me also realize that it's possible we'd have some pre-handshake peers in Pool and we should check for those before doing things like sending them orders. Maybe I will address that in a separate PR, by checking that we're in post handshake state with peers before we attempt to send them packets.

The peer object will exist, but it's not accessible and not on `PeerList`, because peer are getting added there only after they are “opened”, while `initConnection` with its retries is blocking it.

lib/p2p/Peer.ts
```diff
@@ -265,16 +270,45 @@ class Peer extends EventEmitter {
...
 
-      const onTimeout = () => {
+      const onError = (err: Error) => {
         cleanup();
-        reject(new Error('Connection timed out.'));
-      };
 
-      this.socket!.once('connect', onConnect);
+        if (!retry) {
+          this.close();
```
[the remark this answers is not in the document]
moshababo: `this.close()` is not just for closing the connection, but to close the peer object entirely. 
It was indeed previously called from the `close` event which would trigger here, but I changed it to be called from there only if connection was already established. It was necessary for allowing retries.

lib/p2p/Peer.ts
moshababo: I called it `retryConnecting` in all places except in `initConnection` because there it's applied on the connection context. I can change it to have the same name there as well.

lib/p2p/Peer.ts
moshababo: See my comment above.

Before my recent changes, `this.close()` was called even if connection wasn't established yet. I changed that, and that's why i'm calling it here as well.

sangaman:
> It’s not handled at the moment. If the inbound node would suddenly become available for us for outbound as well, then a duplicated connection supposed to be detected. If it’s not available for us, it will continue to retry. I can add some checks to avoid that, hopefully it won’t become to verbose.

Right, I think currently it would get to the point of attempting a handshake and then fail, not a huge deal. It would be nice to not even attempt the outgoing connection if we're already connected, but as you said if it's too complicated or verbose then don't worry about it. Nothing obvious jumps to mind since the `Peer` object isn't aware of other peers and connections.

> The peer object will exist, but it's not accessible and not on PeerList, because peer are getting added there only after they are “opened”, while initConnection with its retries is blocking it.

Yes you are right, I'd overlooked that. Never mind.

lib/p2p/Peer.ts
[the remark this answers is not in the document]
sangaman: OK that makes sense. I think maybe the only minor consideration then would be to only emit the 'close' event from `peer.close()` if the connection had already been successfully opened.

lib/p2p/Peer.ts
sangaman: To me `retry` even in the other contexts makes enough sense, but it's fine to leave as is. @michael1011 any thoughts on the naming here?

lib/p2p/Peer.ts
sangaman: What do you mean it forced you to update the next value?

lib/p2p/Pool.ts
```diff
@@ -155,7 +155,7 @@ class Pool extends EventEmitter {
   private connectNode = async ({ addresses, nodePubKey }: NodeConnectionInfo) => {
     for (let n = 0; n < addresses.length; n += 1) {
       try {
-        await this.addOutbound(addresses[n], nodePubKey);
+        await this.addOutbound(addresses[n], nodePubKey, true);
```
sangaman: One more thought, this looks like we'd wait for the full timeout (currently a week, although I'm thinking we probably should shorten that) before we attempt the next address for a node that has multiple listening addresses. What do you think about making so we retry each address once before delaying and retrying? That would definitely require a bit of refactoring, but wanted to see what you thought.

moshababo: Sounds like a good idea, and shouldn't require serious refactoring, just to do one iteration without retries, and another one with.

@kilrau what do you say?

lib/p2p/Peer.ts
moshababo: I changed it, so you can look now. Nothing serious.

lib/p2p/Peer.ts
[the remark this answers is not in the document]
moshababo: I'd rather leave it for now, since closing is not merely about the connection closing, and this event case might have some usage (specially if we'll want to track non-connected peers).

lib/p2p/Pool.ts
moshababo: Sounds like a good idea, and shouldn't require serious refactoring, just to do one iteration without retries, and another one with.

moshababo:
> Right, I think currently it would get to the point of attempting a handshake and then fail, not a huge deal. It would be nice to not even attempt the outgoing connection if we're already connected, but as you said if it's too complicated or verbose then don't worry about it. Nothing obvious jumps to mind since the Peer object isn't aware of other peers and connections.

This indeed breaks the encapsulation, although technically can be implemented simply with `Pool` passing `Peer` a callback. But I didn't want to do it now, since you already raised some similar issues on gitter, which might require a different solution.  

Other topics:

1) don't retry connecting with a node which we never had a successful connection with
2) when iterating on the node addresses, don't retry connecting until all addresses tried out. If all failed, retry connecting on the first address
3) soft-delete nodes which exceeded the max retry period by deleting their addresses list (although it might turn out to be empty also because the nodes failed to retrieve their external ip address)

@kilrau let's all agree on that before i'll go ahead and implement

lib/p2p/Pool.ts
sangaman: Ok that should work, but then we'd basically only be retrying a single address from that point on right?

sangaman:
I'd agree with all 3. Retrying connections on all their addresses would be nice but not a big deal.

moshababo:
We can do retries on multiple addresses in parallel. Verifying that the pubkey wasn't already connected before each retry shouldn't be a big deal.

lib/p2p/Pool.ts
moshababo: Yes. I continued to discuss the option of doing the retries in parallel on the main comments thread.

sangaman:
I'm actually thinking that maybe we want to keep the connection attempts attempted in a serial fashion though so that we don't have to deal with multiple attempts succeeding on the first attempt thereby creating multiple conncections to the same node, I brought up that issue in #398.

@moshababo Are you planning to make any more changes to this PR or should I review it again as is?

moshababo:
I don't see a problem with it, if we'll manage the pending peers correctly. But we can indeed avoid this complexity for now. 

There's nothing major to review yet. I was waiting for more feedback on my [comment](https://github.com/ExchangeUnion/xud/pull/392#issuecomment-416025884). I'll go ahead and implement it anyway tomorrow morning.

sangaman:
OK, here's my thoughts after reading/thinking some more.

> don't retry connecting with a node which we never had a successful connection with

Yes, and in fact I think ultimately we may want to only retry connections to the last address we've successful connected to.

> when iterating on the node addresses, don't retry connecting until all addresses tried out. If all failed, retry connecting on the first address

Yes, and I think like above maybe the "first address" in this case should be the last successful address we've connected to.

> soft-delete nodes which exceeded the max retry period by deleting their addresses list (although it might turn out to be empty also because the nodes failed to retrieve their external ip address)

Yes, and maybe we can delete particular addresses that have failed from the list.

kilrau:
> 1. don't retry connecting with a node which we never had a successful connection with

agree, once we went through the whole retry cycle and never managed to connect, we don't retry again (even after reboot). So it needs to stored permanently.

> 2. when iterating on the node addresses, don't retry connecting until all addresses tried out. If all failed, retry connecting on the first address

> I'm actually thinking that maybe we want to keep the connection attempts attempted in a serial fashion though so that we don't have to deal with multiple attempts succeeding on the first attempt thereby creating multiple conncections to the same node, I brought up that issue in #398.

Ehm not sure if I get you right, but I would we should try addresses in parallel, otherwise this procedure will take ages if a peer tells me 10 listeningAdresses...

t1: `listeningAddress1`, wait for delay x, try `listeningAddress1` wait for delay x*2
t2: `listeningAddress2`, wait for delay x, try `listeningAddress1` wait for delay x*2

we can just make peers in `pendingPeers` to be `pubKey`@`IP`:`port`, https://github.com/ExchangeUnion/xud/issues/398#issuecomment-416361302

> 3. soft-delete nodes which exceeded the max retry period by deleting their addresses list (although it might turn out to be empty also because the nodes failed to retrieve their external ip address)

Actually soft-deleting makes sense to me now, to store 1. permanently. Not sure if deleting the listening addresses is the way to do it though.

lib/p2p/Pool.ts
kilrau: Answered in the main thread. In short: yes, retry logic PER listening address (otherwise this can take a very long time)

sangaman:
> Ehm not sure if I get you right, but I would we should try addresses in parallel, otherwise this procedure will take ages if a peer tells me 10 listeningAdresses...
> 
> t1: listeningAddress1, wait for delay x, try listeningAddress1 wait for delay x*2
> t2: listeningAddress2, wait for delay x, try listeningAddress1 wait for delay x*2

I'm thinking we try each address only once with no delay and no retry, one at a time, and if they all fail, start the wait-retry routine only with the first address (or preferably last successful address).

kilrau:
... and just don't care about all other listeningAddresses? Or will we try them at some point? If yes, when?

moshababo:
If we want to designate the "first address" to be the one we had a successful connection with, it doesn't work well with deleting it upon connection retries failure.

We can consider trying retries on different addresses in a serial way to be infeasible. So either we'll do it in parallel, or try only one address.
If the later, so either we don't delete the node/address upon failure, or we first fix the "first address" problem (by adding another property to the address object, and maybe also removing the array to a different table).
If we want to do it in parallel, we must delete just the address (and not the entire node). But the parallel mode will also affect #398, so we better finalize on that one first.

I pushed some WIP stuff, basically just implementing suggestions 1 & 2, relying on the first address to be the one which we had a successful connection with (which might not be the case).

Let's talk about this in our weekly dev chat.

kilrau:
As agreed, only try on latest successful. @moshababo

sangaman:
One consideration for adding another property to the address object is that we're updating the addresses in the nodes table each time we get a HelloPacket. A peer could technically remove and add the same address from one handshake to the next, or we could connect to a peer using an address that it doesn't advertise in the handshake. The advantage of tracking the last connected address (or addresses) separately in the database would be to separate it from handshake data.

moshababo:
@sangaman I missed seeing your comment before implementing it. I encounter what you described, and solved it by not overriding existing addresses when they are being updated from the handshake state (add/remove still works). It doesn't solve the problem of xud being able to connect to a peer with an address which wasn't advertised by the peer, but I don't think it's a problem, because we're not re-using this address anyway at the moment.

We can start separating it as you suggested, and this will also give us the ability to re-use this address on future bootstrapping. But i'm not sure if that wouldn't be an overkill for now. Let me know what you think.

Other stuff:
* retrying only on bootstrap nodes, only on latest connected address
* if retrying max period exceeds, address is being removed 
* `Pool.addOutbound` changed to start throwing errors. The string-based interface wasn't a good design, and I wanted to change it before, but now I had to, mostly for `Pool.connectNodes`. This required me to refactor the p2p sanity tests. for gRPC access, the error is caught by `GrpcService`, so it doesn't crash.
* continued to apply the `tryX` naming convention, for functions which swallow errors

kilrau:
* Reconnect to last successful address of a peer first in a separate issue, so taking it out of this one https://github.com/ExchangeUnion/xud/issues/442

> Other stuff:
> 
> retrying only on bootstrap nodes, only on latest connected address
> if retrying max period exceeds, address is being removed
> Pool.addOutbound changed to start throwing errors. The string-based interface wasn't a good design, and I wanted to change it before, but now I had to, mostly for Pool.connectNode. This required me to refactor the p2p sanity tests. for gRPC access, the error is caught by GrpcService, so it doesn't crash.
> continued to apply the tryX naming convention, for functions which swallow errors

Sounds all right to me. Are these questions?

moshababo:
> "Reconnect to last successful address of a peer first" in a separate issue, so taking it out of this one #442

I assumed referred here to initial connection sequence (iterating though all addresses), not to the connection retries. 
We already had the retries working only on the last successful address. So not much was needed to support what you suggested, so I already added it here. 

The other stuff that I wrote described what I already implemented.

lib/p2p/NodeList.ts
```diff
@@ -69,13 +69,36 @@ class NodeList {
   public updateAddresses = async (nodePubKey: string, addresses: Address[] = []) => {
     const node = this.nodes.get(nodePubKey);
     if (node) {
-      node.addresses = addresses;
+      // avoid overriding the `lastConnected` field for existing matching addresses unless a new value was set
+      node.addresses = addresses.map((newAddress) => {
+        const oldAddress = node.addresses.find(address => address.host === newAddress.host && address.port === newAddress.port);
```
sangaman: Might be nice to have an equality method in `addressUtils` for this logic since we're using in a couple of places.

lib/p2p/Peer.ts
```diff
@@ -234,17 +242,22 @@ class Peer extends EventEmitter {
    * Ensure we are connected (for inbound connections) or listen for the `connect` socket event (for outbound connections)
    * and set the [[connectTime]] timestamp. If an outbound connection attempt errors or times out, throw an error.
    */
-  private initConnection = async () => {
+  private initConnection = (retry?: boolean): Promise<void> => {
```
sangaman: Making this function `async` and not returning an explicit `Promise.resolve()` below was an intentional change that @michael1011 and I made in a separate PR. I think the code is a bit cleaner, that way. If you agree, line 252 would just `return` and 255 would return a `Promise<void>`.

Another option is to keep the method signature as is, and move the `if (this.connected)` logic into the `new Promise` logic, and then `resolve()` and `return` in there.

lib/p2p/Peer.ts
```diff
@@ -271,16 +279,41 @@ class Peer extends EventEmitter {
...
+        setTimeout(() => {
+          retryDelay = Math.min(Peer.CONNECTION_RETRIES_MAX_DELAY, retryDelay * 2);
+          retries = retries + 1;
+          this.socket!.connect(address);
+          bind();
+        }, retryDelay);
+      };
 
-      this.connectTimeout = setTimeout(onTimeout, Peer.CONNECTION_TIMEOUT);
+      const bind = () => {
```
sangaman: This would be a bit easier to read if `bind` were declared before `onError` which uses the `bind` method, so just move this block of code up a bit.

lib/p2p/Pool.ts
```diff
@@ -11,6 +11,7 @@ import { OutgoingOrder, OrderIdentifier, StampedPeerOrder } from '../types/order
 import { Models } from '../db/DB';
 import Logger from '../Logger';
 import { HandshakeState, Address, NodeConnectionInfo } from '../types/p2p';
+import { NodeInstance } from '../types/db';
```
sangaman: I don't think `NodeInstance` is actually needed in this file. I pulled down the branch and removed all references and it compiled fine. Is there another reason for this, or can it just be removed?

lib/p2p/Pool.ts
```diff
@@ -166,14 +167,14 @@ class Pool extends EventEmitter {
    * @param ignoreKnown whether to ignore nodes we are already aware of, defaults to false
    * @returns a promise that will resolve when all outbound connections resolve
    */
-  private connectNodes = (nodes: NodeConnectionIterator, ignoreKnown = false) => {
+  private connectNodes = (nodes: NodeConnectionIterator, ignoreKnown?: boolean, retryConnecting?: boolean) => {
```
sangaman: I know it functions the same for purposes of false checks, but this technically contradicts the comments above that says `ignoreKnown` defaults to false. Personally I think it's a bit more clear too to default boolean parameters to `false` rather than `undefined`, unless there's a reason to do it this way that I'm not aware of.

lib/p2p/Pool.ts
```diff
@@ -182,14 +183,27 @@ class Pool extends EventEmitter {
...
+    for (const address of sortedAddresses) {
       try {
-        await this.addOutbound(addresses[n], nodePubKey);
-        break; // once we've successfully established an outbound connection, stop attempting new connections
-      } catch (err) {
-        this.logger.info(err);
-      }
+        await this.addOutbound(address, nodePubKey, false);
+        return; // once we've successfully established an outbound connection, stop attempting new connections
+      } catch (err) {}
```
sangaman: Should we at least log a `trace` level message here in the catch block?

lib/p2p/Pool.ts
```diff
@@ -182,14 +183,27 @@ class Pool extends EventEmitter {
...
+    }
+
+    if (retryConnecting && sortedAddresses.length && sortedAddresses[0].lastConnected) {
+      try {
+        await this.addOutbound(sortedAddresses[0], nodePubKey, true);
+      } catch (err) {}
```
sangaman: Same here for logging.

lib/p2p/Pool.ts
```diff
@@ -331,7 +355,7 @@ class Pool extends EventEmitter {
       }
       case PacketType.NODES: {
         const nodes = (packet as packets.NodesPacket).body!;
-        await this.connectNodes(nodes);
+        await this.connectNodes(nodes, false, false);
```
sangaman: I'd prefer this line without explicitly setting the boolean parameters to false since that's their default, it's just less verbose.

lib/p2p/Pool.ts
```diff
@@ -371,14 +395,27 @@ class Pool extends EventEmitter {
       peer.sendPacket(new packets.GetOrdersPacket());
       peer.sendPacket(new packets.GetNodesPacket());
 
+      // updating the `lastConnected` field for the address we're actually connected to.
+      // if connection is inbound, then the port is discarded. This might result in multiple addresses getting updated
+      const addresses = peer.addresses!.map((address) => {
+        if (peer.socketAddress.host === address.host &&
+            peer.inbound ? true : peer.socketAddress.port === address.port
```
sangaman: I don't think we should be updating `lastConnected` on inbound connections, as you point out it might update multiple addresses, but mainly the inbound connection doesn't actually verify that we're able to connect to their external address, which is what I'm thinking `lastConnected` should track

lib/p2p/errors.ts
```diff
@@ -29,6 +32,10 @@ const errors = {
     message: `could not retrieve external IP: ${err.message}`,
     code: errorCodes.EXTERNAL_IP_UNRETRIEVABLE,
   }),
+  CONNECTING_RETRIES_MAX_PERIOD_EXCEEDED: {
+    message: `Connection retries max period exceeded`,
```
sangaman: Maybe make this a bit more descriptive, like "Connection retry attempts to peer exceeded maximum time allotment"

test/p2p/sanity.spec.ts
```diff
@@ -67,47 +67,53 @@ describe('P2P Sanity Tests', () => {
...
-    expect(result).to.be.equal('Not connected');
+    expect(nodeOne.service.connect({ nodeUri: getUri({ nodePubKey: 'notarealnodepubkey', host: 'localhost', port: 9003 }) }))
+      .to.be.rejected;
   });
 
   after(async () => {
     await Promise.all([nodeOne['db'].models.Node.truncate(), nodeTwo['db'].models.Node.truncate()]);
     await Promise.all([nodeOne['shutdown'](), nodeTwo['shutdown']()]);
   });
+
```
sangaman: Can drop this extra blank line.

lib/grpc/GrpcService.ts
```diff
@@ -119,9 +119,9 @@ class GrpcService {
    */
   public disconnect: grpc.handleUnaryCall<xudrpc.DisconnectRequest, xudrpc.DisconnectResponse> = async (call, callback) => {
     try {
-      const disconnectResponse = await this.service.disconnect(call.request.toObject());
+      await this.service.disconnect(call.request.toObject());
       const response = new xudrpc.DisconnectResponse();
-      response.setResult(disconnectResponse);
+      response.setResult('success');
```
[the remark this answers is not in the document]
moshababo: `Service.disconnect` to throws on failure, so there's actually no need for the response. I left it there to avoid changes in the proto file, which will be changed anyway.

lib/p2p/Peer.ts
moshababo: I overrode this method signature in I resolved conflicts. Will return it to how it was on master.

moshababo: I overrode this method signature when I resolved conflicts. Will return it to how it was on master.

lib/p2p/Pool.ts
moshababo: I used it to make it clear that `connectNodes` can also get a db node entry. I ended up not doing anything with it, but still left this definition. I'll remove it.

lib/p2p/Pool.ts
moshababo: It's already being logged inside `addOutbound` or from more inner methods, which I think is the better place to do it. If we'll add logging here, it will be a duplication. 

In a case of `NODE_ALREADY_CONNECTED` it doesn't log, but I don't think it's necessary - it supposed to happen only due to CLI/gRPC call, so the thrown error will be logged.

lib/p2p/Pool.ts
moshababo: Makes sense, i'll change it

sangaman:
Looks like it just needs to resolve a small conflict and be squashed, then this is ready for merge. Great work.

sangaman:
To get this in for our release I squashed and resolved the conflict myself, merging.

PR #393, opened 2018-08-23
Share only connected peers on GET_NODES

### LINKED ISSUE #386: Share peers

sangaman:
Closes #386.

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -44,7 +44,9 @@ class Pool extends EventEmitter {
   /** The local handshake data to be sent to newly connected peers. */
   public handshakeData!: HandshakeState;
   public swapDeals = new SwapDeals();
+  /** A collection of known nodes on the XU network. */
   private nodes: NodeList;
+  /** A collection of opened, active peers. */
   private peers: PeerList = new PeerList();
   private server?: Server;
   private connected = false;
@@ -285,7 +287,7 @@ class Pool extends EventEmitter {
         break;
       }
       case PacketType.GET_NODES: {
-        peer.sendNodes(this.nodes.toConnectionInfoArray(), packet.header.id);
+        this.handleGetNodes(peer, packet.header.id);
         break;
       }
       case PacketType.NODES: {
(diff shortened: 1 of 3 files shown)


lib/p2p/Pool.ts
```diff
@@ -312,6 +312,18 @@ class Pool extends EventEmitter {
     }
   }
 
+  private handleGetNodes = (peer: Peer, reqId: string) => {
+    const nodesInfo = this.nodes.toConnectionInfoArray();
+    const connectedNodesInfo: NodeConnectionInfo[] = [];
+    nodesInfo.forEach((nodeInfo) => {
```
michael1011: `toConnectionInfoArray` is already iterating over all nodes. This way we iterate over the nodes twice. Is it feasible to check if the node isn't banned, has addresses and is connected in one step?

moshababo:
At the moment you're adding a peer to `PeerList` only after it's "opened", so why not to just use this list?

sangaman:
> At the moment you're adding a peer to PeerList only after it's "opened", so why not to just use this list?

You're right, I think just forgot that this was the behavior (I think at some point peers were added to the list before being opened), so no need to make the explicit `opened` check.

lib/p2p/Pool.ts
sangaman: Thanks, I think I took care of this with the new approach.

sangaman:
Ok, I basically started over from scratch with the knowledge that `Pool.peers` represents only opened peers. I removed the `toConnectionInfoArray` method from `NodesList` because I don't think we'll be using it any longer.

Take another look when you get a chance, thanks.

sangaman:
Right, I'll check to make sure they have a listening address. What do you mean that we are not connected by localhost? Is there a reason we wouldn't want to share a peer running on the same machine?

kilrau:
I think there is: if we are connected to another xud on the same machine (via localhost/127.0.01) we essentially didn't verify its EXTERNAL listening address is working. So we shouldn't share it other xuds. And I agree on that.

Since it's a bit of an edge case, i wouldn't do the effort and check the external listening address.

sangaman:
But we're sharing listening addresses given to us by the peer, not the address we used to connect to them. It won't be their 127.0.01 address (unless they configured their node as such, which is their problem).

offerm:
@sangaman  Say A and B connected over localhost. A configured his address to be `127.0.0.1` and connected to B.  B in his database has `127.0.0.1` as the address of A.
Now C connects to B. C gets from B the nodes list including A with `127.0.0.1`. Clearly, he will not be able to connect to A using this address since they [most probably] not on the same machine.  what you get instead are connection errors.

Note that with the current implementation the only way for a node to listen but to prevent being "automatically connected" is by providing a localhost or wrong address. 

IMHO, in addition to the changes in this PR we should also make the "IP auto detect" an optional feature which is disabled by default.

sangaman:
My thinking is that if a node configures a listening address for themselves that's not reachable from the global internet, that is their issue. If someone wants to run a simulation with multiple nodes running on localhost that share addresses with each other (and that would most likely be in one of our own tests), we shouldn't stop them. Or maybe someone only wants to expose an address that's available on their local subnet, that's fine too.

As for changing the behavior of the IP auto detect feature, I thought the main point of that was to provide a default listening address when the user didn't configure a listening address for themselves. I don't think that's really related to this PR tho so if you want we can discuss more in an issue or on gitter.

kilrau:
Adjusted scope of https://github.com/ExchangeUnion/xud/issues/386, so that this PR can close it.

> But we're sharing listening addresses given to us by the peer, not the address we used to connect to them. It won't be their 127.0.01 address (unless they configured their node as such, which is their problem).

Sharing just ANY listening address a peer gives us, is a problem since this opens up quite an attack vector. Also, we should take into account the very valuable information of **how (using which pubkey@IP:port** we successfully connected to a peer.

Aware, that this takes this PR a step further and I don't want to block the merging this, so I put it separate: https://github.com/ExchangeUnion/xud/issues/403

PR #394, opened 2018-08-24
Add sendPayment method

sangaman:
This adds a wrapper method for the lnd `sendPaymentSync` rpc call which allows us to specify r_hash, amount, and destination as is necessary for lnd swap execution.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -206,6 +206,32 @@ class LndClient extends BaseClient {
   }
 
   /**
+   * Sends a payment over the Lightning Network.
+   * @param r_hash the payment hash to use for the payment
+   * @param amount the number of satoshis (or equivalent) to send
+   * @param destination the pub key of the recipient
+   * @param finalCltvDelta the CLTV delta from the current height for the time lock of the final hop (default: 9)
+   */
+  public sendPayment = (r_hash: Buffer | string, amount: number, destination: Buffer | string, finalCltvDelta = 9):
+  Promise<lndrpc.SendResponse.AsObject> => {
+    const request = new lndrpc.SendRequest();
+    if (typeof r_hash === 'string') {
+      request.setPaymentHashString(r_hash);
+    } else {
+      request.setPaymentHash(r_hash);
+    }
+    request.setAmt(amount);
+    if (typeof destination === 'string') {
+      request.setDestString(destination);
+    } else {
+      request.setDest(destination);
+    }
+    request.setFinalCltvDelta(finalCltvDelta);
+
+    return this.unaryCall<lndrpc.SendRequest, lndrpc.SendResponse.AsObject>('sendPaymentSync', request);
+  }
+
+  /**
    * Get a new address for the internal lnd wallet.
    */
   public newAddress = (addressType: lndrpc.NewAddressRequest.AddressType): Promise<lndrpc.NewAddressResponse.AsObject> => {


sangaman:
A sample for how calling this would look:

```typescript
try {
  const paymentResult = await this.lndbtcClient.sendPayment('', 1, '', 1);
  paymentResult.paymentError;
  paymentResult.paymentPreimage;
  paymentResult.paymentRoute;
} catch (err) {
  this.logger.error(err);
}
```

lib/lndclient/LndClient.ts
```diff
@@ -205,6 +205,32 @@ class LndClient extends BaseClient {
...
+   */
+  public sendPayment = (r_hash: Buffer | string, amount: number, destination: Buffer | string, finalCltvDelta = 9):
+  Promise<lndrpc.SendResponse.AsObject> => {
+    const request = new lndrpc.SendRequest();
+    if (typeof r_hash === 'string') {
+      request.setPaymentHashString(r_hash);
+    } else {
+      request.setPaymentHash(r_hash);
+    }
+    request.setAmt(amount);
```
offerm: Is there really a good reason to replicate function signature and the function handling form LND?

offerm:
So, I already implemented this so no need for this PR.

Still the way we implemented is different and highlights what we already discussed with regards to the gRPC calls.

Lets see:

Your way of calling:
```
try {
  const paymentResult = await this.lndbtcClient.sendPayment(...);
  paymentResult.paymentError;
  paymentResult.paymentPreimage;
  paymentResult.paymentRoute;
} catch (err) {
  this.logger.error(err);
}
```
This looks simple but you missed the parameters for the function (...).   Replacing the ... with real value is not easy since you don't have any guidance from the IDE and you need to hope that you did a proper job. 

My version on the other hand:
```
    const request = new lndrpc.SendRequest();
    request.setAmt(deal.takerAmount);
    request.setDestString(deal.takerPubKey);
    request.setPaymentHash(String(deal.r_hash));

    try {
      const response = await cmdLnd.sendPaymentSync(request);
      paymentResult.paymentError;
      paymentResult.paymentPreimage;
      paymentResult.paymentRoute;
    } catch (err) {
      this.logger.error('Got exception from sendPaymentSync: ' + ' ' + JSON.stringify(request.toObject()));
    }
  }

```

Similar but different. 

On the implementation side:

your code:
 ```
/**
   * Sends a payment over the Lightning Network.
   * @param r_hash the payment hash to use for the payment
   * @param amount the number of satoshis (or equivalent) to send
   * @param destination the pub key of the recipient
   * @param finalCltvDelta the CLTV delta from the current height for the time lock of the final hop (default: 9)
   */
  public sendPayment = (r_hash: Buffer | string, amount: number, destination: Buffer | string, finalCltvDelta = 9):
  Promise<lndrpc.SendResponse.AsObject> => {
    const request = new lndrpc.SendRequest();
    if (typeof r_hash === 'string') {
      request.setPaymentHashString(r_hash);
    } else {
      request.setPaymentHash(r_hash);
    }
    request.setAmt(amount);
    if (typeof destination === 'string') {
      request.setDestString(destination);
    } else {
      request.setDest(destination);
    }
    request.setFinalCltvDelta(finalCltvDelta);
     return this.unaryCall<lndrpc.SendRequest, lndrpc.SendResponse.AsObject>('sendPaymentSync', request);
  }

```

While my code:

```
  /**
   * Send a payment through the Lightning Network.
   * @param lndrpc.SendRequest (see lndrpc.proto)
   */
  public sendPaymentSync = (request: lndrpc.SendRequest): Promise<lndrpc.SendResponse.AsObject> => {
    return this.unaryCall<lndrpc.SendRequest, lndrpc.SendResponse.AsObject>('sendPaymentSync', request);
  }

```

Here we see a big difference. There is simply no need to redefine this method and have all this code. 

And what is the code of it? Another `import` ?

Note that I'm still not happy with `unaryCall` return an object.

lib/lndclient/LndClient.ts
sangaman: What do you mean?

sangaman:
Closing as this is no longer needed.

Issue #297, opened 2018-08-07
Solve docker issues


kilrau:
1. "Segmentation fault"
2. "New release -> trigger travis build -> push to dockerhub -> deploy onto google kubernetics

itsbalamurali:
@kilrau 
Apologies for the delays:
1. Segmentation Faults are not occurring anymore, i am assuming that this is due to latest code changes to `xud`.
2. Docker builds are already automated as you can see here: https://hub.docker.com/r/exchangeunion/xud/builds/
3. Once the docker-ltcd PR is merged and our lnd swap resolvers are setup i will be automating depolyments onto our GCP GKE AKA Kubernetes.

PR #396, opened 2018-08-25
Docker set grpc listen host

itsbalamurali:
This PR adds grpc host to `xud.conf` and sets it to `0.0.0.0` so that grpc can listen on address and startup on the given port. 

And  apparently the `segmentation fault` issue part of #297 is solved, which caused `xud` process in the container to quit with the latest releases of code.


itsbalamurali:
@kilrau docker-compose can be tested now.

sangaman:
Do we want this behavior by default? Listening to `0.0.0.0` for grpc calls means anyone from the internet can control the xud instance, especially until we implement authorization on rpc calls.

kilrau:
Yeah we had this discussion before with @sangaman @itsbalamurali , we can't allow this and I closed the issue. I understand your intention is to allow an exchange to communicate with xud while it sits on a different subsystem with a different IP. How about allowing 

127.0.0.1
10.0.0.0/8
172.16.0.0/12
192.168.0.0/16
::1/128
fc00::/7

only instead?

itsbalamurali:
Apologies for delay, @sangaman & @kilrau i have noticed that `localhost` in the hostname field to grpc is not working for some reason. below is the log.

image

And as @kilrau suggested we can go with `127.0.0.1` since localhost an dns alias of the same.

sangaman:
It looks like grpc is listening successfully in that log you posted? But `127.0.0.1` is perfectly fine if it makes things easier.

itsbalamurali:
@sangaman please refer to this piece of screenshot.

image

kilrau:
So it looks we are back to 127.0.0.1 only? How about https://github.com/ExchangeUnion/xud/pull/396#issuecomment-416309968 ? @itsbalamurali @sangaman

itsbalamurali:
@kilrau since we are not sure about the network interface (not the loopback interfaces) that'll be assigned we cannot pre
 assign to them since we are looking at access via xucli & @sangaman this is exactly what i am face https://github.com/grpc/grpc/issues/9789.

sangaman:
I'm not sure how to proceed, should we just make setting up the rpc listening host in the config file part of the docker instructions? Maybe we can discuss tomorrow.

kilrau:
agreement:
- set to 0.0.0.0
- add to docker documentation how to set my own IP

itsbalamurali:
@kilrau can we review this and merge it?

package-lock.json
```diff
@@ -3640,17 +3640,17 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": false,
+          "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
```
sangaman: Do you know why there are all these changes to the package-lock.json file? I don't think this is related to any of the docker changes, we should probably just discard these changes.

docker/xud/xud.conf
```diff
@@ -4,3 +4,8 @@ password = "xud"
 database = "xud"
 port = 3306
 host = "db"
+[rpc]
+host = "0.0.0.0"
+port = 8886
+[raiden]
+disable = true
```
sangaman: New line EOF.

package-lock.json
```diff
@@ -8806,4 +8806,4 @@
       "dev": true
     }
   }
-}
+}
```
sangaman: Looks like all changes were removed but still have this removing the new line from the end of file.

itsbalamurali:
@kilrau

Issue #340, opened 2018-08-14
Connect: Add port 443 (& 80) fallback


kilrau:
Make `port` specification in `./xucli connect host port` *optional*
- if port is specified in connect command only try this specific port
- if no port is specified, try 8885, then 443, then 80.

Prerequisite/Related: https://github.com/ExchangeUnion/xud/issues/182

kilrau:
Please let me know what you think @moshababo @sangaman @offerm @michael1011

sangaman:
I was thinking the fallback would be for the listening, not for connecting. For connecting we should probably be using confirmed listening addresses from the peer we're connecting to, not trying to guess what port they're using.

kilrau:
This is because I'd want to make the port in the `connect` command *optional*.

Reason again UX: I never specified a port when I used teamviewer to connect to another machine. Just a port.

>TeamViewer connects on port 5938, but also tunnels via ports 80 (HTTP) & 443 (SSL) if that is unavailable.

sangaman:
Ok that should be fine.

kilrau:
Thanks for all your opinions!

I have very strong feelings about this and wouldn't have suggested it if I hadn't spent a long time in a place with very restricted things you can do outside of port 80 & 443. So e.g. I was NOT able to run a reachable bitcoin full node there and had to tunnel my way out to be reachable since you simply a) can't open anything else on ISP (forced) routers and b) even if you can, traffic on non-80/443 ports is heavily disturbed or even completely blocked. In conclusion I don't think we can *always* rely on the role model `lnd` and `btcd`/`bitcoind` to be right and say "it's not common, so we shouldn't do it".

I hope we all agree that we want to make
- `xud` as easy to run and
- as censorship-resistant as possible

for me following the https://community.teamviewer.com/t5/Knowledge-Base/Which-ports-are-used-by-TeamViewer/ta-p/4139#toc-hId-649815939 logic is a way into this direction but definitely is only part of it. It goes far beyond with e.g. making packets on 443 indistinguishable from regular SSL packets...

So yes @offerm you are right, there are some more cases to catch like if there are already services running on 443/80 or xud was not run as root on linux, behind a NAT it still would need forwarding, but I couldn't find any convincing reason why it's entirely a bad idea to do this.

[...]

PR #400, opened 2018-08-27
Verify reachability of advertised addresses

### LINKED ISSUE #182: Verify that my own xud is reachable
If listening to P2P connection, verify that `listenPort` is reachable by the node's external IP. 
If it's not alert that the node is not reachable. 

#181 is a prerequisite.

michael1011:
Closes #182 

This PR takes care of verifying that the node's advertised addresses are reachable. If that is not the case the user is alerted with the message: `Could not connect to advertised address: <host>:<port>`

### DIFF
--- a/lib/Logger.ts
+++ b/lib/Logger.ts
@@ -57,17 +57,22 @@ class Logger {
   private level: string;
   private logDir: string;
   private context: Context;
-  private logger: any;
+  private logger?: any;
   private instanceId: number;
 
   private static defaultLogDir = 'logs';
 
-  constructor({ level, instanceId, logDir, context }: {level: string, instanceId?: number, logDir?: string, context: Context}) {
-    this.level = level;
+  constructor({ instanceId, level, logDir, context, dummy }:
+    {instanceId?: number, level: string, logDir?: string, context: Context, dummy?: boolean}) {
 
+    this.level = level;
     this.logDir = logDir || Logger.defaultLogDir;
     this.context = context || Context.GLOBAL;
     this.instanceId = instanceId || 0;
+
+    if (dummy) {
+      return;
+    }
 
     if (!fs.existsSync(this.logDir)) {
       fs.mkdirSync(this.logDir);
@@ -101,6 +106,10 @@ class Logger {
     };
   }
 
+  public static get dummyLogger(): Logger {
+    return new Logger({ level: Level.ERROR, context: Context.GLOBAL, dummy: true });
+  }
+
   private getLogFormat = (colorize: boolean) => {
     const { format } = winston;
 
@@ -127,7 +136,9 @@ class Logger {
   }
 
   private log = (level: string, msg: string) => {
-    this.logger.log(level, msg);
+    if (this.logger) {
+      this.logger.log(level, msg);
+    }
   }
 
   public error = (msg: Error | string, err?: Error) => {
--- a/lib/Logger.ts
+++ b/lib/Logger.ts
@@ -62,15 +62,15 @@ class Logger {
 
   private static defaultLogDir = 'logs';
 
-  constructor({ instanceId, level, logDir, context, dummy }:
-    {instanceId?: number, level: string, logDir?: string, context: Context, dummy?: boolean}) {
+  constructor({ instanceId, level, logDir, context, disabled }:
+    {instanceId?: number, level?: string, logDir?: string, context?: Context, disabled?: boolean}) {
 
-    this.level = level;
+    this.level = level || Level.TRACE;
     this.logDir = logDir || Logger.defaultLogDir;
     this.context = context || Context.GLOBAL;
     this.instanceId = instanceId || 0;
 
-    if (dummy) {
+    if (disabled) {
       return;
     }
 
@@ -106,8 +106,8 @@ class Logger {
     };
   }
 
-  public static get dummyLogger(): Logger {
-    return new Logger({ level: Level.ERROR, context: Context.GLOBAL, dummy: true });
+  public static get disabledLogger(): Logger {
+    return new Logger({ disabled: true });
   }
 
   private getLogFormat = (colorize: boolean) => {
--- a/lib/Logger.ts
+++ b/lib/Logger.ts
@@ -54,6 +54,8 @@ type Loggers = {
 };
 
 class Logger {
+  public static disabledLogger = new Logger({ disabled: true });
+
   private level: string;
   private logDir: string;
   private context: Context;
@@ -104,10 +106,6 @@ class Logger {
       lnd: new Logger({ instanceId, level, context: Context.LND }),
       raiden: new Logger({ instanceId, level, context: Context.RAIDEN }),
     };
-  }
-
-  public static get disabledLogger(): Logger {
-    return new Logger({ disabled: true });
   }
 
   private getLogFormat = (colorize: boolean) => {
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -144,14 +144,14 @@ class Pool extends EventEmitter {
       const externalAddress = addressUtils.toString(address);
       this.logger.debug(`Verifying reachability of advertised address: ${externalAddress}`);
       try {
-        const peer = Peer.fromOutbound(address, Logger.dummyLogger);
+        const peer = Peer.fromOutbound(address, Logger.disabledLogger);
         await peer.open(this.handshakeData, this.handshakeData.nodePubKey);
         assert(false, errors.ATTEMPTED_CONNECTION_TO_SELF.message);
       } catch (err) {
         if (err.code === errors.ATTEMPTED_CONNECTION_TO_SELF.code) {
           this.logger.verbose(`Verified reachability of advertised address: ${externalAddress}`);
         } else {
-          this.logger.warn(`Could not connect to advertised address: ${externalAddress}`);
+          this.logger.warn(`Could not verify reachability of advertised address: ${externalAddress}`);
         }
       }
     });
@@ -558,6 +558,8 @@ class Pool extends EventEmitter {
     });
 
     peer.on('error', (err) => {
+      // The only situation in which the node should be connected to itself is the
+      // reachability check of the advertised addresses and we don't have to log that
       if (peer.nodePubKey !== this.handshakeData.nodePubKey) {
         this.logger.error(`peer error (${peer.nodePubKey}): ${err.message}`);
       }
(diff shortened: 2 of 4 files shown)


sangaman:
I'd lean towards just creating a plain socket connection and then listening for the `connection` or `error` events, rather than attempting to create a new `Peer`. There's less overhead that way and also less risk that we get some unexpected behavior, I think. This could conflict with connection retry logic, for one thing (currently we're not retrying if we've never connected before, but that could ch), and we wouldn't want to add ourselves to the `nodes` table on a successful connection.

michael1011:
I was thinking that even if the socket and port is open there could be another service listening on that address. That gets even more likely if we want to do #340. Right now it is working as it should (never adding yourself to the nodes table, just checking if the connection to the node works) but I get your point.

I would like to hear some more opinions on this especially from @kilrau because he insisted on #340. @offerm @moshababo @sangaman

sangaman:
Yeah I see what you mean. Shouldn't `await peer.open(this.handshakeData, this.handshakeData.nodePubKey);` fail due to attempting to connect to ourselves though?

michael1011:
It does fail but at a later stage.
```
    peer.on('error', (err) => {
      this.logger.error(`peer error (${peer.nodePubKey}): ${err.message}`);
      if (peer.nodePubKey !== this.handshakeData.nodePubKey) {
        this.logger.error(`peer error (${peer.nodePubKey}): ${err.message}`);
      }
```
I should add an additional check to make sure it fails to the correct reason: the same pubkey.

sangaman:
Ok, yeah it'd be good to have that check and it's also good that the connection with ourselves is never truly opened which I think limits the amount of things that could go wrong. I'd be OK with keeping it this way.

lib/p2p/Pool.ts
```diff
@@ -135,6 +137,25 @@ class Pool extends EventEmitter {
     this.connected = false;
   }
 
+  private verifyReachability = () => {
+    this.handshakeData.addresses!.forEach(async (address) => {
+      const externalAddress = `${address.host}:${address.port}`;
```
sangaman: There's an `addressUtils.toString()` method that does this. It does the same thing, but we might as well use that here.

lib/p2p/Pool.ts
```diff
@@ -135,6 +137,25 @@ class Pool extends EventEmitter {
...
+      this.logger.debug(`Trying to connect to advertised address: ${externalAddress}`);
+      try {
+        const peer = Peer.fromOutbound(address, this.logger);
+        await peer.open(this.handshakeData, this.handshakeData.nodePubKey);
+        peer.close();
+      } catch (err) {
+        if (err === errors.ATTEMPTED_CONNECTION_TO_SELF) {
+          this.logger.verbose(`Verfied advertised address: ${externalAddress}`);
```
sangaman: Typo in "Verified".

lib/p2p/Pool.ts
```diff
@@ -135,6 +137,25 @@ class Pool extends EventEmitter {
...
 
+  private verifyReachability = () => {
+    this.handshakeData.addresses!.forEach(async (address) => {
+      const externalAddress = `${address.host}:${address.port}`;
+      this.logger.debug(`Trying to connect to advertised address: ${externalAddress}`);
+      try {
+        const peer = Peer.fromOutbound(address, this.logger);
+        await peer.open(this.handshakeData, this.handshakeData.nodePubKey);
+        peer.close();
+      } catch (err) {
```
sangaman: One more thing actually. In rereading this code flow, I think we should try to avoid try/catch as part of our *expected* flow and reserve `catch` statements for things that are truly exceptions. In this case since the exception is what we expect, I think the intent of this code would be more clear if we used [assert.rejects()](https://nodejs.org/api/assert.html#assert_assert_rejects_block_error_message). So instead of the try/catch, try something like

```javascript
assert.rejects(peer.open(this.handshakeData, this.handshakeData.nodePubKey), errors.ATTEMPTED_CONNECTION_TO_SELF);
```

I think we can drop the peer.close() statement at the end since that should be handled inside the `open()` method.

michael1011: I cant' get your example working:
```
{ message: 'Cannot attempt connection to self', code: '2.4' }

/home/michael/Git/JavaScript/xud/dist/bootstrap.js:13
        throw err;
        ^
[object Object]
```
and the this error isn't catchable. The `assert.rejects()` doc does exists for Node.js [v10.9.0](https://nodejs.org/api/assert.html#assert_assert_rejects_block_error_message) but not for [v8.11.4](https://nodejs.org/docs/latest-v8.x/api/assert.html). I don't think it is available in the version we should be using but correct me if I am wrong. And are you sure it a good idea to avoid using `catch` by any means?

sangaman: Oh if it's not available for node 8 then maybe this is more trouble than it's worth, as we wouldn't want to lose support for it over this. I think the issue you're seeing though is because it needs an `await` before the `assert.rejects`. Maybe one day if there are other reasons to drop support for node 8 we could revisit this, but for now it's not a big deal.

michael1011: I have had the exact issue with `await`. :thinking: The weird thing is that I am on Node.js v8.11.4 and the method exists (if I install `@types/node`) but it behaves in a weird way.

sangaman: Yeah don't worry about it then, we'll leave it as is.

offerm:
@sangaman @michael1011  can you please explain what we try to achieve in this PR and how we do it? What assumptions do we take when doing it?

lib/p2p/Pool.ts
```diff
@@ -508,7 +527,9 @@ class Pool extends EventEmitter {
     });
 
     peer.on('error', (err) => {
-      this.logger.error(`peer error (${peer.nodePubKey}): ${err.message}`);
+      if (peer.nodePubKey !== this.handshakeData.nodePubKey) {
```
moshababo: I don't see why `error` event handler will ever be called in case of the reachability check, since `bindPeer` is isn't called.

If I missed something, and it's indeed being called, then I think that this condition (`peer.nodePubKey !== this.handshakeData.nodePubKey`) won't be very clear on the long-run.

We need to mark the peer object somehow. We can add a boolean flag to it, and perhaps another static factory method for `Peer` that will activate it. If it's active, `bindPeer` won't bind. This way no event will ever be triggered, and code will be cleaner. 

I'm also seeing logs from `Peer` (`Socket closed prior to handshake`, `Received data`), but since the latest logger refactoring, these can be easily be eliminated by injecting a dummy logger to `Peer` (we can create it on `Logger` as a static instance).

offerm:
Ok. And what is the implication of not being able to connect to an address?

lib/p2p/Pool.ts
```diff
@@ -135,6 +137,23 @@ class Pool extends EventEmitter {
     this.connected = false;
   }
 
+  private verifyReachability = () => {
+    this.handshakeData.addresses!.forEach(async (address) => {
+      const externalAddress = addressUtils.toString(address);
+      this.logger.debug(`Trying to connect to advertised address: ${externalAddress}`);
+      try {
+        const peer = Peer.fromOutbound(address, this.logger);
+        await peer.open(this.handshakeData, this.handshakeData.nodePubKey);
```
moshababo: perhaps mark somehow that this line should never pass without throwing exception. maybe add `assert(false)` after it.

michael1011:
If we don't get a `Hello` packet with the same public key as the node verifying its addresses I think it is reasonable to say that the address is either not yours or unreachable for others.

moshababo:
@offerm there are no implications, except getting an alert message.

lib/p2p/Pool.ts
sangaman: I do like having the check on inbound connections to make sure they're not using our pubkey though. It should never happen, but it doesn't hurt to catch it here either.

moshababo:
Did you pass it with `localhost:8885`?

moshababo:
We need to check it with an actual external IP address.

michael1011:
@moshababo I agree but I don't have access to our remote nodes. Could someone test this on behalf of me?

lib/p2p/Pool.ts
michael1011: The inbound connection is calling `bindPeer` I think. And I added the dummy logger but I am not sure if I implemented it the way you meant it. Please take a look. 

These messages are leftover and sent by the inbound connection I think:
```
2018-8-28 15:39:05 [P2P] debug: Connected pre-handshake to peer 127.0.0.1:60270
2018-8-28 15:39:05 [P2P] verbose: Verified reachability of advertised address: localhost:8885
2018-8-28 15:39:05 [P2P] debug: Received data (127.0.0.1:60270): {"header":{"id":"bb36e170-aac7-11e8-b573-fd57249abaf3","hash":"zh51r1Ey5gL3mUTOM1417A==","type":"HELLO"},"body":{"version":"1.0.0-prealpha.1","pairs":["BTC/LTC","ZRX/GNT"],"nodePubKey":"0284643297670606e6f381b8dff161254a8ea1db927eed6c2d33c2015ab4b4dbd6","addresses":[{"host":"localhost","port":8885}]}}ↂ₿ↂ
2018-8-28 15:39:05 [P2P] info: Peer 0284643297670606e6f381b8dff161254a8ea1db927eed6c2d33c2015ab4b4dbd6 socket closed
```

michael1011:
@moshababo I just rebased this branch and resolved the conflicts. Please take a look and tell me what you think.

michael1011:
I was able to test this branch on a VM and it worked fine:
```
2018-8-29 13:45:40 [P2P] debug: Verifying reachability of advertised address: 40.121.15.58:8885
2018-8-29 13:45:40 [RPC] info: gRPC server listening on localhost:8886
2018-8-29 13:45:40 [P2P] debug: Connected pre-handshake to peer 40.121.15.58:32850
2018-8-29 13:45:40 [P2P] debug: Connected pre-handshake to peer 35.231.171.148:8885
2018-8-29 13:45:40 [P2P] debug: Connected pre-handshake to peer 35.196.118.79:8885
2018-8-29 13:45:40 [P2P] debug: Connected pre-handshake to peer 35.229.81.83:8885
2018-8-29 13:45:40 [P2P] verbose: Verified reachability of advertised address: 40.121.15.58:8885
```

lib/Logger.ts
```diff
@@ -101,6 +106,10 @@ class Logger {
     };
   }
 
+  public static get disabledLogger(): Logger {
+    return new Logger({ disabled: true });
```
moshababo: nothing major, but I wouldn't want a getter to return a new instance on every call. So either create it only once, and return it via a getter, or simply create it as a static field (since it doesn't require any params).

PR #408, opened 2018-08-27
Add channelbalance rpc call

sangaman:
This commit adds a `channelbalance` rpc command which acts as a passthrough for the underlying `channelbalance` lnd command. It accepts a currency ticker as a parameter to indicate which currency we are querying for.

This will be handy for demoing before and after states with swaps.

Here's sample output:

```bash
xucli channelbalance BTC
{
  "balance": 4453344,
  "pendingOpenBalance": 0
}
```

### DIFF
--- a/proto/xudrpc.proto
+++ b/proto/xudrpc.proto
@@ -110,9 +110,9 @@ message ChannelBalanceRequest {
   string currency = 1 [json_name = "currency"];
 }
 message ChannelBalanceResponse {
-  // Sum of channels balances denominated in satoshis
+  // Sum of channels balances denominated in satoshis or equivalent
   int64 balance = 1 [json_name = "balance"];
-  // Sum of channels pending balances denominated in satoshis
+  // Sum of channels pending balances denominated in satoshis or equivalent
   int64 pending_open_balance = 2 [json_name = "pending_open_balance"];
 }
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -111,8 +111,8 @@
 
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
-| balance | [int64](#int64) |  | Sum of channels balances denominated in satoshis |
-| pending_open_balance | [int64](#int64) |  | Sum of channels pending balances denominated in satoshis |
+| balance | [int64](#int64) |  | Sum of channels balances denominated in satoshis or equivalent |
+| pending_open_balance | [int64](#int64) |  | Sum of channels pending balances denominated in satoshis or equivalent |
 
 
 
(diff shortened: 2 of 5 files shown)


proto/xudrpc.proto
```diff
@@ -98,6 +105,17 @@ service Xud {
   }
 }
 
+message ChannelBalanceRequest {
+  // The ticker symbol of the currency to query for 
+  string currency = 1 [json_name = "currency"];
+}
+message ChannelBalanceResponse {
+  // Sum of channels balances denominated in satoshis
```
michael1011: What about litoshis? I think you already phrased this like "denominated in satoshis or equivalent" somewhere else.

sangaman: Yeah we could stick to "satoshis or equivalent" throughout. I was also thinking maybe "smallest unit" - but maybe for now "satoshis or equivalent" is good enough and we'll see what it looks like when more chains have lnd support.

michael1011: "smallest unit" could be confusing because of millisatoshis.

sangaman: True, maybe "smallest blockchain unit" or "smallest discrete unit" but that's getting verbose. We'll deal with it when we have support for more than just a couple of currencies on lnd. I'll make it "or equivalent" for now.

offerm:
@sangaman note that this information may be misleading if you have multiple channels.

sangaman:
You mean for demo purposes? Yes that's true if there's activity on other channels at the same time, but it won't be an issue as I won't have other channels active at the same time.

offerm:
For demo it is almost perfect. Be perfect without the need to provide the ticker (BTC or LTC). Just show the balance in both.

sangaman:
I can always do two calls in succession, I think going forward with support for multiple currencies it would be nice to specify which one you're interested in. Thoughts?

offerm:
You can make it optional. If not specified, show all.

sangaman:
Do you think that will be useful beyond the demo tho? The reason I'm kind of hesitant is because then the response has to have an array and I suppose I'd have to add a currency string to each element in the array. Or if I hardcode BTC and LTC channel balance fields then it's only a temporary solution, and I don't think it's necessary considering calling this twice for the demo is a really minor inconvenience imo. I'll see what other people think or if there's a cleaner approach than what I touched on here.

offerm:
Yes. I think it is needed regardless of the demo. 
As a user you need to be able to check how much you can potentially trade. Note that you should consider multiple channels for this.

sangaman:
Per our discussion I'll leave this as is for now and we can change it to query multiple currencies if/when the need arises.

I don't think though that channel balance is a very good indicator of how much is available to trade, because there are a lot of other factors like potential routes and the balance available on the other side of channels. You might have a ton of balance in one channel but very little routing capacity on any hops beyond that, for instance.

PR #410, opened 2018-08-27
Update instructions-for-devs.md

kilrau:
add `npm run compile` which was taken out of `npm i`

### DIFF
--- a/instructions-for-devs.md
+++ b/instructions-for-devs.md
@@ -10,14 +10,16 @@ Developers & Testers:
 ```bash
 git clone https://github.com/ExchangeUnion/xud
 cd xud
-npm install
+npm run compile
+npm i
 ```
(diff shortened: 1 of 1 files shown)


instructions-for-devs.md
```diff
@@ -10,14 +10,16 @@ Developers & Testers:
...
+npm i
 ```
 For everyone else, we recommend using the [latest release](https://github.com/ExchangeUnion/xud/releases), below an example using the `v1.0.0-prealpha.1` release tag:
 ```bash
 git clone https://github.com/ExchangeUnion/xud 
 cd xud
 git checkout tags/v1.0.0-prealpha.1 -b v1.0.0-prealpha.1
-npm install
+npm run compile
+npm i
```
sangaman: I think for both of these examples, we'd want the install to happen before the compilation. Installing gets the typescript module, which is needed for compilations.

kilrau:
fixed with https://github.com/ExchangeUnion/xud/pull/420 just to not forget about it in your overhaul @sangaman

PR #412, opened 2018-08-28
Relax HelloPacket checks

### LINKED ISSUE #448: Pre-handshake packet handling
At the moment all peer packets are being handled, regardless to whether handshake was completed or not. It means, for example, that we can potentially attempt to execute a swap with a peer before verifying his identity, his xud version compatibility, etc. 

To fix this, some logic need to be added to `Peer.handlePacket`, defining which packets are not allowed before handshake, and which are not allowed after (if any. this is probably just for disallowing multiple `Hello` packets, if we don't want to enable the overriding of the handshake state).

sangaman:
This addresses an edge case whereby we receive a `HelloPacket` before we've started waiting for one, resulting in an "[peer] sent an unsolicited Hello packet" error message. Previously, there was a small disconnect between when a socket was established and when waiting for the HelloPacket began. This changes it so that the wait for the HelloPacket begins immediately after the socket becoming connected and before any packets can be received.

EDIT: Closes #448

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -24,8 +24,7 @@ type PeerInfo = {
 interface Peer {
   on(event: 'packet', listener: (packet: Packet) => void): this;
   on(event: 'error', listener: (err: Error) => void): this;
-  on(event: 'packet', listener: (packet: Packet) => void): this;
-  on(event: 'error', listener: (err: Error) => void): this;
+  on(event: 'handshake', listener: () => void): this;
   once(event: 'open', listener: () => void): this;
   once(event: 'close', listener: () => void): this;
   once(event: 'ban', listener: () => void): this;
@@ -35,6 +34,7 @@ interface Peer {
   emit(event: 'close'): boolean;
   emit(event: 'error', err: Error): boolean;
   emit(event: 'packet', packet: Packet): boolean;
+  emit(event: 'handshake'): boolean;
 }
 
 /** Represents a remote XU peer */
@@ -155,7 +155,7 @@ class Peer extends EventEmitter {
 
     this.finalizeOpen();
 
-    // let the pool know that this peer is ready to go
+    // let listeners know that this peer is ready to go
     this.emit('open');
   }
 
@@ -277,6 +277,7 @@ class Peer extends EventEmitter {
         this.bindSocket(socket);
 
         this.logger.debug(this.getStatus());
+
         this.emit('connect');
 
         cleanup();
@@ -328,15 +329,6 @@ class Peer extends EventEmitter {
     this.stallTimer = setInterval(this.checkTimeout, Peer.STALL_INTERVAL);
   }
 
-  private initHello = async (handshakeData: HandshakeState) => {
-    const packet = this.sendHello(handshakeData);
-
-    if (!this.handshakeState) {
-      // wait for an incoming HelloPacket
-      await this.wait(PacketType.HELLO, Peer.RESPONSE_TIMEOUT);
-    }
-  }
-
   private finalizeOpen = (): void => {
     assert(!this.closed);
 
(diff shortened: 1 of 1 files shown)


lib/p2p/Peer.ts
```diff
@@ -131,7 +133,14 @@ class Peer extends EventEmitter {
 
     await this.initConnection();
     this.initStall();
-    await this.initHello(handshakeData);
+    this.sendHello(handshakeData);
+    if (this.handshakePromise) {
+      // we were already waiting for an incoming hello packet, finish waiting before opening
+      await this.handshakePromise;
+    } else {
+      // we're not already waiting for an incoming hello packet, so wait now
```
moshababo: why this condition would ever occur, if you're waiting for the packet as part of `initConnection`? this makes the previous `initHello` pretty much the same, just that you moved the `this.wait` call to the connection phase.

moshababo:
I noticed now that the problem did exists on `handleHello`, but I think we can solve this with less changes. We're listening to `HelloPacket` since the beginning (when parser is initialized, before connection is established), but `handleHello` check we explicitly called `this.wait` on it, which might not be necessary.

We actually need to solve this in a more general sense - define which packets we are rejecting in a pre-handshake state, and which ones in post-handshake. Because at the moment, for example, I can receive and handle `OrderPacket` before handshake completed.

lib/p2p/Peer.ts
sangaman: It should never occur, although I put a check there just in case. I could just put a `!` operator on `handshakePromise` or maybe still have the check but close the peer if we're not waiting for a handshake at this point for some unexpected reason.

sangaman: I have a cleaner approach almost ready which accepts hello packets at any time (potentially subject to some anti-spam limits down the road) to set or update the handshake state, I'll update the PR tomorrow.

sangaman:
OK, here's the new commit description explaining the approach:

> This allows peers to accept a `HelloPacket` and update handshake state data at any time, and not only when we are explicitly waiting for one. This addresses an edge case whereby we receive a `HelloPacket` before we've started waiting for one, resulting in a "[peer] sent an unsolicited Hello packet" error message. Previously, there was a small disconnect between when a socket was established and when waiting for the HelloPacket began. This changes it so that the wait for the HelloPacket begins immediately after the socket becoming connected and before any packets can be received. It also adds a `handshake` event that can be used going forward for Peer listeners that are interested in changes to the handshake state.

We still `wait` for a hello packet when opening a connection (if we haven't received a hello packet already), but we don't reject hello packets when we're not waiting for them, which could potentially include the time between connecting a socket and formally opening the connection for use in xud.

lib/p2p/Peer.ts
```diff
@@ -131,7 +131,12 @@ class Peer extends EventEmitter {
 
     await this.initConnection();
     this.initStall();
-    await this.initHello(handshakeData);
+    this.sendHello(handshakeData);
+
+    if (!this.handshakeState) {
+      // we must wait to receive handshake data before opening the connection
+      await this.wait(PacketType.HELLO, Peer.RESPONSE_TIMEOUT);
+    }
```
moshababo: I would still keep it inside `initHello`, to make `open` a bit cleaner. Not a must

moshababo:
I opened #448 for the other issue I mentioned.

sangaman:
Oh there must have been a misunderstanding, I was under the impression we did want to allow handshake state updates while connected. We definitely will need to add some handling for when handshake state changes, but it doesn't seem like a big deal and it seems more convenient than having to disconnect and reconnect.

Thanks for opening that other issue, is it ok if I try to handle that in this PR as well?

moshababo:
We can allow to overriding it (this is why I originally called it handshake**State**), it will just require us to remember having the proper support. For example, if I change my supported pairs, it's my responsibility drop orders from the removed pairs, and re-query orders from peers. We might want to add support for querying specific pairs. 

And of course, you can do #448 here.

sangaman:
I took care of discarding packets before a connection is opened. It allows any number of hello packets currently, but I feel restricting the number of hello packets we accept is something that should be done in a more general way since we will need to the restriction both before and after a connection is opened (we wouldn't want someone spamming us hello packets and updating handshake state after open).

sangaman:
This has passed travis, but doesn't show up here due to issues with travis: https://travis-ci.org/ExchangeUnion/xud/builds/426979073

PR #419, opened 2018-08-29
Added logPath to config

michael1011:
This PR adds `logFile` to the config. If it is set to an empty string no logs will be written to a file. 

@moshababo I know that you said that I should keep `contextFileMap` but it adds just unnecessary complexity if we are not planning to use it imo. Tell me what you think.

### DIFF
--- a/lib/Config.ts
+++ b/lib/Config.ts
@@ -13,6 +13,7 @@ class Config {
   public p2p: PoolConfig;
   public xudir: string;
   public logLevel: string;
+  public logPath: string;
   public db: DBConfig;
   public testDb: DBConfig;
   public rpc: { disable: boolean, host: string, port: number };
@@ -50,6 +51,8 @@ class Config {
     // default configuration
     this.initDb = true;
     this.logLevel = this.getDefaultLogLevel();
+    this.logPath = this.getDefaultLogPath();
+
     this.p2p = {
       listen: true,
       port: 8885, // X = 88, U = 85 in ASCII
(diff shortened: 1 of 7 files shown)


lib/Config.ts
```diff
@@ -13,6 +13,7 @@ class Config {
   public p2p: PoolConfig;
   public xudir: string;
   public logLevel: string;
+  public logFile: string;
```
sangaman: `logPath` instead perhaps? It would be consistent with `macaroonpath` and such.

lib/Config.ts
```diff
@@ -23,6 +24,8 @@ class Config {
   public instanceId = 0;
   public initDb: boolean;
 
+  private defaultLogDir = path.resolve(__dirname, '../logs');
```
sangaman: Maybe instead of `__dirname` for the default, use `xudir`?

lib/Logger.ts
```diff
@@ -55,49 +45,46 @@ type Loggers = {
...
   private level: string;
-  private logDir: string;
   private context: Context;
   private logger: any;
   private instanceId: number;
 
-  private static defaultLogDir = 'logs';
-
-  constructor({ level, instanceId, logDir, context }: {level: string, instanceId?: number, logDir?: string, context: Context}) {
+  constructor({ level, file, context, instanceId }: {level: string, file: string, context: Context, instanceId?: number}) {
```
sangaman: Let's rename `file` to `filename` so we use the same name as the winston config option.

lib/Logger.ts
```diff
@@ -55,49 +45,46 @@ type Loggers = {
...
-          level: this.level,
-          filename: path.join(this.logDir, contextFileMap[this.context]),
-          format: this.getLogFormat(false),
-        }),
-      ],
     });
   }
 
-  public static createLoggers = (level: string, instanceId = 0): Loggers => {
+  public static createLoggers = (level: string, file = '', instanceId = 0): Loggers => {
```
sangaman: Here too let's rename `file` to `filename` so we use the same name as the winston config option.

PR #426, opened 2018-08-31
Grpc error consistency

### LINKED ISSUE #324: Consistent gRPC error handling & responses
Currently there's a bit of inconsistency with how our gRPC service deals with failures and with responses that essentially amount to indicating success or failure.

For example, `Connect`, `Disconnect`, and `Shutdown` return strings. `CancelOrder` returns a boolean.

`PlaceOrder` throws an error if the specified PairId does not exist, but `CancelOrder` does not (nor does it throw an error if the specified `order_id` does not exist).

I'm thinking that for all these sorts of calls where the response essentially indicates if the call succeeded or not, we have an empty response to indicate success and throw an error to indicate failure. This seems to be the same approach that lnd takes. We could also tweak `xucli` to print "success" anytime there is an empty response, as opposed to the less descriptive `{}` that would get printed otherwise. 

I can definitely handle these changes, but I just wanted to see if anyone had any opinions before I start on it.

sangaman:
Closes #324. This redefines our protobuf definition for responses that indicate nothing more than whether the call succeeded or failed. All failures now throw a grpc error, and successes return an empty response. 

It also saves the error that causes a peer connection or initilaization to fail. This error can then be retrieved even after it has been caught/handled so it can be rethrown by the `Service` class. It creates a global types class for types that are not specific to any particular module, which contains a type definition for internal xud errors.


lib/p2p/Peer.ts
```diff
@@ -76,6 +79,7 @@ class Peer extends EventEmitter {
   }
 
   public get info(): PeerInfo {
+    this.emitError;
```
michael1011: What is this line doing?

test/p2p/sanity.spec.ts
```diff
@@ -65,43 +65,48 @@ describe('P2P Sanity Tests', () => {
...
     expect(listPeersResult.length).to.equal(0);
   });
 
   it('should fail when connecting to an unexpected node pub key', async () => {
-    const result = await nodeOne.service.connect({ nodeUri: getUri({
+    /*const result = await nodeOne.service.connect({ nodeUri: getUri({
+      nodePubKey: 'thewrongpubkey',
+      host: 'localhost',
+      port: nodeTwoConfig.p2p.port,
+    }) });*/
```
michael1011: You missed to remove a comment here.

lib/types/global.ts
```diff
@@ -0,0 +1,9 @@
+export type Address = {
+  host: string;
+  port: number;
+};
+
+export type XudError = {
```
michael1011: We should migrate the `error` object in the existing error classes to: `{ [key: string]: XudError }`

lib/p2p/Peer.ts
sangaman: Totally inadvertent, thanks for catching.

lib/types/global.ts
sangaman: I liked this idea, but as I started implementing it wound up being way more complicated given that some of our errors are not plain error objects, but rather functions that return error objects. It would have required a lot of changes and resulted in some messy code, so I'm leaving this as is.

lib/cli/command.ts
```diff
@@ -22,6 +22,11 @@ export const callback = (error: Error | null, response: grpcResponse) => {
   if (error) {
     console.error(`${error.name}: ${error.message}`);
   } else {
-    console.log(JSON.stringify(response.toObject(), undefined, 2));
+    const responseObj = response.toObject();
+    if (Object.keys(responseObj).length === 0) {
+      console.log('success');
```
michael1011: Just printing `success` is probably not the the most beautiful thing to do. It is fine for now but we will have to make the cli more beautiful and appealing eventually.

sangaman: Yup, but I figure that can be handled along with other cli user friendliness/aesthetics improvements.

sangaman:
I resolved conflicts and I realized my earlier approach of saving the error on `Peer` is not needed since #392 already made it so `addOutbound` throws an error instead of swallowing it. I simply dropped the first commit.

PR #429, opened 2018-08-31
Bind socket after connection

sangaman:
This commit delays calling the `bindSocket` method until after an outgoing connection is established. It also moves beginning an outgoing peer socket connection from the `fromOutbound` method to the `initConnection` method. This ensures that we bind the `connection` listener immediately after beginning the connection.

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -39,7 +39,7 @@ interface Peer {
 
 /** Represents a remote XU peer */
 class Peer extends EventEmitter {
-  public socketAddress!: Address;
+  // TODO: properties documentation
   public inbound!: boolean;
   public connected = false;
   private opened = false;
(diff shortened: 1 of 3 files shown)


lib/p2p/Peer.ts
```diff
@@ -428,7 +412,7 @@ class Peer extends EventEmitter {
       if (this.connected) {
```
moshababo: Assuming we're already connected, this condition can be removed as well.
The problem is that `this.close` won't be called, but this is solved anyway by the connection retries PR.

lib/p2p/Peer.ts
```diff
@@ -264,6 +268,9 @@ class Peer extends EventEmitter {
       const onConnect = () => {
         this.connectTime = Date.now();
         this.connected = true;
+
+        this.bindSocket(socket);
+
```
moshababo: Are we sure there's no chance of events being called after connection, and prior to this binding?

moshababo:
> This ensures that we bind the `connection` listener immediately after beginning the connection.

Are you referring to `socket.once('connect', onConnect)`? If so, was there any bug with the previous approach, or this is just to improve the coding style?

sangaman:
I don't think there was a bug, so it's partly about style and partly about preventing a bug from surfacing down the road. Basically previously we would start the socket connection for outgoing connections on `fromOutbound`, but we wouldn't actually start listening for the connection event until we called `open` (which in turn calls `initConnection` where the connection listening actually happens).

PR #436, opened 2018-09-04
order book refactoring

moshababo:
Solves #215 

Refactoring was made mostly to avoid order management logic split between `OrderBook` and `MatchingEngine`, pushing it all to the later. 

`MatchingEngine` now contains a merged priority queue (ownOrders + peerOrders), and separated lists (ownOrder | peerOrder). Both of them holds the same order object instance, so the object can be retrieved (and updated) from any of them. 

Tests were added, but more need to be made in order to be production ready. I was thinking on making `OrderBook` more stateless, in the sense of not having to deal with the DB at all (since we dropped the orders persistence), but having the pairs values injected into its constructor (for the higher level class we can use `Xud`, `Service` with some adjustments, or something else). This will help to improve its tests, so they could be plain unit tests. At the moment most of the testing are for `MatchingEngine`.


lib/orderbook/MatchingEngine.ts
```diff
@@ -66,34 +101,140 @@ class MatchingEngine {
...
+    if (list.has(order.id)) {
+      return false;
+    } else {
+      list.set(order.id, order);
+      queue.add(order);
+      return true;
+    }
+  }
+
+  public removePeerOrderQuantity = (orderId: string, quantityToDecrease?: number): StampedPeerOrder | undefined => {
```
michael1011: Don't we have a TSLint for having `public` methods before `private` ones? @sangaman

lib/orderbook/MatchingEngine.ts
```diff
@@ -66,34 +101,140 @@ class MatchingEngine {
...
+    const order = this.peerOrders.buy.get(orderId) || this.peerOrders.sell.get(orderId);
+    if (!order) {
+      return;
+    }
+
+    if (quantityToDecrease && quantityToDecrease < Math.abs(order.quantity)) {
+      order.quantity = order.quantity - quantityToDecrease;
+      return { ...order, quantity: quantityToDecrease };
```
michael1011: Why are you returning `quantityToDecrease`? It is a parameter so the function calling `removePeerOrderQuantity` will know it anyways. Returning what is left is maybe more sensible.

lib/orderbook/MatchingEngine.ts
```diff
@@ -66,34 +101,140 @@ class MatchingEngine {
...
+    const isBuyOrder = order.quantity > 0;
+    const list = isBuyOrder ? lists.buy : lists.sell;
+    const queue = isBuyOrder ? this.queues.buy : this.queues.sell;
+
+    list.delete(order.id);
+    queue.remove(order);
+  }
+
+  private getOrderList = (order: StampedOrder): OrderList<StampedOrder> => {
+    switch (isOwnOrder(order)) {
```
michael1011: A `switch` is kinda overkill for a boolean isn't it?

lib/orderbook/OrderBook.ts
```diff
@@ -147,64 +98,8 @@ class OrderBook extends EventEmitter {
...
-        return true;
-      }
-    }
-
-    this.logger.warn(`Invalid orderId: ${orderId}`);
-    return false;
-  }
-
   private addOwnOrder = (order: orders.OwnOrder, discardRemaining = false): matchingEngine.MatchingResult => {
+
```
michael1011: What about this blank line?

lib/orderbook/OrderBook.ts
```diff
@@ -250,106 +144,77 @@ class OrderBook extends EventEmitter {
...
-    }
-    orderToUpdate.quantity -= quantityToDecrease;
-    if (orderToUpdate.quantity === 0) {
-      if (isOwnOrder) {
-        const { localId } = order as orders.StampedOwnOrder;
-        this.localIdMap.delete(localId);
-      }
-      orderMap.delete(order.id);
+    if (orderId === undefined) {
+      return { removed: false, globalId: orderId };
```
michael1011: What about just:
```
    if (!orderId) {
      return { removed: false };
```

lib/orderbook/OrderBook.ts
```diff
@@ -250,106 +144,77 @@ class OrderBook extends EventEmitter {
...
+      this.emit('peerOrder.invalidation', { orderId, pairId, quantity: order.quantity });
+      return order;
     }
   }
 
-  private isOwnOrdersMap = (ordersMap: OrdersMap) => {
-    return ordersMap === this.ownOrders;
+  private removePeerOrders = async (peer: Peer): Promise<void> => {
+    this.matchingEngines.forEach((matchingEngine) => {
+      const orders = matchingEngine.removePeerOrders(peer.nodePubKey!);
```
michael1011: We should check if the peer supports the pair of the matching engine first.

test/unit/MatchingEngine.spec.ts
```diff
@@ -28,6 +32,11 @@ peerPubKey = '029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345
   pairId: PAIR_ID,
 });
 
+const isEmpty = (engine: MatchingEngine) => {
+  expect(engine.ownOrders.sell).to.be.empty;
+  expect(engine.queues.sell.isEmpty()).to.be.true;
```
michael1011: Why are you just checking the sell queue and list and not the buy ones?

test/unit/MatchingEngine.spec.ts
```diff
@@ -150,96 +159,180 @@ describe('MatchingEngine.splitOrderByQuantity', () => {
...
 
 describe('MatchingEngine.removeOwnOrder', () => {
-  it('should add a new ownOrder and then remove it', async () => {
-    const engine = new MatchingEngine(loggers.orderbook, PAIR_ID);
-    expect(engine.isEmpty()).to.be.true;
+  let engine: MatchingEngine;
 
+  beforeEach(() => {
+    engine = new MatchingEngine(loggers.orderbook, PAIR_ID);
+    isEmpty(engine);
```
michael1011: We have this exact code four times in this file.

lib/orderbook/MatchingEngine.ts
moshababo: This is how it was before my changes. It's basically simulating the order you remove, and it's being used for the peerOrder invalidation (`subscribePeerOrders`).

lib/orderbook/MatchingEngine.ts
moshababo: Indeed, I don't know why I thought it would be cleaner like this. Will change it.

lib/orderbook/OrderBook.ts
moshababo: Haven't touched this row either. I left it like this because I thought that it's a bit safer like that. But I can change it.

test/unit/MatchingEngine.spec.ts
moshababo: Good catch, it was dropped. I'll add it.

test/unit/MatchingEngine.spec.ts
moshababo: I can't put `beforeEach` on the global scope because it will be called on all the tests in the current run (all files), but I can do it for the other lines.

lib/orderbook/OrderBook.ts
moshababo: Right.. I added a TODO, because it's the equivalent to the TODO of sending only supported pairs (line 225).

moshababo:
I did it like in case we'll do need the differentiation for buy/sell, as it currently being used in `getOrders`. If you don't think we'll need it, it can indeed help simplify some logic. If we'll end up needing it, we'll have to do a full scan to separate between them.
The current situation also help to improve performance a bit, because we usually know which side are we, but that's nothing serious. The main thing is that the current approach is more flexible for upcoming features, in the cost of some extra logic. 

If you suggest to change it, i'll need to change `getOrders` response as well. If we wish to finalize this as part of prealpha.3, let's move it to a separate PR (unless you'll gonna have some review comments anyway).

> I agree with removing db-related functions from the Orderbook class for exactly the reasons you stated, probably good for another PR. Maybe it would just exist in OrderbookRepository which would in turn exist separately from Orderbook, and maybe OrderbookRepository could be renamed since it's not actual dealing with orders any more but rather Do you want to open an issue?

You can open an an issue with some design suggestions. This should also include refactoring the order book tests.

moshababo:
I did it like in case we'll do need the differentiation for buy/sell, as it currently being used in `getOrders`. If you don't think we'll need it, it can indeed help simplify some logic. If we'll end up needing it, we'll have to do a full scan to separate between them.
The current situation also help to improve performance a bit, because we usually know which side are we, but that's nothing serious. The main thing is that it's more flexible for upcoming features, in the cost of some extra logic. 

If you suggest to change it, i'll need to change `getOrders` response as well. If we wish to finalize this as part of prealpha.3, let's move it to a separate PR (unless you'll gonna have some review comments anyway).

> I agree with removing db-related functions from the Orderbook class for exactly the reasons you stated, probably good for another PR. Maybe it would just exist in OrderbookRepository which would in turn exist separately from Orderbook, and maybe OrderbookRepository could be renamed since it's not actual dealing with orders any more but rather Do you want to open an issue?

You can open an an issue with some design suggestions. This should also include refactoring the order book tests.

sangaman:
OK I will open an issue for other design suggestions, thanks.

lib/orderbook/MatchingEngine.ts
```diff
@@ -1,23 +1,58 @@
...
+
 /** A class to represent a matching engine responsible for matching orders for a given trading pair according to their price and quantity. */
 class MatchingEngine {
-  public buyOrders: FastPriorityQueue<StampedOrder>;
-  public sellOrders: FastPriorityQueue<StampedOrder>;
+  /** Active buy/sell orders priority queue. */
+  public queues: OrderSidesQueues;
+  /** A map between active own buy/sell order ids and orders. */
+  public ownOrders: OrderSidesLists<orders.StampedOwnOrder>;
+  /** A map between active peer buy/sell order ids and orders. */
```
sangaman: I think as long as we keep this `OrderSides` structure we should clarify in the comments here that these are a pair of maps rather than a single map.

lib/orderbook/MatchingEngine.ts
sangaman: I agree that it's a weird pattern though to return a parameter, maybe add some comments here explaining what's going on. Which (if I understand correctly), is that it's reducing the quantity of the peer order when the peer order is not fully filled, and returning a copy of it with the quantity that was filled.

lib/orderbook/MatchingEngine.ts
```diff
@@ -1,23 +1,58 @@
...
+type OrderSidesQueues = {
+  buy: FastPriorityQueue<StampedOrder>,
+  sell: FastPriorityQueue<StampedOrder>,
+};
+
 /** A class to represent a matching engine responsible for matching orders for a given trading pair according to their price and quantity. */
 class MatchingEngine {
-  public buyOrders: FastPriorityQueue<StampedOrder>;
-  public sellOrders: FastPriorityQueue<StampedOrder>;
+  /** Active buy/sell orders priority queue. */
```
sangaman: Here actually I think the comment can be  bit clearer too, maybe "A pair of priority queues for the buy and sell sides of this trading pair."

lib/orderbook/MatchingEngine.ts
```diff
@@ -66,34 +101,139 @@ class MatchingEngine {
...
    */
-  public match = (takerOrder: StampedOwnOrder): matchingEngine.MatchingResult => {
+  private match = (takerOrder: StampedOwnOrder): matchingEngine.MatchingResult => {
     const isBuyOrder = takerOrder.quantity > 0;
     const matches: matchingEngine.OrderMatch[] = [];
     /** The unmatched remaining taker order, if there is still leftover quantity after matching is complete it will enter the queue. */
     let remainingOrder: StampedOwnOrder | undefined = { ...takerOrder };
 
-    const matchAgainst = isBuyOrder ? this.sellOrders : this.buyOrders;
+    const matchAgainst = isBuyOrder ? this.queues.sell : this.queues.buy;
```
sangaman: Maybe rename this to `queue`? I was never a big fan of `matchAgainst`. You can throw in a comment like `/** The queue to match the taker order against **/`.

lib/orderbook/MatchingEngine.ts
```diff
@@ -106,21 +246,30 @@ class MatchingEngine {
         // there's no match with the best available maker order, so end the matching routine
         break;
       } else {
+        // get the order from the top of the queue, and remove its ref from the list as well
         const makerOrder = matchAgainst.poll()!;
+        const list = this.getOrderList(makerOrder);
+        list.delete(makerOrder.id);
```
sangaman: Maybe something for another PR, but I think it would be to `peek` the order at the top of the queue, and then only `poll()` and `list.delete()` once we confirm that it's getting filled entirely. That should be slightly more performant plus it feels like a cleaner approach.

moshababo:
Had some rough conflicts on the order book, hopefully nothing was screwed. Travis is having it's own issues.

sangaman:
If you want to get this to pass the build checks, merge or rebase the latest from master. I'm also fine with merging it as is tho.

Issue #424, opened 2018-08-31
CLI enhancements


offerm:
For ease of use I suggest the following changes to the CLI:

1. cancelorder - Make pair_id optional.  Order ID is managed globally hence pair_id is not needed to find the order. If provided it should be used only for verification.
2. getorders - make pair_id optional. If not provided it should list all orders. If provided it should filter the list to include that pair_id only. 

Not related to pair_id:
3. rename getorders to listorders. 
4. rename getpairs to listpairs

sangaman:
For `getorders` that actually used to be the default behavior, and more recently we switched it to require a pair. Having orders for multiple pairs in the same collection is very messy, since the prices can vary wildly between pairs and it's not clear what you're looking at or how it should be sorted. However, it might be nice to have a "default" pair, perhaps whatever is the first row in the pairs database, that is used as the default when no pair is specified.

I also prefer `listpairs` to `getpairs` actually - since the result is truly what I would consider a "list". But I prefer `getorders` to `listorders` because you're getting both sides of an orderbook which is not exactly a list. Just my 2 satoshis.

sangaman:
@michael1011 told me that what you were thinking for getorders with optional pair_id is to return orders for all pairs separately, not all orders jumbled together in a single collection. That makes much more sense, and I'd be fine with that.

He also talked me out of having a "default" pair_id. For cancelorder, we might not need a pair_id at all, we just need to restructure our internal mapping of local order id to global order id to map to local order id to a global order id AND a pair id. I think that makes sense as well.

offerm:
> I also prefer listpairs to getpairs actually - since the result is truly what I would consider a "list". But I prefer getorders to listorders because you're getting both sides of an orderbook which is not exactly a list. Just my 2 satoshis.

@sangaman you look at this too technical. Consider the user that is using it and the operation that he is tring to do. Read `list` as a verb and not as the output format. use `get` for a single object and `list` for multiple.

sangaman:
For me the difference between fetching a single object and multiple objects is whether the noun is plural (e.g. `order` vs `orders`). `list` to me as a verb suggests generating a list of things, so if it's in a format other than an array or list I don't think it's the ideal word to use. I'd still prefer sticking to `getorders` but it's obviously not a deal breaker, maybe others can chime in if they have a preference.

kilrau:
Summary:
1) `cancelOrder` - Remove `<pair_id>`
[...]

PR #438, opened 2018-09-04
Enable add/remove pairs & currencies from grpc

### LINKED ISSUE #313: Add database update functions to gRPC and CLI
add pair
remove pair
empty pairs
add currency
remove currency
empty currencies
add host
remove host
empty hosts
add bannedHost
remove bannedHost
empty bannedHosts

sangaman:
Closes #313. Addresses part of #424 - renaming getpairs` to `listpairs`.

This PR adds grpc calls for `addCurrency`, `addPair`, `removeCurrency`, and `removePair`. This allows the user to modify supported currencies and trading pairs without needing to edit the database manually.

It removes the `SwapProtocol` enum and instead tracks the payment channel network client used for swaps on a per-currency basis using a new `NetworkClient` enum. This is a breaking database change that will require the `currencies` and `pairs` tables to be recreated.

New service tests are added to test the various use cases for the newly added calls, along with corresponding cli commands.


lib/grpc/GrpcService.ts
```diff
@@ -46,16 +46,20 @@ class GrpcService {
...
         code = status.INVALID_ARGUMENT;
         break;
-      case orderErrorCodes.INVALID_PAIR_ID:
+      case orderErrorCodes.UNSUPPORTED_PAIR:
         code = status.NOT_FOUND;
         break;
       case orderErrorCodes.DUPLICATE_ORDER:
       case p2pErrorCodes.NODE_ALREADY_CONNECTED:
+      case orderErrorCodes.CURRENCY_ALREADY_SUPPORTED:
+      case orderErrorCodes.PAIR_ALREADY_SUPPORTED:
```
michael1011: I would rename those two errors to `CURRENCY_ALREADY_EXISTS` and `PAIR_ALREADY_EXISTS` because `ALREADY_SUPPORTED` doesn't sound like an error.

lib/orderbook/OrderBook.ts
```diff
@@ -86,14 +94,21 @@ class OrderBook extends EventEmitter {
...
 
-    this.pairs.forEach((pair) => {
-      this.pairIds.push(pair.id);
+    pairs.forEach((pair) => {
       this.matchingEngines.set(pair.id, new MatchingEngine(this.logger, pair.id));
       this.ownOrders.set(pair.id, this.initOrders());
       this.peerOrders.set(pair.id, this.initOrders());
+      this.pairs.set(pair.id, pair);
+    });
+    currencies.forEach((currency) => {
```
michael1011: It doesn't matter but shouldn't the currencies be initialized before the pairs?

lib/orderbook/OrderBook.ts
```diff
@@ -136,6 +151,56 @@ class OrderBook extends EventEmitter {
...
+  public removeCurrency = (currencyId: string) => {
+    const currency = this.currencies.get(currencyId);
+    if (currency) {
+      for (const pairId of this.pairIds) {
+        if (pairId.includes(currencyId)) {
+          throw errors.CURRENCY_CANNOT_BE_REMOVED(currencyId, pairId);
+        }
+      }
+      this.currencies.delete(currencyId);
+      return currency.destroy();
```
michael1011: We should declare functions that return a promise as `async`.

lib/service/Service.ts
```diff
@@ -55,9 +55,13 @@ const argChecks = {
   },
   NON_ZERO_QUANTITY: ({ quantity }: { quantity: number }) => { if (quantity === 0) throw errors.INVALID_ARGUMENT('quantity must not equal 0'); },
   PRICE_NON_NEGATIVE: ({ price }: { price: number }) => { if (price < 0) throw errors.INVALID_ARGUMENT('price cannot be negative'); },
-  SUPPORTED_CURRENCY: ({ currency }: { currency: string }) => {
-    const supportedCurrencies = ['BTC', 'LTC']; // TODO: detect supported currencies automatically instead of hardcoding
-    if (!supportedCurrencies.includes(currency)) throw errors.INVALID_ARGUMENT(`currency ${currency} is not supported`);
+  VALID_CURRENCY: ({ currency }: { currency: string }) => {
+    if (currency.length !== 3 || !currency.match(/^[A-Z]+$/))  {
```
michael1011: There are some currencies that have a ticker symbol with just two letters ([Aeternity](https://etherscan.io/token/0x5ca9a71b1d01849c0a95490cc00559717fcf0d1d) for example).

lib/service/Service.ts
```diff
@@ -55,9 +55,13 @@ const argChecks = {
...
+      throw errors.INVALID_ARGUMENT('currency must consist of exactly 3 upper case English letters');
+    }
+  },
+  VALID_NETWORK_CLIENT: ({ networkClient }: { networkClient: number }) => {
+    if (NetworkClients[networkClient] === undefined) throw errors.INVALID_ARGUMENT('network client is not recognized');
```
michael1011: What about just `if (!NetworkClients[networkClient])`

lib/utils/utils.ts
```diff
@@ -134,3 +135,14 @@ export const groupBy = (arr: object[], keyGetter: (item: any) => string | number
 export const ms = (): number => {
   return Date.now();
 };
+
+export const getEnumNumbers = (enumObj: any) => {
```
michael1011: Is this function used somewhere?

lib/cli/commands/addcurrency.ts
```diff
@@ -0,0 +1,38 @@
...
+  network_client: {
+    description: 'The payment channel network client',
+    type: 'string',
+    choices: ['LND', 'RAIDEN'],
+  },
+  token_address: {
+    description: 'The contract address for layered tokens such as ERC20',
+    type: 'string',
+  },
+  subunits: {
```
michael1011: I am fine with `subunits` but did you consider using the decimals of the currency instead?

lib/orderbook/OrderBook.ts
sangaman: Stylistically I don't think there's a problem with non-async methods returning promises, I believe there are instances of that throughout the codebase and as far as I know it's not bad practice or anything and doesn't affect the compiler or type checking. Correct me if I'm wrong.

In this case specifically it's actually returning a `Bluebird` and not a promise, they function pretty much the same but aren't exactly the same I believe, and making the method `async` gives it a return type of `Promise<void>` instead of `Bluebird<void>` which seems undesirable to me.

lib/service/Service.ts
sangaman: I was not aware, and there are actually a handful that have 4 letters as well, so I'll relax this to be a length of 2-4.

lib/utils/utils.ts
sangaman: Oops nope, I was using it at one point for putting restrictions on the eligible network clients in the database, but abandoned that approach. Thanks.

lib/cli/commands/addcurrency.ts
sangaman: What do you mean? Like the base 10 exponent for the number of subunits in a unit? Such that bitcoin's would be 8? 10^8 satoshis in a bitcoin.

lib/cli/commands/addcurrency.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const builder = {
+  currency: {
+    description: 'The ticker symbol for the currency',
+    type: 'string',
+  },
+  network_client: {
+    description: 'The payment channel network client',
+    type: 'string',
+    choices: ['LND', 'RAIDEN'],
```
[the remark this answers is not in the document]
michael1011: Yeah makes sense. We should add `choices` to `logLevel` too.

lib/orderbook/OrderBook.ts
michael1011: We can leave it as is then. I wasn't aware that not declaring `async` doesn't change its behavior.

lib/service/Service.ts
```diff
@@ -51,9 +52,13 @@ const argChecks = {
   },
   NON_ZERO_QUANTITY: ({ quantity }: { quantity: number }) => { if (quantity === 0) throw errors.INVALID_ARGUMENT('quantity must not equal 0'); },
   PRICE_NON_NEGATIVE: ({ price }: { price: number }) => { if (price < 0) throw errors.INVALID_ARGUMENT('price cannot be negative'); },
-  SUPPORTED_CURRENCY: ({ currency }: { currency: string }) => {
-    const supportedCurrencies = ['BTC', 'LTC']; // TODO: detect supported currencies automatically instead of hardcoding
-    if (!supportedCurrencies.includes(currency)) throw errors.INVALID_ARGUMENT(`currency ${currency} is not supported`);
+  VALID_CURRENCY: ({ currency }: { currency: string }) => {
+    if (currency.length < 2 || currency.length > 4 || !currency.match(/^[A-Z]+$/))  {
```
michael1011: I have to annoy you again but I just noticed that there are tokens with symbols that consist of 5 characters: `STEEM`, `WAVES`, `STRAT`

What about getting rid of the upper limit entirely?

lib/cli/commands/addcurrency.ts
michael1011: Yeah exactly. Especially because all of those ERC20 contracts are using that format and copying `100000000000000000` is harder than just `18`.

sangaman: Yeah I agree that's better, I'll change it.

sangaman:
OK, I changed `subunits` to `decimalPlaces` and I also renamed `NetworkClients` to `SwapClients` as we'd discussed in gitter. See what you think @michael1011, thanks.

lib/cli/commands/addcurrency.ts
```diff
@@ -0,0 +1,38 @@
+import { Arguments } from 'yargs';
+import { callback, loadXudClient } from '../command';
+import { AddCurrencyRequest } from '../../proto/xudrpc_pb';
+import { SwapClients } from '../../types/enums';
+
+export const command = 'addcurrency <currency> <swap_client> [decimal_places] [token_address]';
+
+export const describe = 'add a supported currency';
```
michael1011: `add a supported currency` sounds like you add an already supported currency. What about `add support for a currency` or simply `add a currency`?

lib/cli/commands/addcurrency.ts
```diff
@@ -0,0 +1,38 @@
...
+    type: 'string',
+  },
+  swap_client: {
+    description: 'The payment channel network client for swaps',
+    type: 'string',
+    choices: ['LND', 'RAIDEN'],
+  },
+  decimal_places: {
+    description: 'The places to the right of the decimal point of the smallest subunit (e.g. satoshi)',
+    default: 8,
```
michael1011: Idea for the future: automatically detect `decimal_places` of ERC20s if nothing is specified. The users are running an Ethereum node anyways.

lib/orderbook/OrderBook.ts
```diff
@@ -85,14 +92,21 @@ class OrderBook extends EventEmitter {
...
+    const promises = [await this.repository.getPairs(), await this.repository.getCurrencies()];
+    const results = await Promise.all(promises);
+    const pairs = results[0] as db.PairInstance[];
+    const currencies = results[1] as db.CurrencyInstance[];
 
-    this.pairs.forEach((pair) => {
-      this.pairIds.push(pair.id);
+    currencies.forEach((currency) => {
+      this.currencies.set(currency.id, currency);
+    });
```
michael1011: `currencies.forEach(currency => this.currencies.set(currency.id, currency));`?

lib/orderbook/OrderBook.ts
```diff
@@ -135,6 +149,56 @@ class OrderBook extends EventEmitter {
...
+    const currencyInstance = await this.repository.addCurrency(currency);
+    this.currencies.set(currencyInstance.id, currencyInstance);
+  }
+
+  public removeCurrency = (currencyId: string) => {
+    const currency = this.currencies.get(currencyId);
+    if (currency) {
+      for (const pairId of this.pairIds) {
+        if (pairId.includes(currencyId)) {
+          throw errors.CURRENCY_CANNOT_BE_REMOVED(currencyId, pairId);
```
michael1011: If the currency is in multiple pairs the user has to call `removeCurrency` and `removePair` multiple times to know the blocking pairs and remove them. I think all affected pairs should be returned at once. But that can be done in a separate PR.

Btw: what is more efficient `pairId.includes(currencyId)` or checking if `currencyId` equals either `baseCurrency` or `quoteCurrency`?

lib/orderbook/OrderBook.ts
```diff
@@ -154,7 +218,7 @@ class OrderBook extends EventEmitter {
     const orderId = this.localIdMap.get(localId);
 
     if (orderId === undefined) {
-      throw errors.INVALID_PAIR_ID(pairId);
+      throw errors.UNSUPPORTED_PAIR(pairId);
```
michael1011: Just because the `orderId` can't be found that doesn't mean that the pair is not supported?

lib/service/Service.ts
```diff
@@ -86,6 +91,25 @@ class Service extends EventEmitter {
...
+  /** Adds a supported currency. */
+  public addCurrency = async (args: { currency: string, swapClient: SwapClients | number, tokenAddress?: string, decimalPlaces?: number }) => {
+    argChecks.VALID_CURRENCY(args);
+    argChecks.VALID_SWAP_CLIENT(args);
+    const { currency, swapClient, tokenAddress, decimalPlaces } = args;
+
+    await this.orderBook.addCurrency({
+      tokenAddress,
+      swapClient,
+      decimalPlaces: decimalPlaces || 8,
```
michael1011: I think this default value should be in the orderbook instead of the service class.

lib/cli/commands/addcurrency.ts
sangaman: I'll also change the comments in the proto definition.

lib/orderbook/OrderBook.ts
sangaman: I was curious and it looks like the latter is indeed faster: http://jsben.ch/nIxYb

I'm sure it won't make a difference but I'll have it use the faster way.

lib/orderbook/OrderBook.ts
sangaman: Looks like this was using the wrong error message before this PR, but I'll fix it.

sangaman:
I changed a bit more of the wording around "supported" currencies/pairs in comments and ran through manual tests of all the calls once more. I think this is ready to merge.

sangaman:
The build isn't starting because [travis is having issues](https://www.traviscistatus.com/) but it passes all tests locally.

Issue #416, opened 2018-08-28
Remove `executeSwap` command from CLI


kilrau:
Just for (Berlin) demo purposes

offerm:
Best to leave the command in and hide it from the list.

To be done AFTER Berlin.

sangaman:
To me, the cli command itself is not a big dea. I'd want to remove the `executeSwap` code from xud though so as not to have to maintain it and to be able to pare down the deal packets from repeating data that's specified by the order, and treating the orderId as a non-nullable field for deal requests. I'm in favor of removing the call entirely, but it's not urgen it offer or anyonen else has plans to use it in the foreseeable future.

kilrau:
@offerm you are the decision maker here, please let us know if you'll use it, then we'll keep it

offerm:
@sangaman @kilrau the only reason to keep this has to do with our planned swap protocol change. My concern is that during the change I would need to wait for XUD related changes in order to check the new protocol.
As such I suggest to leave the executeSwap as is for now and remove it once we implement the next version of the swap Protocol. (https://hackmd.io/s/SJ2qZUzdX)

@Daniel, I suggest we accept the PR or close it. This issue can also be assigned to a later milestone so we will not forget it.

sangaman:
@offerm What about removing `executeSwap` from master and leaving it in a branch if you need it for testing purposes?

kilrau:
Leaving it in a `swap_v2` branch sounds like a good idea to me

offerm:
@sangaman I'm OK with removing it from master together with related code.

No need to keep it as a branch. I will pull it from history if needed.

go for it.

Offer

offerm:
PR #465 

@sangaman  please review/accept

PR #444, opened 2018-09-06
Disable executeSwap CLI

offerm:


sangaman:
I don't think this is necessary until we remove the `executeSwap` rpc call altogether. How are you executing this call if not from the command line?

offerm:
@sangaman maybe it is not necessary but #416 is for alpha-1.

There are many ways to execute the call if not from the command line:
- client program
- swagger
- etc

and I can always change it back locally....

sangaman:
Oh I didn't realize #416 said just for the CLI. I don't see any value in removing just the CLI command but leaving the rpc call, unless you're sure you won't use the CLI command in the future.

I have no intentions to use the executeswap call in any capacity going forward so I don't object to removing, but it just doesn't make sense to me to remove the cli command but keep the call. I'll put a comment in the issue.

PR #446, opened 2018-09-06
hash-resolver rpc in lnd.proto

offerm:
This PR implements the rpc.proto of the LND's resolver branch.
In this branch, rpc.proto includes also the hash-resolver service definition.

### DIFF
--- a/lib/grpc/GrpcServer.ts
+++ b/lib/grpc/GrpcServer.ts
@@ -6,7 +6,7 @@ import GrpcService from './GrpcService';
 import Service from '../service/Service';
 import errors from './errors';
 import { XudService } from '../proto/xudrpc_grpc_pb';
-import { HashResolverService } from '../proto/hash_resolver_grpc_pb';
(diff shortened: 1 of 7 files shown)


sangaman:
Just to be clear, is this using the most recent rpc proto definition from lnd? Also, one minor thing is the updates to api.md - can you try running `go get -u github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc` and then regenerating protos?

This looks good to me otherwise, but maybe we should wait a bit to see if we get any feedback on the approach of putting the hash resolver service definition into the main lnd proto file?

offerm:
@sangaman 
Updated after installing `github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc`

Yes, latest LND's rpc.proto is used. Actually, latest + hash-resolver (from ExchangeUnion/Lnd@resolver)

We should not wait with this since the resolver branch (lnd) is already updated and if you use un-updated XUD against it the swap will fail.
If we get feedback we adopt.

PR #449, opened 2018-09-07
CLI enhancements

### LINKED ISSUE #424: CLI enhancements

michael1011:
Closes #424

I don't think that `OrderTypes` is a good for a gRPC message name but I couldn't come up with anything better so feel free to suggest something else.

It is still unclear if `getOrders` should be renamed to `listOrders`. Any decisions on that?

### DIFF
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -2,19 +2,20 @@ import Logger from '../Logger';
 import Pool from '../p2p/Pool';
 import OrderBook from '../orderbook/OrderBook';
 import LndClient, { LndInfo } from '../lndclient/LndClient';
-import RaidenClient, { TokenSwapPayload, RaidenInfo } from '../raidenclient/RaidenClient';
+import RaidenClient, { RaidenInfo } from '../raidenclient/RaidenClient';
 import { EventEmitter } from 'events';
 import errors from './errors';
 import { SwapDealRole, SwapClients } from '../types/enums';
 import { parseUri, getUri, UriParts } from '../utils/utils';
 import * as lndrpc from '../proto/lndrpc_pb';
-import { Pair } from '../types/orders';
+import { Pair, StampedOrder } from '../types/orders';
 import Swaps from '../swaps/Swaps';
+import { OrderSidesArrays } from '../orderbook/MatchingEngine';
 
 /**
  * The components required by the API service layer.
  */
-export type ServiceComponents = {
+type ServiceComponents = {
   orderBook: OrderBook;
   lndBtcClient: LndClient;
   lndLtcClient: LndClient;
@@ -113,12 +114,11 @@ class Service extends EventEmitter {
   /*
    * Cancel placed order from the orderbook.
    */
-  public cancelOrder = async (args: { orderId: string, pairId: string }) => {
-    const { orderId, pairId } = args;
+  public cancelOrder = async (args: { orderId: string }) => {
+    const { orderId } = args;
     argChecks.HAS_ORDER_ID(args);
-    argChecks.HAS_PAIR_ID(args);
 
-    this.orderBook.removeOwnOrderByLocalId(pairId, orderId);
+    this.orderBook.removeOwnOrderByLocalId(orderId);
   }
 
   /** Gets the total lightning network channel balance for a given currency. */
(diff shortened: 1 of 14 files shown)


proto/xudrpc.proto
```diff
@@ -162,14 +160,12 @@ message GetOrdersRequest {
   uint32 max_results = 2 [json_name = "max_results"];
 }
 message GetOrdersResponse {
```
sangaman: Can you also update the description of the `GetOrders` call (line 62) to a better description of what it does, I think it's quite out of date.

proto/xudrpc.proto
```diff
@@ -162,14 +160,12 @@ message GetOrdersRequest {
...
-  // A list of peer orders
-  Orders peer_orders = 1 [json_name = "peer_orders"];
-  // A list of orders placed locally
-  Orders own_orders = 2 [json_name = "own_orders"];
+  // A map between a pair and its orders
```
sangaman: Maybe specify `pair id` or `trading pair id`

proto/xudrpc.proto
```diff
@@ -221,6 +217,13 @@ message Orders {
   repeated Order sell_orders = 2 [json_name = "sell_orders"];
 }
 
+message OrderTypes {
```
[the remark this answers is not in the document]
michael1011: What about two different gRPC methods: `GetOwnOrders` and `GetPeerOrders`? We can combine those two in the CLI if we really want to.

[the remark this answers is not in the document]
michael1011: The problem with an `includeOwnOrders` flag is that you can't get just you own orders. But this is fine for now and should be dealt with in a separate PR.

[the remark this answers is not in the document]
sangaman: True, I'm not sure getting ones own orders will actually be a useful feature though. If it turns out that it is I agree we can address it then.

sangaman:
Thanks, will review again later tonight, in the meantime you can merge or rebase the latest from master to get the build checks to work.

kilrau:
> What about two different gRPC methods: GetOwnOrders and GetPeerOrders? We can combine those two in the CLI if we really want to.

I think `getOrders` with optional `-peer` and `--own` flags are the better way.

proto/xudrpc.proto
```diff
@@ -67,7 +67,7 @@ service Xud {
     };
   }
   
-  /* Get a list of standing orders from the order book. */
+  /* Get a map between pair ids and its orders from the order book. */
```
sangaman: I'd either say "and their orders" or "and orders" since "its" is singular while "pair ids" is plural. It might also be more clear to say "buy and sell orders" since we are returning them in two separate collections.

proto/xudrpc.proto
```diff
@@ -225,12 +223,12 @@ message GetOrdersRequest {
...
   uint32 max_results = 2 [json_name = "max_results"];
+  // Whether own orders should be included in result or not
+  bool include_own_orders = 3 [json_name = "include_own_orders"];
 }
 message GetOrdersResponse {
-  // A list of peer orders
-  Orders peer_orders = 1 [json_name = "peer_orders"];
-  // A list of orders placed locally
-  Orders own_orders = 2 [json_name = "own_orders"];
+  // A map between pair ids and its orders
```
sangaman: Same for wording here.

lib/service/Service.ts
```diff
@@ -265,18 +264,34 @@ class Service extends EventEmitter {
...
+  public getOrders = (args: { pairId: string, maxResults: number, includeOwnOrders: boolean }) => {
+    const { pairId, maxResults, includeOwnOrders } = args;
     argChecks.MAX_RESULTS_NOT_NEGATIVE(args);
 
-    const result = {
-      peerOrders: this.orderBook.getPeerOrders(pairId, maxResults),
-      ownOrders: this.orderBook.getOwnOrders(pairId, maxResults),
+    const result = new Map<string, OrderArrays>();
+    const getOrderTypes = (pairId: string) => {
+      const result = this.orderBook.getPeerOrders(pairId, maxResults);
```
sangaman: Maybe call this `peerOrders` or `orders` instead of `result`, mainly to eliminate any confusion between it and the `result` variable in the scope above it.

PR #451, opened 2018-09-07
Npm install

kilrau:


instructions-for-devs.md
```diff
@@ -13,13 +13,9 @@ cd xud
 npm i
 npm run compile
 ```
-For everyone else, we recommend using the [latest release](https://github.com/ExchangeUnion/xud/releases), below an example using the `v1.0.0-prealpha.2` release tag:
+For everyone else, we recommend installing the latest release from the npm:
```
sangaman: Just "from npm" without "the"

PR #461, opened 2018-09-11
Feat: Switch from mysql to sqlite

### LINKED ISSUE #460: Move to SQLite

sangaman:
Closes #460.

This PR transitions from myql to sqlite which means the database is now stored as a file. This removes an external dependency on a running mysql server and simplifies the logic around creating new databases and test databases.

I removed the `db` configuration since it no longer applies to sqlite, and added a `dbPath` option for specifying a specific path to a db file, which can also be set to `:memory:` to use a temporary in-memory database (useful for testing).  The default dbPath is `[xudir]/xud.db`.

In working on this I uncovered a pair of bugs. The first is that `initDb` was not being checked before populating a new database with default values - I fixed this along with the migration. This led me to find a second bug from #438 which prevented a newly added trading pair from being able to accept new orders. No new `MatchingEngine` instance was being created for the newly added pair. The tests previously did not catch this because they were initialized with the pairs we were trying to trade in tests, but switching those pairs to the pairs we had just added via the service layer would cause them to fail. I fixed this in a separate commit, since it's not related to the database. The tests would not pass without this fix, so I included it in this PR.


kilrau:
Gotta check travis tho @sangaman @rsercano 
>The command "eval npm ci  " failed 3 times.
>The command "npm ci " failed and exited with 1 during .

sangaman:
Yup, Travis is having some serious issues with github integration. Builds not working at all now. All tests pass locally for me though.

rsercano:
Hello, I just fixed the build with #464

bin/xud
```diff
@@ -4,16 +4,16 @@ const Xud = require('../dist/Xud').default;
 
 const { argv } = require('yargs')
   .options({
+    dbPath: {
+      describe: 'The file path for the database, or ":memory:" if the database should be kept in memory',
```
michael1011: Do we really need `or ":memory:" if the database should be kept in memory`? A normal user or exchange would not even consider using this and if you think it is beneficial to developers you should write it into either the contribution guidelines or some other document.

lib/db/DB.ts
```diff
@@ -31,31 +19,35 @@ class DB {
   public sequelize: Sequelize.Sequelize;
   public models: Models;
 
-  constructor(private config: DBConfig, private logger: Logger) {
-    assert(Number.isInteger(config.port) && config.port > 1023 && config.port < 65536, 'port must be an integer between 1024 and 65535');
-
-    this.sequelize = this.createSequelizeInstance(this.config);
+  /**
+   * @param storage the file path for the sqlite database file, if not specified the db is stored in memory
```
michael1011: Is the database stored in the memory when `storage` is set to `:memory:`, not specified or both?

lib/db/DB.ts
```diff
@@ -31,31 +19,35 @@ class DB {
...
+   */
+  constructor(private logger: Logger, private storage?: string) {
```
michael1011: `storage` is a bit confusing. What about just `filename`?

bin/xud
sangaman: Agreed.

lib/orderbook/OrderBook.ts
```diff
@@ -191,6 +194,9 @@ class OrderBook extends EventEmitter {
     const pair = this.pairs.get(pairId);
     if (pair) {
       this.pairs.delete(pairId);
+      this.matchingEngines.delete(pairId);
```
moshababo: Again, update the handshake data, re-send `Hello` packet to peers. Otherwise we'll penalize them for keep sending orders of this pair. 

Also, Send order invalidation for all ownOrders in this matchingEngine instance (we might want to introduce invalidating an entire pair instead). Can be done in a separate PR, after merging from master latest order book changes. 

Also, we probably want to trigger order removal events for the client. We shouldn't assume that he's smart enough to trigger his removal logic after calling this. He's relying on these events to fully sync the order book state. Should be done after #123. Can just add a TODO for now.

bin/xud
```diff
@@ -23,26 +23,10 @@ const { argv } = require('yargs')
       describe: 'Path to the log file',
       type: 'string',
     },
-    'db.database': {
```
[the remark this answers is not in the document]
sangaman: That file is gone w/ the latest readme changes and was moved to the wiki, I was going to update the wiki after this is merged but thanks for the reminder.

lib/orderbook/OrderBook.ts
```diff
@@ -161,6 +161,9 @@ class OrderBook extends EventEmitter {
 
     const pairInstance = await this.repository.addPair(pair);
     this.pairs.set(pairInstance.id, pairInstance);
+    this.matchingEngines.set(pairInstance.id, new MatchingEngine(this.logger, pairInstance.id));
```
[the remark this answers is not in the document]
sangaman: Yeah, I'll put in a TODO

lib/orderbook/OrderBook.ts
sangaman: I'll put in a TODO for both of these, thanks.

sangaman:
I need to make a bit of changes to the orderbook fix commit given the latest refactoring recently merged to master, will push and comment when that's done.

sangaman:
Done, it was actually very simple. Merging.

PR #466, opened 2018-09-12
Chore: fixing travis build

sangaman:
This fixes an issue that was causing travis builds to fail by installing `node-pre-gyp`. It also removes mysql queries that are obsolete after the transition to sqlite.

Credit to @rsercano.


michael1011:
No wait a second. The commit description is no accurate: `It also removes mysql queries that are
obsolete after the transition to sqlite.` is true for #464 but not this PR

sangaman:
Ah I'm sorry I should've removed that part but didn't, and didn't see your message until after I'd merged. My mistake.

Issue #431, opened 2018-09-01
SWAP: Proper timelock calculation


offerm:
Swap involves two sendPayments on two different chains.

The timeLock of the first chain should include the time lock of the second chain.

PR #468, opened 2018-09-12
SWAP protocol V2 

offerm:
hash and preimage created by taker.
taker starts the swap.
dealRequest packet removed.


lib/p2p/packets/types/SwapCompletePacket.ts
```diff
@@ -0,0 +1,19 @@
+import Packet, { PacketDirection } from '../Packet';
+import PacketType from '../PacketType';
+
+export type SwapCompletePacketBody = {
+  r_hash: string;
+  r_preimage?: string;
```
sangaman: Why is this field optional? If the swap is complete, shouldn't the preimage be shared knowledge at this point?

lib/swaps/Swaps.ts
```diff
@@ -141,116 +132,152 @@ class Swaps extends EventEmitter {
...
 
     peer.sendPacket(packet);
-    return dealId;
+    return r_hash;
+  }
+
+  /**
+   * resolveHash resolve hash to preimage.
+   */
+  public resolveHash = async (args: { hash: string }) => {
```
sangaman: I don't think we need to keep the `args` parameter from the Service class. This can just take a single string parameter, and better to call it `r_hash` than `hash` I think to stay consistent.

lib/swaps/Swaps.ts
```diff
@@ -12,32 +12,29 @@ import { StampedPeerOrder, StampedOwnOrder } from '../types/orders';
...
   /** The role of the local node in the swap. */
+  // TODO: replace maker/taker with initiator/responder
   myRole: SwapDealRole;
-  /** The xud node pub key of the counterparty to this swap deal. */
-  peerPubKey: string;
   /** Global order id in the XU network. */
   orderId: string;
-  dealId: string;
   /** The quantity of the order to execute as proposed by the taker. Negative when the taker is selling. */
+  // TODO: is it needed here? if yes, should be in satoshis
```
sangaman: My thinking here was to have this quantity be in the same full units we use in the order book for easy subtraction/comparison. Proposed quantity is proposed by the taker, maker can respond with a lesser amount (which becomes the `quantity` field below.

lib/swaps/Swaps.ts
```diff
@@ -56,7 +53,7 @@ interface Swaps {
 }
 
 class Swaps extends EventEmitter {
-  /** A mapping containing two entries for each deal, one using dealId as a key and one using r_hash as a key. */
+  /** A mapping containing an entry for each deal, using r_hash as a key. */
```
sangaman: Now that there's only one key used, this can just say "A map between r_hash and swap deals."

lib/swaps/Swaps.ts
```diff
@@ -141,207 +132,218 @@ class Swaps extends EventEmitter {
...
-      peerPubKey: this.pool.handshakeData.nodePubKey,
+      ...swapRequestBody,
+      r_preimage: preimage.toString('hex'),
+      myRole: SwapDealRole.Taker,
       createTime: Date.now(),
     };
 
     this.addDeal(deal);
-    this.logger.debug('Initiating swap deal: ' + JSON.stringify(deal));
+    this.logger.debug('[SWAP] Initiating deal: ' + JSON.stringify(deal));
```
sangaman: No need for the `[SWAP]` here as this is already inserted into the logs from the swaps context.

lib/swaps/Swaps.ts
```diff
@@ -141,207 +132,218 @@ class Swaps extends EventEmitter {
...
+      takerPubKey: peer.lndPubKey(requestBody.takerCurrency),
+      r_hash: requestBody.r_hash,
+      myRole: SwapDealRole.Maker,
+      createTime: Date.now(),
+    };
 
-      this.addDeal(deal);
-      this.logger.debug('[SWAP] New deal: ' + JSON.stringify(deal));
+    this.addDeal(deal);
+    this.logger.debug('[SWAP] New deal: ' + JSON.stringify(deal));
```
sangaman: Same here for `[SWAP]`

lib/swaps/Swaps.ts
```diff
@@ -141,207 +132,218 @@ class Swaps extends EventEmitter {
...
       } else if (quantity < deal.proposedQuantity) {
-        // TODO: the maker accepted only part of our deal request, adjust the deal amounts with Swaps.calculateSwapAmounts
+        // TODO: the maker accepted only part of our swap request, adjust the deal amounts with Swaps.calculateSwapAmounts
       }
       this.emit('swap.accepted', deal, quantity);
     }
-    deal.makerPubKey = makerPubKey;
-    deal.r_hash = r_hash;
-    this.deals.set(r_hash, deal);
     this.logger.debug('[SWAP] Updated deal: ' + JSON.stringify(deal));
```
sangaman: Same here for `[SWAP]`

lib/swaps/Swaps.ts
```diff
@@ -141,207 +132,218 @@ class Swaps extends EventEmitter {
...
-    let makerPubKey: string | undefined;
-    switch (requestBody.makerCurrency){
-      case 'BTC':
-        makerPubKey = this.pool.handshakeData.lndbtcPubKey;
-        break;
-      case 'LTC':
-        makerPubKey = this.pool.handshakeData.lndltcPubKey;
-        break;
-    }
+    let responsePacket: packets.SwapResponsePacket;
```
sangaman: Is there a reason to declare the `responsePacket` up here and not just when it's assigned a value further down?

lib/p2p/Peer.ts
```diff
@@ -112,6 +112,20 @@ class Peer extends EventEmitter {
     return peer;
   }
 
+  public lndPubKey(chain: string): string | undefined {
```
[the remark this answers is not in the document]
offerm: done

lib/p2p/packets/types/SwapCompletePacket.ts
offerm: WIP. Not there anymore.

lib/swaps/Swaps.ts
offerm: well, you may be right but resolveHash takes few parameters and not just the hash. 
Don't you find it wrong that we don't know about it? 
This has to do with the toObject discussion. 

I will take care of it when doing #431

lib/swaps/Swaps.ts
offerm: quantity and proposedquantity will be handled later. I don't see how this can/should be done without proper integration to the order book.

sangaman: Yes for sure, I didn't mean to suggest it has to be implemented in this PR, this fields are just placeholders. I was just explaining why I put it there and why I wasn't planning on using satoshis.

lib/swaps/Swaps.ts
sangaman: I'm not sure what you mean by other parameters? Which ones?

I'm just saying that the method signature should be `public resolveHash = async (hash: string)  => {`. There's no need to pass this method an object instead of just passing the value directly to be consistent with most other methods that accept only a handful of parameters.

offerm:
@sangaman OK to review now.

I think this is in good shape to be merged into master.

There are many TODOs in code and we need to decide how/when to handle them. Need to open different issues as we think (@kilrau ).

What was not implemented in this PR (@kilrau ):

1. emit of swap.accepted - this has to do with orderbook integration on the maker side. Looks like integration should be done before the swap is accepted. 
2. emit swap.rejected - currently, the maker has no execution path that leads to rejection. To be implemented with some TODOs
3. quantity/proposedQuantity - This also has to do with orderBook integration. TBD
4. handling timeouts/peer disconnection. 

I plan to continue the work tomorrow so I look forward to your feedback.

Enjoy.

lib/swaps/Swaps.ts
offerm: check the HashResolver proto definition:
```
service HashResolver {
    // ResolveHash is used by LND to request translation of Rhash to a pre-image.
    // the resolver may return the preimage and error indicating that there is no
    // such hash/deal
    rpc ResolveHash(ResolveRequest) returns (ResolveResponse) {}
}

message ResolveRequest {
    string hash = 1 [json_name = "hash"];
    uint32 timeout = 2 [json_name = "timeout"];
    uint32 height_now = 3 [json_name = "height_now"];
    int64  amount = 4 [json_name = "amount"];
}

message ResolveResponse {
    string preimage = 1 [json_name = "preimage"];
}
```
`ResolveRequest` has other fields in addition to the `hash`. But we don't know it.

It is really wrong.

lib/swaps/Swaps.ts
```diff
@@ -88,46 +89,46 @@ class Swaps extends EventEmitter {
...
     this.pool.on('packet.swapResponse', this.handleSwapResponse);
+    this.pool.on('packet.swapComplete', this.handleSwapComplete);
+    this.pool.on('packet.swapError', this.handleSwapError);
   }
 
   /**
-   * Gets a deal by its dealId or r_hash value.
-   * @param dealIdOrHash The dealId or r_hash value of the deal to get.
+   * Gets a deal by its r_hash value.
+   * @param hash The r_hash value of the deal to get.
```
sangaman: The param should be `r_hash`

lib/swaps/Swaps.ts
sangaman: Should this TODO be removed then?

lib/swaps/Swaps.ts
```diff
@@ -1,67 +1,68 @@
...
 }
 
 class Swaps extends EventEmitter {
-  /** A mapping containing two entries for each deal, one using dealId as a key and one using r_hash as a key. */
+  /** A map between r_hash and swap deals. */
   private deals = new Map<string, SwapDeal>();
 
   constructor(private logger: Logger, private pool: Pool, private lndBtcClient: LndClient, private lndLtcClient: LndClient) {
     super();
-
```
sangaman: I'd prefer keeping this blank line after `super()`, that's also consistent with how the spacing is in the rest of the project.

lib/swaps/Swaps.ts
```diff
@@ -141,218 +142,284 @@ class Swaps extends EventEmitter {
...
+        break;
+      case SwapDealStatus.AmountSent:
+        assert(deal.myRole === SwapDealRole.Taker && deal.status === SwapDealStatus.SwapRequested ||
+          deal.myRole === SwapDealRole.Maker && deal.status === SwapDealStatus.SwapAgreed,
+            'AmountSent can only be set after SwapRequested (taker) or SwapAgreed (maker)');
+        break;
+      case SwapDealStatus.AmountReceived:
+        assert(deal.status === SwapDealStatus.AmountSent, 'AmountReceived can be only be set after AmountSent');
+        break;
+      case SwapDealStatus.SwapDone:
```
sangaman: I'd prefer a status called `SwapComplete` - to match the packet name - but it can stay `SwapDone` if you prefer that.

lib/swaps/Swaps.ts
```diff
@@ -141,218 +142,284 @@ class Swaps extends EventEmitter {
...
+        assert(deal.status === SwapDealStatus.AmountReceived, 'SwapDone can be only be set after AmountReceived');
+        break;
+      default:
+        assert(false, 'unknown deal status');
+    }
+
+    deal.status = newStatus;
+    switch (deal.status) {
```
sangaman: Why two switch statements? This logic can be combined with the switch statement above I believe without any change in behavior.

lib/swaps/Swaps.ts
```diff
@@ -1,67 +1,68 @@
...
 
 type SwapDeal = {
   /** The role of the local node in the swap. */
   myRole: SwapDealRole;
-  /** The xud node pub key of the counterparty to this swap deal. */
-  peerPubKey: string;
+  /** The most updated deal status */
+  status: SwapDealStatus;
+  /** The most updated deal state. Works together with status. */
+  state: SwapDealState;
```
sangaman: Having both `state` and `status` is confusing to me, can you combine the logic into a single field or at least explain in comments specifically what they represent and how they are used differently? It seems right now state is only being used for setting the error state - can we just have a `SwapError` status instead?

lib/swaps/Swaps.ts
```diff
@@ -141,218 +142,284 @@ class Swaps extends EventEmitter {
...
     }
 
-    const responsePacket = new packets.SwapResponsePacket(responseBody, requestPacket.header.id);
+  }
 
-    this.logger.debug('Sending swap response to peer: ' + JSON.stringify(responseBody));
+  private setDealState = (deal: SwapDeal, newState: SwapDealState, newStateReason: string): void => {
+    deal.state = newState;
+    deal.stateReason = newStateReason;
+    this.emit('swap.failed', deal);
```
sangaman: Do we want to emit `swap.failed` every time the deal state is updated? If `state` is only used to track the error status, maybe we just rename the field to `errored` and this method to `setError`?

lib/swaps/Swaps.ts
offerm: change to `swapCompleted`

lib/swaps/Swaps.ts
offerm: Yes, it can be combined but I don't see what is wrong with two switch statements. 

The first switch is more for logic validation.

Do see a reason to combine these two switches.

lib/swaps/Swaps.ts
offerm: State and Status are not the same. 
The idea is that know that the deal is in error state and at which phase it was while having the error.  If we set SwapError status we lose the deal status at the time of the error. 
I updated the comment.

lib/swaps/Swaps.ts
offerm: Currently, there is only an Error state. So we emit every time (actually once) this is set.

In the future, we may have other states and once we add these we will also update this code. 

I see no reason to change it.

offerm:
@sangaman updated where agreed.

Please merge so I can continue with other related issues

lib/swaps/Swaps.ts
sangaman: Because it adds extra comparisons and I think makes the code unnecessarily verbose and harder to follow - requiring us to look in two separate places to see what happens for each case. It's also just not a pattern I'm used to seeing anywhere.

lib/swaps/Swaps.ts
sangaman: OK that's what I was thinking too and I see the value in preserving where we were in the swap flow (the status), but I still think the naming is confusing here given that state and status are synonyms.  But will elaborate in the other comment regarding the `setDealState` method.

lib/swaps/Swaps.ts
sangaman: What other states might we have? Unless there's something we'd need to handle in the foreseeable future, I'd prefer that we deal with it when/if a new "state" is actually needed. Since this is only being used to indicate whether a swap has errored or not, I think a boolean `errored` field and `errorReason` (instead of `stateReason`, which likewise is only being used to indicate why a swap failed) makes more sense. If there's another state to handle then it's a different story, but I'm not thinking of anything besides "had an error" and "haven't had an error" that's needed for our purposes.

About the `emit` firing every time, my main concern there is that if we were to have other states and someone called `setDealState()` they could unintentionally emit an error event. It also just doesn't follow from the method name. But really my preference would just be to rename this to something like `setError` - at which point emitting `swap.failed` every time makes perfect sense.

lib/swaps/Swaps.ts
offerm: Ok. I merged the two so you will have it as you like and used to.

Now we have an ugly `if` after the assignment of the new `status` to check if `swap.paid` should be emitted.  This can't get into the switch since the `swap.paid` listeners may execute before the `deal .status` changed (after the switch).

I hope you like it like this better. I don't but don't consider this important at all.

lib/swaps/Swaps.ts
offerm: Changed the code to make sure that `swap.failed` emitted only when setting the state to `Error` (which can happend only once).

We actually have 3 states for the swapdeal. Active, Error, and Completed. With these we can expose a public method named `isFinalState` which can be used to check if all activities ended related to the `swap` or if we are expecting more `events`.

we already discussed about `removing objects` like `swapdeal` and `order`. This should not be done. These can be removed from memory if backed to the storage knowing that you can always fetch them back if needed. As such you need to have a mechanism to idenfify active swaps (and orders) from completed/errored ones. The state is for that.

This is relevant to `subscribedOrders` RPC call and buy/sell with wait option. Will add a comment there (#123) shortly.

lib/swaps/Swaps.ts
```diff
@@ -349,59 +349,55 @@ class Swaps extends EventEmitter {
...
-        this.logger.debug('Amount received for preImage ' + deal.r_preimage);
-        this.emit('swap.paid', deal);
-        break;
-      case SwapDealStatus.SwapCompleted:
-        deal.competionTime = Date.now();
-        this.logger.debug('Swap completed. preimage = ' + deal.r_preimage);
-        break;
+
+    if (deal.status === SwapDealStatus.AmountReceived) {
+      this.emit('swap.paid', deal);
```
sangaman: This can also go in the `case SwapDealStatus.AmountReceived:` logic right, or is there a reason it's separate? Again my thinking is just to have all logic in one place and not make more checks than needed.

lib/types/enums.ts
```diff
@@ -24,6 +24,7 @@ export enum SwapDealStatus {
 }
 
 export enum SwapDealState {
-  Normal = 0,
+  Active = 0,
   Error = 1,
+  Completed = 2,
```
sangaman: Is there a reason we'd want a "completed" state *and* a completed status? It seems like it's not telling us anything different from `SwapDealStatus.SwapCompleted` and I don't think we'd want the redundancy.

lib/swaps/Swaps.ts
sangaman: Just saw this. I don't think the `if` after the assignment is necessary. This method will finish executing before the listener executes.

lib/swaps/Swaps.ts
sangaman: Also didn't see until after reviewing the latest changes, I made a comment about this below, over redundnancy between status completed and state completed.

As for removing swap deals, I agree w/ removing from memory but persisting to the database. For orders we are deliberately not persisting them to the database *at all*, that should be a separate discussion from swaps if you think that should change though.

lib/types/enums.ts
offerm: As I wrote, a state is more `external` while status is internal and deals with the lifecycle of the object. I think we should implement a `state` also in other objects (mainly `order`).  It can be that state and status have similar value.

offerm:
Maybe. Maybe today. Maybe even not today.
But if we add something between the switch and the assignment,  that is
blocking for some reason this may break.  When this happened to trace it
would be a nightmare.
So, the code may look nicer with one switch but clearly not robust.

lib/swaps/Swaps.ts
offerm: Maybe. Maybe today. Maybe even not today.
But if we add something between the switch and the assignment,  that is
blocking for some reason this may break.  When this happened to trace it
would be a nightmare.
So, the code may look nicer with one switch but clearly not robust.

I don't see swap different from orders. swap is a mechanism which is part of order execution. IMHO both must be persistent.

lib/swaps/Swaps.ts
offerm: A possible race between the listener and the code that change the status.
Maybe low risk today but if some code added in the future this may break.

lib/types/enums.ts
sangaman: Ok, what do you think about renaming `status` to something like `phase` (and `SwapDealPhase`). That I think would more clearly differentiate it from `state`.

offerm: No objection.

sangaman: OK cool let's go with that. I'm just looking into that possible race condition you mentioned, then should be good to go.

lib/swaps/Swaps.ts
sangaman: OK, if you want to leave it this way to clearly set the status before any listeners are informed that's fine. We *shouldn't* even be looking at the status from listeners - if we receive `swap.paid` we already know that it's been paid - and as you said this is supposed to be more of an internal indicator, but this way doesn't hurt. So I'd say either way is fine, up to you.

sangaman:
Tests are failing to compile now, can you took a look at these @offerm 

```
lib/p2p/packets/types/index.ts (12,71): Cannot find module './SwapCompletePacket'. (2307)
lib/p2p/packets/types/index.ts (13,65): Cannot find module './SwapErrorPacket'. (2307)
```

PR #469, opened 2018-09-13
Automatically create sample config file

### LINKED ISSUE #335: Add 'sample-xud.conf'
This issue is about auto-generating a `sample-xud.conf` file in the repo (e.g. root folder) with all available options (on every build). This sample config file is NOT used by XUD in any way. We can link to it from the Configuration wiki and from any other place. The reason we want this is, because like this we can keep the sample config option always up-to-date.

rsercano:
Hello,

I've created a new npm script with name `createExampleConfig`, it basically calls an exported method (`convertConfigToToml`) from `Config.ts`

The method `convertConfigToToml` constructs a new `Config` object and then recursively converts it into `TOML` file format. Currently it only excludes function types, if we need to exclude anything from config file such as `instanceId` please let me know, I can improve the code to exclude requested fields.

Closes #335


lib/Config.ts
```diff
@@ -150,4 +150,36 @@ class Config {
   }
 }
 
+export const convertConfigToToml = () => {
```
moshababo: I think it would be better to have this in our `tasks` folder, and add it to gulp.
It will also make the npm script more straightforward, and not relying on a previous compilation.

lib/Config.ts
```diff
@@ -150,4 +150,36 @@ class Config {
...
+      result += value[0] + ' = ' + value[1] + '\n';
+    });
+    nestedPairs.forEach((value: any) => {
+      recursivelyConvertJsonToToml(value[1], (prefix === null || prefix === '') ? value[0] : [prefix, value[0]].join('.'));
+    });
+  };
+
+  recursivelyConvertJsonToToml(new Config(), '');
+
+  fs.writeFileSync('../xud.conf', result);
```
moshababo: Do we want to add this file to our source repo? if not, let's add it to `.gitignore`

lib/utils/utils.ts
```diff
@@ -142,3 +142,17 @@ export const ms = (): number => {
 export const derivePairId = (pair: Pair) => {
   return `${pair.baseCurrency}/${pair.quoteCurrency}`;
 };
+
+export const isPlainObject = (obj: any) => {
```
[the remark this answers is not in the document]
sangaman: Agreed, some comments explaining what this does and why some of these checks are being performed would be nice because it wasn't obvious to me either.

lib/Config.ts
sangaman: Yeah I'm thinking it would make sense not to have it in source given we have this script. It'd be a bit of a hassle to update the file in source each time the default config changes, and if anyone wants this for themselves they can just run the script.

sangaman: Elsewhere we're using `path.join` and `__dirname` for getting relative paths so let's stick to that if possible, like `path.join(__dirname, '..', 'xud.conf');`

package.json
```diff
@@ -8,6 +8,7 @@
     "xucli": "./bin/xucli"
   },
   "scripts": {
+    "createExampleConfig":"cd dist && node -e 'require(\"./Config\").convertConfigToToml()'",
```
sangaman: `createExampleConfig` is a lot to type, I think just `config` is fine.

moshababo:
@sangaman did you compile the project before?

sangaman:
Yes I did, did it work for you?

moshababo:
Yes.

lib/utils/utils.ts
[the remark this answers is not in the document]
rsercano: Sure, will do tomorrow !

lib/Config.ts
rsercano: Will check that tomorrow too, thanks for the comment

rsercano:
I've implemented all requested changes, removed `npm` script, and added a `gulp` task. But current gulp tasks were wrong and preventing me to execute `gulp "config.create"` so I commented out all gulp tasks.

tasks/config/sample.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (!json) return;
+
+    Object.keys(json).sort().forEach((key: any) => {
+      if (Object.prototype.toString.call(json[key]) !== '[object Function]') {
+        (isPlainObject(json[key]) ? nestedPairs : simplePairs).push([key, json[key]]);
+      }
+    });
+
+    if (!(prefix === '' || simplePairs.length === 0)) {
+      result += '[' + prefix + ']\n';
```
sangaman: I believe here is where we'd want to add a blank line (per my other comment above)?

lib/utils/utils.ts
```diff
@@ -142,3 +142,33 @@ export const ms = (): number => {
 export const derivePairId = (pair: Pair) => {
   return `${pair.baseCurrency}/${pair.quoteCurrency}`;
 };
+
+/**
+ * A simplified copy of lodash's isPlainObject;
+ *
+ * Checks is the object is a plain object;
+ * A plain object's;
```
sangaman: I think more clear to say "A plain object is:"

Otherwise great comment block!

sample-xud.conf
```diff
@@ -0,0 +1,42 @@
+initDb = true
```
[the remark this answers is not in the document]
rsercano: Both works for me, changing it back, I actually agree with you on this

sample-xud.conf
```diff
@@ -0,0 +1,42 @@
...
+instanceId = 0
+logLevel = debug
+logPath = /home/sercan/.xud/logs/xud.log
+xudir = /home/sercan/.xud/
+[db]
```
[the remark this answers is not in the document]
rsercano: consider it done

gulpfile.ts
```diff
@@ -1,7 +1,8 @@
...
 import tasks from './tasks';
 
-gulp.task('db.dropTables', () => tasks.db.dropTables()) ;
-gulp.task('db.init', () => tasks.db.init());
-gulp.task('db.restart', () => tasks.db.restart());
-gulp.task('db.truncate', () => tasks.db.truncate());
+// gulp.task('db.dropTables', () => tasks.db.dropTables()) ;
+// gulp.task('db.init', () => tasks.db.init());
+// gulp.task('db.restart', () => tasks.db.restart());
+// gulp.task('db.truncate', () => tasks.db.truncate());
```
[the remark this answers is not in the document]
rsercano: done

tasks/db/dropTables.ts
```diff
@@ -1,3 +1,4 @@
+/*
```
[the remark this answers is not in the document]
rsercano: done

tasks/index.ts
```diff
@@ -1,5 +1,7 @@
-import * as db from './db';
+// import * as db from './db';
```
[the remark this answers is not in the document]
rsercano: done

tasks/config/sample.ts
rsercano: indeed

lib/utils/utils.ts
rsercano: done

rsercano:
Adding `npm script` back too additionally `gulp` task as requested by @sangaman

tasks/config/sample.ts
```diff
@@ -0,0 +1,36 @@
...
+        (isPlainObject(json[key]) ? nestedPairs : simplePairs).push([key, json[key]]);
+      }
+    });
+
+    if (!(prefix === '' || simplePairs.length === 0)) {
+      result += '\n[' + prefix + ']\n';
+    }
+
+    simplePairs.forEach((value: any) => {
+      result += value[0] + ' = ' + value[1] + '\n';
```
sangaman: I think we need to handle two special cases here. The simpler one is if `typeof value[1] === 'string'` then we should surround `value[1]` with `""` in the result, because TOML expects strings to be in quotation marks.

Then we also need to check if the value is an array, in which case we need to use `[]`. We only have one array config property, `p2p.addresses` and it defaults to an empty array. I don't think we'll ever have default values in an array, so if you just want to detect if a value is an array using `isArray()` or something and then output `[]` that should be fine.

lib/Config.ts
sangaman: Sorry for the confusion over keeping the file in the repo per #335, I wasn't considering the ease with which someone who wanted it could generate it for themselves.

tasks/config/sample.ts
rsercano: done

rsercano:
Changed as you described @sangaman

kilrau:
Updated https://github.com/ExchangeUnion/xud/wiki/Configuration

sangaman:
Oh @kilrau we went with not actually keeping the sample config in the repo, because it can easily be generated by the user. Do you want to update the wiki to say that a sample config file can be generated by running `npm run config`? Otherwise I can update it later.

kilrau:
Updated the wiki for now, but don't see the downside of just generating it. Less work for the user. Could you explain?

sangaman:
My thinking was just so we don't have to maintain the file in source, having to update the file each time the default config changes and risking letting it become outdated.

Although one option I just thought of would be to run the script automatically when we publish to npm, and include the sample config among the published files. That ensures it will always be updated, and provides it pre-made to end users.

kilrau:
> Although one option I just thought of would be to run the script automatically when we publish to npm, and include the sample config among the published files. That ensures it will always be updated, and provides it pre-made to end users.

That sounds like a good option! Can we do that?

sangaman:
Yup I'll take care of it.

sangaman:
@kilrau see #489.

Issue #457, opened 2018-09-11
Track node reputation events & score


sangaman:
Offshoot of #185. We want to track node behavior scores for reputation and banning purposes.

- Track reputation events and record each instance of a negative event, including the type of event and time of occurrence.
- Track positive reputation events as well - at this point I'm thinking this would consist of successfully completed swaps/trades, maybe we could also award points for sustained uptime with low pings or something along those lines. I think it would be excessive to record every swap strictly for reputation purposes, although maybe if we track each successful swap/trade for historical/info purposes we could simply use that same data for calculating reputation.
- Have hardcoded (or maybe configurable) scores assigned to each event.
- On startup, load the reputation events for each unbanned node and calculate a score. For example, if timing out on a packet request is -10 points, and completed swaps are each +1 point, a peer with 2 timeouts and 5 swaps would have a score of -15 points.
- If a peer's score falls below a certain threshold, we ban the peer and notify the peer of the negative events that caused us to ban them.
- We can also use this score to determine which nodes get shared in response to a `GET_NODES` request.
- We might want to set a "statute of limitations" for reputation events as well. We can periodically purge events older than a certain date (maybe ~1 month), this would have scores based more on recent than historical behavior and also prevent db bloat.

kilrau:
Dependency: https://github.com/ExchangeUnion/xud/issues/336

michael1011:
Closed by #470

Issue #152, opened 2018-06-27
p2p packet: disconnecting


moshababo:
Inform a peer that disconnection is imminent, and specify the reason (enum) for that.
If the reason is that the ban threshold exceeded, send him the penalization history as well.
If received, packet payload should be saved to the database.

kilrau:
Other reasons are stalling (peer became unresponsive, socket closed etc) or shutdown (`xucli shutdown`).

PR #470, opened 2018-09-13
feat(p2p): add reputation events

michael1011:
Closes #470 

This PR implements reputation events as described in issue #470. Those events track the history of the behaviour of a node and are stored in the database.

I decided to move the pruning of old events and informing nodes that they got banned into separate follow up PRs.

### DIFF
--- a/lib/db/models/ReputationEvent.ts
+++ b/lib/db/models/ReputationEvent.ts
@@ -1,0 +1,26 @@
+import Sequelize from 'sequelize';
+import { db } from '../../types';
+
+export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes) => {
+  const attributes: db.SequelizeAttributes<db.ReputationEventAttributes> = {
+    id: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
+    event: { type: DataTypes.INTEGER, allowNull: false },
+    nodeId: { type: DataTypes.INTEGER, allowNull: false },
+  };
+
+  const options: Sequelize.DefineOptions<db.ReputationEventInstance> = {
+    tableName: 'reputationEvents',
+    timestamps: true,
+    updatedAt: false,
+  };
+
+  const ReputationEvent = sequelize.define<db.ReputationEventInstance, db.ReputationEventAttributes>('ReputationEvent', attributes, options);
+
+  ReputationEvent.associate = (models: Sequelize.Models) => {
+    models.ReputationEvent.belongsTo(models.Node, {
+      foreignKey: 'nodeId',
+    });
+  };
+
+  return ReputationEvent;
+};
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -1,17 +1,39 @@
+import { EventEmitter } from 'events';
 import P2PRepository from './P2PRepository';
 import { NodeInstance, NodeFactory } from '../types/db';
 import { Address } from '../types/p2p';
 import addressUtils from '../utils/addressUtils';
+import { ReputationEvent } from '../types/enums';
+
+const reputationEventWeight = {
+  [ReputationEvent.ManualBan]: Number.NEGATIVE_INFINITY,
+  [ReputationEvent.ManualUnban]: 0,
+  [ReputationEvent.PacketTimeout]: -1,
+  [ReputationEvent.SwapFailure]: -10,
+  [ReputationEvent.SwapSuccess]: 1,
+};
+
+// TODO: inform node about getting banned
+// TODO: remove reputation events after certain amount of time
+
+interface NodeList {
+  on(event: 'node.ban', listener: (nodePubKey: string) => void): this;
+  emit(event: 'node.ban', nodePubKey: string): boolean;
+}
 
 /** Represents a list of nodes for managing network peers activity */
-class NodeList {
+class NodeList extends EventEmitter {
   private nodes = new Map<string, NodeInstance>();
+
+  private static readonly banThreshold = -50;
 
   public get count() {
     return this.nodes.size;
   }
 
-  constructor(private repository: P2PRepository) {}
+  constructor(private repository: P2PRepository) {
+    super();
+  }
 
   /**
    * Check if a node with a given nodePubKey exists.
@@ -29,13 +51,7 @@ class NodeList {
    * @returns true if the node was banned, false otherwise
    */
   public ban = async (nodePubKey: string): Promise<boolean> => {
-    const node = this.nodes.get(nodePubKey);
-    if (node) {
-      node.banned = true;
-      await node.save();
-      return true;
-    }
-    return false;
+    return this.addReputationEvent(nodePubKey, ReputationEvent.ManualBan);
   }
 
   public isBanned = (nodePubKey: string): boolean => {
@@ -49,8 +65,14 @@ class NodeList {
   public load = async (): Promise<void> => {
     const nodes = await this.repository.getNodes();
 
-    nodes.forEach((node) => {
-      this.nodes.set(node.nodePubKey, node);
+    nodes.forEach(async (node) => {
+      if (!node.banned) {
+        this.nodes.set(node.nodePubKey, node);
+        const events = await this.repository.getReputationEvents(node);
+        events.forEach(({ event }) => {
+          this.updateReputationScore(node, event);
+        });
+      }
     });
   }
 
@@ -67,7 +89,7 @@ class NodeList {
    * Update a node's addresses.
    * @return true if the specified node exists and was updated, false otherwise
    */
-  public updateAddresses = async (nodePubKey: string, addresses: Address[] = [], lastAddress?: Address) => {
+  public updateAddresses = async (nodePubKey: string, addresses: Address[] = [], lastAddress?: Address): Promise<boolean> => {
     const node = this.nodes.get(nodePubKey);
     if (node) {
       // avoid overriding the `lastConnected` field for existing matching addresses unless a new value was set
@@ -91,6 +113,36 @@ class NodeList {
     return false;
   }
 
+  /**
+   * Add a reputation event to the node's history
+   * @return true if the specified node exists and the event was added, false otherwise
+   */
+  public addReputationEvent = async (nodePubKey: string, event: ReputationEvent): Promise<boolean> => {
+    const node = this.nodes.get(nodePubKey);
+    if (node) {
+      const promises: PromiseLike<any>[] = [
+        this.repository.addReputationEvent({ event, nodeId: node.id }),
+      ];
+
+      this.updateReputationScore(node, event);
+
+      if (node.reputationScore < NodeList.banThreshold) {
+        promises.push(this.setBanStatus(node, true));
+        this.emit('node.ban', nodePubKey);
+      } else if (node.banned) {
+        // If the reputationScore is not below the banThreshold but node.banned
+        // is true that means that the node was unbanned
+        promises.push(this.setBanStatus(node, false));
+      }
+
+      await Promise.all(promises);
+
+      return true;
+    }
+
+    return false;
+  }
+
   public removeAddress = async (nodePubKey: string, address: Address) => {
     const node = this.nodes.get(nodePubKey);
     if (node) {
@@ -108,6 +160,22 @@ class NodeList {
     }
 
     return false;
+  }
+
+  private updateReputationScore = (node: NodeInstance, event: ReputationEvent) => {
+    switch (event) {
+      case (ReputationEvent.ManualBan):
+      case (ReputationEvent.ManualUnban): {
+        node.reputationScore = reputationEventWeight[event];
+        break;
+      }
+      default: node.reputationScore += reputationEventWeight[event];
+    }
+  }
+
+  private setBanStatus = (node: NodeInstance, status: boolean) => {
+    node.banned = status;
+    return node.save();
   }
 }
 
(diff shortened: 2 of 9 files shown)


lib/p2p/NodeList.ts
```diff
@@ -1,17 +1,38 @@
...
 import { NodeInstance, NodeFactory } from '../types/db';
-import { Address } from '../types/p2p';
+import { Address, ReputationEvent } from '../types/p2p';
 import addressUtils from '../utils/addressUtils';
+import { ReputationEventType } from '../types/enums';
+import { ms } from '../utils/utils';
+import { EventEmitter } from 'events';
+
+const ReputationEventWeight = {
+  [ReputationEventType.ManualBan]: -100000000,
```
sangaman: It might be good to use `Number.NEGATIVE_INFINITY` for this.

lib/db/models/Node.ts
```diff
@@ -20,7 +20,24 @@ export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes)
         }
       },
     },
-    banned: { type: DataTypes.BOOLEAN, allowNull: true },
```
sangaman: I also think we can keep the banned flag on the nodes table. I figure this can be an easy way to check if a node is banned without even having to load and calculate their reputation upon startup.

lib/p2p/NodeList.ts
```diff
@@ -100,6 +138,20 @@ class NodeList {
...
+    let score = 0;
+
+    reputationEvents.forEach((occurrence) => {
+      score += ReputationEventWeight[occurrence.type];
+    });
+
+    return score;
+  }
+
+  private nodeIsBanned = (node: NodeInstance) => {
```
sangaman: If we keep the `banned` db field, we shouldn't need this method,

lib/p2p/NodeList.ts
```diff
@@ -1,17 +1,38 @@
...
+
+interface NodeList {
+  on(event: 'node.ban', listener: (nodePubKey: string, events: ReputationEvent[]) => void): this;
+  emit(event: 'node.ban', nodePubKey: string, events: ReputationEvent[]): boolean;
+}
 
 /** Represents a list of nodes for managing network peers activity */
-class NodeList {
+class NodeList extends EventEmitter {
+  private banThreshold = -50;
```
sangaman: This should be a static field

lib/p2p/NodeList.ts
```diff
@@ -50,6 +65,7 @@ class NodeList {
     const nodes = await this.repository.getNodes();
 
     nodes.forEach((node) => {
+      node.reputationScore = this.calculateReputationScore(node.reputationEvents);
```
sangaman: I'm thinking we only need to do this for nodes that aren't already banned.

lib/p2p/NodeList.ts
```diff
@@ -87,6 +103,28 @@ class NodeList {
     return false;
   }
 
+  /**
+   * Add a reputation event to the nodes history
+   * @return true if the node was banned, false otherwise
+   */
+  public addReputationEvent = async (nodePubKey: string, type: ReputationEventType) => {
```
sangaman: If we go ahead with renaming `ReputationEventType` to `ReputationEvent`, then we can rename `type` to `event`.

lib/p2p/NodeList.ts
```diff
@@ -1,17 +1,38 @@
...
 import addressUtils from '../utils/addressUtils';
+import { ReputationEventType } from '../types/enums';
+import { ms } from '../utils/utils';
+import { EventEmitter } from 'events';
+
+const ReputationEventWeight = {
+  [ReputationEventType.ManualBan]: -100000000,
+  [ReputationEventType.PacketTimeout]: -5,
+  [ReputationEventType.SwapTimeout]: -15,
+  [ReputationEventType.SwapSuccess]: 1,
```
sangaman: I think eventually we'll want something like a `MANUAL_UNBAN` or `MANUAL_CONNECT` event for when we intentionally reconnect to a peer we have already banned. That should maybe be a special case where we reset the score to 0, rather than adding or subtracting. Just a thought.

lib/types/p2p.ts
```diff
@@ -21,7 +23,7 @@ export type HandshakeState = {
   lndltcPubKey?: string;
 };
 
-export function isHandshakeState(obj: any): obj is HandshakeState {
```
sangaman: I know we don't use this method currently but I *think* we may use it when it comes to validating that a HELLO packet we've received is valid, so I don't think I'd remove it yet.

lib/p2p/NodeList.ts
```diff
@@ -87,6 +103,28 @@ class NodeList {
...
+   * Add a reputation event to the nodes history
+   * @return true if the node was banned, false otherwise
+   */
+  public addReputationEvent = async (nodePubKey: string, type: ReputationEventType) => {
+    const node = this.nodes.get(nodePubKey);
+    if (node) {
+      node.reputationScore += ReputationEventWeight[type];
+
+      if (this.nodeIsBanned(node)) {
+        this.emit('node.ban', nodePubKey, node.reputationEvents);
```
sangaman: I figure we should only be sending events with negative scores for the ban reasons.

lib/p2p/NodeList.ts
```diff
@@ -87,6 +103,28 @@ class NodeList {
...
+      }
+
+      node.reputationEvents = [...node.reputationEvents, { type, date: ms() }];
```
sangaman: I find this syntax a little unusual for adding an element to an array vs just using `push()`. And out of curiosity I ran some benchmarks and using `push` is also much much more performant, I'm guessing that this way is cloning the `reputationEvents` array each time. http://jsben.ch/JpVcR

sangaman:
Oh and commit message is good I believe. You could maybe mention #457 in it, I believe this should close it right? Thanks!

lib/types/enums.ts
```diff
@@ -4,6 +4,13 @@ export enum SwapClients {
   RAIDEN,
 }
 
+export enum ReputationEventType {
+  ManualBan = 'manualBan',
```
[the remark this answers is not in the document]
michael1011: What about adding new events though? When you add a new field to an enum those automatic 0, 1, 2, 3 move accordingly. Adding, removing or just changing the position of a single event is a breaking change that way.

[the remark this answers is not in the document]
sangaman: That's true. If we add new events to the end of the list of enums then we don't interfere with the existing ones, BUT if we ever delete an event it could mess things up, which I hadn't really considered. I think then maybe the best approach would be to explicitly assign each enum an integer value. `MANUAL_BAN = 0` and so on... That should prevent unintentional breaking changes, but also give us flexibility in terms of being able to rename events plus reducing the amount of data we need to store in the database for each event.

sangaman:
This PR (or a follow up to this one) should look at addressing #152.

lib/types/enums.ts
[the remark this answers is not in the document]
michael1011: I am still in favor of using `MANUAL_BAN = 'manualBan'` instead of a number because I think the few bytes more to store are negligible and if we ever remove an event without adding a new one we skip a number which looks messy. Having to change the `manualBan` is very unlikely.

I will implement it the way you described it but I am not 100% convinced of that approach.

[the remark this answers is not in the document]
michael1011: Also should the `SwapDealPhase` isn't using UPPER_SNAKE_CASE. Should I update that enum in this PR?

lib/p2p/NodeList.ts
michael1011: Why `static`? Right now this variable is just used in the instance of `NodeList`. Setting it to `readonly` would make sense

lib/types/enums.ts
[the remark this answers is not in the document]
sangaman: So as I see it, it's true that saving bytes in the db is not a big deal. Aside from that the advantage of string-valued enums is that there's no risk of removed events leading to out-of-sequence enum values (0, 1, 3, 4, etc..), which I also don't think would be a deal breaker since the code puts these values behind the scenes but I agree it's not ideal. The advantage of number-valued enums is we can rename the enums internally. I'd guess that we're probably not going to be either renaming reputation events nor removing them all that often, so it's kind of a wash. I'd still lean to using numbers although I suppose either way is fine. Maybe somebody else can chime in.

lib/p2p/NodeList.ts
sangaman: I think `static` makes more sense because it's an unchanging value that pertains to the class itself (e.g. all `NodeList`s should use this value) rather than a particular instance of `NodeList` (granted we're only using one). `readonly` from what I understand is for member fields that should be initialized once during the constructor and then not changed again.

lib/p2p/NodeList.ts
michael1011: After thinking a bit about this: why should we bother telling a node why we banned them? The banned node isn't important anymore and should just be informed about the fact that it got banned.

sangaman: @moshababo can probably answer this better than I can, but I think it's useful for debugging/testing purposes and also so that a node that was banned for honest reasons (something like bad connectivity) they can know what the problem is and possibly correct it. How would a peer know why it got banned without us telling it?

lib/p2p/NodeList.ts
```diff
@@ -87,6 +111,43 @@ class NodeList {
...
+        default: node.reputationScore += reputationEventWeight[event];
+      }
+
+      if (this.nodeIsBanned(node)) {
+        promises.push(this.setBanStatus(node, true));
+        this.emit('node.ban', nodePubKey);
+      } else if (node.banned) {
+        // If the reputationScore is not below the banThreshold but node.banned
+        // is true that means that the node was unbanned
+        promises.push(this.setBanStatus(node, false));
```
sangaman: Does it make sense to also set `node.banned` to false here?

lib/p2p/NodeList.ts
```diff
@@ -100,6 +161,15 @@ class NodeList {
...
     return false;
   }
+
+  private setBanStatus = (node: NodeInstance, status: boolean) => {
+    node.banned = status;
+    return node.save();
+  }
+
+  private nodeIsBanned = (node: NodeInstance) => {
+    return node.reputationScore < NodeList.banThreshold;
```
sangaman: Unless you envision using this check elsewhere in the future, I'd just put the logic directly in the `addReputationEvent` method above. The logic is simple and just a line, so I don't see a need to have a method for it if it's only used once.

lib/p2p/NodeList.ts
```diff
@@ -49,8 +65,16 @@ class NodeList {
   public load = async (): Promise<void> => {
     const nodes = await this.repository.getNodes();
 
-    nodes.forEach((node) => {
-      this.nodes.set(node.nodePubKey, node);
+    nodes.forEach(async (node) => {
+      if (!node.banned) {
+        this.nodes.set(node.nodePubKey, node);
+        const events = await this.repository.getReputationEvents(node);
+        events.forEach((event) => {
```
sangaman: If you make this `forEach(({ event }) => {` then below you can just use `event` and not the slightly awkward `event.event`.

lib/types/db.ts
```diff
@@ -41,3 +44,16 @@ export type PairAttributes = PairFactory & {
...
+  event: ReputationEvent;
+  nodeId: number;
+};
+
+export type ReputationEventAttributes = ReputationEventFactory & {
+  id: number;
+};
+
+export type ReputationEventInstance = ReputationEventAttributes & Sequelize.Instance<ReputationEventAttributes> & {
+  createdAt: number;
```
sangaman: Do we need this here? Or is it fine to just leave `createdAt` in the database, retrievable by query if we need it. I believe that's similar to our approach with `Nodes`.

michael1011:
I did test the events a little bit manually and it worked pretty well. Should I add some tests (would those be p2p or integration ones) in this PR or will a followup one be sufficient?

sangaman:
Some tests would be good, although since we're not penalizing for most things yet I figure it would have to be pretty limited. Maybe you can test just the basic manual ban in the p2p tests?

lib/p2p/NodeList.ts
michael1011: `this.setBanStatus` is doing that and also saving that change to the database

lib/types/db.ts
michael1011: If we want to prune old events on startup I would leave it as is in order to get all needed information in one query.

lib/db/models/ReputationEvent.ts
```diff
@@ -0,0 +1,26 @@
...
+  const options: Sequelize.DefineOptions<db.ReputationEventInstance> = {
+    tableName: 'reputationEvents',
+    timestamps: true,
+    updatedAt: false,
+  };
+
+  const ReputationEvent = sequelize.define<db.ReputationEventInstance, db.ReputationEventAttributes>('ReputationEvent', attributes, options);
+
+  ReputationEvent.associate = (models: Sequelize.Models) => {
+    models.ReputationEvent.belongsTo(models.Node, {
```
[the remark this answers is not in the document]
michael1011: I get what you are saying but I couldn't get it working the way you described it. Removing `foreignKey` resulted in:
```
error: SequelizeDatabaseError: SQLITE_ERROR: duplicate column name: NodeId
```

and removing the `nodeId` attribute in the value no being accessible at all.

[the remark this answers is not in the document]
sangaman: Never mind then, looks like it's just not well documented.

michael1011:
@sangaman although the commit history looks messed up here you can see that I added [just a single commit](https://github.com/ExchangeUnion/xud/pull/470/commits) in the `reputation-events` branch

PR #481, opened 2018-09-14
refactor(tasks): remove obsolete db tasks

sangaman:
Removes obsolete, outdated db tasks that are not necessary after the migration to sqlite from mysql.

With sqlite, we can reset the database simply by moving or deleting the `xud.db` file. I left the tasks and db folders in place in case we want to have other db-related tasks in the future.


rsercano:
@sangaman this will actually conflict with #469, if you wish I can add these changes into #469 ?

sangaman:
Oh I did not realize that #469 touched these files. They should be deleted I think, not commented out, since I don't expect we'll ever be using them again. Normally I'd say to keep them separate, better for PRs to focus on a single concern, but since the gulp tasks won't work with these files I'll close this and comment in the other PR.

PR #483, opened 2018-09-14
chore: let Travis compile TypeScript

michael1011:
This commit lets Travis compile TypeScript to JavaScript from now on. I
think this makes sense because we are not compiling when
installing anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ cache:
     - "node_modules"
 
 script:
+  - npm run compile
   - npm run lintNoFix
   - npm run test
 


sangaman:
I'm not sure this is actually needed, and it might be better that we're no longer explicitly compiling in travis. I double checked, and a compilation error in our source code will cause the tests to fail with an error like `TSError: Γ¿» Unable to compile TypeScript`. Since compilation errors are detected as is, is there a reason to add in a compile step?

michael1011:
I am not sure whether we actually need to compile in Travis. Is ts-node compiling the whole source or just the needed parts when executing the tests? @sangaman 

If you are that we don't need it just close this PR.

michael1011:
You are right we will probably never catch a compilation error that way since our implementation tests *should* import all files in one way or another.

I would say just close this PR and delete the branch.

PR #489, opened 2018-09-17
chore: generate sample config on new version

### LINKED ISSUE #498: 'sample-xud.conf' v2
- auto-update `sample-xud.conf` in github repo on version bump, in root folder
- auto-update `sample-xud.conf` in npm repo

sangaman:
This commit updates the publish configuration to generate a sample config file before publishing and to include it among the published files.

Edit: The new approach generates the sample config file on the `npm version` script and checks it in via git. Closes #498.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,4 @@ proto/xudrpc.proto linguist-vendored=false
 lib/proto/* linguist-generated
 package-lock.json linguist-generated
 docs/api.md linguist-generated
+sample-xud.conf linguist-generated
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,3 @@ dist/
 # typedoc
 typedoc/
 .DS_Store
-
-# example config
-sample-xud.conf
--- a/package.json
+++ b/package.json
@@ -27,7 +27,9 @@
     "test:unit": "mocha -r ts-node/register test/unit/*",
     "test:unit:watch": "mocha -r ts-node/register test/unit/*  --watch --watch-extensions ts",
     "test:p2p": "mocha -r ts-node/register test/p2p/*",
-    "test:p2p:watch": "mocha -r ts-node/register test/p2p/*  --watch --watch-extensions ts"
+    "test:p2p:watch": "mocha -r ts-node/register test/p2p/*  --watch --watch-extensions ts",
+    "preversion": "npm run lintNoFix && npm test && npm run compile",
+    "version": "npm run config && git add sample-xud.conf"
   },
   "cross-os": {
     "postcompile": {
--- a/sample-xud.conf
+++ b/sample-xud.conf
@@ -1,0 +1,65 @@
+# Sample configuration file for xud
+#
+# This sample file contains the default values for all configuration
+# options for xud. Directories and file path options are platform &
+# user specific and are not included, but are explained below.
+#
+# 'xudir' is the directory for data stored by xud including logs,
+# keys, config and its database. Individual paths can be overridden
+# by settings such as 'logpath' and 'dbpath'.
+#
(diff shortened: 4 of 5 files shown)


kilrau:
Agree with @michael1011 , either on publishing or better even travis build (if that's possible since it'd need write permissions)

sangaman:
I'm still leaning against committing it to source, so that we don't have to maintain it by updating it each time the default config changes, nor risk forgetting to update it and having it be out of sync.

Another option is to run `npm run config` after *install* - so that every time you install `xud` whether it be from npm or from github clone or docker, this file will be generated for you and it's guaranteed to be up to date. That seems just as good as having it in source, right? @michael1011

sangaman:
@michael1011 See the new approach per our discussion on the weekly call. I've tested this and it works like a charm, committing any changes to the sample config file along with the version change to package.json and package-lock.json when runing `npm version`.

sangaman:
Ahhhh it did not even occur to me that the default values differ between windows and linux. We can have 3 separate files or maybe figure out a way to get values for all 3 in the same file - it will definitely take some tweaking of the config script itself tho. Right now it will generate it for whatever OS you're running when you generate the file, which is definitely not ideal.

sangaman:
@michael1011 I'm thinking of tweaking the config job a bit to make 3 separate files. `sample-linux-xud.conf`, `sample-win-xud.conf`, and `sample-mac-xud.conf`. Let me know if that sounds good to you.

michael1011:
@sangaman  what do you think of putting the default values in comments above the actual value:
```
# Directory for all data stored by LightningTip (config and log file)
# Gets overwritten by individual settings and flags like: "logfile" and "config"
#
# Default values:
# Darwin (macOS): /Users/<user>/Library/Application Support/LightningTip
# Linux: /home/<user>/.lightningtip
# Windows: C:\Users\<user>\AppData\Local\LightningTip
#
# datadir =
```

Having three separate config files seems a bit messy and overkill to me since only a few values are affected by the OS.

sangaman:
I thought about that but still have some concerns. There are quite a few actually when you count the multiple lnd `macaroonpath`/`tlscertpath` which are not derived from `xudir`. Plus it might be harder to maintain if we add more path-based config variables, my approach would be to look for "path" in the config property name but that wouldn't be 100% reliable.

Maybe I could generate only a linux file - since I figure that will be the most commonly used platform - and then put a preamble in the header similar to the one you've posted and list the various `xudir` values as well as the default directory for `lnd`. What do you think of that?

kilrau:
I think having 3 separate samples are the best option since this doesn't require the user to think much. Don't see where it's messy. I'm ok with it. @sangaman @michael1011

michael1011:
Yeah if you don't think it is worth the effort to put everything into one config I am also fine with one for every OS (but don't forget to run the sample config script on a Linux and Mac VM before you publish the NPM package). 

One thing that came just to my mind: the sample config would contain the user name of the one publishing the NPM package. Is that intended by you @sangaman?

sangaman:
> One  thing that came just to my mind: the sample config would contain the user name of the one publishing the NPM package. Is that intended by you @sangaman?

Nope it's not - that's another thing I overlooked. Will definitely have to make all user-related info generic.

sangaman:
Take a look at this again when you can, it has a single config file without any of the default paths, but with comments explaining the key path options and their defaults based on the platform.

kilrau:
since you use `<user>` for Linux and MacOS, also do this for Win and not `Daniel` in the description:

`# macaroonpath = "C:\Users\Daniel\AppData\Local\Lnd\data\chain\<currency`

After this fine for me.

sangaman:
Hah thanks for catching that @kilrau, thought I'd replaced it everywhere. Will fix and merge, thanks!

PR #492, opened 2018-09-17
refactor: remove maxResults from getOrders

sangaman:
This PR removes `maxResults` and `limit` from the `getOrders` rpc call and all related internal calls. The reason for this is because orders are not ordered according to price internally - limiting the
number of orders returned simply picks an arbitrary set of orders rather than the orders with the best price. It is up to the client to sort and discard orders after they've been retrieved from `xud`.

I left a `max_results` argument on the cli call for now in case we want the cli itself to limit the number of orders displayed, which seems like a desirable feature.

### DIFF
--- a/lib/cli/commands/getorders.ts
+++ b/lib/cli/commands/getorders.ts
@@ -2,7 +2,7 @@ import { Arguments } from 'yargs';
 import { callback, loadXudClient } from '../command';
 import { GetOrdersRequest } from '../../proto/xudrpc_pb';
 
-export const command = 'getorders [pair_id] [max_results]';
+export const command = 'getorders [pair_id]';
 
 export const describe = 'get orders from the order book';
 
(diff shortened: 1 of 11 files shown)


lib/cli/commands/getorders.ts
```diff
@@ -12,7 +12,7 @@ export const builder = {
   },
   max_results: {
```
michael1011: This should be removed.

michael1011: I think we remove it for now if it is not used.

sangaman: That's OK with me.

sangaman:
We're not returning them sorted because we don't keep them sorted within xud, and keeping them sorted would have meaningful performance implications. I figure since the sorting is only needed for display by clients, we should leave the sorting up to clients.

PR #493, opened 2018-09-17
test: prevent loading local config

sangaman:
This commit prevents the tests from loading the local configuration file. It also prevents loading the local node pub key from disk. This is unnecessary after the sqlite migration since we no longer may depend
on the local config to provide mysql connection info. Rather, errors in the local config file could cause tests to fail, and should no longer be a dependency for tests to pass.

I think a next step would be to stop the p2p simulation from reading & writing its node keys to disk. This could be done a couple of ways, one way would be to store the node key in the database (since tests use temporary databases stored only in memory), which I know we've discussed a bit and maybe should revisit.

### DIFF
--- a/test/integration/OrderBook.spec.ts
+++ b/test/integration/OrderBook.spec.ts
@@ -1,6 +1,5 @@
 import { expect } from 'chai';
 import uuidv1 from 'uuid/v1';
-import Config from '../../lib/Config';
 import DB from '../../lib/db/DB';
 import OrderBook from '../../lib/orderbook/OrderBook';
 import OrderBookRepository from '../../lib/orderbook/OrderBookRepository';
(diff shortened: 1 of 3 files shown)


sangaman:
Normally I'd agree with you re: not storing a single file in a sql database (and with mysql I did feel the same way), but I think the convention is a little different with sqlite.

Their official website brags about their performance improvement with [reading and writing small blobs vs the filesystem](https://www.sqlite.org/fasterthanfs.html). Also this [stack overflow question](https://stackoverflow.com/questions/3010930/is-there-a-way-to-store-a-one-unique-data-element-in-sqlite) seems to cover a similar use case.

We can discuss some more, maybe I'll change my mind again but it seems like a reasonable approach to me.

michael1011:
@sangaman I am cool with storing the private key in the database if it is a common thing to do with SQLite.

Issue #365, opened 2018-08-20
Store an external address if we successfully make an outgoing connection 


kilrau:
Store an external address if we successfully make an outgoing connection even if the other peer doesn't advertise a listening address, add "notAdvertised" or similar in this case -> don't share this address with others

sangaman:
One thought that occurred to me is that if we store the last connected address along with the other listening addresses, we'd be sharing that node and address to others. Maybe we should have a separate column in the table `lastAddress` or something that we don't share to others. We try that address first, if one exists, then if it that doesn't work we try to connect to their advertised addresses.

kilrau:
Agree, we should take care that we don't share any non-advertised addresses with others, but don't you already do this check with `connectedPeer.addresses.length > 0)` in https://github.com/ExchangeUnion/xud/pull/393/files ? @sangaman

kilrau:
Tag last successful address of a peer in separate issue: https://github.com/ExchangeUnion/xud/issues/442

rsercano:
Sure, I want to check that do I need to know anything before I start @moshababo @sangaman ?

moshababo:
Just take some time to go through our usage of `NodeList` / `PeerList` inside of `Pool`. If you have any questions, feel free to hassle.

PR #496, opened 2018-09-18
Storing lastAddress in node for outgoing connection for peer

rsercano:
Hello,

This PR is for #365 

I actually need a hand here, I added `lastAddress` field as we discussed and set it to the `peer.address` since it's the latest connected address inside of `handleOpen` method. 

I just need your opinions here, is my path wrong ? What I think from here is (to-do); try to connect to `lastAddress` before trying any other one. Is there anything else do I need to do ?

**EDIT:** I just added connecting to `lastAddress` first into `tryConnectNode` method.

### DIFF
--- a/lib/db/DB.ts
+++ b/lib/db/DB.ts
@@ -62,14 +62,17 @@ class DB {
         {
           nodePubKey: '02b66438730d1fcdf4a4ae5d3d73e847a272f160fee2938e132b52cab0a0d9cfc6',
           addresses: [{ host: 'xud1.test.exchangeunion.com', port: 8885 }],
+          lastAddress: { host: 'xud1.test.exchangeunion.com', port: 8885 },
         },
         {
           nodePubKey: '028599d05b18c0c3f8028915a17d603416f7276c822b6b2d20e71a3502bd0f9e0a',
           addresses: [{ host: 'xud2.test.exchangeunion.com', port: 8885 }],
+          lastAddress: { host: 'xud2.test.exchangeunion.com', port: 8885 },
         },
         {
           nodePubKey: '03fd337659e99e628d0487e4f87acf93e353db06f754dccc402f2de1b857a319d0',
           addresses: [{ host: 'xud3.test.exchangeunion.com', port: 8885 }],
+          lastAddress: { host: 'xud3.test.exchangeunion.com', port: 8885 },
         },
       ]);
 
--- a/lib/db/DB.ts
+++ b/lib/db/DB.ts
@@ -62,17 +62,14 @@ class DB {
         {
           nodePubKey: '02b66438730d1fcdf4a4ae5d3d73e847a272f160fee2938e132b52cab0a0d9cfc6',
           addresses: [{ host: 'xud1.test.exchangeunion.com', port: 8885 }],
-          lastAddress: { host: 'xud1.test.exchangeunion.com', port: 8885 },
         },
         {
           nodePubKey: '028599d05b18c0c3f8028915a17d603416f7276c822b6b2d20e71a3502bd0f9e0a',
           addresses: [{ host: 'xud2.test.exchangeunion.com', port: 8885 }],
-          lastAddress: { host: 'xud2.test.exchangeunion.com', port: 8885 },
         },
         {
           nodePubKey: '03fd337659e99e628d0487e4f87acf93e353db06f754dccc402f2de1b857a319d0',
           addresses: [{ host: 'xud3.test.exchangeunion.com', port: 8885 }],
-          lastAddress: { host: 'xud3.test.exchangeunion.com', port: 8885 },
         },
       ]);
 
--- a/lib/db/models/Node.ts
+++ b/lib/db/models/Node.ts
@@ -7,6 +7,16 @@ export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes) 
     id: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
     nodePubKey: { type: DataTypes.STRING, unique: true, allowNull: false },
     addressesText: { type: Sequelize.TEXT, allowNull: false },
+    lastAddressText: { type: Sequelize.TEXT, allowNull: false },
+    lastAddress: {
+      type: Sequelize.VIRTUAL,
+      get(this: db.NodeInstance) {
+        return JSON.parse(this.lastAddressText);
+      },
+      set(this: db.NodeInstance, value: Address) {
+        this.setDataValue('lastAddressText', value ? JSON.stringify(value) : '');
+      },
+    },
     addresses: {
       type: Sequelize.VIRTUAL,
       get(this: db.NodeInstance) {
--- a/lib/db/models/Node.ts
+++ b/lib/db/models/Node.ts
@@ -20,14 +20,17 @@ export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes) 
         }
       },
     },
-    lastAddressText: { type: Sequelize.TEXT, allowNull: false },
+    lastAddressText: { type: Sequelize.TEXT, allowNull: true },
     lastAddress: {
       type: Sequelize.VIRTUAL,
+      allowNull: true,
       get(this: db.NodeInstance) {
-        return JSON.parse(this.lastAddressText);
+        return this.lastAddressText ? JSON.parse(this.lastAddressText) : undefined;
       },
       set(this: db.NodeInstance, value: Address) {
-        this.setDataValue('lastAddressText', value ? JSON.stringify(value) : '');
+        if (value) {
+          this.setDataValue('lastAddressText', JSON.stringify(value));
+        }
       },
     },
     banned: { type: DataTypes.BOOLEAN, allowNull: true },
--- a/lib/db/models/Node.ts
+++ b/lib/db/models/Node.ts
@@ -7,16 +7,6 @@ export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes) 
     id: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
     nodePubKey: { type: DataTypes.STRING, unique: true, allowNull: false },
     addressesText: { type: Sequelize.TEXT, allowNull: false },
-    lastAddressText: { type: Sequelize.TEXT, allowNull: false },
-    lastAddress: {
-      type: Sequelize.VIRTUAL,
-      get(this: db.NodeInstance) {
-        return JSON.parse(this.lastAddressText);
-      },
-      set(this: db.NodeInstance, value: Address) {
-        this.setDataValue('lastAddressText', value ? JSON.stringify(value) : '');
-      },
-    },
     addresses: {
       type: Sequelize.VIRTUAL,
       get(this: db.NodeInstance) {
@@ -28,6 +18,16 @@ export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes) 
         } else {
           this.setDataValue('addressesText', '[]');
         }
+      },
+    },
+    lastAddressText: { type: Sequelize.TEXT, allowNull: false },
+    lastAddress: {
+      type: Sequelize.VIRTUAL,
+      get(this: db.NodeInstance) {
+        return JSON.parse(this.lastAddressText);
+      },
+      set(this: db.NodeInstance, value: Address) {
+        this.setDataValue('lastAddressText', value ? JSON.stringify(value) : '');
       },
     },
     banned: { type: DataTypes.BOOLEAN, allowNull: true },
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -101,12 +101,9 @@ class NodeList {
         return true;
       }
 
+      // if the lastAddress is removed, then re-assigning lastAddress with the latest connected advertised address
       if (node.lastAddress && addressUtils.areEqual(address, node.lastAddress)) {
-        node.lastAddress = [...node.addresses].sort((a, b) => {
-          if (!a.lastConnected) return 1;
-          if (!b.lastConnected) return -1;
-          return b.lastConnected - a.lastConnected;
-        })[0];
+        node.lastAddress = addressUtils.sortByLastConnected(node.addresses)[0];
       }
     }
 
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -67,7 +67,7 @@ class NodeList {
    * Update a node's addresses.
    * @return true if the specified node exists and was updated, false otherwise
    */
-  public updateAddresses = async (nodePubKey: string, addresses: Address[] = [], lastAddress: Address) => {
+  public updateAddresses = async (nodePubKey: string, addresses: Address[] = [], lastAddress?: Address) => {
     const node = this.nodes.get(nodePubKey);
     if (node) {
       // avoid overriding the `lastConnected` field for existing matching addresses unless a new value was set
@@ -79,7 +79,10 @@ class NodeList {
           return newAddress;
         }
       });
-      node.lastAddress = lastAddress;
+
+      if (lastAddress) {
+        node.lastAddress = lastAddress;
+      }
 
       await node.save();
       return true;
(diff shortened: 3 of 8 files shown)


lib/db/DB.ts
```diff
@@ -62,14 +62,17 @@ class DB {
         {
           nodePubKey: '02b66438730d1fcdf4a4ae5d3d73e847a272f160fee2938e132b52cab0a0d9cfc6',
           addresses: [{ host: 'xud1.test.exchangeunion.com', port: 8885 }],
+          lastAddress: { host: 'xud1.test.exchangeunion.com', port: 8885 },
```
sangaman: I don't think we should be initializing `lastAddress` like this. To me it should really mean the last address we've successfully connected with, if we haven't connected yet it should be blank.

lib/db/models/Node.ts
```diff
@@ -7,6 +7,16 @@ export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes)
     id: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
     nodePubKey: { type: DataTypes.STRING, unique: true, allowNull: false },
     addressesText: { type: Sequelize.TEXT, allowNull: false },
+    lastAddressText: { type: Sequelize.TEXT, allowNull: false },
+    lastAddress: {
+      type: Sequelize.VIRTUAL,
```
sangaman: Can you move these new lines down so they're below `addresses` - it'd be nice to keep `addressesText` and `addresses` adjacent.

lib/p2p/Pool.ts
```diff
@@ -183,9 +183,28 @@ class Pool extends EventEmitter {
...
+    } catch (err) {}
+
+    return false;
+  }
+
+  private connectWithAdvertisedAddresses = async(node: NodeConnectionInfo, retryConnecting = false) => {
     const { addresses, nodePubKey } = node;
 
-    // sort by lastConnected desc
+      // sort by lastConnected desc
```
sangaman: Remove extra indentation.

lib/db/models/Node.ts
```diff
@@ -7,6 +7,16 @@ export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes)
...
+      get(this: db.NodeInstance) {
+        return JSON.parse(this.lastAddressText);
+      },
+      set(this: db.NodeInstance, value: Address) {
+        this.setDataValue('lastAddressText', value ? JSON.stringify(value) : '');
```
sangaman: I'd have the null case here either set an empty object `{}` or `undefined`, for the latter I think we'd need `lastAddressText` to `allowNull: true`.  Undefined would be better I think, I would expect nodes we've never had an outgoing connection to to not have a value and I suspect empty string might give us parse errors.

lib/p2p/Pool.ts
```diff
@@ -412,10 +431,11 @@ class Pool extends EventEmitter {
         await this.nodes.createNode({
           addresses,
           nodePubKey: peer.nodePubKey!,
+          lastAddress: peer.address,
         });
       } else {
         // the node is known, update its listening addresses
-        await this.nodes.updateAddresses(peer.nodePubKey!, addresses);
+        await this.nodes.updateAddresses(peer.nodePubKey!, addresses, peer.address);
```
sangaman: We only want to set a value for `lastAddress` if the connection is outbound (similar to the check above for updating `lastConnected` on `addresses`). Same for the new node case below.

lib/db/DB.ts
rsercano: So, would it be okay to put host and port empty or should I make `lastAddress` field optional (which could change more parts of the code) ?

lib/db/models/Node.ts
rsercano: done :)

lib/p2p/Pool.ts
rsercano: done

lib/db/models/Node.ts
rsercano: so as I asked in the initiation part, I guess I need to make `lastAddress` field optional ?

lib/p2p/Pool.ts
rsercano: Oh right, I guess I overlooked it, since I need to set it for only outbound connections, I have  to make it optional.

rsercano:
added changes you mentioned @sangaman

lib/p2p/Pool.ts
```diff
@@ -183,6 +183,25 @@ class Pool extends EventEmitter {
    * Attempt to create an outbound connection to a node using its known listening addresses.
    */
   private tryConnectNode = async (node: NodeConnectionInfo, retryConnecting = false) => {
+    if (!this.connectWithLastAddress(node)) {
+      this.connectWithAdvertisedAddresses(node, retryConnecting);
+    }
+  }
+
+  private connectWithLastAddress = async(node: NodeConnectionInfo) => {
```
sangaman: Can you put a space after `async`? That's the style throughout most of the codebase. It looks like there's a lint rule for this so I'll open a separate PR for it.

lib/p2p/Pool.ts
```diff
@@ -183,6 +183,25 @@ class Pool extends EventEmitter {
...
+
+    try {
+      await this.addOutbound(lastAddress, nodePubKey, false);
+      return true;
+    } catch (err) {}
+
+    return false;
+  }
+
+  private connectWithAdvertisedAddresses = async(node: NodeConnectionInfo, retryConnecting = false) => {
```
sangaman: Space after async here as well please.

lib/p2p/Pool.ts
```diff
@@ -431,6 +451,7 @@ class Pool extends EventEmitter {
         connectedNodesInfo.push({
           nodePubKey: connectedPeer.nodePubKey!,
           addresses: connectedPeer.addresses,
+          lastAddress: connectedPeer.address,
```
sangaman: I actually don't think we want to share *our* `lastAddress` with other peers - let them connect and determine which address works for them. It's possible `lastAddress` is reachable for us but not fro the external internet. Plus if it's not an advertised address, we don't want to share it.

lib/types/p2p.ts
```diff
@@ -9,6 +9,7 @@ export type Address = {
 export type NodeConnectionInfo = {
   nodePubKey: string;
   addresses: Address[];
+  lastAddress: Address | undefined;
```
sangaman: Rather than explicitly put `| undefined`, just call this `lastAddress?` to be consistent with our other type definitions.

test/integration/OrderBook.spec.ts
```diff
@@ -30,7 +30,7 @@ describe('OrderBook', () => {
     const p2pRepository = new P2PRepository(loggers.p2p, db.models);
 
     await p2pRepository.addNode(
-      { nodePubKey: nodeKey.nodePubKey, addresses: [] },
+      { nodePubKey: nodeKey.nodePubKey, addresses: [], lastAddress: { host: '', port: 0 } },
```
sangaman: Can we not just leave `lastAddress` as `undefined` here?

lib/db/models/Node.ts
```diff
@@ -20,6 +20,19 @@ export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes)
...
+    lastAddressText: { type: Sequelize.TEXT, allowNull: true },
+    lastAddress: {
+      type: Sequelize.VIRTUAL,
+      allowNull: true,
+      get(this: db.NodeInstance) {
+        return this.lastAddressText ? JSON.parse(this.lastAddressText) : undefined;
+      },
+      set(this: db.NodeInstance, value: Address) {
+        if (value) {
+          this.setDataValue('lastAddressText', JSON.stringify(value));
```
sangaman: Can we `setDataValue` to `null` or `undefined` if `value` is `undefined`?

lib/p2p/Pool.ts
rsercano: Ah okay, currently there's no lint rule for it.

lib/p2p/Pool.ts
rsercano: done

lib/p2p/Pool.ts
rsercano: Then we actually don't want to share it at all ? Then I'm setting undefined to `lastAddress` here

test/integration/OrderBook.spec.ts
rsercano: since I'll change it as you described `lastAddress?` I guess it'll be better to remove it there.

lib/db/models/Node.ts
rsercano: Actually not setting it wouldn't be sufficient, isn't it actually the same thing with setting is as undefined since we already defined it as `lastAddress?` ? Also trying to set undefined to a `TEXT` field is not allowed I guess

lib/p2p/Pool.ts
sangaman: Yup, but I just opened #497.

lib/p2p/Pool.ts
sangaman: Yeah I believe so.

lib/db/models/Node.ts
sangaman: OK, I wasn't too sure.  Thanks.

lib/p2p/Pool.ts
```diff
@@ -183,6 +183,25 @@ class Pool extends EventEmitter {
    * Attempt to create an outbound connection to a node using its known listening addresses.
    */
   private tryConnectNode = async (node: NodeConnectionInfo, retryConnecting = false) => {
+    if (!this.connectWithLastAddress(node)) {
+      this.connectWithAdvertisedAddresses(node, retryConnecting);
```
sangaman: We need to `await` both of these calls since they are async functions, this the reason the build is currently failing.

rsercano: ah again overlooked, silly me thanks !

moshababo:
Sorry for the delay. Will review this tomorrow morning.

lib/p2p/NodeList.ts
```diff
@@ -96,6 +100,14 @@ class NodeList {
         await node.save();
         return true;
       }
+
+      if (node.lastAddress && addressUtils.areEqual(address, node.lastAddress)) {
+        node.lastAddress = [...node.addresses].sort((a, b) => {
+          if (!a.lastConnected) return 1;
+          if (!b.lastConnected) return -1;
+          return b.lastConnected - a.lastConnected;
+        })[0];
```
moshababo: the sorting logic also exists on `Pool.connectWithAdvertisedAddresses`. We should extract it to `addressUtils`

lib/p2p/Pool.ts
```diff
@@ -183,6 +183,25 @@ class Pool extends EventEmitter {
    * Attempt to create an outbound connection to a node using its known listening addresses.
    */
   private tryConnectNode = async (node: NodeConnectionInfo, retryConnecting = false) => {
+    if (!await this.connectWithLastAddress(node)) {
+      await this.connectWithAdvertisedAddresses(node, retryConnecting);
```
moshababo: Let's stick to the `tryX` naming convention for functions which are swallowing errors.

lib/p2p/Pool.ts
```diff
@@ -183,6 +183,25 @@ class Pool extends EventEmitter {
...
+    try {
+      await this.addOutbound(lastAddress, nodePubKey, false);
+      return true;
+    } catch (err) {}
+
+    return false;
+  }
+
+  private connectWithAdvertisedAddresses = async (node: NodeConnectionInfo, retryConnecting = false) => {
     const { addresses, nodePubKey } = node;
```
moshababo: This method contain the old logic, where retries are applies on the last connected address from the advertised addresses. As @sangaman mentioned, we should apply the retry logic on `lastAddress` instead - which might not have been advertised. It also means that we should extract the retries section from here to a different method, to be called from  `tryConnectNode`.

In addition, there might be a case where `lastAddress` was advertised, so you will try connecting to it twice - once in `connectWithLastAddress`, and then again in `connectWithAdvertisedAddresses`. This wouldn't have been if we saved `lastAddress` solely for an unadvertised address. There are other ways to solve this as well. I'll leave it up to you, let us know if you need help with it.

lib/p2p/NodeList.ts
```diff
@@ -96,6 +100,14 @@ class NodeList {
         await node.save();
         return true;
       }
+
+      if (node.lastAddress && addressUtils.areEqual(address, node.lastAddress)) {
```
moshababo: adding a small comment here would be nice

lib/types/p2p.ts
```diff
@@ -9,6 +9,7 @@ export type Address = {
 export type NodeConnectionInfo = {
   nodePubKey: string;
   addresses: Address[];
+  lastAddress?: Address | undefined;
```
moshababo: Is the `| undefined` necessary here, after you made `lastAddress` optional?

lib/p2p/NodeList.ts
rsercano: done

lib/p2p/NodeList.ts
rsercano: sure

lib/types/p2p.ts
rsercano: I guess you're right, probably I overlooked it

lib/p2p/Pool.ts
rsercano: But if I put retry logic to the connecting last address part, wouldn't that prevent connecting to the advertised addresses ? Is this what we want ? 

I'll handle connecting twice issue by checking the address in connecting to advertised addresses method, if the address equals with lastAddress then I won't try to connect it in advertised address method.

rsercano: And wouldn't it be sufficient to send `retryConnecting = true` for `addOutbound` method inside of `tryConnectWithLastAddress` method

rsercano: Would you please check current implementation ?

rsercano:
@moshababo just added the changed you requested, would you please check again ?

lib/p2p/Pool.ts
```diff
@@ -183,27 +183,42 @@ class Pool extends EventEmitter {
...
+    if (!lastAddress) return false;
+
+    try {
+      await this.addOutbound(lastAddress, nodePubKey, false);
+      return true;
+    } catch (err) {}
+
+    if (retryConnecting) {
+      try {
+        await this.addOutbound(lastAddress, nodePubKey, true);
```
sangaman: Don't we want a `return true;` after this line? It looks like it will always return false if we only manage to connect on retry. If so, this can actually be combined with the block above, just:

```typescript
try {
  await this.addOutbound(lastAddress, nodePubKey, retryConnecting);
  return true;
} catch (err) {}

return false;
```

lib/p2p/Pool.ts
```diff
@@ -183,27 +183,42 @@ class Pool extends EventEmitter {
    * Attempt to create an outbound connection to a node using its known listening addresses.
    */
   private tryConnectNode = async (node: NodeConnectionInfo, retryConnecting = false) => {
+    if (!await this.tryConnectWithLastAddress(node, retryConnecting)) {
+      await this.tryConnectWithAdvertisedAddresses(node);
```
sangaman: I believe the logic we want here is to first attempt the last address, then the advertised addresses, and then (optionally) retry the last address again. Otherwise we'd immediately go into the long retry loop before trying their other addresses. So we probably want another if statement here:

```typescript
// first try connecting with the last address once, then try connecting with the advertised
// addresses, then if retryConnecting is true keep retrying the last address
if (!await this.tryConnectWithLastAddress(node, false)) {
  if (!await this.tryConnectWithAdvertisedAddresses(node) && retryConnecting) {
    await this.tryConnectWithLastAddress(node, true);
  }
}
```

rsercano: Yes that was my first approach, will do !

rsercano:
I believe this's done @sangaman @moshababo ?

rsercano:
@moshababo do you want me to add check into `tryConnectWithAdvertisedAddresses` for; if the address equals with `lastAddress `then we'll not try to connect it in tryConnectWithAdvertisedAddresses method.

rsercano:
added the statement I mention @moshababo 

` if (node.lastAddress && addressUtils.areEqual(address, node.lastAddress)) continue;`

moshababo:
Great, thanks.

PR #499, opened 2018-09-18
chore: modify list of published files

sangaman:
This commit removes unnecessary/informational files from our list of published files. This is because users that install from npm won't be looking in their `node_modules` folder for documentation, but rather in
our GitHub repository and similar online resources. Proto files are not needed because we use the static, generated files instead. The README is needed for use in the npm website. The LICENSE should be included with all copies of the code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,11 +96,10 @@
   "files": [
     "bin",
     "dist",
-    "docs",
-    "proto",
     "tasks",
     ".env",
-    "*.md"
+    "README.md",
+    "LICENSE"
   ],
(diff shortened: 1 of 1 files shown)


sangaman:
Good question. I was unaware of that recommendation but the [GNU website](https://www.gnu.org/licenses/gpl-howto.html) also instructs people who use their licenses to put a notice at the top of each file. I'd say we should follow the recommended guidelines, so maybe in a separate PR we can make sure we're following the how-to I linked to above (and adjust the published files, if needed).

Issue #479, opened 2018-09-14
placeOrder / placeOrderSync


offerm:
It would be nice to have a `wait` option of  buy/sell market order. 
These can block until the trade is done and the result is available. 
The result is not just the match(s) but also the outcome of the execution (swap for example).
If can also show the impact on balance once we have it.

michael1011:
#123 is doing that.

offerm:
Thanks. Noticed. 
Not sure why we mix the discussion of Subscribe calls with placeOrder call. 
I envisioned something simple. clearly not complicated as the discussion in #123. 
Will take some time to go over #123 and digest it.

sangaman:
We can maybe use this issue to track the synchronous `placeOrder` mode which waits for all swaps to complete (or fail), as an offshoot of #123. @moshababo have you worked on this any more or thought more about how that would look like? I know you'd discussed using an approach similar to the peer `ResponseEntry` model.

sangaman:
@moshababo's most recent post:

> Designing the new API should take #145 into consideration. I'd suggest to discuss it first before implementing.
> 
> In regards to having a synchronous version for placeOrder, i'd suggest to utilize Peer.wait in order to create a synchronous code flow in Swap.beginSwap (which could be awaited from OrderBook.handleMatch, and so to affect MatchingResults). I think this change would be good for our code base, regardless of placeOrderSync.
> The actual LND call is synchronous, so we don't need extra work there.
> 
> The other option is to manage a global state for the swap procedures, which I think would be a far worse solution.
> 
> I think this should be implemented separately from the high-level API changes. Let me know if you want me to work on it.

If we go forward with `placeOrder` and `pIaceOrderSync` calls, which I still think is best, we can have only `placeOrder` active for non-matching mode. It will behave the same, returning an empty response immediately as long as the order is valid, the only difference is there won't be any streaming events following it.

You should take a look at #468 if you get a chance, it has a new swap flow and it does use a swap status to track which point of the swap process it's at currently. It has fewer steps and is less reliant on peer packets to advance each step of the flow, so I'm not sure if we can use `peer.wait` all that much, but it does fire events Basically what I'm thinking is:

- on `swap.rejected` or `swap.failed` abandon the match, remove the limit order from the order book, and try to match with a new order.
- on `swap.paid` consider the match final and add it to our list of matches for `sendPaymentSync` - or publish an update to the grpc order subscription for `sendPayment`.

kilrau:
[...]

Issue #223, opened 2018-07-30
set order localId (currently generated separately) = globalId


moshababo:
From a discussion at #175. @sangaman wrote:

> I'm also thinking now we might have a case where an exchange is letting XUD do all the matching and doesn't want to have to create and track their own Ids. Which makes me think maybe localId should be optional when placing an order. Maybe even tracking localIds at all can be something that's configurable? Also something that can be done in a separate PR, but I wanted to see what you thought.

kilrau:
For now we can assume that the vast majority of exchanges will need to have their own localIDs. Something to be implemented in a much later stage. Moved to beta.

sangaman:
In thinking about this more, maybe rather than generating a new UUID for `localId` when none is specified by the client, we just set `localId` equal to the global order id. I'm not thinking of any reason why they can't have the same value, and it would simplify things a bit. I can handle that in a PR and close this issue if that makes sense to you.

kilrau:
localID=global ID in matching mode makes sense to me, a PR of this can close this issue  @sangaman @moshababo

kilrau:
Something for you? @ImmanuelSegol

kilrau:
`localId` is already optional, but momentarily got it's own UUID is generated. This issue is about setting it equal to the `globalId` UUID. Something for you? @erkarl

erkarl:
@kilrau possibly - would be nice to clean this up sooner rather than later.

kilrau:
go for it!

sangaman:
I was actually just thinking about this last night and was going to change it today, I definitely think this would be nice to have and it should be easy to do. Please go ahead if you get a chance.

PR #500, opened 2018-09-19
feat(rpc): add new order events rpc subscriptions

### LINKED ISSUE #123: Add streaming gRPC call to listen to orderbook events

sangaman:
Closes #123.

This PR replaces the old `subscribePeerOrders` and `subscribeSwaps` rpc calls with three new streaming rpc calls:

- `subscribeNewOrders`: Notifies when new incoming peer orders are added.
- `subscribeExpiredOrders`: Notifies when peer orders become expired due to being canceled or filled remotely.
- `subscribeExecutedOrders`: Notifies when local orders are executed via a successful swap.

It also restructures and updates the `Order` grpc message type.

Feedback/critique welcome as I'm not 100% certain on the approach/structure and naming. I went with the three separate rpc calls above because each uses different message types, and I figure clients who subscribe to these will have very distinct handling for all three. For naming I went with "expired" rather than "removed," "invalidated", "canceled", etc... because from the perspective of the client that's more or less what's happening. An order (or part of it) that was previously good is now no longer good, a.k.a it expired. If we like the name, maybe we can also apply it to the internal "invalidation" naming. I was also torn on the names for the other calls, `subscribeNewOrders` could be `subscribeIncomingOrders` (I kind of like this better actually) or `subscribeAddedOrders` and `subscribeExecutedOrders` could be `subscribeFilledOrders`, `subscribeSwappedOrders`, or even just reuse the old `subscribeSwaps` name. I figure it's worth trying to get the names right since these will be hard to change down the road and users/exchanges will be dealing with them and seeing them a lot.

I do think the subscriptions tie in nicely to the `peerOrder.incoming`, `peerOrder.invalidation`, and `swap.paid` events, respectively. I had to add a couple of fields to `SwapDeal` (`peerPubKey` actually existed previously and will be useful for not sending an order invalidation packet to the taker) but I think they represent relevant info and will be handy to persist to the database going forward.

From the cli perspective, I only put the subscription for incoming orders there and used it mainly for testing purposes. I didn't put the other two in there because I don't really see the cli tool as a good fit for subscription rpc calls, the console just goes into a holding pattern waiting for the streaming messages to come in. If there is indeed a use case I can certainly include the other two calls, but I can't think of one right now.

Lastly, I think these will be useful for the sort of tests being added in #491. It will be useful for our test suites to detect when an `xud` instance receives an order, invalidates an order, or registers a successful swap.


lib/grpc/GrpcServer.ts
```diff
@@ -30,8 +30,7 @@ class GrpcServer {
       removeCurrency: grpcService.removeCurrency,
       removePair: grpcService.removePair,
       shutdown: grpcService.shutdown,
-      subscribePeerOrders: grpcService.subscribePeerOrders,
-      subscribeSwaps: grpcService.subscribeSwaps,
+      subscribeNewOrders: grpcService.subscribeNewOrders,
```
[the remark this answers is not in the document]
sangaman: Yeah the more I think about it the more I agree, I'll wait a bit for other comments before changing it though.

sangaman:
I did actually have it as `SubscribeOrderExpirations` at one point and then switched it, I'm fine with switching it back. `SubscrbibeOrderExecutions` as well then?

michael1011:
I would say so but there is no equivalent for incoming orders.

sangaman:
I think `SubscribeOrderExpirations`, `SubscrbibeOrderExecutions`, and `SubscribeIncomingOrders` works, only in the last one are you really getting the full order itself.

moshababo:
* I think we should keep the incoming / invalidation terminology solely for the p2p context. The order book listens to the p2p layer, applies add / remove, and should emit added / removed, hence -`peerOrder.added` instead of `peerOrder.incoming`, `peerOrder.removed` instead of `peerOrder.invalidated`. ownOrder events should be emitted as well, see the following clause.
The gRPC events are order book events, so we should favor the order book terminology over the p2p or over inventing something new. So i'm not in favor of `subscribeNewOrders` / `subscribeExpiredOrders`. 

* ownOrders events are not emitted. You are relying on `placeOrder` response (internal matches / taker swaps) and the new `subscribeExecutedOrders` (maker swaps), but I thought we [agreed](https://github.com/ExchangeUnion/xud/issues/123#issuecomment-421769253) on not doing so. Current situation:
   * `subscribeExecutedOrders` implicitly announces the removal of an ownOrder.
   * `placeOrder` response `remainingOrder` implicitly announces the adding of an ownOrder.
   * `placeOrder` response will (#479) contain internal matches and taker swaps resolution array/s. The swap resolution should probably be identical to the maker swap's `OrderExecution` struct (`subscribeExecutedOrders`), but in this case it implicitly announce the removal of a peerOrder, instead. Internal match, however, announces ownOrder removal. I think we should emit these events separately.

moshababo:
* I think we should keep the incoming / invalidation terminology solely for the p2p context. The order book listens to the p2p layer, applies add / remove, and should emit added / removed, hence -`peerOrder.added` instead of `peerOrder.incoming`, `peerOrder.removed` instead of `peerOrder.invalidated`. ownOrder events should be emitted as well, see the following clause.
The gRPC events are order book events, so we should favor the order book terminology over the p2p or over inventing something new. So i'm not in favor of `subscribeNewOrders` / `subscribeExpiredOrders`. 

* ownOrders events are not emitted. You are relying on `placeOrder` response (internal matches / taker swaps) and the new `subscribeExecutedOrders` (maker swaps), but I thought we [agreed](https://github.com/ExchangeUnion/xud/issues/123#issuecomment-421769253) on not doing so. Current situation:
   * `subscribeExecutedOrders` implicitly announces the removal of an ownOrder.
   * `placeOrder` response `remainingOrder` implicitly announces the adding of an ownOrder.
   * `placeOrder` response will (#479) contain internal matches and taker swaps resolution array/s. The swap resolution should probably be identical to the maker swap's `OrderExecution` struct (`subscribeExecutedOrders`), but in this case it implicitly announce the removal of a peerOrder, instead. Internal match, however, announces ownOrder removal. I think we should emit these events separately. 

* If we'll indeed separate the order book state events from the matches / swaps results, I also think that `subscribeExecutedOrders` won't be appropriate, and if we want to announce a maker swap (which we said we won't, but I think it's a good idea), we better stick to `subscribeSwaps`.

sangaman:
> I think we should keep the incoming / invalidation terminology solely for the p2p context. The order book listens to the p2p layer, applies add / remove, and should emit added / removed, hence -peerOrder.added instead of peerOrder.incoming, peerOrder.removed instead of peerOrder.invalidated. ownOrder events should be emitted as well, see the following clause.
The gRPC events are order book events, so we should favor the order book terminology over the p2p or over inventing something new. So i'm not in favor of subscribeNewOrders / subscribeExpiredOrders.

OK, `subscribeAddedOrders` then? I'm ok with `subscribeRemovedOrders` too - or `subscribeOrderRemovals`?

> ownOrders events are not emitted. You are relying on placeOrder response (internal matches / taker swaps) and the new subscribeExecutedOrders (maker swaps), but I thought we agreed on not doing so.

I guess I misunderstood you. I thought you were proposing the `placeOrder` calls to return their matches, whether in a single response or streaming. I'd actually prefer that, because again from the client's perspective there'd be quite a difference in handling logic between a taker order swap/execution and a maker order swap/execution.

> If we'll indeed separate the order book state events from the matches / swaps results, I also think that subscribeExecutedOrders won't be appropriate, and if we want to announce a maker swap (which we said we won't, but I think it's a good idea), we better stick to subscribeSwaps.

As it's set up in this PR though, what do you think about the naming? I'm having a hard time picturing the separate order book state events.

offerm:
How do you suggest this should be used?

Should I open 3 terminal and execute and RPC call in each of them?

Or maybe 6 since I have exchange-a and exchange-b?

Also, every time I restart the XUD process I need to restart the RPC too?

And to check what happened I need to check these 6 termnials?

sangaman:
From the point of an exchange or any application that wants to have an up to date list of the available orders (to display to users/traders for instance), they would subscribe to these calls and get events every time an order is added or removed from the order book. Also for any client that wants to have a real time feed of when their orders are filled (again maybe to display to users or keep up to date balances).

As I said in my first post I don't think these calls make a whole lot of sense from the cli/terminal, I only put it a cli command for one of them and it's mainly for manual testing/smoke check purposes. It makes more sense form the perspective of an application that's directly consuming our API. Yes when xud restarts you'd need to resubscribe, similar to how you have to resubscribe to `SubscribeInvoices` or `SubscribeTransactions` in lnd.

sangaman:
One thought is having a single `xucli` command that subscribes to all three of these calls- that would probably be best. I can put that in this PR.

offerm:
Agree that this is for an app and not for user/cli.

as a user, I would like to see what is going on with a command line
(checking, not stream).

moshababo:
Sorry if i'm not clear enough. 

In "order book state events" i refer only to Added / Removed events. Other events:
* `placeOrder` matches (internal match): pairs of orders (no Added / Removed)
* `placeOrder` (taker) / `subscribeExecutedOrders ` (maker) swaps results: what you defined as the  `OrderExecution` struct.

I did proposed `placeOrder` to return the matches/swaps results, but also that we should differ them from the order book state events, which we should emit separately. 
So on a maker swap, for example, `subscribeExecutedOrders` will emit the swap result. If successful, `ownOrder.removed` should be emitted as well. That's what I mean by "separation".

My last comment about it I referred to the name `subscribeExecutedOrders`. I don't think it works well, because it implies a reference to an order, while we (should) emit the order state separately (`ownOrder.removed` in this case), and stream here the swap result, which is different. So for me, this name marks again a confusion on how we differentiate between these.
In addition, this event is a bit confusing in general, because it applies only to maker swaps. That's because taker swaps are supposed to be emitted to `placeOrder`. I don't think we should emit the taker swap results in two places, because unlike the separation with the order book state events, it's the exact same event. I think that the `subscribeSwaps` name can work fine, defining it as a stream solely for maker swaps results, and perhaps also renaming `OrderExecution` as `DealResult` / `SwapResult` / `SwapResolution`, etc.

In regards to the orders events - I personally in favor on separating based on ownOrder / peerOrder:
* it will allow us to teach the client this terminology, which is a bit weird but is the basis to what we're doing
* the Added / Removed payload is pretty similar - the latter is just more narrow. There's no conflict like between ownOrder (has localId) / peerOrder (has peerPubKey)
* on no-matching mode, ownOrders subscription is disabled. It will work nicely if we can say that this entire subscription is disabled on that mode

sangaman:
Ahh I see, that makes sense now. I'd be fine with having all order removals come through a single stream as you described, so `placeOrder` returns the matching results (for informational purposes) but a streaming call that's used for actually removing orders.

So now I'm thinking:

- `subscribeAddedOrders` - notifies of all incoming peer orders that get added to the orderbook
- `subscribeRemovedOrders` - notifies of all peer orders . Although you suggest splitting this into two subscriptions, one for peer orders and one for own orders? I think I'd prefer this as one - although it's true that it would be nice to disable the own orders subscription entirely for non-matching mode. Also `subscribeRemovedPeerOrders` and `subscribeRemovedOwnOrders` are a mouthful in my opinion. 
- `subscribeSwaps` (or how about `subscribeTrades`? or maybe `subscribeExecutions`?) - notifies whenever one of our maker orders has a completed swap.

moshababo:
I actually meant having `subscribePeerOrders` & `subscribeOwnOrder`, each to apply "added" / "removed" field per event, in addition to the order payload.

moshababo:
And i'm in favor of `subscribeSwaps`, I found it the most explicit and less confusing compared to others (trades, executions).

sangaman:
Ohhh I see. Although `subscribeOwnOrders` will only ever have removal events, right?

moshababo:
I think it should include an "add" of `remainingOrder` from `placeOrder` response, which I also see as information which is distinct from the order book state events. That's mainly to avoid confusion - the order payload, and thus the event, will be the same. We can even drop `remainingOrder` from the response altogether, but I think it's beneficial to have it directly on the `placeOrder` response as well.

sangaman:
Ok. that would work for me. I'd also want to keep `remainingOrder` (although I suppose only on `placeOrderSync` call?) or at least some minimal info like `remainingQuantity` plus the order id.

moshababo:
We agreed that `placeOrderSync` response will be identical to `placeOrder` one, just that the latter will be async. So I don't see why `placeOrder` can't get `remainingOrder` back as well.

sangaman:
I may have overlooked this fact earlier, but we have to define a message type for the streaming `placeOrder` call, I'd expect that message type to contain match information similar to what we return for `matches` currently. But I'm figuring the `remainingOrder` would have a significantly different structure. I suppose we could manage and define (maybe with generous use of `oneOf`) a message to handle both though. Either way, something to deal with in a separate PR.

moshababo:
Why different?

As far as I see, `matches` is the result of internal matching, and the array item is simply a maker/taker pair of ownOrders (maker/taker), and so is `remainingOrder` - an ownOrder.
If non-internal matches were found, we don't return matches, but swaps results, and that indeed will require a different structure, and a probably a separate array.

moshababo:
Not related to this PR - The internal matches (`matches`) array could be returned immediately, so if combined with non-internal matches (swaps results), allowing immediate return will require some tweaking.  We can discuss that later.

moshababo:
Not related to this PR - The internal matches (`matches` array) could be returned immediately, so if combined with non-internal matches (swaps results), allowing immediate return will require some tweaking.  We can discuss that later.

offerm:
What is the plan if the swap fails? What's the impact on the matches?
Remainingorder?

moshababo:
If the swap fails we just try to rematch the taker order (and emitting `peerOrder.removed` event separately), until an eventual swap success or no matching scenario, which will result in a `remainingOrder`.

moshababo:
We can return the swap failures info in the response. Is it necessary?

offerm:
So placeorder blocks until all this is done?

moshababo:
`placeOrderSync` yes. `placeOrder` can and should emit an event for each progressed step.

offerm:
@sangaman @moshababo What we discuss here is a new API to the XUD system. I suggest that you leave aside how the system is internally built and think only on the user of the API to make sure what we give him can be used and has a good UI/UX.

If you think we should provide these 3/4 rpc calls, I suggest that you start with a short description that explains **the user** (exchange) what each API is, when and how it should be used.
In such a description, you can't use the internal terms unless you expect the client to understand them.

And let me voice for the last time - what you suggest is very complicated to use/understand. you are building it from the inside out and not from the outside in which is not good. KISS.
As a user, I expect to see something like `ordersStatus` call (note that "subscribe" is an RPC term that the user should not care about). I expect this call to give my all important events regarding selected orders (and I can select which order I would like to see).

my 2 cents.

moshababo:
@offerm since this is a pretty important stuff, i'll be happy to meet with you next week (or after Sukkot) so you'll explain me in details your alternative approach. Let me know on gitter if you're available.

sangaman:
I can rename the cli call if that's what you mean. For the actual RPC call I think "subscribe" make sense and follows the terminology lnd uses. I'll add detailed descriptions for all the calls. End users should not need to even be aware of any API (and certainly not internal) terminology, but I think here we are focusing not on end-users/consumers but rather developers who are integrating xud into an exchange or other application.

I don't think there's anything inherently less complicated about a single call through which everything streams, and in fact I think it could very well be more complicated since the explanation for that call would have to encompass everything.

So I will refactor this per the discussions so far and write up detailed descriptions of each new call, then we can review.

lib/grpc/GrpcServer.ts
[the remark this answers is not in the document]
kilrau: Agree on `subscribeIncomingOrders`

sangaman:
OK I spent a lot of time working/thinking about this. I changed my mind again partly (again) about the best way to structure the calls while actually working on implementing the approach discussed by me and @moshababo in the most recent comments here.

Making `subscribeOwnOrders` and `subscribePeerOrders` work for both adds and removes turned out to be challenging due to the different data structures of an order and an update to remove part (or all) of an order. Our internal events for order portions removed, matched, etc... only use the orderId/pairId/quantity triplet to identify orders (rightfully I think). Fitting both of those into a single call felt like it was getting messy and going against one of the principles of grpc which is having well-structured message types for effective serialization/deserialization.

So what I wound up going with is `subscribeAddedOrders` and `subscribeRemovedOrders` which cover the same events we've been discussing but do so with a different separation of interests. Whether the order in question is a peer or own order can be trivially inferred by checking whether `localId` has a value. All `quantity` values in question are always positive - any quantity on an `OrderRemoval` message is to be subtracted from the existing order it refers to.

Other things to note:

- I added an `ownOrder.filled` event for the order book to notify the service layer that it should send a `OrderRemoval` message. `ownOrder.filled` should emitted both on an internal match of two own orders as well as on a successful swap. That logic can be refined when reworking the `placeOrder` call in a separate PR, but I just wanted to create an event to listen to for now at least.

- Should we notify via `subscribeRemovedOrders` when `cancelOrder` is called? I'm thinking no - although wanted to see what people thought.

- I created a new type called `OrderPortion` to represent a portion of an order. Previously we were using `OrderIdentifier` both to identify an order (as is done with the `localId` to `orderId` mapping) as well as to identify a portion of an order. I felt these were separate and that the code is cleaner with having separate types for the two concepts.

I'm still definitely open to change or feedback, I'd like this to be done right.

Left to do: Detailed documentation/descriptions in our proto file of what the new calls do and how they should be used. I didn't want to write this up before getting the green light on the current approach. Also a more detailed commit message.

proto/xudrpc.proto
```diff
@@ -141,15 +141,22 @@ service Xud {
...
     option (google.api.http) = {
-      get: "/v1/subscribepeerorders"
+      get: "/v1/subscribeorders"
     };
   }
   
-  /* Subscribe executed swaps. */
-  rpc SubscribeSwaps(SubscribeSwapsRequest) returns (stream SubscribeSwapsResponse) {
+  /* Subscribe to orders being removed from the order book. */
+  rpc SubscribeRemovedOrders(SubscribeRemovedOrdersRequest) returns (stream OrderRemoval) {
```
moshababo: Segregating on removed/added instead of ownOrder/peerOrder is fine, but less optimal IMO. 
I already [specified](https://github.com/ExchangeUnion/xud/pull/500#issuecomment-423241655) the benefits of the later.
 The former seems to have the same problem - ownOrder and peerOrder structures are not equal, and the current `Order` protobuf type contains them both. I would actually like to see a separate `OwnOrder` + `PeerOrder` protobuf types. It would be useful for `placeOrder` response. 

If order removal events are missing some fields, which are actually not necessary, we can either retrieve them (if applicable) and emit anyway, or not populate them in the response (same like `peerPubKey` is not populated for ownOrder, and `localId` for peerOrder).  If i'm missing some technical challenge here, let me know.

In addition, relying on `localId` for differentiation is problematic. It's not very self evident, and will raise issues for #223.

With ownOrder/peerOrder segregation we'll have the same problem with quantity, but the positive/negative solution that you suggested sounds like a good solution.

lib/types/orders.ts
```diff
@@ -48,7 +48,20 @@ export type OutgoingOrder = Pick<StampedOwnOrder, Exclude<keyof StampedOwnOrder,
...
   pairId: string;
-  quantity?: number;
+};
+
+export type OrderPortion = OrderIdentifier & {
+  quantity: number;
+  localId?: string;
+};
+
+export type Swap = {
```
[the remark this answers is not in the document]
sangaman: 1. That's my goal.
2. I thought about `SwapResult` too actually, I thought `Swap` was probably clear enough but I'll change it.
3. Good question on swap failures, I wasn't thinking we would because it's not really an actionable item for the client. I'd probably lean towards leaving it as is and if we see a need for announcing swap failures in the future we can change it.

lib/service/Service.ts
```diff
@@ -319,22 +319,39 @@ class Service extends EventEmitter {
...
-      quantity: order.quantity,
-    }));
+    // TODO: send message on remaining order from placeOrder
+  }
+
+  /**
+   * Subscribe to orders being removed from the order book.
+   */
+  public subscribeRemovedOrders = async (callback: (order: OrderPortion) => void) => {
+    this.orderBook.on('peerOrder.invalidation', order => callback(order));
```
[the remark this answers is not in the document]
sangaman: Right, that's definitely better. I'll probably rename it in a separate PR.

proto/xudrpc.proto
sangaman: In regards to the `Order` protobuf and possibly splitting the types based on own or peer order, I actually think a [oneof](https://developers.google.com/protocol-buffers/docs/proto3#oneof) field fits perfectly. Their structures are identical except for a pair of fields, only one of which applies to each type. That's also why I feel the a single message type can be cleanly used for the added orders. Also I think splitting them would complicate the `getOrders` call - unless we went back to splitting them into `repeated own_orders` and `repeated peer_orders` which I don't think is ideal.

As for whether relying on `localId` is problematic, it doesn't seem like a big deal to me however I'm only basing that on my own intuition and maybe it really isn't self evident enough (although I would certainly document it as best I can). To me as well, it's the same sort of logic that's behind peer/own orders being combined in `getOrders` for instance, where clients can also infer which orders are local by the `localId`. Maybe we could use another field to make this clearer? `is_local`? Regarding #223, my most recent thoughts there weren't necessarily to have`localId` undefined or removed but rather just set it equal to `orderId`, so this wouldn't necessarily conflict although it would limit our options.

It's definitely possible to have the order removals use the `Order` message type, leaving the non-applicable fields like `price` with default values and using negative values for quantity. That's what I was going for at first, and while there wasn't a technical challenge really I did feel like the logic started to get messy when it came to dealing with `StampedOwnOrder | OrderPortion` variables passed to the GrpcService callback. From purely the protobuf message standpoint I don't think it's actually a big deal to leave a handful of fields unused.

I guess some key questions to me are:

- Are we generally comfortable with the concept of combining peer & own orders on the rpc layer? If not, maybe we should revisit `getOrders`?

- Is a field like `is_local` on orders something we'd want, perhaps even useful for the `getOrders` call without consideration for these new calls? Vs the alternative of checking for a value on `local_id`.

Thanks for all the thought you've put into this.

moshababo: I wasn’t aware of this `oneof`, looks like it can work well. 

In this regard, I don’t see a technical issue with combining peerOrder & ownOrder. 
I’m more inclined towards separation mostly for the purpose of “educating” the client. I think that the concept of ownOrder/peerOrder is weird and uncommon, so having it clearly defined in the API as much as possible will help users/developers to understand what is xud about. 
Having it merely mentioned in the documentation won't have the same effect. 
This separation will also lead to better understanding of `placeOrder` response, which isn't trivial.
We should definitely not decide against it because of the current `getOrders` implementation, which is of minor importance. 

In regards to having #223 with `localId` equals to the global id - It can work technically, but i’m not sure it’s the best solution. 

in regards to having a field like `is_local` - this name isn’t so self explanatory as well IMO. if we’ll do that, perhaps simply use `is_own_order`. So I'd say it's either we'll use an extra field like that, or separate the methods/types for ownOrder/peerOrder instead. Feel free to decide about this yourself, or wait for @kilrau @michael1011 input as well.

sangaman: `is_own_order` I agree is more clear. I think that might be the way to go, it would also be a nice enhancement to `getOrders` I think. That all seems clear to enough to me. I will write up detailed documentation tonight and then we can see if it makes sense.

michael1011:
I agree on the new approach of yours @sangaman. But I don't think a `is_own_order` is needed unless we want to implement #223 by leaving `local_id` undefined. Then we will need such a field eventually.

sangaman:
I'll probably keep `is_own_order` in there for now at least to give us more flexibility with how to solve #223 and err on the side of being explicit, although I'm also not totally sure it's worth having. If we do choose to remove it at some point in the future, it should be easy enough.

@moshababo re: `Swap` vs `SwapResult` - should I rename that on the rpc layer as well?

sangaman:
Addressed the feedback and added some more substantial comments to the proto file.

One more thing that occurred to me - earlier we had talked about being able to subscribe to added orders and getting all standing orders streamed to you up front. That way if you subscribe after orders have already started being collected, you can get the most recent state of the order book up front. I was thinking we could either make this up-front stream of preexisting orders happen by default, or set a flag on the `SubscribeAddedOrdersRequest` message that specifies whether we want existing orders up front. This also doesn't have to be done in this PR necessarily, but I think it's related and I know what needs to be done.

kilrau:
> In addition, relying on `localId` for differentiation is problematic. It's not very self evident, and will raise issues for #223.
> But I don't think a is_own_order is needed unless we want to implement #223 by leaving local_id undefined.

If we just set localID=globalID when XUD is in matching mode (not undefined) to make it optional, this should solve it. Don't see any issue with that @sangaman @moshababo 

> Are we generally comfortable with the concept of combining peer & own orders on the rpc layer? If not, maybe we should revisit getOrders?

Yes, in most cases an exchange wants all orders. I'm in favor for distinguishing them via the `is_own_order` field, not via separate calls. And yes please add it to `getOrders` @sangaman

sangaman:
Sounds good @kilrau, that's how it's currently set up for this PR.

kilrau:
And I see the final naming is

```
      subscribeAddedOrders: grpcService.subscribeAddedOrders,
      subscribeRemovedOrders: grpcService.subscribeRemovedOrders,
      subscribeSwaps: grpcService.subscribeSwaps,	      subscribeSwaps: grpcService.subscribeSwaps,
```

that's also what I was thinking. So good from my side. Updated https://github.com/ExchangeUnion/xud/issues/530#issuecomment-425169108 with what we discussed here.

lib/swaps/Swaps.ts
```diff
@@ -629,7 +633,16 @@ class Swaps extends EventEmitter {
...
+      const swapResult = {
+        orderId: deal.orderId,
+        localId: deal.localOrderId,
+        pairId: deal.pairId,
+        amountReceived: deal.makerAmount,
+        amountSent: deal.takerAmount,
+        r_hash: deal.r_hash,
+        peerPubKey: deal.peerPubKey,
+      };
+      this.emit('swap.paid', swapResult, deal.myRole);
```
moshababo: Maybe add `myRole` to `SwapResult` as well, in addition to having it in the grpc response, so the client won't need to guess what's `myRole` from the context (`placeOrder` response vs `subscribeSwaps`)

proto/xudrpc.proto
```diff
@@ -370,25 +404,28 @@ message RemovePairRequest {
...
+
+message SubscribeRemovedOrdersRequest {}
 
 message SubscribeSwapsRequest {}
-message SubscribeSwapsResponse {
-  // The order which was executed for the swap with updated remaining quantity
-  string order = 1 [json_name = "order"];
-}
+
+message Swap {
```
moshababo: Yes, we should rename `SwapResult` here as well.

sangaman:
Thanks @moshababo, just updated the PR for your latest comments.

lib/grpc/GrpcService.ts
```diff
@@ -382,17 +386,45 @@ class GrpcService {
...
-  public subscribePeerOrders: grpc.handleServerStreamingCall<xudrpc.SubscribePeerOrdersRequest, xudrpc.SubscribePeerOrdersResponse> = (call) => {
-    this.service.subscribePeerOrders((order: StampedPeerOrder) => call.write({ order }));
+  public subscribeRemovedOrders: grpc.handleServerStreamingCall<xudrpc.SubscribeRemovedOrdersRequest, xudrpc.OrderRemoval> = (call) => {
+    this.service.subscribeRemovedOrders((order: OrderPortion) => {
+      const orderRemoval = new xudrpc.OrderRemoval();
+      orderRemoval.setPairId(order.pairId);
+      orderRemoval.setOrderId(order.orderId);
+      orderRemoval.setQuantity(order.quantity);
+      orderRemoval.setLocalId(order.localId || '');
+      orderRemoval.setIsOwnOrder(!!order.localId);
```
michael1011: The `!!order.localId` markup is still confusing me. What about `(order.localId === undefined)` for the sake of readability?

proto/xudrpc.proto
```diff
@@ -370,25 +404,34 @@ message RemovePairRequest {
...
+  int64 amount_sent = 6 [json_name = "amount_sent"];
+  // The node pub key of the peer that executed this order.
+  string peer_pub_key = 7 [json_name = "peer_pub_key"];
+  enum Role {
+    TAKER = 0;
+    MAKER = 1;
+  }
+  // Our role in the swap, either MAKER or TAKER
+  Role role = 8 [json_name = "role"];
+}
```
michael1011: Missing `EOF`

sangaman:
Addresses your comments @michael1011, looks good to you to merge?

PR #503, opened 2018-09-19
feat(SWAPS): cltv-delta

offerm:
Both maker and taker can now specify via command line or configuration file a value for final hope cltv-delta. This value is used to calculate the cltv-delta when sending to the maker and to the taker. the maker's cltv-delta includes the taker's cltv-delta converted by the ration between the two cltv-deltas.

Both maker and taker now verify the received amount to make sure it matches the agreed amount.
In additon both take and maker verify the received cltv-delta time lock against the agreed one.

These checks are done as part of the resolver work. In case of a wrong value, the preImage is not discloused and error is raised.

Left to be done: cltv-delta in case of multi hops route.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -5,6 +5,7 @@ import BaseClient, { ClientStatus } from '../BaseClient';
 import errors from './errors';
 import { LightningClient } from '../proto/lndrpc_grpc_pb';
 import * as lndrpc from '../proto/lndrpc_pb';
+import assert from 'assert';
 
 /** The configurable options for the lnd client. */
 type LndClientConfig = {
@@ -13,6 +14,7 @@ type LndClientConfig = {
   macaroonpath: string;
   host: string;
   port: number;
+  cltvdelta: number;
 };
 
 /** General information about the state of this lnd client. */
@@ -41,6 +43,7 @@ interface LightningMethodIndex extends LightningClient {
 
 /** A class representing a client to interact with lnd. */
 class LndClient extends BaseClient {
+  public readonly cltvDelta: number = 0;
   private lightning!: LightningClient | LightningMethodIndex;
   private meta!: grpc.Metadata;
   private uri!: string;
@@ -57,7 +60,7 @@ class LndClient extends BaseClient {
     super(logger);
 
     let shouldEnable = true;
-    const { disable, certpath, macaroonpath } = config;
+    const { disable, certpath, macaroonpath, cltvdelta } = config;
 
     if (disable) {
       shouldEnable = false;
@@ -71,7 +74,10 @@ class LndClient extends BaseClient {
       shouldEnable = false;
     }
     if (shouldEnable) {
+      this.cltvDelta = cltvdelta;
+      assert(this.cltvDelta > 0, 'cltvdelta must be a positive number');
       this.uri = `${config.host}:${config.port}`;
+      this.cltvDelta = cltvdelta;
       const lndCert = fs.readFileSync(certpath);
       this.credentials = grpc.credentials.createSsl(lndCert);
 
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -325,8 +325,8 @@ class Service extends EventEmitter {
   /**
    * resolveHash resolve hash to preimage.
    */
-  public resolveHash = async (args: { hash: string }) => {
-    return this.swaps.resolveHash(args);
+  public resolveHash = async (request: lndrpc.ResolveRequest) => {
+    return this.swaps.resolveHash(request);
   }
 }
 export default Service;
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -1,4 +1,4 @@
-import { SwapDealRole, SwapDealState, SwapDealPhase } from '../types/enums';
+import { SwapDealPhase, SwapDealRole, SwapDealState } from '../types/enums';
 import Peer from '../p2p/Peer';
 import * as packets from '../p2p/packets/types';
 import { createHash, randomBytes } from 'crypto';
@@ -21,7 +21,7 @@ type SwapDeal = {
    */
   state: SwapDealState;
   /** The reason for being in current state */
-  stateReason: String;
+  stateReason: string;
   /** Global order id in the XU network. */
   orderId: string;
   /** The quantity of the order to execute as proposed by the taker. Negative when the taker is selling. */
@@ -37,10 +37,14 @@ type SwapDeal = {
   takerCurrency: string;
   /** Taker's lnd pubkey on the taker currency's network. */
   takerPubKey?: string;
+  /** The CLTV delta from the current height that should be used to set the timelock for the final hop when sending to taker. */
+  takerCltvDelta: number;
   /** The number of satoshis (or equivalent) the maker is expecting to receive. */
   makerAmount: number;
   /** The currency the maker is expecting to receive. */
   makerCurrency: string;
+  /** The CLTV delta from the current height that should be used to set the timelock for the final hop when sending to maker. */
+  makerCltvDelta?: number;
   /** The hash of the preimage. */
   r_hash: string;
   r_preimage?: string;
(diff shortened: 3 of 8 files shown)


lib/swaps/Swaps.ts
```diff
@@ -307,6 +341,23 @@ class Swaps extends EventEmitter {
...
+    return deal.stateReason;
+  }
+
+      // verify taker's takerCltvDelta
+      if (deal.makerCltvDelta! > resolveRequest.getTimeout() - resolveRequest.getHeightNow()) {
+        this.logger.error('got timeout ' + resolveRequest.getTimeout() + ' at height ' + resolveRequest.getHeightNow());
+        this.logger.error('cltvDelta is ' + (resolveRequest.getTimeout() - resolveRequest.getHeightNow()) +
+            'expected delta of ' + deal.makerCltvDelta);
+        this.setDealState(deal, SwapDealState.Error, 'cltvDelta sent from Taker to Maker is too small');
+        return deal.stateReason;
```
sangaman: Doesn't have to be addressed in this PR, but I feel this should be treated as an error case from the grpc perspective. Right now calling `resolveHash` will never result in a grpc error response with a code and description, even when the operation has clearly failed. From the perspective of lnd, I guess it would receive back this string message, see that it's not the preimage for the payment hash, and then fail the payment?

lib/service/Service.ts
```diff
@@ -325,8 +325,8 @@ class Service extends EventEmitter {
   /**
    * resolveHash resolve hash to preimage.
    */
-  public resolveHash = async (args: { hash: string }) => {
-    return this.swaps.resolveHash(args);
+  public resolveHash = async (request: lndrpc.ResolveRequest) => {
+    return this.swaps.resolveHash(request);
```
sangaman: Also doesn't have to be in this PR, but it would be good to get some argument checks in here - like amount has to be > 0 and hash should match a certain length, etc...

lib/lndclient/LndClient.ts
```diff
@@ -48,6 +49,7 @@ class LndClient extends BaseClient {
   private invoiceSubscription?: ClientReadableStream<lndrpc.InvoiceSubscription>;
   private reconnectionTimer?: NodeJS.Timer;
   private identityPubKey?: string;
+  private cltvdelta!: number;
```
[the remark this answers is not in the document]
offerm: I prefer not to deal with the impact of this being nullable. Fixed differently. 

Made sure it is not null.

lib/lndclient/LndClient.ts
```diff
@@ -87,6 +90,10 @@ class LndClient extends BaseClient {
     return this.identityPubKey;
   }
 
+  public get cltvDelta() {
```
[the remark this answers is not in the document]
offerm: don't see a need for it.

anyway the method is now getCltvDelta().

lib/swaps/Swaps.ts
```diff
@@ -307,6 +341,23 @@ class Swaps extends EventEmitter {
     // TODO: calculate CLTV
     if (deal.myRole === SwapDealRole.Maker) {
 	  this.logger.debug('Executing maker code');
+
+      // Verify that taker got right amount.
+	  if (amount < deal.makerAmount * 1000) {
+    this.logger.error('received ' + amount + ' mSat, expected ' + deal.makerAmount + 'mSat');
```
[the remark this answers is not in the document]
offerm: Yes. fixed.

Do you see the indentation problem?  lines 345-350 ?

lib/swaps/Swaps.ts
offerm: Yes.  Not for this PR and highly dependent on the LND change.
I did it like this to keep the LND change smaller and less compliacted. The preImage is checked against the hash for validity on LND side so I can return whatever I wish.

lib/service/Service.ts
offerm: hash is check  - if not valid you will not find the deal.

the amount is checked - if not valid it will not match to the deal amount.

Same for the other 2 values. If these are wrong we will catch it by one of the checks.

I see no need to enhance it (until we see an example of such a need).

sangaman: Basically I just mean to return a [INVALID_ARGUMENT] error if the argument is just completely invalid. Not for this PR, just a thought.

lib/lndclient/LndClient.ts
```diff
@@ -87,6 +97,10 @@ class LndClient extends BaseClient {
     return this.identityPubKey;
   }
 
+  public getCltvDelta() {
```
sangaman: I still find this method unnecessary, and it still technically allows `cltvDelta` to be changed after the fact when from what I understand that shouldn't be possible. It's also just a deviation from the style/structure of the rest of the codebase I think. If you just declare `cltvDelta` as `public readonly cltvDelta = 0` you can read it from anywhere but never overwrite it and without the need for an extra method.

lib/swaps/Swaps.ts
```diff
@@ -192,9 +207,21 @@ class Swaps extends EventEmitter {
     // TODO: check that we have route to taker.
     // TODO: extract data needed for proper timelock calculation and share with taker
 
+    let makerCltvDelta = 0;
+    const cltvDeltaFactor = this.lndLtcClient.getCltvDelta() / this.lndBtcClient.getCltvDelta();
```
sangaman: Are we risking a divide by zero error here? We might want to put in a check first.

lib/lndclient/LndClient.ts
```diff
@@ -48,6 +50,7 @@ class LndClient extends BaseClient {
   private invoiceSubscription?: ClientReadableStream<lndrpc.InvoiceSubscription>;
   private reconnectionTimer?: NodeJS.Timer;
   private identityPubKey?: string;
+  private cltvDelta!: number;
```
sangaman: Minor thing but you can assign this a default value here `cltvDelta = 0` and drop the `!` operator and the assignment in the constructor. Fwiw I still think it'd be better just to have this be nullable particularly if we need truthy checks where it's used to prevent divide by zero errors.

lib/swaps/Swaps.ts
sangaman: Never mind, this would put us back to where we were before which is checking every time which is unnecessary since it's a value that shouldn't change, I was just concerned about the default value being 0, but there's a check to prevent that case if the client is enabled. Maybe then we can leave this as is and just make the `cltvDelta` property readonly so it can't be possibly be changed 0.

lib/lndclient/LndClient.ts
offerm: ok

lib/swaps/Swaps.ts
offerm: Added a comment to make it clear that it can't be zero

lib/lndclient/LndClient.ts
offerm: OK

lib/swaps/Swaps.ts
```diff
@@ -342,13 +394,37 @@ class Swaps extends EventEmitter {
     } else {
       // If we are here we are the taker
       this.logger.debug('Executing taker code');
+
+      // Verify that taker got right amount.
```
michael1011: We have `Verify that taker got right amount.` and `verify taker's takerCltvDelta` twice in this file. The logic of those things should be put into a function to avoid redundancy.

lib/lndclient/LndClient.ts
```diff
@@ -70,8 +73,13 @@ class LndClient extends BaseClient {
       this.logger.error('could not find lnd macaroon, is lnd installed?');
       shouldEnable = false;
     }
+    if (cltvdelta) {
+      this.cltvDelta = cltvdelta;
```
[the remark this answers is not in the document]
offerm: Yes, back to as it was originally.

lib/swaps/Swaps.ts
```diff
@@ -37,10 +37,14 @@ type SwapDeal = {
   takerCurrency: string;
   /** Taker's lnd pubkey on the taker currency's network. */
   takerPubKey?: string;
+  /** The CLTV delta from the current height that should be used to set the timelock for the final hop when sanding to taker. */
```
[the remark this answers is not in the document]
offerm: thanks!

lib/swaps/Swaps.ts
```diff
@@ -307,6 +341,23 @@ class Swaps extends EventEmitter {
     // TODO: calculate CLTV
     if (deal.myRole === SwapDealRole.Maker) {
 	  this.logger.debug('Executing maker code');
+
+      // Verify that taker got right amount.
+	  if (amount < deal.makerAmount * 1000) {
```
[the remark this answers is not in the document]
offerm: yep. fixed.

lib/swaps/Swaps.ts
offerm: OK boss.

offerm:
@sangaman @michael1011 anything else?

Issue #398, opened 2018-08-25
Track pending outgoing connections to peers


sangaman:
Right now, on startup it's common to get a flood of `NODES` packets all at once which have a lot of overlap, which then results in multiple outgoing connection attempts to those nodes. There's no mechanism currently to check if we're attempting to connect to a given node. For this issue I suggest that we track a `pendingPeers` Map in Pool which maps peer's node pub keys to the unopened `Peer` object. When the peer connection either succeeds and the open event is fired, or fails and the close event is fired, we remove the peer from `pendingPeers` and add to `peers` if successful.

moshababo:
So you're suggesting to ignore nodes which their pubkey is already listed on `pendingPeers`?

sangaman:
Exactly.

moshababo:
Sounds good to me. If we'll want to support parallelism (specially due to connection retries, which can take ages), we'll need to introduce `address` in addition or instead of `pubKey`.

kilrau:
Introducing `pendingPeers` will ensure we only try to connect to a given `pubKey` once throughout the whole lifetime, right (not only on startup...)? @sangaman 

If yes, that's what we need!

sangaman:
Yeah, well we already have checks in place to not attempt a connection to a peer we're already connected to. This will check to make sure we're not attempting a connection a peer we're already *attempting* to connect to.

kilrau:
Ok thx. As per `pendingPeers` contains `pubKey`@`IP`:`port`, otherwise

sangaman:
I think just nodepubkey would be enough actually, since all outgoing connections reference a peer nodepubkey, we just check if we have a pending peer for the given nodepubkey.

sangaman:
Yeah that's true. But I'm pretty sure I prefer the serial approach, otherwise I think we'd see this same burst of failing connection attempts should anyone have a lot of listening addresses configured.

robertsdotpm:
I'll take this issue

sangaman:
That would be great, I haven't started any work on it yet. Let me know if you have any questions at all.

PR #512, opened 2018-09-22
Prevent connecting to a node if we're already attempting a pending connection

robertsdotpm:
Hello, this is some basic code to prevent making duplicate connections to nodes we're already in the process of talking to.

I also refactored some of the add connection checks because they were getting hard to read and added a new script that allows you to setup a number of virtual interfaces for testing p2p code. I'll probably write some docs on that at some point in the future.


lib/p2p/Pool.ts
```diff
@@ -169,11 +171,28 @@ class Pool extends EventEmitter {
...
       // and that either we haven't heard of it, or we're not ignoring known nodes and it's not banned
-      if (node.nodePubKey !== this.handshakeData.nodePubKey && node.addresses.length > 0 &&
-        (!this.nodes.has(node.nodePubKey) || (!ignoreKnown && !this.nodes.isBanned(node.nodePubKey))) &&
-        !this.peers.has(node.nodePubKey)) {
+      const is_unknown_node = !this.nodes.has(node.nodePubKey);
+      if (!is_unknown_node) {
+        const is_not_ignored = !ignoreKnown && !this.nodes.isBanned(node.nodePubKey);
+
+        // Check we're not already trying to connect to this node.
+        if (node.nodePubKey in this.outgoingConnections) {
```
sangaman: I believe we want to do this check whether we "know" the node or not, right now I don't think this check would work if we have a pending connection to a node we've never connected to before.

lib/p2p/Pool.ts
```diff
@@ -456,6 +475,14 @@ class Pool extends EventEmitter {
     });
   }
 
+  private clearOutgoingConnection = (nodePubKey: any) => {
```
sangaman: `nodePubKey` should be a `string` rather than `any`.

Edit: Although actually, if we go with a `Set` we can just use the `delete()` function it provides and wouldn't need this one.

lib/p2p/Pool.ts
```diff
@@ -49,6 +49,7 @@ interface NodeConnectionIterator {
 class Pool extends EventEmitter {
   /** The local handshake data to be sent to newly connected peers. */
   public handshakeData!: HandshakeState;
+  public outgoingConnections: any = {};
```
sangaman: I think the best data type for this is a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) like `public outgoingConnections = new Set<string>();`. All we're interested in knowing is whether a given `nodePubKey` exists within the set.

I also feel like `pendingConnections` or maybe `pendingOutgoingConnections` would be a better name, since `outgoingConnections` to me could mean established outbound connections. Either way a brief comment explaining what this property is and what it's used for would be good.

Lastly I see this is designated as `public` - is it actually going to be used outside this class? It doesn't look like it is currently, if not it should probably be `private`.

lib/p2p/Pool.ts
```diff
@@ -471,12 +498,14 @@ class Pool extends EventEmitter {
...
     peer.once('open', async () => {
       await this.handleOpen(peer);
+      this.clearOutgoingConnection(peer.nodePubKey);
     });
 
     peer.once('close', () => {
       if (peer.nodePubKey) {
         this.peers.remove(peer.nodePubKey);
       }
+      this.clearOutgoingConnection(peer.nodePubKey);
```
sangaman: I think this line should be inside the `if` statement above, if we don't have a `nodePubKey` value this line won't do anything.

lib/p2p/Pool.ts
```diff
@@ -169,11 +171,28 @@ class Pool extends EventEmitter {
   private connectNodes = (nodes: NodeConnectionIterator, ignoreKnown = false, retryConnecting = false) => {
     const connectionPromises: Promise<void>[] = [];
     nodes.forEach((node) => {
-      // check that this node is not ourselves, that it has listening addresses,
+      const is_not_ignored = false;
```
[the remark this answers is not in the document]
robertsdotpm: Looks like I overwrote some changes and the code to update the dictionary wasn't added. I think I'm going to change editors. I am using Sublime at the moment but I prefer Geany.

moshababo:
@robertsdotpm Thanks for the PR!

I'm not sure I understand the reason for the python script. If it's not related to #398, it should be done in a separate PR. But adding python code to the project isn't a trivial thing, so we should discuss it before (preferably on gitter).

scripts/setup_virtual_network_interfaces.py
```diff
@@ -0,0 +1,120 @@
...
+    replace_network_interface_file(iface_file)
+
+    # Use new config file.
+    restart_network_interfaces(physical_iface_name)
+
+    print("")
+    print("Virtual interfaces should now be up.")
+    print("Type ifconfig to see them listed at their new IP addresses.")
+
+setup_virtual_network_interfaces()
```
sangaman: If we keep this python script, can you add a new line at the end of the file just to be consistent with the rest of the repo?

test/p2p/connection_reuse.ts
```diff
@@ -0,0 +1 @@
+console.log('Test');
```
sangaman: Is this test going to be fleshed out in this PR? If not I figure it can be removed.

lib/p2p/Pool.ts
```diff
@@ -456,6 +476,14 @@ class Pool extends EventEmitter {
     });
   }
 
+  private clearPendingOutgoingConnection = (nodePubKey?: string) => {
```
sangaman: I don't think we really need this method, just using the `pendingOutgoingConnections.delete` method should be fine. You can use `pendingOutgoingConnections.delete(peer.nodePubKey!)` with the `!` non-null assertion operator in the `'open'` event handler, since we already assert elsewhere that on open we'll have a value for `nodePubKey`.

lib/p2p/Pool.ts
```diff
@@ -49,6 +49,7 @@ interface NodeConnectionIterator {
 class Pool extends EventEmitter {
   /** The local handshake data to be sent to newly connected peers. */
   public handshakeData!: HandshakeState;
+  private pendingOutgoingConnections = new Set<string>();
```
sangaman: Maybe add a comment here to clarify what this is for like `/** A set of pub keys of nodes for which we have pending outgoing connections. */`.

sangaman:
I think the python script is intended for developers only and not as a dependency for end users. Agree that it should probably be in a separate PR though.

scripts/setup_virtual_network_interfaces.py
robertsdotpm: Removed for now. Will do a new PR in the future if its needed

test/p2p/connection_reuse.ts
robertsdotpm: removed

lib/p2p/Pool.ts
robertsdotpm: Nice, I was wondering how to do that :+1:

lib/p2p/Pool.ts
robertsdotpm: Done :speech_balloon:  :+1:

tslint-alt.json
```diff
@@ -6,5 +6,10 @@
     "no-unnecessary-type-assertion": false,
     "await-promise": false,
     "no-circular-imports": false
+  },
+  "linterOptions": {
```
sangaman: Could you also take this out of the PR? It should go with the python code.

lib/p2p/Pool.ts
```diff
@@ -169,11 +172,29 @@ class Pool extends EventEmitter {
   private connectNodes = (nodes: NodeConnectionIterator, ignoreKnown = false, retryConnecting = false) => {
     const connectionPromises: Promise<void>[] = [];
     nodes.forEach((node) => {
-      // check that this node is not ourselves, that it has listening addresses,
-      // and that either we haven't heard of it, or we're not ignoring known nodes and it's not banned
-      if (node.nodePubKey !== this.handshakeData.nodePubKey && node.addresses.length > 0 &&
-        (!this.nodes.has(node.nodePubKey) || (!ignoreKnown && !this.nodes.isBanned(node.nodePubKey))) &&
-        !this.peers.has(node.nodePubKey)) {
+      let isNotIgnored = false;
```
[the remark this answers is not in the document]
michael1011: @sangaman yeah lets leave it as is then.

kilrau:
Ready to merge?

PR #515, opened 2018-09-22
feat(SWAPS): cltv-delta Part 2

### LINKED ISSUE #413: Swap: failures & other exceptions
Successful swaps are covered in https://github.com/ExchangeUnion/xud/issues/422, this issue is about implementing swap failure events (https://github.com/ExchangeUnion/xud/pull/380 prepared events for this).

* `cancelSwap` packet, which can be sent by maker or taker any time during the 3s interval to 
cancel the swap as a bilateral agreement. -> remove peer order, continue matching
* swap time-out (3s from DEAL_RESPONSE to Pre-image release) -> remove peer order, dummy penalize event and continue matching
* order invalidated -> remove peer order, dummy penalize event and continue matching
* check if rate of incoming HTLC matches DEAL rate, if not cancel swap and try next
* generic error -> ?

### LINKED ISSUE #421: SWAP: protocol improvements
1. Eliminate ` SWAP_REQUEST` packet: taker creates the preimage and hash and initiates the swap
2. Eliminate `SWAP_RESPONSE` by adding a function to lnd to stream / be queried for preimage. So the maker's `xud` can forward the preimage from one `xud` to the other without waiting for the taker to send it over the wire --> check required changes

Summary: in order to support partial matches we decided to go for:

> Swap protocol V2
> Swap version 2 reduces the need for swapRequest message from the taker to the maker and allows the actual swap to start sooner. To do that the following changes take place:
> 
> The r_hash and the r_preImage are set by the taker and not by the maker.
> The taker sends to the maker aswapRequest message which includes the r_hash.
> The maker confirms with swapResponse message
> The taker initiates the swap
> upon success the taker issue a swapResponse message (to be renamed to swapConmpleted) to the maker.
> Impact on Deal Status
> swapRequested status is not needed
> amountSent - set when sending amounts with sendPaymentSync. Prior status dealAgreed when maker and dealRequested when taker
> swapDone - set by taker when getting a response from sendPaymentSync which includes the preImage. Prior status should be preImagerevealed. Set by maker when getting from maker the swapResponse message with the preImage.

### LINKED ISSUE #431: SWAP: Proper timelock calculation

offerm:
closes #431 
closes #421
closes #413 (with the exception of RPC timeouts which deserve a separate issue)

Maker to Taker routes
Place holder for isOrderOnBook - checks by maker before accepting a swap
Place holder for checkAndHoldAmount - uses by maker before accepting a swap. currently can't return partial amount (all or nothing).
Verify LND setup - checks that peer has LNDs PubKeys for BTC and LTC
Queryroutes - use queryroutes to check if there is a route between the maker and the taker and the implication of the route on the maker's cltvDelta.
Maker's cltvDelta better calcualtion.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -293,6 +293,20 @@ class LndClient extends BaseClient {
   }
 
   /**
+   * List all routes to destination.
+   */
+  public queryRoutes = (request: lndrpc.QueryRoutesRequest): Promise<lndrpc.QueryRoutesResponse> => {
+    return this.unaryCallNative<lndrpc.QueryRoutesRequest, lndrpc.QueryRoutesResponse>('queryRoutes', request);
+  }
+
+  /**
+   * Send amount to destination using pre-defined routes.
+   */
+  public sendToRouteSync = (request: lndrpc.SendToRouteRequest): Promise<lndrpc.SendResponse> => {
+    return this.unaryCallNative<lndrpc.SendToRouteRequest, lndrpc.SendResponse>('sendToRouteSync', request);
+  }
+
+  /**
    * Attempt to close an open channel.
    */
   public closeChannel = (fundingTxId: string, outputIndex: number, force: boolean): void => {
--- a/lib/p2p/packets/types/SwapErrorPacket.ts
+++ b/lib/p2p/packets/types/SwapErrorPacket.ts
@@ -12,7 +12,11 @@ class SwapErrorPacket extends Packet<SwapErrorPacketBody> {
     return PacketType.SwapError;
   }
 
-  public get direction() {
+  public get direction(): PacketDirection {
+    // SwapErrorPacket may serve as a response to SwapRequest packet
+    if (this.header.reqId) {
+      return PacketDirection.Response;
+    }
     return PacketDirection.Unilateral;
   }
 }
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -48,6 +48,8 @@ type SwapDeal = {
   /** The hash of the preimage. */
   r_hash: string;
   r_preimage?: string;
+  /** The routes the maker should use to send to the taker. */
+  makerToTakerRoutes?: lndrpc.Route[];
   createTime: number;
   executeTime?: number;
   competionTime?: number
@@ -105,6 +107,63 @@ class Swaps extends EventEmitter {
   }
 
   /**
+   * Verifies that the order still exists on orderBook
+   */
+  // TODO: replace the dummy implementation with real order book integration
+  private isOrderOnBook = (pairId: string, orderId: string): boolean => {
+    this.logger.debug('checking order book for ' + pairId + ':' + orderId);
+    return true;
+  }
+
+  /**
+   * Verifies that the order still exists on orderBook
+   */
+  // TODO: replace the dummy implementation with real order book integration
+  private checkAndHoldAmount = (deal: SwapDeal): number => {
+    this.logger.debug('checking and holding amount for swap ' + deal.pairId + ':' + deal.orderId + ':' + deal.takerAmount);
+    return deal.takerAmount;
+  }
+
+  /**
+   * Sends an error to peer. set reqId if packet is a response to a request.
+   */
+  private sendErrorToPeer = (peer: Peer, r_hash: string, errorMessage: string, reqId?: string) => {
+    const errorBody: packets.SwapErrorPacketBody = {
+      r_hash,
+      errorMessage,
+    };
+    this.logger.debug('Sending swap error to peer: ' + JSON.stringify(errorBody));
+    peer.sendPacket(new packets.SwapErrorPacket(errorBody, reqId));
+    return;
+  }
+
+  /**
+   * Verifies LND setup. Make sure we are connected to BTC and LTC and that
+   * the peer is also connected to these networks. Returns an error message
+   * or undefined in case all is good.
+   */
+  private verifyLndSetup = (deal: SwapDeal, peer: Peer) => {
+    if (!peer.getLndPubKey(deal.takerCurrency)) {
+      return 'peer did not provide an LND PubKey for ' + deal.takerCurrency;
+    }
+
+    if (!peer.getLndPubKey(deal.makerCurrency)) {
+      return 'peer did not provide an LND PubKey for ' + deal.makerCurrency;
+    }
+
+    // verify that this node is connected to BTC and LTC networks
+    if (!this.lndLtcClient.isConnected()) {
+      return 'Can not swap. Not connected to LTC network';
+    }
+
+    if (!this.lndBtcClient.isConnected()) {
+      return 'Can not swap. Not connected to BTC network';
+    }
+
+    return;
+  }
+
+  /**
    * Gets a deal by its r_hash value.
    * @param r_hash The r_hash value of the deal to get.
    * @returns A deal if one is found, otherwise undefined.
@@ -131,7 +190,6 @@ class Swaps extends EventEmitter {
   public beginSwap = (maker: StampedPeerOrder, taker: StampedOwnOrder) => {
     const peer = this.pool.getPeer(maker.peerPubKey);
 
-    // TODO: check lndPubKeys. Error if we or maker do not have full set of keys
     // TODO: check route to peer. Maybe there is no route or no capacity to send the amount
     // TODO: what is the status of the order here? is it off the book? What if partial match
     //       do we create another order which has the same orderId?
@@ -187,6 +245,14 @@ class Swaps extends EventEmitter {
 
     this.addDeal(deal);
 
+    // Verify LND setup. Make sure we are connected to BTC and LTC and that
+    // the peer is also connected to these networks.
+    const errMsg = this.verifyLndSetup(deal, peer);
+    if (errMsg) {
+      this.logger.error(errMsg);
+      this.setDealState(deal, SwapDealState.Error, errMsg);
+      return;
+    }
     peer.sendPacket(new packets.SwapRequestPacket(swapRequestBody));
 
     this.setDealPhase(deal, SwapDealPhase.SwapRequested);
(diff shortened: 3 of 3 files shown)


lib/p2p/packets/types/SwapErrorPacket.ts
```diff
@@ -12,7 +12,11 @@ class SwapErrorPacket extends Packet<SwapErrorPacketBody> {
     return PacketType.SWAP_ERROR;
   }
 
-  public get direction() {
+  public get direction(): PacketDirection {
+    // SwapErrorPacket may server as a response to SwapRequest packet
```
michael1011: I think this should be `may serve as a`

lib/swaps/Swaps.ts
```diff
@@ -104,6 +106,60 @@ class Swaps extends EventEmitter {
     this.pool.on('packet.swapError', this.handleSwapError);
   }
 
+  /**
+   * verify that the order still exists on orderBook
+   */
+  private isOrderOnBook = (pairId: string, orderId: string): boolean => {
+    this.logger.debug('checking order book for ' + pairId + ':' + orderId);
+    return true;
```
michael1011: A followup PR should replace this dummy method with an actual orderbook check. Please add a `TODO` comment

lib/swaps/Swaps.ts
```diff
@@ -104,6 +106,60 @@ class Swaps extends EventEmitter {
...
+   * verify LND setup. Make sure we are connected to BTC and LTC and that
+   * the peer is also connected to these networks.
+   */
+  private verifyLNDSetup = (deal: SwapDeal, peer: Peer) => {
+    if (!peer.getLndPubKey(deal.takerCurrency)) {
+      return 'peer did not provide an LND PubKey for ' + deal.takerCurrency;
+    }
+
+    if (!peer.getLndPubKey(deal.makerCurrency)) {
+      return 'peer did not provide an LND PubKey for ' + deal.makerCurrency;
```
michael1011: The log messages in this method are redundant. Could you put them into a method?

lib/swaps/Swaps.ts
```diff
@@ -104,6 +106,60 @@ class Swaps extends EventEmitter {
...
+    this.logger.debug('Sending swap error to peer: ' + JSON.stringify(errorBody));
+    peer.sendPacket(new packets.SwapErrorPacket(errorBody, reqId));
+    return;
+  }
+
+  /**
+   * verify LND setup. Make sure we are connected to BTC and LTC and that
+   * the peer is also connected to these networks.
+   */
+  private verifyLNDSetup = (deal: SwapDeal, peer: Peer) => {
```
sangaman: Can you name this `verifyLndSetup` to stick to camelCase naming?

lib/swaps/Swaps.ts
```diff
@@ -187,6 +243,14 @@ class Swaps extends EventEmitter {
 
     this.addDeal(deal);
 
+    // verify LND setup. Make sure we are connected to BTC and LTC and that
+    // the peer is also connected to these networks.
+    const err = this.verifyLNDSetup(deal, peer);
```
sangaman: I think it's a bit clear to to rename this to `errMsg` or similar to clarify that this is a message and not an "error" in the sense of a NodeJS `Error` or one of our internal error objects with a corresponding code.

lib/swaps/Swaps.ts
```diff
@@ -233,10 +290,89 @@ class Swaps extends EventEmitter {
       createTime: Date.now(),
     };
 
+    // add the deal. Going forward we can "record" errors related to this deal.
     this.addDeal(deal);
 
+    // verify LND setup. Make sure we are connected to BTC and LTC and that
+    // the peer is also connected to these networks.
+    const err = this.verifyLNDSetup(deal, peer);
```
sangaman: Same here regarding the naming of `err`.

lib/swaps/Swaps.ts
```diff
@@ -233,10 +290,89 @@ class Swaps extends EventEmitter {
...
+    if (err) {
+      this.setDealState(deal, SwapDealState.Error, err);
+      this.sendErrorToPeer(peer, deal.r_hash, deal.stateReason,  requestPacket.header.id);
```
sangaman: I see the back to back `setDealState` then `sendErrorToPeer` many times and it seems like we will always do it this way. Maybe the two can be combined in a single function that handles everything - simplest would just be to set the deal state when we send the error to peer. 

Also very minor nit but there are two spaces instead of one before `requestPacket` everywhere this line appears, fix at your leisure.

lib/swaps/Swaps.ts
```diff
@@ -233,10 +290,89 @@ class Swaps extends EventEmitter {
...
+      return;
+    }
+
+    const heldAmount = this.checkAndHoldAmount(deal);
+    assert(heldAmount === 0 || heldAmount === deal.takerAmount,
+        'Partial amount is not yet supported');
+    if (heldAmount === 0) {
+      this.setDealState(deal, SwapDealState.Error, 'Can not swap. No capacity is available');
```
sangaman: "Insufficient capacity" would be more accurate right?

lib/swaps/Swaps.ts
```diff
@@ -233,10 +290,89 @@ class Swaps extends EventEmitter {
...
+        this.sendErrorToPeer(peer, deal.r_hash, deal.stateReason,  requestPacket.header.id);
+        return;
+    }
+
+    let height: number;
+    try {
+      const req = new lndrpc.QueryRoutesRequest();
+      req.setAmt(requestBody.takerAmount);
+      req.setFinalCltvDelta(requestBody.takerCltvDelta);
+      req.setNumRoutes(1);
```
sangaman: Do we want to query for more than 1 route in case payment on the first route fails for some reason? Or would this be a future enhancement, and if so maybe add a TODO?

lib/swaps/Swaps.ts
```diff
@@ -233,10 +290,89 @@ class Swaps extends EventEmitter {
...
+      const routes = await lndclient.queryRoutes(req);
+      deal.makerToTakerRoutes = routes.getRoutesList();
+      this.logger.debug('got ' + deal.makerToTakerRoutes.length + ' routes to destination: ' + deal.makerToTakerRoutes);
+      if (deal.makerToTakerRoutes.length === 0) {
+        this.setDealState(deal, SwapDealState.Error, 'Can not swap. Zero routes to destination found.');
+        this.sendErrorToPeer(peer, deal.r_hash, deal.stateReason,  requestPacket.header.id);
+        return;
+      }
+    } catch (err) {
+      this.setDealState(deal, SwapDealState.Error, 'Can not swap. No route to destination found: ' + err.message);
```
sangaman: I'd prefer the "unable to find route to destination" wording - "no route found" and "zero routes found" are synonymous to me but if we are in this catch block it's probably due to some error unrelated to the number of routes that exist.

lib/swaps/Swaps.ts
```diff
@@ -233,10 +290,89 @@ class Swaps extends EventEmitter {
...
+      this.logger.debug('got block height of ' + height);
+    } catch (err) {
+      this.setDealState(deal, SwapDealState.Error, 'Can not swap. Unable to fetch block height: ' + err.message);
+      this.sendErrorToPeer(peer, deal.r_hash, deal.stateReason,  requestPacket.header.id);
+      return;
+    }
+
+    const routeCltvDelta = deal.makerToTakerRoutes[0].getTotalTimeLock() - height;
+
+    // cltvDelta can't be zero for both the LtcClient and BtcClient (constractor)
```
sangaman: typo in "constractor" - and maybe I'd also expand a little to say "(checked in constructor)".

lib/p2p/packets/types/SwapErrorPacket.ts
offerm: OK

lib/swaps/Swaps.ts
offerm: Will add a TODO.
Will keep the method as it marks where handling should be and allows Error handling.

Now that we are in good shape with regards to Swap, @sangaman can implement the order book integration.

as a side note: No need to be fanatic about a circular reference.  You need to avoid it with fast moving objects to allow the GC to do a good job. IMHO This is not the case here. Clearly, it is better to have a circular reference here than moving the code to the order book....

Anyway, I will leave it to you to design how the order book changes should be done.  You may like to break the code of Swaps.ts between the different modules.

lib/swaps/Swaps.ts
offerm: Don't see real value in making it a method now. will leave it like this.

lib/lndclient/LndClient.ts
```diff
@@ -208,8 +208,8 @@ class LndClient extends BaseClient {
    * Return general information concerning the lightning node including it’s identity pubkey, alias, the chains it
    * is connected to, and information concerning the number of open+pending channels.
    */
-  private getInfo = (): Promise<lndrpc.GetInfoResponse.AsObject> => {
-    return this.unaryCall<lndrpc.GetInfoRequest, lndrpc.GetInfoResponse.AsObject>('getInfo', new lndrpc.GetInfoRequest());
+  public getInfo = (): Promise<lndrpc.GetInfoResponse> => {
```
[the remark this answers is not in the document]
offerm: @sangaman so you like me to revert the getInfo code to use object? I can do it. I made the change since I need now to call getInfo to get blockHeight when checking CLTV for the maker to taker route.
note that getInfo was a private method with no external usage.


Do you think I should also do the same for queryRoutes? transform the routes to objects and rebuild them for sendToRoutes? Big NO on this.

lib/lndclient/LndClient.ts
```diff
@@ -292,6 +292,20 @@ class LndClient extends BaseClient {
...
+   * List all routes to destination.
+   */
+  public queryRoutes = (request: lndrpc.QueryRoutesRequest): Promise<lndrpc.QueryRoutesResponse> => {
+    return this.unaryCallNative<lndrpc.QueryRoutesRequest, lndrpc.QueryRoutesResponse>('queryRoutes', request);
+  }
+
+  /**
+   * Send amount to destination using pre-defined routes.
+   */
+  public sendToRouteSync = (request: lndrpc.SendToRouteRequest): Promise<lndrpc.SendResponse> => {
```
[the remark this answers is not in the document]
offerm: We are still using it when making the taker payment.

lib/swaps/Swaps.ts
```diff
@@ -104,6 +106,60 @@ class Swaps extends EventEmitter {
...
+   */
+  private isOrderOnBook = (pairId: string, orderId: string): boolean => {
+    this.logger.debug('checking order book for ' + pairId + ':' + orderId);
+    return true;
+  }
+
+  /**
+   * verify that the order still exists on orderBook
+   */
+  private checkAndHoldAmount = (deal: SwapDeal): number => {
```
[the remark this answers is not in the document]
offerm: Added TODO. 
remove it as part of #495

lib/swaps/Swaps.ts
offerm: Fixed the 2 spaces. Thanks.

setDealState may be used without sending Error to Peer.
Don't think these should be combined.

lib/swaps/Swaps.ts
offerm: not really.

This case is really "no capacity". If there is some capacity we should try a reduced amount.

lib/swaps/Swaps.ts
offerm: done

lib/swaps/Swaps.ts
offerm: done

lib/lndclient/LndClient.ts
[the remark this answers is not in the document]
sangaman: No, I wouldn't say you have to change `queryRoutes` since it's new code anyway and a core piece of this PR. If you even just want to split this into a separate commit but keep it in the PR that'd be better, but refactoring everything in a single PR would be cleaner imo. I'm mainly just thinking about maintaining a sensible commit history which I think is becoming increasingly important.

lib/swaps/Swaps.ts
sangaman: OK I was thinking this referred to the lightning routing capacity but for available order quantity yes that is correct, nm.

PR #518, opened 2018-09-23
feat: update handshake on lnd pubkey change

### LINKED ISSUE #513: Peer connection and LND info
For executing a swap the taker and maker should be ower of the peer PubKeys of BTC/LTC.

This information is now provided as part of the handshake when the nodes connect.

Problem: Assume A and B (nodes). A connects to B. at the time A connects to B it's LND processes are down so A does not have PubKeys to share with B. Later the LNDs processes are up and A connects to them but B is unaware of PubKeys and can't swap with A.

sangaman:
Closes #513.

This commit creates an event for when a LightningClient verifies (or re-verifies) a connection to an lnd server. It includes the lnd identity pub key if it changed. When the pub key changes, the new value
is broadcast to all currently connected peers within a new Hello packet. This allows for switches of lnd servers without requiring a restart of xud, and also will inform peers of lnd pub keys when lnd comes back online should lnd be offline at the time of peer connection.

I've tested this locally and it seems to be working well. Here's what the log output looks like on the remote peer receiving the handshake update, below is the original and the updated handshake.

```
2018-9-23 15:46:03 [P2P] verbose: received hello packet from 127.0.0.1:9885: {"version":"1.0.0-prealpha.4","pairs":["LTC/BTC"],"nodePubKey":"029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345","addresses":[{"host":"75.75.58.83","port":8885},{"host":"127.0.0.1","port":8885}]}
...
2018-9-23 15:47:06 [P2P] verbose: received hello packet from 029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345: {"version":"1.0.0-prealpha.4","pairs":["LTC/BTC"],"nodePubKey":"029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345","lndbtcPubKey":"036c566039991ed0203667b9d9ac19cd67a8230de3761adbca80a6d54ab4ed7a54","addresses":[{"host":"75.75.58.83","port":8885},{"host":"127.0.0.1","port":8885}]}

### DIFF
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -12,7 +12,6 @@ import NodeKey from './nodekey/NodeKey';
 import Service from './service/Service';
 import { EventEmitter } from 'events';
 import Swaps from './swaps/Swaps';
-import path from 'path';
 
 const version: string = require('../package.json').version;
 
@@ -141,9 +140,24 @@ class Xud extends EventEmitter {
       } else {
         this.logger.warn('RPC server is disabled.');
       }
+
+      this.bind();
     } catch (err) {
       this.logger.error(err);
     }
+  }
+
+  private bind = () => {
+    this.lndbtcClient.on('connectionVerified', (newPubKey) => {
+      if (newPubKey) {
+        this.pool.updateHandshake({ lndbtcPubKey: newPubKey });
+      }
+    });
+    this.lndltcClient.on('connectionVerified', (newPubKey) => {
+      if (newPubKey) {
+        this.pool.updateHandshake({ lndltcPubKey: newPubKey });
+      }
+    });
   }
 
   private shutdown = async () => {
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -558,6 +558,15 @@ class Peer extends EventEmitter {
   }
 
   private handleHello = (packet: packets.HelloPacket): void => {
+    const helloBody = packet.body!;
+    this.logger.verbose(`received hello packet from ${this.nodePubKey || addressUtils.toString(this.address)}: ${JSON.stringify(helloBody)}`);
+    if (this.nodePubKey && this.nodePubKey !== helloBody.nodePubKey) {
+      // peers cannot change their nodepubkey while we are connected to them
+      // TODO: penalize?
+      this.close();
+      return;
+    }
+
     this.handshakeState = packet.body;
 
     const entry = this.responseMap.get(PacketType.Hello);
@@ -628,4 +637,4 @@ class Job {
 }
 
 export default Peer;
-export { HandshakeState, PeerInfo };
+export { PeerInfo };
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: lib/Xud.ts [hunk 1]
```diff
@@ -141,11 +142,26 @@ class Xud extends EventEmitter {
...
         this.logger.warn('RPC server is disabled.');
       }
+
+      this.bindXud();
     } catch (err) {
       this.logger.error(err);
     }
   }
 
+  private bindXud = () => {
```
michael1011: Why `bindXud`? Wouldn't `bindLnd(s)` be more appropriate?

sangaman: I was thinking to make it more generic for handling future events we need to listen to. But maybe I should just call this `bind`? That's the naming in Swaps.ts. Elsewhere we have "bind" functions they're binding a particular object, like `bindPeer` or `bindPool`.

michael1011: :+1: for `bind`

offerm:
Maybe it is even better to disconnect from all peers and stop P2p until
LNDs are connected. Making LNDs mandatory if not  disabled .

sangaman:
I think requiring connectivity to LND is definitely a good idea. I don't think we need to entirely disconnect from peers necessarily, but dropping all orders for pairs that involve a network where we've lost connectivity to lnd makes sense. I think it makes more sense for a future PR though, and maybe only once the swaps functionality is a bit more mature.

lib/p2p/Peer.ts
```diff
@@ -558,6 +558,14 @@ class Peer extends EventEmitter {
   }
 
   private handleHello = (packet: packets.HelloPacket): void => {
+    const handshakeData = packet.body!;
+    this.logger.verbose(`received hello packet from ${this.nodePubKey || addressUtils.toString(this.address)}: ${JSON.stringify(handshakeData)}`);
+    if (this.nodePubKey && this.nodePubKey !== handshakeData.nodePubKey) {
+      // peers cannot change their nodepubkey while we are connected to them
+      // TODO: penalize?
```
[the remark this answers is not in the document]
sangaman: I can call `close()` from hereDisconnecting makes sense, although I think penalizing in addition makes sense too right?

lib/Xud.ts
```diff
@@ -141,11 +142,26 @@ class Xud extends EventEmitter {
...
+    this.lndbtcClient.on('connectionVerified', (newPubKey) => {
+      if (newPubKey) {
+        this.pool.updateHandshake({ lndbtcPubKey: newPubKey });
+      }
+    });
+    this.lndltcClient.on('connectionVerified', (newPubKey) => {
+      if (newPubKey) {
+        this.pool.updateHandshake({ lndltcPubKey: newPubKey });
+      }
+    });
```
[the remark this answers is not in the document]
sangaman: Yeah, I was thinking that it would be a bit much for Pool having to know about the LndClients

PR #523, opened 2018-09-24
test: use dynamic port for web proxy tests

sangaman:
This changes the web proxy tests to use a dynamic, unused port instead of the default 8080 port, which could cause the tests to fail if 8080 is being used.


test/integration/WebProxy.spec.ts
```diff
@@ -11,7 +12,7 @@ describe('WebProxy', () => {
     config = {
       webproxy: {
         disable: false,
-        port: 8080,
+        port: await getUnusedPort(),
```
moshababo: I'd rather have this value be assigned to a var in the beginning of the file/test, in case more usages will pop out.

PR #525, opened 2018-09-25
refactor: use positive quantity for buys & sells

sangaman:
This PR refactors how we represent buy and sell orders. Previously, sell orders were represented with a negative quantity. However, I believe this to be potentially confusing especially in regards to the API and quantity updates to existing orders, and this was echoed in our discussions. Instead, orders now have a `side` property to indicate whether they are buys or sells, and all standing orders and order matches are represented with positive quantities.

This passes all tests as well as some limited local testing of mine, namely manually placing buy and sell orders via the cli and ensuring they match as expected.


sangaman:
Per our discussions, I went with representing the side internally via an `isBuy` boolean property on orders.

PR #532, opened 2018-09-25
perf: improve parallel await on init

sangaman:
This fixes several instances in the code during the initialization process for `xud` where we're unnecessarily awaiting for promises to resolve serially, instead awaiting for them in parallel to shave a bit of time off the startup process.

### DIFF
--- a/lib/db/DB.ts
+++ b/lib/db/DB.ts
@@ -57,8 +57,9 @@ class DB {
 
     if (newDb && initDb) {
       // populate new databases with default data
+      const firstStagePromises: Bluebird<any>[] = [];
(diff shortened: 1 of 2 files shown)


lib/db/DB.ts
```diff
@@ -57,8 +57,9 @@ class DB {
 
     if (newDb && initDb) {
       // populate new databases with default data
+      const firstStagePromises: Bluebird<any>[] = [];
```
michael1011: Why not just `promises`?

sangaman: Oops, that's kind of a leftover from when I was using a different approach with two arrays of promises, will correct it.

PR #535, opened 2018-09-28
chore: add VS Code settings

michael1011:
This commit adds Visual Studio Code debug configurations for the tests and XUD itself and a few settings that should make creating new TypeScript files easier. It also removes those configuration files from the `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,9 +54,6 @@ dist/
 # Yarn Integrity file
 .yarn-integrity
 
-# Visual Studio Code config
-.vscode/
-
 #  Webstorm config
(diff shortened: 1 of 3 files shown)


sangaman:
I'm a bit hesitant to check this in to the repository - for example my config is quite a bit different from this yet it's not entirely stuff I'd want to check in, but it's also stuff I wouldn't want to lose when checking out code. What about just putting it in the wiki?

michael1011:
I am fine with not merging and putting them into the wiki or just *somewhere* where they can be found. @sangaman

PR #536, opened 2018-09-28
chore: rename cancelOrder as removeOrder

### LINKED ISSUE #533: rename cancelOrder as removeOrder
https://github.com/ExchangeUnion/xud/pull/500/files#r220511799

Renaming should be applied to gRPC / CLI / `Service.ts`

ImmanuelSegol:
Closes #533


proto/xudrpc.proto
```diff
@@ -42,9 +42,9 @@ service Xud {
   }
 
   /* Cancel placed order from the orderbook. */
```
sangaman: Can you update this comment as well? I think just something simple like "Removes an order from the order book by its local id." would be fine for now.

proto/xudrpc.proto
```diff
@@ -187,11 +187,11 @@ message AddPairRequest {
 }
 message AddPairResponse {}
 
-message CancelOrderRequest {
+message RemoveOrderRequest {
   // The local id of the order to cancel
```
sangaman: "remove" instead of "cancel"

lib/service/Service.ts
```diff
@@ -111,7 +111,7 @@ class Service extends EventEmitter {
   /*
    * Cancel placed order from the orderbook.
```
sangaman: Update to the same comment as in the proto file.

test/integration/Service.spec.ts
```diff
@@ -94,7 +94,7 @@ describe('API Service', () => {
     const args = {
```
sangaman: Also update the description of this test case.

sangaman:
Thanks! I'm thinking it might be good to do a find through the codebase for "cancel" to see if there are any lingering references to the old cancel order terminology.

sangaman:
Oh and one minor housekeeping item for the commit message, I think "chore" is supposed to be for build/project related items, this should probably be a "refactor". I can take care of it when squashing tho.

PR #538, opened 2018-09-28
fix :Deprecation warning 

ImmanuelSegol:
This is a simple fix we used at [`walli-server`](https://github.com/dopetard/walli-server) it simply changes `new Buffer` to `Buffer.from`.


michael1011:
@sangaman We did test is for all three operating systems before merging it in to walli-server.

Issue #114, opened 2018-06-08
Generate and use TLS certificates for gRPC layer


sangaman:
Currently, our gRPC layer is not using a TLS certificate. We'll need to generate one as part of our initialization process if none exists, and use it for all rpc calls similar to how we use the `lnd` cert for calls in `LndService.ts`.

kilrau:
Includes Macaroons authentication on RPC Calls

sangaman:
Closing this and will open a separate issue for macaroons.

PR #540, opened 2018-09-30
feat(grpc): add SSL to gRPC

michael1011:
`This commit adds SSL encryption to the gRPC server. If the needed certificate doesn't exist a new one will be generated and stored in the data directory of XUD.`

This PR addresses #114 just partially. Initially I wanted to do the SSL and macaroon stuff in one PR but I wasn't able to find a good library for generating and verifying macaroons. The best I found was [`macaroons.js`](https://github.com/nitram509/macaroons.js) which is written in TypeScript but the typings for that library don't work for some reason. Should we just fork `macaroons.js`, generate proper typings and publish those to npm?


.env
```diff
@@ -1 +1 @@
-GRPC_SSL_CIPHER_SUITES='HIGH+ECDSA'
+GRPC_SSL_CIPHER_SUITES='HIGH+ECDSA:ECDHE-RSA-AES128-GCM-SHA256'
```
sangaman: I'm not too familiar with this, but is using the `HIGH+ECDSA` for our cert an option?

bin/xucli
```diff
@@ -17,6 +22,11 @@ require('yargs')
       describe: 'RPC service hostname',
       type: 'string',
     },
+    'rpc.tlscertpath': {
```
sangaman: I think we can drop the `rpc.` prefix, that's more for the config structure and here it's clear that the cli is interacting with our rpc api.

bin/xucli
```diff
@@ -17,6 +22,11 @@ require('yargs')
...
+      alias: 'c',
+      describe: 'Path to the TLS certificate of XUD',
```
sangaman: I think we ought to standardize on either "XUD" or "xud" for the name in regular text, my preference is probably "xud" but lets try to discuss whenever it's convenient.

lib/grpc/GrpcServer.ts
```diff
@@ -70,6 +85,56 @@ class GrpcServer {
       });
     });
   }
+
+  private generateCertificate = (tlsCertPath: string, tlsKeyPath: string) => {
```
sangaman: Maybe add a brief method comment block here explaining that this generates a certificate and saves it to disk, the latter part is not obvious from the name.

lib/grpc/GrpcServer.ts
```diff
@@ -70,6 +85,56 @@ class GrpcServer {
...
+    const keys = pki.rsa.generateKeyPair(1024);
+    const cert = pki.createCertificate();
+
+    cert.publicKey = keys.publicKey;
+    cert.serialNumber = String(Math.floor(Math.random() * 1024) + 1);
+
+    const date = new Date();
+    cert.validity.notBefore = date;
+    cert.validity.notAfter = new Date(date.getFullYear() + 1, date.getMonth(), date.getDay());
```
sangaman: We're making it valid for a year, will we need a validity check to regenerate once it expires? How did you choose a year?

lib/grpc/GrpcServer.ts
```diff
@@ -70,6 +85,56 @@ class GrpcServer {
...
+      {
+        name: 'subjectAltName',
+        altNames: [
+          {
+            type: 2,
+            value: 'localhost',
+          },
+          {
+            type: 7,
+            ip: '127.0.0.1',
```
sangaman: At some point I think we'll definitely want to support additional IP addresses/domain names in the cert, similar to how lnd has `tlsextradomain` and `tlsextraip`. Doesn't have to be in this PR tho, but maybe add a TODO and lets make sure to create an issue after this is merged.

lib/grpc/GrpcServer.ts
```diff
@@ -44,10 +47,22 @@ class GrpcServer {
...
    * @returns true if the server started listening successfully, false otherwise
    */
-  public listen = (port: number, host: string) => {
+  public listen = async (port: number, host: string, tlsCertPath: string, tlsKeyPath: string): Promise<boolean> => {
     assert(Number.isInteger(port) && port > 1023 && port < 65536, 'port must be an integer between 1024 and 65535');
 
-    const bindCode = this.server.bind(`${host}:${port}`, grpc.ServerCredentials.createInsecure());
+    if (!fs.existsSync(tlsCertPath)) {
+      this.logger.debug('Could not find gRPC TLS certificate. Generating new one');
+      this.generateCertificate(tlsCertPath, tlsKeyPath);
```
sangaman: `generateCertificate` writes files to disk, and then we read the files from disk later in this method. It could maybe return a  `{ cert_chain, private_key }` object to then be used below when we just generated the cert, otherwise read from disk.

lib/grpc/webproxy/GrpcWebProxyServer.ts
```diff
@@ -23,10 +25,15 @@ class GrpcWebProxyServer {
...
    */
-  public listen = (proxyPort: number, grpcPort: number, grpcHost: string): Promise<void> => {
+  public listen = (proxyPort: number, grpcPort: number, grpcHost: string, tlsCertPath: string): Promise<void> => {
     // Load the proxy on / URL
     const protoPath = path.join(__dirname, '..', '..', '..', 'proto');
-    const gateway = grpcGateway(['xudrpc.proto'], `${grpcHost}:${grpcPort}`, grpc.credentials.createInsecure(), protoPath);
+    const gateway = grpcGateway(
+      ['xudrpc.proto'],
+      `${grpcHost}:${grpcPort}`,
+      grpc.credentials.createSsl(fs.readFileSync(tlsCertPath)),
```
sangaman: Right now we're reading this file twice, once when starting the grpc server and once for the web proxy. It would be nice if we only read it once, but no particularly clean way to do that jumps out at me. Maybe the `listen` method for GrpcServer could return the loaded cert, to then be used optionally for the web proxy? Not a big deal though so do you what you think is best.

.env
michael1011: I used [`node-forge`](https://github.com/digitalbazaar/forge) which is purely written in JavaScript for generating the certificate and that library has unfortunately no support for ECDSA signatures

But if we *really* want to use ECDSA we could use a wrapper for the CLI of `openssl` (which is tricky to install on Windows) or write a simple C++ application that generates the certificate with the `openssl` library (shouldn't be a problem to compile since we need `node-gyp` for other dependencies anyways).

What's your opinion? @sangaman

lib/grpc/GrpcServer.ts
michael1011: I wasn't sure about that one so I checked what LND is doing and they are making the certificate valid for 14 months. So I just made it valid for a year.

But maybe we should remove remove the validity completely for now? @sangaman

.env
sangaman: Gotcha, doesn't seem like it's worth the trouble. I figure this should be fine, but your idea sounds good if we ever do want to implement it that way.

lib/grpc/GrpcServer.ts
sangaman: Yeah, if you can make it valid indefinitely (or like 5+ years) for now that's probably fine. You can also add a TODO here to revisit how we want to handle expired certificates perhaps.

bin/xucli
michael1011: We should drop it for all options (`port` and `host`) then. But I think we should keep it. It makes more clear what is meant and the user doesn't have to type more because they will be using the alias `-c` anyways

lib/grpc/webproxy/GrpcWebProxyServer.ts
michael1011: That way the gRPC server would take the path to the certificate as argument but the web proxy the certificate itself and I don't want to move the generation of the certificate into `Xud.ts` because it doesn't belong there.

I think that reading a file of the size of 800 bytes a second time on startup isn't that big of a overhead tbh.

bin/xucli
sangaman: Oh I had forgotten that we had `rpc.` for port and host. That was probably me that made it that way originally, ironically, but now I still think we don't need it for those options either. The `.` is a bit non-standard. I'd be ok with `rpcport`, `rpchost`, and `tlscertpath` options. That also seems more consistent with other services (bitcoind uses `rpcport` and `rpcconnect` while lnd uses `rpcserver` for both port and host).

lib/grpc/webproxy/GrpcWebProxyServer.ts
sangaman: OK with me.

PR #543, opened 2018-09-30
 Refactor: add LND information to listpeers output

### LINKED ISSUE #487: CLI - add LND information to listpeers output
Today, listpeers does not show any LND data about a remote peer.

We do have some information available (like pubKeys on BTC and LTC) that we should add to the listeners output

This can help understanding swap issues.

ImmanuelSegol:
Closes #487

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -130,7 +130,7 @@ class Peer extends EventEmitter {
       case 'LTC':
         return this.handshakeState.lndltcPubKey;
       default:
-        return undefined;
+        return;
     }
   }
 
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -19,6 +19,8 @@ type PeerInfo = {
   pairs?: string[],
   xudVersion?: string,
   secondsConnected: number,
+  lndbtcPubKey?: string;
+  lndltcPubKey?: string;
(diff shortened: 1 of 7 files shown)


lib/p2p/Peer.ts
```diff
@@ -126,7 +130,7 @@ class Peer extends EventEmitter {
       case 'LTC':
         return this.handshakeState.lndltcPubKey;
       default:
-        return;
+        return undefined;
```
sangaman: I think the `undefined` is implied  with a blank `return;` statement. I'd leave it as is unless you think it's more clear this way or otherwise better.

proto/xudrpc.proto
```diff
@@ -356,14 +356,18 @@ message Peer {
   string address = 1 [json_name = "address"];
   // The node pub key to uniquely identify this peer
   string node_pub_key = 2 [json_name = "node_pub_key"];
+  // The LND BTC pub key associated with this peer
```
sangaman: I see "LND" usually capitalized as lower case "lnd" - including on the lnd main readme - so I'd stick with that.

lib/p2p/Peer.ts
ImmanuelSegol: It just seemed more aesthetic to me, but I think you are right.

PR #544, opened 2018-10-01
feat(proto): ban node via gRPC

### LINKED ISSUE #336: Permanently disconnect from a node
As discussed in https://github.com/ExchangeUnion/xud/issues/305#issuecomment-411606536 we should permanently disconnect from a peer on manual `disconnect` via gRPC/CLI.

Easiest proposed solution: add to `bannedHosts`.

michael1011:
`This commit adds a method for banning a XUD node to the gRPC and cli as described in #336.`

Closes #336


sangaman:
We were thinking of essentially renaming `disconnect` to `ban` right? I don't think we need both.

michael1011:
Oh sorry I completely forgot about that. Are you still sure that we don't want to have both? Not even for testing purposes? @sangaman

sangaman:
Personally I can't see a purpose for it, I never use `disconnect` anyway. But an `unban` call would be nice to have soon, if we want to reverse a ban without starting over with a new database.

michael1011:
Yeah you are right. I am going to remove `disconnect` in this PR and open an issue for `unban`.

proto/xudrpc.proto
```diff
@@ -236,10 +236,10 @@ message ConnectRequest {
 }
 message ConnectResponse {}
 
-message DisconnectRequest {
-  string node_pub_key = 1 [json_name = "node_pub_key"];
+message BanRequest {
+  string node_pub_key = 1 [json_name = "node_pub_key"];
```
sangaman: Looks like there's an extra whitespace here at the end of the line.

lib/p2p/errors.ts
```diff
@@ -42,6 +43,10 @@ const errors = {
     message: `could not connect to peer at ${addressUtils.toString(address)}: ${err.message}`,
     code: errorCodes.COULD_NOT_CONNECT,
   }),
+  NODE_NOT_FOUND: (nodePubKey: string) => ({
+    message: `could not find node ${nodePubKey}`,
```
sangaman: For the wording here, what do you think of `NODE_UNKNOWN`? I think that's a bit more descriptive, although "not found" is also correct.

lib/cli/commands/ban.ts
```diff
@@ -0,0 +1,20 @@
+import { callback, loadXudClient } from '../command';
+import { Arguments } from 'yargs';
+import { BanRequest } from '../../proto/xudrpc_pb';
+
+export const command = 'ban <node_pub_key>';
+
+export const describe = 'ban a xu node';
```
sangaman: ban "an" xu node is what I'd normally say/write.

lib/cli/commands/ban.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const builder = {
+  node_pub_key: {
+    description: 'nodePubKey of node to ban',
```
sangaman: Maybe "of *the* node to ban"?

PR #553, opened 2018-10-02
refactor(lndclient): remove asObject types

sangaman:
This PR attempts to standardize the way `LndClient` methods return values, going with native gRPC objects rather than the converted plain `AsObject` types. It also removes methods that are no longer expected to be used.

It may also make sense to standardize the way parameters are passed to LndClient methods - either as a native object or just as standard method parameters with the native request object being created inside the LndClient class. I think i prefer the latter - it seems like the much cleaner way to handle empty request objects and overall it reduces the potential for boiler plate code I believe.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -34,10 +34,6 @@ type ChannelCount = {
   pending: number,
 };
 
-interface GrpcResponse {
-  toObject: Function;
-}
-
 interface LightningMethodIndex extends LightningClient {
   [methodName: string]: Function;
 }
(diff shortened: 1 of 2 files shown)


moshababo:
I'm also in favor of passing standard parameters to `LndClient` methods. The native gRPC objects are just a part of the deserialization. 

What was the reason for removing the `AsObject` conversion?

sangaman:
It's @offerm's preference and we were at a point where both the `AsObject` conversion approach and the native gRPC object approach was in use, and my main preference is just to have a single approach.

moshababo:
Should we merge this?

Issue #545, opened 2018-10-01
Unban node via gRPC/CLI


michael1011:
Basically the reversing the `ban` method:
- adding the reputation event `ManualUnban` and
- connecting to the node again

ImmanuelSegol:
I will take this :)

PR #555, opened 2018-10-02
feat(grpc) add unban to grpc/cli 

ImmanuelSegol:
Issue #545 
Should I call the function `removeban` , `unban`, `revokeban` or something else ?


ImmanuelSegol:
@sangaman in `xudrpc.proto` both `Ban` and `UnBan` request\response share the same `message`, should we create a single request\response for both?

proto/xudrpc.proto
```diff
@@ -241,6 +249,11 @@ message BanRequest {
 }
 message BanResponse {}
 
+message UnBanRequest {
```
sangaman: Oh and one more thing, I've been trying to keep the messages and calls in alphabetic order respectively. Could you move these messages down to the bottom of the file, and also move the `rpc Unban` to the bottom of the call definitions?

proto/xudrpc.proto
```diff
@@ -71,6 +71,14 @@ service Xud {
       body: "*"
     };
   }
+
+  /* Remove ban from XU node manually and connect to it. */
```
sangaman: Also noticed this, the comment says we connect to a node after unbanning it but it doesn't look like we actually do that in the code. Am I missing where we connect? If not, we should either update the description or add the logic to connect. Personally, I'd probably lean to not automatically connecting after unbanning if the logic isn't already there to do so.

ImmanuelSegol: @sangaman On the issue of connecting after uban, I was thinking we could make the cli command as follows: unban <node_pub_key> <reconnect>
Where <reconnect> is a boolean that lets the user decide to reconnect or not.
Dose is that on any use?

lib/p2p/Pool.ts
```diff
@@ -341,11 +341,29 @@ class Pool extends EventEmitter {
...
+    if (this.nodes.isBanned(nodePubKey)) {
+      const unbanned = await this.nodes.addReputationEvent(nodePubKey, ReputationEvent.ManualUnban);
+
+      if (!unbanned) {
+        throw errors.NODE_UNKNOWN(nodePubKey);
+      }
+    } else {
+      throw errors.NODE_NOT_BANNED(nodePubKey);
     }
+
```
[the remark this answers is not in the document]
ImmanuelSegol: VS code warns me about that but I will drop it.

lib/p2p/NodeList.ts
```diff
@@ -130,6 +138,7 @@ class NodeList extends EventEmitter {
         promises.push(this.setBanStatus(node, true));
         this.emit('node.ban', nodePubKey);
       } else if (node.banned) {
+
```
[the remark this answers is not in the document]
ImmanuelSegol: VS code warns me about that but I will drop it.

proto/xudrpc.proto
sangaman: Yeah I think that's a good idea. If the boolean is true, then we also reconnect to the node.

ImmanuelSegol:
@sangaman Im not sure I correctly reconnected to the node after unbanning, could you point me in the right direction?
I was thinking that perhaps I have to retrieve the `node` from the `db` using the `nodePubKey` and then constructing a new `HandShakeState`, is that correct?

ImmanuelSegol:
@sangaman Could you review I think I finished this.

lib/p2p/Pool.ts
```diff
@@ -341,10 +351,39 @@ class Pool extends EventEmitter {
...
+      }
+
+      if (uri) {
+        const nodeUri: UriParts & {reconnect: boolean} = {
+          reconnect,
+          nodePubKey,
+          host: uri.lastAddress.host,
+          port: uri.lastAddress.port,
+        };
+        this.nodes.emit('node.unban', nodeUri);
```
sangaman: I think instead of this block of code, you can just check `if (reconnect)` and then call `this.tryConnectNode`, which will use existing logic to attempt to connect to a peer first by the last working address and then by advertised addresses. This would be instead of emitting the event. I'd lean towards specifying `retryConnecting` to be true for this.

sangaman:
Thanks, this is great work.

lib/p2p/Pool.ts
```diff
@@ -341,10 +343,43 @@ class Pool extends EventEmitter {
...
+        const Node: NodeConnectionInfo = {
+          nodePubKey,
+          addresses: node.addresses,
+          lastAddress: node.lastAddress,
+        };
+
+        this.logger.info(`node ${nodePubKey} was unbanned`);
+        if (reconnect) {
+          await this.tryConnectNode(Node, false);
+          // await this.addOutbound({ host, port }, nodePubKey, false);
```
sangaman: Is there a reason to keep this commented out line?

Issue #434, opened 2018-09-01
Make P2P functionality configurable


kilrau:
As per consensus between @offerm @sangaman @michael1011 make this the default behavior:
* Don't initialize seed nodes db.init (enabled separately via `p2p.initializeSeeds` CLI & rpc call)
* Don't auto-detect own IP (enabled separately via `p2p.detectMyIP` CLI & rpc call)
* Don't advertise own listening address (enabled separately via `p2p.advertise` CLI & rpc call)
* Don't call `GET_NODES` (enabled separately via `p2p.discover` CLI & rpc call)

NOTE: still auto-connects to all in `nodes` table

`xud.conf`:
```
[P2P Auto-Pilot (EXPERIMENTAL)]
p2p.auto = true
```

or via CLI arg `p2p.auto` behavior reverts to:
* Initialize seed nodes via db.init (`p2p.initializeSeeds`)
* Auto-detect own IP (`p2p.detectMyIP`)
* Advertise own listening address (`p2p.advertise`)
* Call `GET_NODES` (`p2p.discover`)

moshababo:
1) We already have a `p2p` config clause, so we can have it as `discovery` below it (`p2p.discovery`, or rather `p2p.discover`). But I don't think it represent well all this behavior. 
It's suitable only for the 3rd behavior (don't call `GET_NODES`). 
For the 2nd (don't advertise own addresses), the user actually controls the addresses, but we auto-detect his external IP and add it for him. The user can always keep his addresses list empty, and we should add a config field to enable the adding the external IP to the list. 
For the 1st (don't init seed nodes) - we already have `initDb`, which applies on all values. We can have this one as `p2p.initSeedNodes`, and make sure it would work also if the the database already exist.

2) Not advertising listening address will result in connected peer not being able to create a future outbound connection to us. Is that the goal?

3) If flag is disabled (default), you said `Don't advertise own listening address`. if enabled, you wrote the same thing. Is that a mistake?

offerm:
@moshababo agree to most of your comments.

1. p2p.discover  can controle `GET_NODES` if false, the node will not learn about additional peers and will not try to connect them.
2. Need to add another option to control auto detect of external IP. Should be `false` by default. This will cases the address list to be empty hence no advertise)
3. by default, initDB should not add the seed nodes to the database. Note that `p2p.discover = false` reduce the need for this cases (but still I think  that it should be done). 

Still, some functionality is missing:
[...]

PR #556, opened 2018-10-03
p2p.discover & p2p.detectExternalIp options #434

rsercano:
Solves #434 

Hello guys, just added the `p2p.discover` and `p2p.detectExternalIp` options


bin/xud
```diff
@@ -87,6 +87,16 @@ const { argv } = require('yargs')
       type: 'boolean',
       default: undefined,
     },
+    'p2p.discover': {
+      describe: 'Do not send GET_NODES if false, when connecting to a peer',
```
sangaman: I don't think we need the internal packet terminology here, we can just say "Automatically discover and connect to new nodes"

bin/xud
```diff
@@ -87,6 +87,16 @@ const { argv } = require('yargs')
...
+      type: 'boolean',
+      default: false,
```
sangaman: The way defaults arguments work from the command line is they overwrite both the hardcoded defaults and what's in the config file. The default for boolean types should be `undefined` to avoid this behavior.

bin/xud
```diff
@@ -87,6 +87,16 @@ const { argv } = require('yargs')
...
+    },
+    'p2p.detectExternalIp': {
+      describe: 'Turn on/off discovering and sharing current external ip address on startup',
```
sangaman: How about "Detect and advertise external ip address to peers"?

bin/xud
```diff
@@ -87,6 +87,16 @@ const { argv } = require('yargs')
       type: 'boolean',
       default: undefined,
     },
+    'p2p.discover': {
+      describe: 'Do not send GET_NODES if false, when connecting to a peer',
+      type: 'boolean',
+      default: false,
+    },
+    'p2p.detectExternalIp': {
```
sangaman: I think we decided on all lower case config options, so this should be `detectexternalip`.

lib/Config.ts
```diff
@@ -59,6 +59,8 @@ class Config {
 
     this.p2p = {
       listen: true,
+      discover: false,
```
sangaman: Default should be `true` I think, I think the typical user would want to connect to peers automatically.

lib/p2p/Pool.ts
```diff
@@ -17,10 +17,18 @@ import assert from 'assert';
 import { ReputationEvent } from '../types/enums';
 
 type PoolConfig = {
+  /** Turn on/off automatically detecting and sharing current external ip address on startup. */
```
sangaman: I like the longer description here, but how about instead "turn on/off automatically..." we say "Whether to automatically..."

lib/p2p/Pool.ts
```diff
@@ -494,9 +505,12 @@ class Pool extends EventEmitter {
       this.logger.verbose(`opened connection to ${peer.nodePubKey} at ${addressUtils.toString(peer.address)}`);
       this.peers.add(peer);
 
-      // request peer's orders and known nodes
+      // request peer's orders
       peer.sendPacket(new packets.GetOrdersPacket({ pairIds: this.handshakeData.pairs }));
-      peer.sendPacket(new packets.GetNodesPacket());
+      if (this.config.discover) {
+        // request peer's known nodes if only p2p.discover option is true
```
sangaman: "only if" instead of "if only"

lib/p2p/Pool.ts
```diff
@@ -137,6 +133,21 @@ class Pool extends EventEmitter {
     this.connected = true;
   }
 
+  private detectExternalIPAddress = async () => {
```
sangaman: Keep this camel case so make it `detectExternalIpAddress` with lower case "p"

lib/p2p/Pool.ts
```diff
@@ -96,26 +107,11 @@ class Pool extends EventEmitter {
...
-
       await this.listen();
       this.bindServer();
 
-      if (externalIp) {
-        this.addresses.push({
-          host: externalIp,
-          port: this.listenPort!,
-        });
+      if (this.config.detectExternalIp) {
```
sangaman: The new method doesn't keep the `if (this.addresses.length === 0)` check before detecting external ip. We should probably keep that, since if we've specified a listening address then we don't need to automatically detect one as well I figure. I think this line would make the most sense to have it, so `this.config.detectexternalip && this.addresses.length === 0`.

bin/xud
rsercano: Whatever you say :)

bin/xud
rsercano: Oh didn't notice this, thanks

bin/xud
rsercano: Alright, changing it

lib/Config.ts
rsercano: Oh okay, I thought you wanted vice versa.

lib/p2p/Pool.ts
rsercano: consider it done

lib/p2p/Pool.ts
rsercano: True

lib/p2p/Pool.ts
rsercano: Sure

bin/xud
offerm: @sangaman can you please explain why we need `undefined` here? what problem is this solving?

lib/p2p/Pool.ts
rsercano: I thought, we put that configuration option to get rid of it since the config is false it doesn't matter to have an address or not to detect external ip. Putting it back.

bin/xud
offerm: So if `detectexternalip` is false we will not advertise out ip? sound unrelated. 
We should allow the user to tell us which ip+port to use for advertising and separately from this option.

lib/Config.ts
offerm: @kilrau ?

lib/Config.ts
```diff
@@ -59,6 +59,8 @@ class Config {
...
+      detectExternalIp: true,
```
offerm: I think the default should be false. In most cases, this will run behind NAT and the detection will fail. Better to have this as an advanced option.

lib/p2p/Pool.ts
offerm: I see these two options working together. The user can provide listening addresses and can also ask to detect external addresses.

lib/Config.ts
rsercano: I would gladly make changes after conclusion @sangaman ?

lib/Config.ts
sangaman: This was discussed this on the weekly call last week I believe. The default behavior we want is a fully connected network.

lib/p2p/Pool.ts
sangaman: The idea of the external ip detection originally was to provide a listening address when none is provided. I suppose we could change that, but then at a minimum we'd want to check to make sure that it's not duplicating a value that's already specified in the listening addresses.

lib/Config.ts
sangaman: That would change the original intention of having this feature, I would prefer to keep it that way. If it fails, it's not a problem, it's no different than if we didn't try in the first place. Fwiw, personally I am behind NAT and it detects the external IP for my router fine.

bin/xud
sangaman: We'll still advertise any addresses that are specified manually as you described.

bin/xud
sangaman: Yeah it has to do with how yargs works. All booleans in yargs have a "default" default value of false - as in if you don't manually specify a default value, it will pass false for that arguent if `type` is `boolean`. The way our config works, command line arguments override the hardcoded defaults and what's in the config file. If we didn't have an `undefined` default here, `p2p.discover` would always be `false` unless we set `--p2p.discover true` from the command line.

lib/p2p/Pool.ts
rsercano: adding a duplicate check then before adding the external ip to the addresses

lib/p2p/Pool.ts
```diff
@@ -137,6 +133,24 @@ class Pool extends EventEmitter {
     this.connected = true;
   }
 
+  private detectExternalIpAddress = async () => {
+    let externalIp: string | undefined;
+    try {
+      externalIp = await getExternalIp();
+      this.logger.info(`retrieved external IP: ${externalIp}`);
+
+      const externalIpExists = this.addresses.some((address) =>  { return address.host === externalIp; });
```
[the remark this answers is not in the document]
kilrau: I'd want to stick with the current behaviors:
listening address specified = no auto detection
listening address NOT specified = auto detection

lib/Config.ts
```diff
@@ -59,6 +59,8 @@ class Config {
 
     this.p2p = {
       listen: true,
+      discover: true,
+      detectexternalip: true,
```
sangaman: Can we make this `false`?

lib/p2p/Pool.ts
```diff
@@ -96,26 +107,11 @@ class Pool extends EventEmitter {
...
-
       await this.listen();
       this.bindServer();
 
-      if (externalIp) {
-        this.addresses.push({
-          host: externalIp,
-          port: this.listenPort!,
-        });
+      if (this.config.detectexternalip && this.addresses.length === 0) {
```
sangaman: Let's drop the `length === 0` check here.

lib/Config.ts
rsercano: Ofcourse

lib/p2p/Pool.ts
rsercano: sure thing

sangaman:
@offerm, are the changes good with you? Also @rsercano, can you resolve the conflict? I can merge afterwards.

lib/p2p/Pool.ts
```diff
@@ -17,10 +17,18 @@ import assert from 'assert';
 import { ReputationEvent } from '../types/enums';
 
 type PoolConfig = {
+  /** Whether or not to automatically detect and share current external ip address on startup. */
+  detectexternalip: boolean;
+
+  /** If false, don't send GET_NODES when connecting, defaults to false. */
```
offerm: Is the default true or false?
it is true in config.ts line 62 but comment here says that the default is false.

lib/p2p/Pool.ts
```diff
@@ -528,9 +542,12 @@ class Pool extends EventEmitter {
       this.logger.verbose(`opened connection to ${peer.nodePubKey} at ${addressUtils.toString(peer.address)}`);
       this.peers.add(peer);
 
-      // request peer's orders and known nodes
+      // request peer's orders
       peer.sendPacket(new packets.GetOrdersPacket({ pairIds: this.handshakeData.pairs }));
-      peer.sendPacket(new packets.GetNodesPacket());
+      if (this.config.discover) {
+        // request peer's known nodes only if p2p.discover option is true
```
offerm: comment mention p2p.discover while if is using this.config.discover

lib/p2p/Pool.ts
rsercano: Oh sorry my bad, fixing the comment

lib/p2p/Pool.ts
rsercano: I couldn't get what you mean exactly @offerm  ?

rsercano: this.config has a comment which tells that;
>   /** Points to config comes during construction. */
And in constructor it's a `PoolConfig`

offerm: Ignore. I'm good.

PR #557, opened 2018-10-04
feat: add network command line argument

sangaman:
This PR adds a command line argument for the recently added `network` config option, which specifies the network for all lnd clients.

### DIFF
--- a/bin/xud
+++ b/bin/xud
@@ -23,6 +23,12 @@ const { argv } = require('yargs')
       describe: 'Path to the log file',
       type: 'string',
     },
+    network: {
+      describe: 'The network the lnd clients are using',
+      type: 'string',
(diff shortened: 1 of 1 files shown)


bin/xud
```diff
@@ -23,6 +23,12 @@ const { argv } = require('yargs')
       describe: 'Path to the log file',
       type: 'string',
     },
+    network: {
+      describe: 'The network to use for lnd clients',
```
michael1011: Shouldn't it rather be `The network the lnd clients are using`? Because we are not telling the LNDs what they should do but just get information about what they are already doing on their own?

sangaman: Yup, that's definitely more accurate, I'll update it.

Issue #554, opened 2018-10-02
Refactor swap clients to be generic, remove hardcoded btclnd/ltclnd


sangaman:
This issue is to track refactoring code to make lnd swap clients generic and configurable. Currently we hardcode `btclnd` and `ltclnd` as the only supported networks, but longer term we'd need to support more and we don't want code changes for each one.

ImmanuelSegol:
Can I take this?

PR #559, opened 2018-10-04
fix(config): properly set default macaroon paths

sangaman:
This PR fixes the derivation of the default lnd macaroon paths based on the active network. Previously, changing the `network` config value from `testnet` would not change the default macaroon paths, they would need to be manually specified. This also correctly sets the testnet directory for the litecoin network as `testnet4` instead of `testnet`.

### DIFF
--- a/lib/Config.ts
+++ b/lib/Config.ts
@@ -82,7 +82,8 @@ class Config {
     this.lndltc = {
       disable: false,
       certpath: path.join(lndDefaultDatadir, 'tls.cert'),
-      macaroonpath: path.join(lndDefaultDatadir, 'data', 'chain', 'litecoin', this.network, 'admin.macaroon'),
+      macaroonpath: path.join(lndDefaultDatadir, 'data', 'chain', 'litecoin',
+        this.network === Network.TestNet ? 'testnet4' : this.network, 'admin.macaroon'),
       host: 'localhost',
       port: 10010,
       cltvdelta: 576,
@@ -95,8 +96,13 @@ class Config {
   }
 
   public load(args?: { [argName: string]: any }): Config {
-    if (args && args.xudir) {
-      this.updateDefaultPaths(args.xudir);
+    if (args) {
+      if (args.xudir) {
+        this.updateDefaultPaths(args.xudir);
+      }
+      if (args.network) {
+        this.updateMacaroonPaths(args.network);
+      }
     }
     const configPath = path.join(this.xudir, 'xud.conf');
     if (!fs.existsSync(this.xudir)) {
@@ -108,6 +114,16 @@ class Config {
 
         if (props.xudir && (!args || !args.xudir)) {
           this.updateDefaultPaths(props.xudir);
+        }
+
+        if (props.network !== undefined && (!args || !args.network)) {
+          // first check that network is a valid value
+          if (typeof props.network !== 'string' || !Object.values(Network).includes(props.network)) {
+            // delete the invalid network value
+            delete props.network;
+          } else {
+            this.updateMacaroonPaths(props.network);
+          }
         }
 
         // merge parsed json properties from config file to the default config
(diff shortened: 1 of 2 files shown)


lib/Config.ts
```diff
@@ -110,6 +116,16 @@ class Config {
           this.updateDefaultPaths(props.xudir);
         }
 
+        if (props.network !== undefined && (!args || !args.network)) {
+          // first check that network is a valid value
+          if (typeof props.network !== 'string' || !Object.keys(Network).some((key: string) => key.toLowerCase() === props.network)) {
```
michael1011: Why not `!Object.keys(Network).includes(`?

lib/Config.ts
```diff
@@ -151,6 +167,13 @@ class Config {
     this.dbpath = this.getDefaultDbPath();
   }
 
+  private updateMacaroonPaths = (network: string) => {
+    this.network = network as Network;
+    this.lndbtc.macaroonpath = path.join(this.lndbtc.macaroonpath, '..', '..', this.network, 'admin.macaroon');
+    this.lndltc.macaroonpath = path.join(this.lndltc.macaroonpath, '..', '..',
+      this.network === Network.TestNet ? 'testnet4' : this.network, 'admin.macaroon');
```
michael1011: I would put `this.network === Network.TestNet ? 'testnet4' : this.network` into a method since it is used twice and pretty easy to get wrong.

lib/Config.ts
sangaman: Good question, I suppose I could add a comment because it's a bit complicated (although not very important).

The `Network` enum is basically an object with keys for all the enum values (`mainnet, `testnet`, etc..) as well as the enum names (`MainNet`, `TestNet`, etc...).  I only want to check for a match against the lower case `enum` values, and this was the only way I could figure out how to do it. `includes` would match with a value of `network = "MainNet"` which I'd want to avoid.

lib/Config.ts
sangaman: It sort of is in a method, `updateMacaroonsPath`. I can't use it in the constructor because it's the first time we're setting a value for `macaroonpath`, but if the logic is ever needed to be used again it can call this method again. Also I think longer term this may need to change for a more generic way to handle things - e.g. what if there's a `testnet5` one day or a separate currency supported by lnd one day has its own numbering system. I was thinking reading from the directory and using whichever directory starts with "testnet" maybe.

FYI, I think it's actually pretty annoying that the testnet folder is different for litecoin in lnd, especially given that within the code and config (namely the `--litecoin.testnet` option) it just uses "testnet" most places, so I'm also hoping that one day lnd just uses a plain "testnet" folder. Was thinking of opening an issue over there about this.

Long story short, I can definitely still put this in its own method if you'd like, but I feel it's OK for now as is.

offerm:
What about supporting --nomacaroon. 

?

sangaman:
I'll put `nomacaroons` in a separate PR.

lib/Config.ts
michael1011: No it is fine as is then.

#554 should take care of the completely dynamic approach then.

lib/Config.ts
michael1011: If you want to compare against the values of the `enum` (which are all lower case) why are you using `Object.keys` then?

`!Object.values(Network).includes(props.network)` would do the exact same thing or am I missing something? @sangaman

sangaman: Same thing. In an enum, the set of keys is the same as the set of values. Took me a while to figure this out.

michael1011: I am not sure I get what you mean:

`Object.keys(Network)` is equal: `[ 'MainNet', 'TestNet', 'SimNet', 'RegTest' ]` but
`Object.values(Network)` is: `[ 'mainnet', 'testnet', 'simnet', 'regtest' ]`

So `Object.keys(Network).some((key: string) => key.toLowerCase() === props.network)` should always be equal to `Object.values(Network).includes(props.network)`? @sangaman

sangaman: Wow well I tested this out and I was totally wrong, either this has changed in TypeScript from older versions or I just misinterpreted it off of answers like this one: https://stackoverflow.com/a/39439520/9137842

Will change this to use `values` and `includes`. Thanks for the lesson.

PR #560, opened 2018-10-04
feat(rpc): add option to disable tls

sangaman:
This commit adds an option to disable TLS on gRPC calls, which may be useful for testing and debugging purposes.

This is necessary as a short term fix for swap resolver communication with lnd until the resolver branch can be modified to properly use the xud tls.cert file.

This works well in my tests, below is how xucli commands look with TLS disabled on the server:

```
xucli getinfo
E1004 13:21:38.417000000 10260 ssl_transport_security.cc:1228] Handshake failed with fatal error SSL_ERROR_SSL: error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER.
E1004 13:21:38.418000000 10260 ssl_transport_security.cc:1228] Handshake failed with fatal error SSL_ERROR_SSL: error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER.
Error: 14 UNAVAILABLE: Connect Failed

xucli getinfo -d
{
  "version": "1.0.0-prealpha.4",
  "nodePubKey": "029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345",
  "urisList": [
    "029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345@75.75.58.83:8885",
    "029a96c975d301c1c8787fcb4647b5be65a3b8d8a70153ff72e3eac73759e5e345@127.0.0.1:8885"
  ],
  "numPeers": 3,
  "numPairs": 2,
  "orders": {
    "peer": 0,
    "own": 0
  },
  "lndbtc": {
    "error": "lnd is not connected",
    "chainsList": [],
    "blockheight": 0,
    "urisList": [],
    "version": "",
    "alias": ""
  },
  "lndltc": {
    "error": "",
    "channels": {
      "active": 1,
      "inactive": 0,
      "pending": 0
    },
    "chainsList": [
      "litecoin"
    ],
    "blockheight": 788671,
    "urisList": [],
    "version": "0.5.0-beta commit=",
    "alias": "Exchange A LTC on 10001/10011"
  }
}
```

### DIFF
--- a/bin/xucli
+++ b/bin/xucli
@@ -27,6 +27,11 @@ require('yargs')
       describe: 'Path to the TLS certificate of xud',
       type: 'string',
     },
+    disabletls: {
+      alias: 'd',
+      describe: 'Whether to disable TLS for gRPC calls',
+      type: 'boolean',
+    },
   })
   .commandDir('../dist/cli/commands/', { recurse: true })
   .demandCommand(1, '')
--- a/bin/xud
+++ b/bin/xud
@@ -110,6 +110,11 @@ const { argv } = require('yargs')
       type: 'number',
       alias: 'r',
     },
+    'rpc.usetls': {
+      describe: 'Whether to use TLS for gRPC calls',
+      type: 'boolean',
+      default: undefined,
+    },
     'webproxy.disable': {
       describe: 'Disable web proxy server',
       type: 'boolean',
--- a/lib/Config.ts
+++ b/lib/Config.ts
@@ -15,7 +15,12 @@ class Config {
   public loglevel: string;
   public logpath: string;
   public network: Network;
-  public rpc: { disable: boolean, host: string, port: number };
+  public rpc: {
+    disable: boolean,
+    host: string,
+    port: number,
+    disabletls: boolean,
+  };
   public lndbtc: LndClientConfig;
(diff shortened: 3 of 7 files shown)


bin/xucli
```diff
@@ -27,6 +27,11 @@ require('yargs')
       describe: 'Path to the TLS certificate of xud',
       type: 'string',
     },
+    disabletls: {
```
michael1011: Why not set `tlscertpath` to `undefined` or just to an empty string to disable TLS instead of having an own field for it?

sangaman: I thought about this, my main reason to go with it's own option is for the easy `-d` arg. I don't know of a way to pass an empty string or `undefined` for a string argument, if you do then I'd be fine with that approach.

sangaman:
Noticed myself that I hadn't applied this logic to the web proxy, will do so now.

offerm:
@sangaman I added TLS support for the resolver.

Not sure there is a need for it now. Suggest not to proress.

sangaman:
Ok, we can close this PR then but lets keep the branch for now in case it turns out we see a need for this feature in the future.

PR #564, opened 2018-10-04
refactor(lndclient): remove invoice subscription

sangaman:
This removes code that subscribes to lnd invoices. This is currently unused and as far as I know there are no plans to use it in the foreseeable future given the hash resolver approach we have taken for swaps.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -54,7 +54,6 @@ class LndClient extends BaseClient {
   private meta!: grpc.Metadata;
   private uri!: string;
   private credentials!: ChannelCredentials;
-  private invoiceSubscription?: ClientReadableStream<lndrpc.InvoiceSubscription>;
   private reconnectionTimer?: NodeJS.Timer;
   private identityPubKey?: string;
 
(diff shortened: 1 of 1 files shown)


offerm:
Don't we use it internally to check if LND is still running? With this change we will not detect it.

Regardless of the above, I like this change. I don't think we need to "manage" the LND state and if LND is not available we should detect it when we need it. 

Also, with the current code (master) we  properly detect LND shotdown but not restart (it may take long time). This PR fixes the problem....

@sangaman if you are happy with it, I'm too.

sangaman:
Yeah I don't think we need the invoice subscription as a way to check when lnd goes down, like you said we should handle certain errors on calls to lnd by marking it as `Disconnected` and then starting the timer to recheck periodically for connectivity.

offerm:
Don't think we need the timer and the disconnectes status ... When we need
lnd we should access  it and if there is an error we should hanlde it.

sangaman:
I think it's needed longer term because we don't want to advertise orders for trading pairs that we can not swap. If we do so, peers who try and fill those orders will fail repeatedly and will ban us.

Issue #562, opened 2018-10-04
Store swapdeals to database, remove from memory


sangaman:
We want to persist swap results (whether successful or failed) to the database for historical/informational purposes. This will require a new table in the database. Probably not every property of a `SwapDeal` needs to be persisted, but certainly enough to know what the swap was for, how much, with who, when it completed, etc...

Once a result has been persisted, we can remove it from the collection of swap deals stored in memory. However, I'm thinking we probably want to maintain in memory a list of payment hashes that we've used in the past, so that we don't reuse one (see #547). For that reason, we may also want to load all payment hashes from the database on startup.

ImmanuelSegol:
I think I will give this a shot if that's ok.

sangaman:
Fine with me, thanks.

kilrau:
Can this issue be closed? @sangaman 

Do we store failures? @moshababo  Could lead to a bloated database in the long run. I'd put a limit on storing failed swaps, like e.g. only the last 1000 failed ones.

sangaman:
@offerm was in favor of storing everything including failed swaps and I agreed, I think it's fine for now and eventually we can add scripts to prune old failed swaps from the database I figure.

kilrau:
Ok, so good to close here?

kilrau:
What's the primary key? r_hash? @sangaman @offerm

Issue #551, opened 2018-10-01
Minor swap refactoring/renaming


sangaman:
Below is a list of minor and/or superficial refactorings to the swaps related code that I think would be desirable:

- Rename `SwapErrorPacket` to `SwapFailedPacket`. My thinking is that a swap failure is not always due to an "error" per se, it could just be rejected even when everything is functioning as expected. This also aligns with the `swap.failed` event name.
- On error/failure responses, send an enum failure reason rather than a string. This involves less data transmission between peers, and more importantly it makes checking the error or failure reason easier since we don't need to compare exact strings where a typo could mess things up and go undetected by the compiler.
- Rename `SwapResponsePacket` to `SwapAcceptedPacket`. As it's set up currently, the swap response always indicates acceptance of the swap, so I think this name is a bit more descriptive. 
- Rename `SwapDealRole` to `SwapRole`. I think the "Deal" part is redundant and the role applies not just to the deal.
- Rename `SwapDealPhase` to `SwapPhase`. Same as above.
- Rename `SwapDealState` to `SwapState`. Same.
- Rename `stateReason` to `errorReason`. It seems like the only way this field is being used is when swaps fail, and the reason for the "Active" and "Completed" states seems obvious. Also if we go with an enum failure reason, this field can possibly use the same enum type instead of a string.
- Rename `myRole` to `role`. I think the "my" is implied.

Let me know if you agree or disagree on anything in particular.

sangaman:
Another minor thing to change is switching `r_hash` and `r_preimage` to camel case `rHash` and `rPreimage` to be consistent with the project-wide variable naming. This is also consistent with how those variables are generally named internally in lnd, it's mainly the rpc layer where snake_case is used.

moshababo:
Sounds fine to me.

PR #569, opened 2018-10-06
refactor(SwapDeals) Store swapdeals in database

ImmanuelSegol:
Issue #562 
@sangaman In this PR I created a new table in the DB for swap-deals, and on initialization, we load the `r_hash`'s from the DB and load them into a Set for easy and quick access. 

1. What info should we store about the `swapdeals`?
2. Every time we attempt to get a `swapdeal` should we check if the `r_hash` is already in existence in the DB? 
3. Should we store all `swapdeals` event failed ones?

### DIFF
--- a/test/unit/Swaps.spec.ts
+++ b/test/unit/Swaps.spec.ts
@@ -1,5 +1,6 @@
 import { expect } from 'chai';
-import Swaps, { SwapDeal } from '../../lib/swaps/Swaps';
+import Swaps from '../../lib/swaps/Swaps';
+import { SwapDeal } from '../../lib/swaps/types';
 import Logger, { Level } from '../../lib/Logger';
 import { SwapPhase, SwapState, SwapRole } from '../../lib/types/enums';
 
@@ -18,7 +19,7 @@ describe('Swaps', () => {
     state: SwapState.Active,
     peerPubKey: '03029c6a4d80c91da9e40529ec41c93b17cc9d7956b59c7d8334b0318d4a86aef8',
     orderId: 'f8a85c66-7e73-43cd-9ac4-176ff4cc28a8',
-    localOrderId: '1',
+    localId: '1',
     proposedQuantity: quantity,
     pairId: 'LTC/BTC',
     takerCurrency: 'LTC',


sangaman:
Will review this after alpha.1 is tagged, thanks.

sangaman:
1. I'm not exactly sure but I would go with:

- myRole
- state
- stateReason
- peerPubKey
- orderId
- localOrderId
- proposedQuantity
- quantity
- pairId
- takerAmount
- makerAmount
- takerCltvDelta
- makerCltvDelta
- price
- r_hash
- r_preimage
- createTime (this would likely be in place of the automatic timestamping feature provided by Sequelize)
- executeTime
- completionTime

I'm also thinking it might make sense to handle #551 first just so we can get the names of some of those fields changed before doing so is a breaking DB change, I'll open a PR for that soon.

sangaman:
@ImmanuelSegol Are you going to update this PR? I think this is definitely a feature we would want for the next milestone, so I don't want it to be forgotten.

sangaman:
@ImmanuelSegol Looks like there are a number of minor linting issues, I think `npm run lint` should clear them up. I'll take another look at this tomorrow, thanks!

ImmanuelSegol:
@sangaman Im not sure that I properly defined `SwapDeal` in the data base modals. Could you check that for me. Thanks.

lib/Xud.ts
```diff
@@ -91,7 +91,8 @@ class Xud extends EventEmitter {
 
       this.pool = new Pool(this.config.p2p, loggers.p2p, this.db.models);
 
-      this.swaps = new Swaps(loggers.swaps, this.pool, this.lndbtcClient, this.lndltcClient);
+      this.swaps = new Swaps(loggers.swaps, this.db.models, this.pool, this.lndbtcClient, this.lndltcClient);
+      await this.swaps.init();
```
sangaman: I think we can add this to the `initPromises` array rather than awaiting it separately.

lib/db/models/SwapDeal.ts
```diff
@@ -0,0 +1,33 @@
...
+    orderId: { type: DataTypes.STRING, allowNull: false },
+    localOrderId: { type: DataTypes.STRING, allowNull: false },
+    proposedQuantity: { type: DataTypes.INTEGER, allowNull: false },
+    quantity: { type: DataTypes.INTEGER, allowNull: false },
+    pairId: { type: DataTypes.STRING, allowNull: false },
+    price: { type: DataTypes.INTEGER, allowNull: false },
+    takerCltvDelta: { type: DataTypes.INTEGER, allowNull: false },
+    makerCltvDelta: { type: DataTypes.INTEGER, allowNull: false },
+    createTime: { type: DataTypes.INTEGER, allowNull: false },
+    executeTime: { type: DataTypes.INTEGER, allowNull: true },
```
sangaman: We need to use `BIGINT` for the times, they won't fit in 32-bit integers.

lib/db/models/SwapDeal.ts
```diff
@@ -0,0 +1,33 @@
+import Sequelize from 'sequelize';
+import { db } from '../../types';
+
+export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes) => {
+  const attributes: db.SequelizeAttributes<db.SwapDealAttributes> = {
+    myRole: { type: DataTypes.TINYINT, allowNull: false },
+    state: { type: DataTypes.TINYINT, allowNull: false },
+    stateReason: { type: DataTypes.STRING, allowNull: false },
+    r_hash: { type: DataTypes.STRING, allowNull: false },
+    r_preimage: { type: DataTypes.STRING, allowNull: false },
```
sangaman: I know we haven't changed the names of these variables yet, but lets go with camelCase for these two so that we don't need a breaking db change later on.

lib/db/models/SwapDeal.ts
```diff
@@ -0,0 +1,33 @@
...
+    peerPubKey: { type: DataTypes.STRING, allowNull: false },
+    orderId: { type: DataTypes.STRING, allowNull: false },
+    localOrderId: { type: DataTypes.STRING, allowNull: false },
+    proposedQuantity: { type: DataTypes.INTEGER, allowNull: false },
+    quantity: { type: DataTypes.INTEGER, allowNull: false },
```
sangaman: The quantity fields are decimals, I'd suggest going with 8 decimal places.

lib/db/models/SwapDeal.ts
```diff
@@ -0,0 +1,33 @@
...
+    pairId: { type: DataTypes.STRING, allowNull: false },
+    price: { type: DataTypes.INTEGER, allowNull: false },
```
sangaman: Price is also a decimal, maybe go with 8 decimal places too.

lib/db/models/SwapDeal.ts
```diff
@@ -0,0 +1,33 @@
+import Sequelize from 'sequelize';
+import { db } from '../../types';
+
+export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes) => {
+  const attributes: db.SequelizeAttributes<db.SwapDealAttributes> = {
```
sangaman: The `allowNull` values should all match whether each of these fields is nullable on the `SwapDeal` type, there are a few that don't allow null that can be null on the deal itself currently. (stateReason/errorReason should be nullable but it's not currently, I'll open a PR for that).

lib/swaps/SwapRepository.ts
```diff
@@ -0,0 +1,29 @@
...
+        r_hash,
+      },
+    });
+  }
+
+  public addSwapDeal = (swapDeal: db.SwapDealFactory): Bluebird<db.SwapDealInstance> => {
+    return this.models.SwapDeal.create(<db.SwapDealAttributes>swapDeal);
+  }
+
+  public addSwapDeals = (swapDeals: db.SwapDealFactory[]): Bluebird<db.SwapDealInstance[]> => {
```
sangaman: It doesn't seem we're using this method currently and I can't envision a use for it, I would drop it.

lib/swaps/SwapRepository.ts
```diff
@@ -0,0 +1,29 @@
+import { db } from '../types';
+import Bluebird from 'bluebird';
+import { Models } from '../db/DB';
+
+class SwapRepository {
+
+  constructor(private models: Models) {}
+
+  public getSwaps = (): Bluebird<db.SwapDealInstance[]> => {
```
sangaman: I'd call this `getSwapDeals` to be consistent with naming.

lib/swaps/SwapRepository.ts
```diff
@@ -0,0 +1,29 @@
...
+    return this.models.SwapDeal.findAll();
+  }
+
+  public getSwap = async (r_hash: string): Promise<db.SwapDealInstance | null> => {
```
sangaman: Likewise this should be `getSwapDeal`

lib/swaps/Swaps.ts
```diff
@@ -76,11 +79,13 @@ interface Swaps {
 class Swaps extends EventEmitter {
   /** A map between r_hash and swap deals. */
   private deals = new Map<string, SwapDeal>();
+  private hashs = new Set<string>();
```
sangaman: I'd prefer calling this `usedHashes` so it's more clear what this is for.

lib/swaps/Swaps.ts
```diff
@@ -76,11 +79,13 @@ interface Swaps {
...
+  private repository: SwapRepository;
 
-  constructor(private logger: Logger, private pool: Pool, private lndBtcClient: LndClient, private lndLtcClient: LndClient) {
+  constructor(private logger: Logger, private models: Models, private pool: Pool, private lndBtcClient: LndClient, private lndLtcClient: LndClient) {
     super();
+    this.repository = new SwapRepository(this.models);
 
-    this.bind();
```
sangaman: I'd leave this in the constructor, that's how it's done in other classes with a similar bind function.

lib/swaps/Swaps.ts
```diff
@@ -96,6 +101,16 @@ class Swaps extends EventEmitter {
...
   }
 
+  public init = async () => {
+    // Load Swaps from data base
+    const promises: PromiseLike<any> = this.repository.getSwaps();
+    const result = await Promise.resolve(promises);
+    result.map((deal: SwapDeal) => {
+      this.hashs.add(deal.r_hash);
+    });
+    this.bind();
```
sangaman: This can go back to the constructor.

lib/swaps/Swaps.ts
```diff
@@ -96,6 +101,16 @@ class Swaps extends EventEmitter {
     return { takerAmount, makerAmount };
   }
 
+  public init = async () => {
+    // Load Swaps from data base
+    const promises: PromiseLike<any> = this.repository.getSwaps();
+    const result = await Promise.resolve(promises);
```
sangaman: Can we collapse these two lines into one by just calling `const swapDeals = await this.repository.getSwaps();`?  There's only a single promise here so we don't really need a `promises` variable.

lib/db/models/SwapDeal.ts
```diff
@@ -0,0 +1,33 @@
...
+    myRole: { type: DataTypes.TINYINT, allowNull: false },
+    state: { type: DataTypes.TINYINT, allowNull: false },
+    stateReason: { type: DataTypes.STRING, allowNull: false },
+    r_hash: { type: DataTypes.STRING, allowNull: false },
```
sangaman: We want this to be `unique: true`

lib/swaps/SwapRepository.ts
sangaman: Better yet, since all we really need are the hashes for now, we'd be reading way more data than needed by pulling back everything. It would be better to query just for the r hash values and call this something like `getRHashes`.

lib/swaps/Swaps.ts
```diff
@@ -96,6 +101,16 @@ class Swaps extends EventEmitter {
...
+    result.map((deal: SwapDeal) => {
```
sangaman: Is `map` the right function here? I would think we'd want `forEach`. Either way this would be moot if we go with my suggestion to pull back an array of hashes from the database rather than the full instances.

lib/types/db.ts
```diff
@@ -25,6 +26,17 @@ export type CurrencyAttributes = CurrencyFactory & {
 
 export type CurrencyInstance = CurrencyAttributes & Sequelize.Instance<CurrencyAttributes>;
 
+// SwapDeals
+export type SwapDealFactory = SwapDeal;
+
+export type SwapDealAttributes = SwapDealFactory & {
```
sangaman: I think the attributes is simply equal to the factory type in this case, those two time fields can already be undefined.

lib/types/db.ts
```diff
@@ -25,6 +26,17 @@ export type CurrencyAttributes = CurrencyFactory & {
 
 export type CurrencyInstance = CurrencyAttributes & Sequelize.Instance<CurrencyAttributes>;
 
+// SwapDeals
+export type SwapDealFactory = SwapDeal;
```
sangaman: I believe we want everything in `SwapDeal` *except* `makerToTakerRoutes` for now. You can do this with `type SwapDealFactory = Pick<SwapDeal, Exclude<keyof SwapDeal, 'makerToTakerRoutes'>>;`

lib/swaps/Swaps.ts
```diff
@@ -150,6 +165,40 @@ class Swaps extends EventEmitter {
...
   }
 
+  /**
+   * Saves deal to database and deletes from memory.
+   * @param deal The deal to persist.
+   */
+  private persistDeal = (deal: SwapDeal) => {
+    const { r_hash } = deal;
+    if (this.hashs.has(r_hash)) {
+      const persistDeal: SwapDealAttributes = {
```
sangaman: Do we need this separate object? I think you can just pass the `deal` to the `addSwapDeal` method.

lib/swaps/Swaps.ts
```diff
@@ -150,6 +165,40 @@ class Swaps extends EventEmitter {
     return;
   }
 
+  /**
+   * Saves deal to database and deletes from memory.
+   * @param deal The deal to persist.
+   */
+  private persistDeal = (deal: SwapDeal) => {
```
sangaman: I kind of prefer `archiveDeal` or something to that effect. That makes it more clear that we are removing the deal from memory, persist sounds like we might only be saving it.

lib/types/db.ts
```diff
@@ -17,6 +17,7 @@ export type SequelizeAttributes<T extends { [key: string]: any }> = {
 * "xInstance" is the type definition of a fetched record as a Sequelize row instance, which contains some util properties.
 */
 
+// Currencys
```
sangaman: Can these comments just be in singular form? That would match the model name/type names.

lib/types/orders.ts
```diff
@@ -91,6 +91,24 @@ export type Pair = {
   quoteCurrency: string;
 };
 
+export type SwapDeal = {
```
sangaman: We already define this in `Swaps.ts`, why define it again?

lib/swaps/Swaps.ts
```diff
@@ -150,6 +165,40 @@ class Swaps extends EventEmitter {
...
+        quantity: deal.quantity,
+        pairId: deal.pairId,
+        price: deal.price,
+        takerCltvDelta: deal.takerCltvDelta,
+        makerCltvDelta: deal.makerCltvDelta,
+        createTime: deal.createTime,
+        executeTime: deal.executeTime,
+        completionTime: deal.completionTime,
+      };
+      this.repository.addSwapDeal(persistDeal)
```
sangaman: We try to use the async/await pattern wherever possible so you can just await `addSwapDeal` and then await `removeDeal`.

lib/types/db.ts
ImmanuelSegol: I changed the comments to `/* */` it just seems more aesthetic to me.

lib/types/orders.ts
sangaman: I was actually thinking of leaving this in `Swaps.ts`, although in #597 I'm moving it to its own types file just for swap-related types. So if that gets merged it would be good to use that commit here and drop this type definition.

lib/swaps/Swaps.ts
```diff
@@ -96,6 +50,14 @@ class Swaps extends EventEmitter {
     return { takerAmount, makerAmount };
   }
 
+  public init = async () => {
+    // Load Swaps from data base
+    const result = await this.repository.getSwapDeals();
+    result.forEach((deal: SwapDeal) => {
```
sangaman: I think the simplest fix here is just to have this be `deal: SwapDealInstance`?

ImmanuelSegol: I think we should have 2 types of `SwapDeal`s one for saving in memory and one for saving in the `DB`, currently Im getting errors here for example [error](https://github.com/ExchangeUnion/xud/blob/a8f7812b27fe466a5b365c2b40a5fb11a988fada/lib/swaps/Swaps.ts#L291).
The errors are because [of](https://github.com/ExchangeUnion/xud/blob/a8f7812b27fe466a5b365c2b40a5fb11a988fada/lib/types/orders.ts#L116) the use of `<type> | undefined`, for some reason [`attributes`](https://github.com/ExchangeUnion/xud/blob/a8f7812b27fe466a5b365c2b40a5fb11a988fada/lib/db/models/SwapDeal.ts#L36) will throw an error if I use for example `makerCltvDelta?: number;`

sangaman: `SwapDealInstance` *is* the type for what gets saved to the database. I haven't tried pulling this branch down yet, but would using that type for this line fix the issues?

ImmanuelSegol: Yeah if I use 2 separate types it solves the issue.

sangaman: Ok cool, change this line when you get a chance and I'll review again.

ImmanuelSegol:
@sangaman  I don't think we should have moved `swapdeal` type out of `orders.ts`, it creates a circular import warning in the linter and also we use it in multiple places.

sangaman:
Sorry I just saw your comment about the circular import. It doesn't look like there should be any issue if you pull the latest from master and use the `SwapDeal` and `SwapResult` definitions from `swaps/types.ts`.

lib/db/models/SwapDeal.ts
```diff
@@ -0,0 +1,39 @@
...
+    errorReason: { type: DataTypes.STRING, allowNull: true },
+    r_hash: { type: DataTypes.STRING, allowNull: false, unique: true },
+    r_preimage: { type: DataTypes.STRING, allowNull: true },
+    peerPubKey: { type: DataTypes.STRING, allowNull: false },
+    orderId: { type: DataTypes.STRING, allowNull: false },
+    localId: { type: DataTypes.STRING, allowNull: false },
+    proposedQuantity: { type: DataTypes.DECIMAL(8), allowNull: false },
+    quantity: { type: DataTypes.DECIMAL(8), allowNull: true },
+    pairId: { type: DataTypes.STRING, allowNull: false },
+    takerAmount: { type: DataTypes.NUMBER , allowNull: false },
```
sangaman: There is no Data Type called `NUMBER` in sequelize, which is why the tests are failing. `takerAmount` should be a `BIGINT`.

lib/db/models/SwapDeal.ts
```diff
@@ -0,0 +1,39 @@
...
+    takerCurrency: { type: DataTypes.STRING , allowNull: false },
+    takerPubKey: { type: DataTypes.STRING , allowNull: true },
+    price: { type: DataTypes.DECIMAL(8), allowNull: false },
+    takerCltvDelta: { type: DataTypes.NUMBER, allowNull: false },
+    makerCltvDelta: { type: DataTypes.NUMBER, allowNull: true },
```
sangaman: The CLTV deltas can probably be `SMALLINT`.

lib/db/models/SwapDeal.ts
```diff
@@ -0,0 +1,39 @@
...
+    makerAmount: { type: DataTypes.NUMBER, allowNull: false },
```
sangaman: `makerAmount` also should be `BIGINT`.

lib/db/models/SwapDeal.ts
```diff
@@ -0,0 +1,39 @@
...
+import { db } from '../../types';
+
+export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes) => {
+  const attributes: db.SequelizeAttributes<db.SwapDealAttributes> = {
+    role: { type: DataTypes.TINYINT, allowNull: false },
+    state: { type: DataTypes.TINYINT, allowNull: false },
+    phase: { type: DataTypes.TINYINT, allowNull: false },
+    errorReason: { type: DataTypes.STRING, allowNull: true },
+    r_hash: { type: DataTypes.STRING, allowNull: false, unique: true },
+    r_preimage: { type: DataTypes.STRING, allowNull: true },
```
sangaman: Also please make these `rHash` and `rPreimage` since we'll be changing those names internally and it'd be nice not to have to change the schema too.

sangaman:
I'm going to fix the remaining build issues and merge.

sangaman:
@ImmanuelSegol Please don't merge commits without the right commit message template in the future, this was missing a semicolon, and it's definitely a new feature (`feat`) and not a refactoring. We have a [commit template](https://github.com/ExchangeUnion/xud/blob/master/.git-commit-template.txt) that has the proper format. I was about to merge it as I'd said and was writing a proper commit message as below, but now it's too late. I'll try to manually add this in the changelog because it's a significant feature.

```
This commit adds a database table to persist swapdeals. Deals that have
completed, whether due to error or being completed, are stored and then
removed from memory.

Closes #562.
```

PR #570, opened 2018-10-06
feat(getNodeInfo) get information about node

### LINKED ISSUE #529: Add API method to retrieve information from node
If XUD is running in `disableMatching` mode the exchange (or whatever application integrates the API) matches the orders and also triggers swap executions. There are situations in which the reputation of the node could and should affect the behavior of the matching engine when deciding with whom to execute swaps.

Therefore XUD should expose a gRPC method that allows external applications to *get* the reputation of a node.

ImmanuelSegol:
Close #529


sangaman:
At a high level, do we want to return all reputation events? Or just a score? The former seems like it might be a bit excessive, and it would be a bit of a hassle to have to update the reputation event enum in our proto definition every time there's a change there.

Thoughts @michael1011?

michael1011:
@ImmanuelSegol @sangaman I would say just the score. Handling the whole list of events and figuring out how to evaluate the is not something an exchange should take care of. 

But an on the other hand if an exchange disagrees with our evaluation of the importance of a reputation event type it might be good to allow querying all events.

I am not sure and didn't expect this to be implemented so quickly tbh. But I think we need more input of everyone in #529 @offerm @moshababo

ImmanuelSegol:
@michael1011I agree I don't see a reason to show all of the whole lists of events.

sangaman:
Yeah then let's go with just the score. Can you update this PR @ImmanuelSegol ?

@michael1011 I'd say that for the importance of a reputation event type, we should consider making scores for various events configurable at some point. But that would really be an advanced feature and not a priority.

ImmanuelSegol:
> Yeah then let's go with just the score. Can you update this PR @ImmanuelSegol ?
> 
> @michael1011 I'd say that for the importance of a reputation event type, we should consider making scores for various events configurable at some point. But that would really be an advanced feature and not a priority.

@sangaman What do you mean by score? currently we return a list of all reputation events in the nodes history example:
`[ { event: 0, createdAt: 2018-10-16T07:07:34.002Z },{ event: 1, createdAt: 2018-10-16T07:07:51.181Z}]`

Should we just represent the last event? That makes most sense.
But what if we want to evaluate the nodes history for trust reasons?

sangaman:
@ImmanuelSegol By score I mean the `reputationScore` that's a property of all `NodeInstance` objects. I think returning every event is too much information.

I am also thinking now that it might make sense to roll this into a `GetNodeInfo` call that returns basic information about the node, including the reputation score, although it wouldn't hurt to start with just the reputation score and add more info in future PRs. Maybe lets try to discuss some on gitter to get on the same page, I'll bring it up tomorrow.

ImmanuelSegol:
@sangaman What other info besides `reputation` should I add?
I have some ideas but would like to hear your opinion.

lib/cli/commands/getnodereputation.ts
```diff
@@ -0,0 +1,20 @@
+import { Arguments } from 'yargs';
```
sangaman: Rename this file to getnodeinfo.ts.

lib/grpc/GrpcService.ts
```diff
@@ -294,6 +294,21 @@ class GrpcService {
     }
   }
 
+  /**
+   * See [[Service.getNodeReputation]]
```
sangaman: Change to `getNodeInfo`

lib/grpc/GrpcService.ts
```diff
@@ -294,6 +294,21 @@ class GrpcService {
...
+   */
+  public getNodeInfo: grpc.handleUnaryCall<xudrpc.GetNodeInfoRequest, xudrpc.GetNodeInfoResponse> = async (call, callback) => {
+    try {
+      const { banned, reputationScore } = await this.service.getNodeInfo(call.request.toObject());
+      const response = new xudrpc.GetNodeInfoResponse();
+      response.setBanned(banned);
+      response.setReputationscore(reputationScore);
+      callback(null, response);
+    } catch (error) {
```
sangaman: Minor nit but elsewhere in the file we are using `err` instead of `error` if you wouldn't mind changing.

lib/p2p/Pool.ts
```diff
@@ -290,6 +296,25 @@ class Pool extends EventEmitter {
     }
   }
 
+  /**
+   * Get node info
```
sangaman: This comment should line up with the method name, so something like "Gets a node's reputation score and whether it is banned."

lib/p2p/Pool.ts
```diff
@@ -290,6 +296,25 @@ class Pool extends EventEmitter {
...
+   * @param nodePubKey to get reputation events
```
sangaman: Can you change this to "The node pub key of the node for which to get reputation information"?

lib/service/Service.ts
```diff
@@ -187,6 +187,14 @@ class Service extends EventEmitter {
     await this.pool.unban(args);
   }
 
+  /**
+   * Get node info.
+   */
+  public getNodeInfo = async (args: {nodePubKey: string}) => {
```
sangaman: I'm actually surprised this didn't throw a lint error (I'll look into that), but every where else we have a space after the open bracket `{` and before the close bracket `}` if you don't mind changing that.

lib/p2p/Pool.ts
```diff
@@ -36,6 +37,11 @@ type PoolConfig = {
   addresses: string[];
 };
 
+type NodeInfo = {
```
sangaman: I think we should either rename this to `NodeReputationInfo`, or just drop this type and let the return value for `getNodeReputation` be implied, since we're not using the type anywhere else.

proto/xudrpc.proto
```diff
@@ -86,7 +86,13 @@ service Xud {
       get: "/v1/info"
     };
   }
-  
+
+  /*Get node reputation. */
```
sangaman: We can either drop this comment for now or change this to something like `/* Gets general information about a node. */`.

lib/service/Service.ts
ImmanuelSegol: Yeah, I don't know why that happend normally the `npm run lint` saves me from such mistakes.

PR #575, opened 2018-10-09
Track ban

### LINKED ISSUE #565: Track banned nodes in NodeList
Right now when xud starts up, it does not track any nodes that have been previously banned. That means that we could theoretically accept a connection with a banned node, resulting in a uniqueness constraint violation when we try to add the banned node to the database.

Instead, when we read from the nodes table and come across a banned node, we should track it in a separate list of banned node pub keys. Then in `Pool.openPeer` we should check to make sure the node pub key of the newly connected peer does not appear in the banned list.

This also means that, when a node is banned, we should remove it from the `nodes` map in `NodesList` and add its pub key to the list of banned pub keys.

ImmanuelSegol:
Close #565


lib/p2p/NodeList.ts
```diff
@@ -39,10 +40,26 @@ class NodeList extends EventEmitter {
   /**
    * Check if a node with a given nodePubKey exists.
    */
-  public has = (nodePubKey: string) => {
+  public has = (nodePubKey: string): boolean => {
     return this.nodes.has(nodePubKey);
   }
 
+  public remove = (nodePubKey: string): void => {
```
sangaman: Do we need this method? It doesn't look like it's being used and I don't think we'd want to remove/forget about nodes in this way.

lib/p2p/NodeList.ts
```diff
@@ -39,10 +40,26 @@ class NodeList extends EventEmitter {
...
+    this.nodes.delete(nodePubKey);
+  }
+
+  public inBannedList = (nodePubKey: string): boolean  => {
```
sangaman: I think we should use/modify the existing `isBanned` method for this purpose.

lib/p2p/NodeList.ts
```diff
@@ -39,10 +40,26 @@ class NodeList extends EventEmitter {
...
+    return this.bannedNodes.has(nodePubKey);
+  }
+
+  public addBannedList = (nodePubKey: string): void  => {
+    this.bannedNodes.add(nodePubKey);
+  }
+
+  public removeBannedList = (nodePubKey: string): void  => {
```
sangaman: Similarly for these remove and add banned list methods, I think it makes most sense just to add/remove to the list when we're banning/unbanning nodes. In other words, we wouldn't want to remove or add to the list unless we're actually banning or unbanning.

lib/p2p/NodeList.ts
```diff
@@ -141,6 +160,9 @@ class NodeList extends EventEmitter {
       } else if (node.banned) {
         // If the reputationScore is not below the banThreshold but node.banned
         // is true that means that the node was unbanned
+        if (this.inBannedList(node.nodePubKey)) {
+          this.removeBannedList(node.nodePubKey);
```
sangaman: Here for example, I think we can remove the node from the list within the `setBanStatus` method, and we shouldn't need to check twice if the node is banned.

lib/p2p/Pool.ts
```diff
@@ -330,19 +331,22 @@ class Pool extends EventEmitter {
...
-    try {
-      await peer.open(this.handshakeData, nodePubKey, retryConnecting);
-    } catch (err) {
+    const isBanned = nodePubKey ? this.nodes.inBannedList(nodePubKey) : false;
+    if (!isBanned) {
+      this.bindPeer(peer);
+      try {
+        await peer.open(this.handshakeData, nodePubKey, retryConnecting);
+      } catch (err) {
       // we don't have `nodePubKey` for inbound connections, which might fail on handshake
```
sangaman: This comment line should be indented as well.

ImmanuelSegol:
@sangaman I recompiled the proto files as well because someone didn't compile them properly and the deprecation warning came up.

lib/p2p/NodeList.ts
```diff
@@ -64,8 +65,8 @@ class NodeList extends EventEmitter {
   }
 
   public isBanned = (nodePubKey: string): boolean => {
-    const node = this.nodes.get(nodePubKey);
-    return node ? node.banned : false;
+    const node = this.bannedNodes.has(nodePubKey);
+    return node ? true : false;
```
sangaman: I think this is cleaner in one line: `return this.bannedNodes.has(nodePubKey);`

lib/p2p/NodeList.ts
```diff
@@ -137,10 +140,13 @@ class NodeList extends EventEmitter {
 
       if (node.reputationScore < NodeList.banThreshold) {
         promises.push(this.setBanStatus(node, true));
+        this.bannedNodes.add(node.nodePubKey);
         this.emit('node.ban', nodePubKey);
       } else if (node.banned) {
+
```
sangaman: Can you drop this blank line?

PR #577, opened 2018-10-10
refactor: treat failed swaps as invalidations

sangaman:
This PR changes the event emitted for failed swaps during matching from `peerOrder.filled` to `peerOrder.invalidated`. It also doesn't add back peer orders to the book if they could not be swapped, and it makes the `verifyExecution` method private. This is so that we don't confuse orders that are removed from the order book because they were successfully swapped with those that were removed without being swapped.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -257,17 +257,6 @@ class OrderBook extends EventEmitter {
     // instantiate a container for failed swaps, for retry purposes
     const swapFailures: StampedOwnOrder[] = [];
 
-    // append the taker quantity to the remaining order, after making sure it's initialized.
-    // if the maker is given, re-add it to the repository
-    const rejectNonInternalMatch = (taker: StampedOwnOrder, maker?: StampedPeerOrder) => {
-      result.remainingOrder = result.remainingOrder || { ...order, quantity: 0 };
-      result.remainingOrder.quantity += taker.quantity;
-
-      if (maker) {
-        matchingEngine.addPeerOrder(maker);
-      }
-    };
-
     // iterate over the matches
     for (const { maker, taker } of matchingResult.matches) {
       const portion: OrderPortion = { orderId: maker.id, pairId: maker.pairId, quantity: maker.quantity };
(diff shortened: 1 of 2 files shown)


lib/orderbook/OrderBook.ts
```diff
@@ -279,19 +268,19 @@ class OrderBook extends EventEmitter {
...
 
-        this.emit('peerOrder.filled', portion); // make sure we emit this event on every case in which we don't re-add the maker order
         try {
           const swapResult = await this.swaps.executeSwap(maker, taker);
+          this.emit('peerOrder.filled', portion);
           result.swapResults.push(swapResult);
           onUpdate && onUpdate({ case: PlaceOrderEventCase.SwapResult, payload: swapResult });
         } catch (err) {
-          // we can either push to swapFailures, or reject in case of non-retry errors
+          this.emit('peerOrder.invalidation', portion);
```
michael1011: Although it is nice to have for testing you might want to put in a todo so that we don't forget to remove this line in the future.

sangaman: Remove the `this.emit('peerOrder.invalidation', portion);`? In my mind this is deliberate, we want to get rid of peer orders that we can't swap.

lib/orderbook/OrderBook.ts
```diff
@@ -279,19 +268,19 @@ class OrderBook extends EventEmitter {
         this.emit('ownOrder.filled', portion);
         onUpdate && onUpdate({ case: PlaceOrderEventCase.InternalMatch, payload: maker });
       } else {
-        if (!this.swaps || !this.swaps.verifyExecution(maker, taker)) {
-          rejectNonInternalMatch(taker, maker);
+        if (!this.swaps) {
           continue;
```
moshababo: You are missing here the part of appending the taker order to the remaining order. It will cause a portion of the taker order to disappear.

moshababo: We're already doing it when handling the swap failures. Instead, we can do it here and also after an exception is thrown (line 280), so it will be more clear that it's happening on both of these cases.

sangaman: OK, I was mentally treating the case of no swaps module differently because that should only happen during testing, but I'll treat it the same as a swap failure for now.

moshababo: This is not optimal IMO, because you will have a few futile swap retries iterations untill reaching the timeout, where none can actually succeed.
But since this condition is solely for testing, i'm fine with not over-optimizing it, also because this whole area is subject to future changes anyway.

sangaman:
@moshababo I resolved the conflicts, I'm still not sure about the optimal approach for when `this.swaps` is undefined. If you can clarify that's fine, or if indeed it's not a big deal since it doen't affect actual operation (and will change in the future anyway) you can approve or merge.

PR #579, opened 2018-10-10
feat: check if order is swappable before broadcast

sangaman:
This PR adds a check before broadcasting an order to peers to ensure that we have active swap clients for both sides of the trading pair for that order. In other words, we check to make sure that an order can be swapped before advertising it to peers. This is done to avoid polluting peers' order books with orders that can't possibly be swapped, which can result in unnecessary swap failures and banning.

### DIFF
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -103,21 +103,12 @@ class Swaps extends EventEmitter {
   }
 
   /**
-   * Verifies that the order still exists on orderBook
+   * Checks if there exist active swap clients for both currencies in a given trading pair.
+   * @returns `true` if the pair has swap support, `false` otherwise
    */
-  // TODO: replace the dummy implementation with real order book integration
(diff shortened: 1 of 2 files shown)


lib/swaps/Swaps.ts
```diff
@@ -103,21 +103,12 @@ class Swaps extends EventEmitter {
...
-  }
-
-  /**
-   * Verifies that the order still exists on orderBook
-   */
-  // TODO: replace the dummy implementation with real order book integration
-  private checkAndHoldAmount = (deal: SwapDeal): number => {
-    this.logger.debug('checking and holding amount for swap ' + deal.pairId + ':' + deal.orderId + ':' + deal.takerAmount);
-    return deal.takerAmount;
+  public isPairSupported = (pairId: string): boolean => {
```
moshababo: this should to be united with the temp `verifyExecution` method I recently added

sangaman: Makes sense. You mean in addition to being used here right? I'll add it.

PR #581, opened 2018-10-12
Minor swap refactoring/renaming

### LINKED ISSUE #551: Minor swap refactoring/renaming

ImmanuelSegol:
Closes #551 

@sangaman I didn't understand what you meant by: 

> On error/failure responses, send an enum failure reason rather than a string. This involves less data transmission between peers, and more importantly it makes checking the error or failure reason easier since we don't need to compare exact strings where a typo could mess things up and go undetected by the compiler.

Where you refering to `swap.failed`?

### DIFF
--- a/lib/p2p/packets/types/SwapAcceptedPacket.ts
+++ b/lib/p2p/packets/types/SwapAcceptedPacket.ts
@@ -3,7 +3,7 @@ import PacketType from '../PacketType';
 
 // TODO: proper error handling
 export type SwapResponsePacketBody = {
-  rHash: string;
+  r_hash: string;
   /** Specifies the accepted quantity (which may be less than the proposed quantity). */
   quantity: number;
   makerCltvDelta: number;
--- a/lib/p2p/packets/types/SwapFailedPacket.ts
+++ b/lib/p2p/packets/types/SwapFailedPacket.ts
@@ -3,7 +3,7 @@ import PacketType from '../PacketType';
 
 // TODO: proper error handling
 export type SwapErrorPacketBody = {
-  rHash: string;
+  r_hash: string;
   errorMessage: string;
 };
 
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -103,21 +103,12 @@ class Swaps extends EventEmitter {
   }
 
   /**
-   * Verifies that the order still exists on orderBook
+   * Checks if there exist active swap clients for both currencies in a given trading pair.
+   * @returns `true` if the pair has swap support, `false` otherwise
    */
-  // TODO: replace the dummy implementation with real order book integration
-  private isOrderOnBook = (pairId: string, orderId: string): boolean => {
-    this.logger.debug('checking order book for ' + pairId + ':' + orderId);
-    return true;
-  }
-
-  /**
-   * Verifies that the order still exists on orderBook
-   */
-  // TODO: replace the dummy implementation with real order book integration
-  private checkAndHoldAmount = (deal: SwapDeal): number => {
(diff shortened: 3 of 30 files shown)


proto/lndrpc.proto
```diff
@@ -1625,10 +1625,10 @@ message Invoice {
     The hex-encoded preimage (32 byte) which will allow settling an incoming
     HTLC payable to this preimage
     */
-    bytes r_preimage = 3 [json_name = "r_preimage"];
+    bytes rPreimage = 3 [json_name = "rPreimage"];
```
sangaman: FYI we're using snake_case for the proto definition, I was thinking of only changing the names internally.

lib/p2p/packets/types/SwapFailedPacket.ts
```diff
@@ -7,18 +7,18 @@ export type SwapErrorPacketBody = {
   errorMessage: string;
 };
 
-class SwapErrorPacket extends Packet<SwapErrorPacketBody> {
+class SwapFailedPacket extends Packet<SwapErrorPacketBody> {
```
sangaman: Per my last comment, let's leave the packet renaming changes out for now because they are a breaking change for p2p communication.

lib/p2p/packets/types/SwapAcceptedPacket.ts
```diff
@@ -9,7 +9,7 @@ export type SwapResponsePacketBody = {
   makerCltvDelta: number;
 };
 
-class SwapResponsePacket extends Packet<SwapResponsePacketBody> {
+class SwapAcceptedPacket extends Packet<SwapResponsePacketBody> {
```
sangaman: Let's leave this as the old name until we're closer to the next release.

lib/swaps/Swaps.ts
```diff
@@ -57,7 +57,7 @@ type SwapDeal = {
   makerToTakerRoutes?: lndrpc.Route[];
   createTime: number;
   executeTime?: number;
-  competionTime?: number
+  completionTime?: number
```
sangaman: Can we call this `completeTime`? Just to be consistent with `createTime` and `executeTime` above.

ImmanuelSegol:
@sangaman Do you know how I can fix the travis build ? I think there is some conflicts this PR was made before I got access to the repo.

PR #583, opened 2018-10-16
Track ban

### LINKED ISSUE #565: Track banned nodes in NodeList

ImmanuelSegol:
Close #565 

This is a continuation of #575.

### DIFF
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -140,11 +140,13 @@ class NodeList extends EventEmitter {
       if (node.reputationScore < NodeList.banThreshold) {
         promises.push(this.setBanStatus(node, true));
         this.bannedNodes.add(node.nodePubKey);
+        this.nodes.delete(node.nodePubKey);
         this.emit('node.ban', nodePubKey);
       } else if (node.banned) {
         // If the reputationScore is not below the banThreshold but node.banned
         // is true that means that the node was unbanned
         this.bannedNodes.delete(node.nodePubKey);
+        this.nodes.set(node.nodePubKey, node);
         promises.push(this.setBanStatus(node, false));
       }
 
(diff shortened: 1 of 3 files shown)


lib/p2p/NodeList.ts
```diff
@@ -137,10 +139,12 @@ class NodeList extends EventEmitter {
 
       if (node.reputationScore < NodeList.banThreshold) {
         promises.push(this.setBanStatus(node, true));
+        this.bannedNodes.add(node.nodePubKey);
         this.emit('node.ban', nodePubKey);
       } else if (node.banned) {
         // If the reputationScore is not below the banThreshold but node.banned
         // is true that means that the node was unbanned
+        this.bannedNodes.delete(node.nodePubKey);
```
sangaman: We want to add the unbanned node to `this.nodes` here.

lib/p2p/NodeList.ts
```diff
@@ -137,10 +139,12 @@ class NodeList extends EventEmitter {
 
       if (node.reputationScore < NodeList.banThreshold) {
         promises.push(this.setBanStatus(node, true));
+        this.bannedNodes.add(node.nodePubKey);
```
sangaman: We want to delete the banned node from `this.nodes` here. This is to ensure we don't share the banned node with other peers and to be consistent with the loading logic which only adds unbanned nodes to `this.nodes`.

PR #585, opened 2018-10-16
P2P integration for added/removed trading pairs

sangaman:
This PR sends handshake state updates to peers when trading pairs are added or removed, and also broadcasts order invalidations for any orders belonging to a trading pair that we are removing.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -171,6 +171,9 @@ class OrderBook extends EventEmitter {
     this.pairs.set(pairInstance.id, pairInstance);
     this.matchingEngines.set(pairInstance.id, new MatchingEngine(this.logger, pairInstance.id));
 
+    if (this.pool) {
+      this.pool.updateHandshake({ pairs: this.pairIds });
+    }
     return pairInstance;
   }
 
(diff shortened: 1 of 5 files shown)


lib/orderbook/OrderBook.ts
```diff
@@ -62,9 +62,9 @@ class OrderBook extends EventEmitter {
   /** Max time for addOwnOrder iterations (due to swaps failures retries). */
   private static MAX_ADD_OWN_ORDER_ITERATIONS_TIME = 10000; // 10 sec
```
michael1011: Not related to this PR but since when are we using this capitalizing standard? And why it is not `readonly`?

sangaman: I couldn't find a standard convention for TypeScript, but I know that the UPPER_SNAKE_CASE is common across programming languages for final constants like this one.

Still, `readonly` wpuld make sense. I checked and right now the compiler will let me reassign these static fields, which clearly we don't want. Also the naming style isn't totally consistent. I'll open an issue.

lib/orderbook/OrderBook.ts
```diff
@@ -201,9 +204,20 @@ class OrderBook extends EventEmitter {
...
       this.pairs.delete(pairId);
-      this.matchingEngines.delete(pairId);
-      // TODO: invalidate all orders for this pair
-      // TODO: update handshake state
+      const matchingEngine = this.matchingEngines.get(pairId);
+
+      if (matchingEngine) {
+        if (this.pool) {
+          matchingEngine.ownOrders.buy.forEach(this.pool.broadcastOrderInvalidation);
+          matchingEngine.ownOrders.sell.forEach(this.pool.broadcastOrderInvalidation);
```
moshababo: didn't we say we want to allow to invalidate a pair, instead of broadcasting a packet for each order separately?

lib/orderbook/OrderBook.ts
```diff
@@ -201,9 +204,20 @@ class OrderBook extends EventEmitter {
     const pair = this.pairs.get(pairId);
     if (pair) {
       this.pairs.delete(pairId);
-      this.matchingEngines.delete(pairId);
-      // TODO: invalidate all orders for this pair
-      // TODO: update handshake state
+      const matchingEngine = this.matchingEngines.get(pairId);
+
+      if (matchingEngine) {
```
moshababo: should we assert this, instead of having a silent error?

we also have `getMatchingEngine` which throws exception, but in this case this error is not expected to ever happen.

lib/orderbook/OrderBook.ts
moshababo: Good idea!

sangaman:
I started working on removing orders when a peer updates handshake state and drops a pair, but I realized it would conflict with - and be much easier to do with - the functionality from #589. So I'll create an issue and handle it after that's merged.

PR #589, opened 2018-10-17
refactor(p2p/orderbook): group peer orders by peerPubKey

erkarl:
Adds logic to check that `peerPubKey` on the order matches the pub key
of the peer who sent the invalidation packet.

Closes: https://github.com/ExchangeUnion/xud/issues/242

### DIFF
--- a/lib/orderbook/MatchingEngine.ts
+++ b/lib/orderbook/MatchingEngine.ts
@@ -11,9 +11,7 @@ type SplitOrder = {
   remaining: StampedOrder;
 };
 
-type orderId = string;
-
-type OrderList<T extends orders.StampedOrder> = Map<orderId, T>;
+type OrderList<T extends orders.StampedOrder> = Map<string, T>;
 
 type OrderSidesLists<T extends orders.StampedOrder> = {
   buy: OrderList<T>,
@@ -34,10 +32,10 @@ type OrderSidesQueues = {
 class MatchingEngine {
   /** A pair of priority queues for the buy and sell sides of this trading pair */
   public queues: OrderSidesQueues;
-  /** a pair of maps between active own orders ids and orders for the buy and sell sides of this trading pair. */
+  /** A pair of maps between active own orders ids and orders for the buy and sell sides of this trading pair. */
   public ownOrders: OrderSidesLists<orders.StampedOwnOrder>;
-  /** a pair of maps between active peer orders ids and orders for the buy and sell sides of this trading pair. */
-  public peerOrders: OrderSidesLists<orders.StampedPeerOrder>;
+  /** A map between peerPubKey and a pair of maps between active peer orders ids and orders for the buy and sell sides of this trading pair. */
+  public peersOrders: Map<string, OrderSidesLists<orders.StampedPeerOrder>>;
 
   constructor(private logger: Logger, public pairId: string) {
     this.queues = {
@@ -46,14 +44,11 @@ class MatchingEngine {
     };
 
     this.ownOrders = {
-      buy: new Map<orderId, StampedOwnOrder>(),
-      sell: new Map<orderId, StampedOwnOrder>(),
+      buy: new Map<string, StampedOwnOrder>(),
+      sell: new Map<string, StampedOwnOrder>(),
     };
 
-    this.peerOrders = {
-      buy: new Map<orderId, StampedPeerOrder>(),
-      sell: new Map<orderId, StampedPeerOrder>(),
-    };
+    this.peersOrders = new Map<string, OrderSidesLists<orders.StampedPeerOrder>>();
   }
 
   private static createPriorityQueue = (orderingDirection: OrderingDirection): FastPriorityQueue<StampedOrder> => {
@@ -105,7 +100,15 @@ class MatchingEngine {
    * @returns false if it's a duplicated order, otherwise true
    */
   public addPeerOrder = (order: StampedPeerOrder): boolean => {
-    return this.addOrder(order, this.peerOrders);
+    let peerOrdersList = this.peersOrders.get(order.peerPubKey);
+    if (!peerOrdersList) {
+      peerOrdersList = {
+        buy: new Map<string, StampedPeerOrder>(),
+        sell: new Map<string, StampedPeerOrder>(),
+      };
+      this.peersOrders.set(order.peerPubKey, peerOrdersList);
+    }
+    return this.addOrder(order, peerOrdersList);
   }
 
   // TODO: remove method
@@ -155,11 +158,8 @@ class MatchingEngine {
       ...this.queues.sell.removeMany(callback),
     ] as StampedPeerOrder[];
 
-    // remove from lists. for further optimization, we can maintain a separate list for each peer pubKey
-    removedOrders.forEach((order: StampedPeerOrder) => {
-      const list = order.isBuy ? this.peerOrders.buy : this.peerOrders.sell;
-      list.delete(order.id);
-    });
+    // remove from lists.
+    this.peersOrders.delete(peerPubKey);
 
     return removedOrders;
   }
@@ -170,8 +170,12 @@ class MatchingEngine {
    * quantity then the entire order is removed
    * @returns the removed order or order portion, otherwise undefined if the order wasn't found
    */
-  public removePeerOrder = (orderId: string, quantityToRemove?: number): { order: StampedOwnOrder, fullyRemoved: boolean} => {
-    return this.removeOrder(orderId, this.peerOrders, quantityToRemove);
+  public removePeerOrder = (orderId: string, peerPubKey: string, quantityToRemove?: number): { order: StampedOwnOrder, fullyRemoved: boolean} => {
+    const peerOrdersList = this.peersOrders.get(peerPubKey);
+    if (!peerOrdersList) {
+      throw errors.ORDER_NOT_FOUND(orderId);
+    }
+    return this.removeOrder(orderId, peerOrdersList, quantityToRemove);
   }
 
   /**
@@ -208,11 +212,13 @@ class MatchingEngine {
     }
   }
 
-  private getOrderList = (order: StampedOrder): OrderList<StampedOrder> => {
+  private getOrderList = (order: StampedOrder): OrderList<StampedOrder> | undefined => {
     if (isOwnOrder(order)) {
       return order.isBuy ? this.ownOrders.buy : this.ownOrders.sell;
     } else {
-      return order.isBuy ? this.peerOrders.buy : this.peerOrders.sell;
+      const peerOrdersList = this.peersOrders.get(order.peerPubKey);
+      if (!peerOrdersList) return;
+      return order.isBuy ? peerOrdersList.buy : peerOrdersList.sell;
     }
   }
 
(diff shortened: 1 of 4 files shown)


lib/orderbook/errors.ts
```diff
@@ -45,6 +46,10 @@ const errors = {
     message: `order with id ${orderId} could not be found`,
     code: errorCodes.ORDER_NOT_FOUND,
   }),
+  UNAUTHORIZED: (orderId: string, peerPubKey: string) => ({
+    message: `order with id ${orderId} does not belong to ${peerPubKey}`,
```
sangaman: I'd be inclined to either rename this error to something like `UNAUTHORIZED_INVALIDATION` or to make the message more generic, like `peer attempted to perform an unauthorized action`.

lib/orderbook/OrderBook.ts
```diff
@@ -79,7 +79,8 @@ class OrderBook extends EventEmitter {
   private bindPool = () => {
     if (this.pool) {
       this.pool.on('packet.order', this.addPeerOrder);
-      this.pool.on('packet.orderInvalidation', order => this.removePeerOrder(order.orderId, order.pairId, order.quantity));
+      this.pool.on('packet.orderInvalidation', order =>
+        this.removePeerOrder(order.orderId, order.pairId, order.peerPubKey, order.quantity));
```
sangaman: We'd probably want a try/catch here, and the TODO for penalizing the peer should be in the case we get an unauthorized error.

sangaman:
So @moshababo had actually just asked me about changing the way we structure our orders so that we group them by peers. In other words instead of having two maps of `orderId` to `StampedPeerOrder` for buys and sells, we'd have multiple pairs of buy/sell maps like that nested within a map of `peerPubKeys`. It has some performance/structural benefits and I think this would be one of them.

Basically with a structure like that, we could just attempt to remove the invalidated order from that peer's orders, resulting in an `ORDER_NOT_FOUND` error - which we would still want to penalize a peer for. I think that would be a bit cleaner and would require fewer changes to the order removal logic. 

If that makes sense to you, I'd probably prefer that approach. I'm thinking we'd add `peerPubKey` as a nullable field to `OrderIdentifier`, and change `removePeerOrder` in MatchingEngine.ts to get the `OrderSidesList` for the particular peer and pass that to `removeOrder`. It would be two birds with one stone I believe.

erkarl:
@sangaman thanks for the comments.

I agree with your recommended approach - will make the necessary changes.

erkarl:
@sangaman updated.

I noticed @moshababo also implemented those changes in his PR. If we get this PR merged first we'll reduce the amount of changes to review in https://github.com/ExchangeUnion/xud/pull/594 as it's quite a huge PR. I've aligned the variable names etc. so it would be easier to rebase.

Alternatively we could close this PR and open a separete one later just for penalizing the peer/reputation event.

sangaman:
Thanks, will probably discuss with @moshababo  tomorrow about the best way to approach but I'd probably agree that it makes more sense in this PR to reduce the scope of #594. At a glance your changes look good, I'll do a more detailed review tomorrow if we go forward with keeping this PR and rebasing #594 on top of this.

lib/orderbook/MatchingEngine.ts
```diff
@@ -37,7 +38,7 @@ class MatchingEngine {
   /** a pair of maps between active own orders ids and orders for the buy and sell sides of this trading pair. */
   public ownOrders: OrderSidesLists<orders.StampedOwnOrder>;
   /** a pair of maps between active peer orders ids and orders for the buy and sell sides of this trading pair. */
```
moshababo: You can use the updated comment: `  /** a map between peerPubKey and a pair of maps between active peer orders ids and orders for the buy and sell sides of this trading pair. */`

lib/orderbook/OrderBook.ts
```diff
@@ -391,14 +392,19 @@ class OrderBook extends EventEmitter {
...
+  private removePeerOrder = (orderId: string, pairId: string, peerPubKey: string, quantityToRemove?: number) => {
     const matchingEngine = this.getMatchingEngine(pairId);
     try {
-      const removeResult = matchingEngine.removePeerOrder(orderId, quantityToRemove);
+      const removeResult = matchingEngine.removePeerOrder(orderId, peerPubKey, quantityToRemove);
       this.emit('peerOrder.invalidation', { orderId, pairId, quantity: removeResult.order.quantity });
       return true;
     } catch (err) {
-      this.logger.error(`attempted to remove non-existing orderId (${orderId})`);
+      let errMsg = `attempted to remove non-existing orderId (${orderId})`;
```
moshababo: we don't know anymore that the orderId doesn't exists. just that it doesn't exist for this specific peer

lib/orderbook/OrderBook.ts
```diff
@@ -391,14 +392,19 @@ class OrderBook extends EventEmitter {
...
+      if (err && err.code === errorCodes.PEER_ORDER_NOT_FOUND) {
```
moshababo: I think we can assume `err` is not nullable

moshababo: at the current situation, you will get this error code only if there are no orders at all for the peer. If there are any, but this specific order is missing, you will get a different error. This is quite confusing and problematic IMO, although at the moment it's affecting just the logging message. 

My suggestion: 
1) `MatchingEngine.removeOrder` should throw the specific order type error (own/peer), or instead `MatchingEngine.removePeerOrder`/`removeOwnOrder` should throw the generic order error. I prefer the former. 
2) I don't think we should use the underlying error message, which doesn't give the full context that we are aware of at this point. I would rather remove the error handling from `OrderBook.removePeerOrder`, and create `OrderBook.handleOrderInvalidation` method (for the `Pool` event listener) to catch the error and to log a better error message.

erkarl: @moshababo thanks for the feedback.

1. Good catch. I think we want to distinguish between own/peer order not found errors. I propose we change the signature of `MatchingEngine.removeOrder` to accept an `order` instead of `orderId` as param. This way we can fetch throw more specific error messages in `MatchingEngine.removePeerOrder` and `MatchingEngine.removeOwnOrder`.

2. Agreed. Will make the changes.

lib/orderbook/MatchingEngine.ts
sangaman: Minor nit if we're changing this anyway, but these comments should start with capital letters.

lib/orderbook/MatchingEngine.ts
```diff
@@ -152,11 +158,14 @@ class MatchingEngine {
       ...this.queues.sell.removeMany(callback),
     ] as StampedPeerOrder[];
 
-    // remove from lists. for further optimization, we can maintain a separate list for each peer pubKey
-    removedOrders.forEach((order: StampedPeerOrder) => {
-      const list = order.isBuy ? this.peerOrders.buy : this.peerOrders.sell;
-      list.delete(order.id);
-    });
+    // remove from lists.
+    const peerOrdersList = this.peersOrders.get(peerPubKey);
```
sangaman: I think we can just do `this.peersOrders.delete(peerPubKey)` in stead of looping through and deleting them one by one, right?

lib/orderbook/OrderBook.ts
sangaman: For `MatchingEngine.removeOrder` to accept `order instead of `orderId`, I was toying with that a week or two ago and I forget the specifics but I came away thinking it was a lot cleaner to stick to `orderId`. I'm not necessarily opposed, but I remember having this same thought and then thinking better of it.

But I doubt whether we need separate error types for peer or own order not found. Why not just have one, and then let the caller of the function determine how to handle it? The caller will be calling some either `removeOwnOrder` or `removePeerOrder` and therefore will already know if it's a peer order or an own order that wasn't found, even with a single error type for both of them.

lib/orderbook/MatchingEngine.ts
```diff
@@ -209,7 +222,15 @@ class MatchingEngine {
...
       return order.isBuy ? this.ownOrders.buy : this.ownOrders.sell;
     } else {
-      return order.isBuy ? this.peerOrders.buy : this.peerOrders.sell;
+      let peerOrdersList = this.peersOrders.get(order.peerPubKey);
+      if (!peerOrdersList) {
+        peerOrdersList = {
+          buy: new Map<orderId, StampedPeerOrder>(),
+          sell: new Map<orderId, StampedPeerOrder>(),
+        };
+        this.peersOrders.set(order.peerPubKey, peerOrdersList);
```
sangaman: I don't think we're trying to add the order to the list in this method. I'm thinking that the list really *should* already exist at this point (having been created when the order was added to it). In fact, the only time we call this method, we call `delete` on the the order immediately after.

I'd probably just have this return `undefined` if `peerOrdersList` is undefined.

lib/orderbook/MatchingEngine.ts
```diff
@@ -220,8 +241,30 @@ class MatchingEngine {
...
+  private getOrder = <T extends StampedOrder>(orderId: string, maps: OrderSidesLists<T>): T | undefined => {
+    return maps.buy.get(orderId) || maps.sell.get(orderId);
+  }
+
+  public getPeerOrders = (): OrderSidesArrays<StampedPeerOrder> => {
+    const res: OrderSidesArrays<StampedPeerOrder> = { buy: [], sell: [] };
+    this.peersOrders.forEach((peerOrders) => {
+      const peerOrdersArrs = this.getOrders(peerOrders);
+      res.buy.push(...peerOrdersArrs.buy);
+      res.sell.push(...peerOrdersArrs.sell);
```
sangaman: I believe we get a performance edge here using `concat` over `push` with the rest operator. I did a quick test and it seems to confirm that: http://jsben.ch/SZYnZ

You should be able to do `res.buy = res.buy.concat(peerOrdersArrs.buy);`.

lib/p2p/Pool.ts
```diff
@@ -466,7 +466,7 @@ class Pool extends EventEmitter {
       case PacketType.OrderInvalidation: {
         const order = (packet as packets.OrderInvalidationPacket).body!;
         this.logger.verbose(`canceled order from ${peer.nodePubKey}: ${JSON.stringify(order)}`);
-        this.emit('packet.orderInvalidation', order);
+        this.emit('packet.orderInvalidation', { ...order, peerPubKey: peer.nodePubKey } as OrderInvalidation);
```
sangaman: I also have a performance concern here, where with the rest operator again we're creating a new object which is a minor performance hit but one that we'd have to do many times. I think it would be just as clear if we used two parameters, the `OrderPortion` and the `peer.nodePubKey` as a separate string. We just turn around and pass them in as separate parameters to `removePeerOrder` anyway, so we're not gaining much by having these combing into a single object as I see it.

test/unit/MatchingEngine.spec.ts
```diff
@@ -248,12 +248,27 @@ describe('MatchingEngine.removePeerOrders', () => {
...
-    let removedOrder = engine.removePeerOrder(order.id, quantityToRemove).order;
+    let removedOrder = engine.removePeerOrder(order.id, order.peerPubKey, quantityToRemove).order;
     expect(removedOrder.quantity).to.be.equal(quantityToRemove);
 
-    removedOrder = engine.removePeerOrder(order.id).order;
+    removedOrder = engine.removePeerOrder(order.id, order.peerPubKey).order;
     expect(removedOrder.quantity).to.be.equal(quantity - quantityToRemove);
   });
+
+  it('should add a new peerOrder and then not allow other peer to remove it', () => {
```
sangaman: Since this test suite doesn't really have knowledge of other peers, maybe just describe this test case as "should not remove a peer order using the wrong peer pub key". Thanks for adding this test case though.

lib/orderbook/OrderBook.ts
moshababo: I think that accepting `order` instead of `orderId` is an overkill to this problem. 

I was thinking of either inspecting the T value somehow (by the injected lists instance, if it's viable), or injecting the error to throw. Other option what @sangaman suggested - to drop the separated errors entirely, and to use only `ORDER_NOW_FOUND`.

erkarl:
Updated. I went with the generic error message approach.

lib/orderbook/MatchingEngine.ts
```diff
@@ -12,6 +12,7 @@ type SplitOrder = {
 };
 
 type orderId = string;
+type PeerPubKey = string;
```
sangaman: Not a big deal at all but I'm not sure we need a type alias for `PeerPubKey`. I didn't know we had an `orderId` type actually, I don't think we're using it in most places. My main concern is that using it inconsistently would be bit confusing.

erkarl:
@sangaman updated + squashed. I also removed `orderId` basic type to keep it consistent of not having a `peerPubKey: string` type.

moshababo:
@erkarl Thanks.. i'm going to merge this so we can continue to develop on top of it.

moshababo:
Travis is stuck for some reason..

erkarl:
@moshababo looks like the builds have been rejected because of "Abuse detected". I'll try opening another PR to see if it triggers a build.

erkarl:
Unfortunately, it didn't work :(.

moshababo:
haha.. ok, let's see what they're up to.

erkarl:
@moshababo the build is now passing :). Travis support is fast!

Issue #145, opened 2018-06-27
Add disableMatching mode


moshababo:
`xud` should have only 2 matching modes:  do no matching, or all matching.
 represent this by adding a boolean config property, `disableMatching`, which its default value will be `false`. It should be a top-level config (not under a specific module). Should be passed, at least, to the order book module. 

DONE
* disableMatching = false:
  * placeOrder (returns matchingResult) (matching result= successfullyMatched[orderID,quantity],  remaining[orderID,quantity]
  * cancelOrder (returns success/error)
  * subscribePeerOrders [streaming]
  * subscribeSwaps [streaming]

TODO
I'd avoid the double negation `disableMatching = false`, also unified gRPC as per https://github.com/ExchangeUnion/xud/issues/123#issuecomment-419176541 with following events:

1) ownOrder added (`placeOrder`)
2) ownOrder removed (`cancelOrder`)
3) ownOrder removed (due to internal match)
4) ownOrder removed (due to a successful swap that was instantiated remotely)
5) peerOrder added
6) peerOrder removed (due to non-internal match)
7) peerOrder removed (due to peer order invalidation)
8) ownOrder incomingSwap (incoming swap that was instantiated remotely)
9) peerOrder initiateSwap (due to non-internal match)

DONE (https://github.com/ExchangeUnion/xud/issues/123)
`matching = true;`
* placeOrder (sync, returns success/error) event `(1)`
* cancelOrder (sync, returns success/error) event `(2)`
* subscribeOrders [streaming] events `(3), (4), (5), (6), (7)`

TODO (this issue)
`matching = false;`
* placeOrder (sync, returns success/error) event `(1)` (propagates ownOrder to peers)
* cancelOrder (sync, returns success/error) event `(2)` (sends order invalidation packet for ownOrder to peers)
* subscribeOrders (streaming) events `(5), (7), (8), (9)`
* executeSwap (async, returns success/error) event, swap result via (9) in subscribe orders
* executeSwapSync (returns swap result) 

Reminder: [matching scenarios](https://github.com/ExchangeUnion/xud/issues/277#issuecomment-416370196)

kilrau:
Related: https://github.com/ExchangeUnion/xud/issues/146

itsbalamurali:
@moshababo I've updated the issue description since #146 is closed and internal matching is implemented as part of #196

kilrau:
I think this task is rather about implementing the non-matching mode which needs some fleshing out on implementation side.

[...]

PR #594, opened 2018-10-21
feat: nomatching mode

moshababo:
Solves #145 

* supporting `nomatching` mode
* introduced `placeOrder` to the order book API, so it could be distinguished from `addOwnOrder`
* `MatchingEngine` was renamed to `TradingPair`
* `executeSwap` high-level API method was added
* peers orders are now grouped by `peerPubKey` 
* made sure we don't throw errors on `Pool` event handlers (otherwise peers could crash local `xud` process)

I didn't commit the tests refactoring, I'd like to get the initial feedback first.


moshababo:
@sangaman you’re definitely right, I overlooked that the main use-case will be for executing taker orders which do not exists in xud. 

Do we even want to make the other way optional? As you said, it’s going against the protocol (at least the one that is implemented in the matching mode), and if the order was placed in xud, it’s considered a maker order, and the swap is expected to be initiated by a remote peer. Allowing to trigger a swap anyway gives the exchange more flexibility, but i'm not sure that we want to support it, at least at this stage.

moshababo:
In regards to #589, I agree that I should wait for it to be merged first.

kilrau:
I think we discussed a while back to not support "triggering a swap" from a remote peer for now. So IMO `takerID` not needed *optionally*.

moshababo:
@sangaman take another look, before i'll add the tests.

lib/orderbook/TradingPair.ts
```diff
@@ -28,31 +28,36 @@ type OrderSidesQueues = {
...
-  public ownOrders: OrderSidesLists<orders.StampedOwnOrder>;
-  /** A map between peerPubKey and a pair of maps between active peer orders ids and orders for the buy and sell sides of this trading pair. */
-  public peersOrders: Map<string, OrderSidesLists<orders.StampedPeerOrder>>;
-
-  constructor(private logger: Logger, public pairId: string) {
-    this.queues = {
-      buy: MatchingEngine.createPriorityQueue(OrderingDirection.Desc),
-      sell: MatchingEngine.createPriorityQueue(OrderingDirection.Asc),
-    };
+  public queues: OrderSidesQueues | undefined;
```
sangaman: Rather than `| undefined` we should just say `queues?`

lib/orderbook/TradingPair.ts
```diff
@@ -28,31 +28,36 @@ type OrderSidesQueues = {
...
+  /** a pair of maps between active own orders ids and orders for the buy and sell sides of this trading pair. */
+  public ownOrders: OrderSidesMaps<orders.StampedOwnOrder>;
+  /** a map between peerPubKey and a pair of maps between active peer orders ids and orders for the buy and sell sides of this trading pair. */
```
sangaman: Could you capitalize the start of this comment and the one above?

lib/orderbook/TradingPair.ts
```diff
@@ -28,31 +28,36 @@ type OrderSidesQueues = {
...
+  public peersOrders: Map<string, OrderSidesMaps<StampedPeerOrder>>;
+
+  constructor(private logger: Logger, public pairId: string, private matching = true) {
+    if (matching) {
```
sangaman: I think I'd prefer sticking with `nomatching`, I know it raises the possibility for double negatives but I think it's better to call it the same thing everywhere, also this way we could actually get rid of the double negatives when passing in `!this.nomatching` to the constructor. It's fine if you prefer it this way though.

lib/orderbook/errors.ts
```diff
@@ -4,12 +4,14 @@ const codesPrefix = errorCodesPrefix.ORDERBOOK;
...
   DUPLICATE_ORDER: codesPrefix.concat('.2'),
-  OWN_ORDER_NOT_FOUND: codesPrefix.concat('.3'),
-  CURRENCY_DOES_NOT_EXIST: codesPrefix.concat('.4'),
-  CURRENCY_CANNOT_BE_REMOVED: codesPrefix.concat('.5'),
-  CURRENCY_ALREADY_EXISTS: codesPrefix.concat('.6'),
-  PAIR_ALREADY_EXISTS: codesPrefix.concat('.7'),
-  ORDER_NOT_FOUND: codesPrefix.concat('.8'),
+  ORDER_NOT_FOUND: codesPrefix.concat('.3'),
+  OWN_ORDER_NOT_FOUND: codesPrefix.concat('.4'),
+  PEER_ORDER_NOT_FOUND: codesPrefix.concat('.5'),
```
sangaman: In #589 @erkarl and I discussed going with just a single `ORDER_NOT_FOUND` error type, with the reasoning that the caller of the method to get an order will know whether they were trying to get a peer or own order already. Would it make sense to you to stick to that approach as well? If so we can drop the peer and own order not found errors, it actually looks like `PEER_ORDER_NOT_FOUND` is not currently being used.

lib/service/Service.ts
```diff
@@ -330,6 +330,34 @@ class Service extends EventEmitter {
...
+      quantity: quantity || maker.quantity
+    });
+
+    try {
+      const swapResult = await this.swaps.executeSwap(maker, taker);
+      this.emit('peerOrder.filled', maker);
+      return swapResult;
+    } catch (err) {
+      this.emit('peerOrder.invalidation', maker);
+      // TODO: penalize peer for failed swap? penalty severity should depend on reason for failure
```
sangaman: Emitting these events and penalizing the peer from `Service` is a problem, because nothing is listening to `Service` for these events. This logic is copied from the `placeOrder` (around line 311) method in OrderBook.ts, so I figure we should move that into is own method and call that method directly from here. Maybe `OrderBook.executeSwap`?

lib/types/orders.ts
```diff
@@ -64,6 +64,8 @@ export type OrderPortion = OrderIdentifier & {
   localId?: string;
 };
 
+export type OrderInvalidation = OrderPortion & Remote;
```
sangaman: I don't think we're using this type anywhere so we can drop it for now.

proto/xudrpc.proto
```diff
@@ -145,6 +145,14 @@ service Xud {
   }
 
 
+  /* Execute a swap between a maker peer order and a taker own order */
```
sangaman: I think with the new approach, this should just say "Execute a swap on a maker peer order" or something like that.

lib/service/Service.ts
```diff
@@ -330,6 +330,34 @@ class Service extends EventEmitter {
     return this.orderBook.removePair(pairId);
   }
 
+  public executeSwap = async (args: { orderId: string, pairId: string, peerPubKey: string, quantity: number }): Promise<SwapResult> => {
```
sangaman: Another minor thing, I believe this file also tries to follow alphabetical order for methods to align with the proto file and GrpcService, so could you move this method up a bit as well?

lib/cli/commands/executeswap.ts
```diff
@@ -0,0 +1,32 @@
+import { callback, loadXudClient } from '../command';
+import { Arguments } from 'yargs';
+import { ExecuteSwapRequest } from '../../proto/xudrpc_pb';
+
+export const command = 'executeswap <pair_id> <maker_order_id> <maker_peer_pub_key> <taker_order_id>';
+
+export const describe = 'execute a swap on between a maker peer order and a taker own order (nomatching mode only)';
```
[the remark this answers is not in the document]
moshababo: > This still needs to be addressed I believe. Let's just go with <pair_id> <order_id> <peer_pub_key> [quantity] as you've done below.

Sorry, forgot to change that. 

> As an aside, this might actually be a use case where I'd want that flat, not-grouped-by-peer, mapping of peer orders, so that the peer_pub_key wouldn't need to be specified on this call. But it's fine as is for now, just something to think about.

Yes, I was thinking of that too, but wasn't sure it's important enough for us, if that will be its only benefit. 

> With the recent change it'd make more sense to say something like "execute a swap on a peer order".

I left the `(nomatching mode only)` suffix, let me know if you want me to remove it.

lib/orderbook/OrderBook.ts
```diff
@@ -49,25 +49,25 @@ interface OrderBook {
 class OrderBook extends EventEmitter {
   /** A map of supported currency tickers to currency instances. */
   public currencies = new Map<string, CurrencyInstance>();
-  /** A map of supported trading pair tickers to pair instances. */
+  /** A map of supported pair tickers to pair instances. */
```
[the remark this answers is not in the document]
moshababo: I changed it because of confusing with the new `TradingPair` class, but I think that specifying that it's a database instance is better.

lib/orderbook/OrderBook.ts
```diff
@@ -108,51 +108,64 @@ class OrderBook extends EventEmitter {
...
+    return tp.getOwnOrder(orderId);
+  }
+
+  private tryGetOwnOrder = (orderId: string, pairId: string): StampedOwnOrder | undefined => {
     try {
-      const matchingEngine = this.getMatchingEngine(pairId);
-      order = matchingEngine.ownOrders.buy.get(orderId) || matchingEngine.ownOrders.sell.get(orderId);
+      return this.getOwnOrder(orderId, pairId);
     } catch (err) {
-      this.logger.debug(`tried to find order for unsupported pairId ${pairId}`);
```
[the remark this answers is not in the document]
moshababo: the problem can be either that it's an unsupported pairId (like previously), or that the order doesn't exists. 
I usually prefer to do the logging in the higher context, so in this case it's in `handleSwapRequest`. in this case, it should use `getOwnOrder` directly, and catch/log the error. Do you want me to do this change?

lib/orderbook/TradingPair.ts
```diff
@@ -100,47 +105,36 @@ class MatchingEngine {
...
 
-  private addOrder = (order: StampedOrder, lists: OrderSidesLists<StampedOrder>): boolean => {
-    const list = order.isBuy ? lists.buy : lists.sell;
-    const queue = order.isBuy ? this.queues.buy : this.queues.sell;
-
-    if (list.has(order.id)) {
+  private addOrder = (order: StampedOrder, maps: OrderSidesMaps<StampedOrder>): boolean => {
+    const map = order.isBuy ? maps.buy : maps.sell;
+    if (map.has(order.id)) {
       return false;
```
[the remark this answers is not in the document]
moshababo: If we'll throw an error, we need to be careful with a peerOrder case. At the moment we're having the relaxed `boolean` result, and it's applied on `OrderBook.addOwnOrder` with an assert, because it's an unexpected behavior (id supposed to be unique).

I'm fine with the current behavior, let me know if you think otherwise.

lib/orderbook/OrderBook.ts
```diff
@@ -300,41 +327,52 @@ class OrderBook extends EventEmitter {
...
+    /**
+   * Add own order
+   * @returns false if it's a duplicated order or with an invalid pair id, otherwise true
+   */
+  private addOwnOrder = (order: orders.StampedOwnOrder): boolean => {
+    const tp = this.getTradingPair(order.pairId);
+    const result = tp.addOwnOrder(order);
+    assert(result, 'own order id is duplicated');
+
+    this.emit('ownOrder.added', order);
```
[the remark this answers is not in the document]
moshababo: Indeed, thanks for the catch.

lib/orderbook/errors.ts
moshababo: Yes, I overlooked that. 

The `peerPubKey`/`localId` differentiation issue isn't very neat, but we can fix it later.

lib/service/Service.ts
moshababo: Yes, I was thinking of putting it inside `OrderBook`, to be used also from `placeOrder`, but the latter has some other logic as well and I thought it would be confusing to separate it from its flow. 

But I overlooked events issue you mentioned, so obviously it does need to be inside `OrderBook`.

I ended up putting it inside `OrderBook`, but not to be used from `placeOrder`, so code is duplicated. Let me know if it looks fine to you. Otherwise I can try to make it work without code duplication.

lib/types/orders.ts
moshababo: right, it's from a previous iteration of mine.

moshababo:
@sangaman Thanks for the extensive review. I made the necessary fixes or commented on your notes. Will add the tests soon.

lib/orderbook/OrderBook.ts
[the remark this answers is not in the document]
sangaman: Nope that's a good enough explanation for me.

lib/orderbook/OrderBook.ts
```diff
@@ -207,27 +220,42 @@ class OrderBook extends EventEmitter {
...
 
-  public addLimitOrder = async (order: orders.OwnOrder, onUpdate?: (e: PlaceOrderEvent) => void): Promise<PlaceOrderResult> => {
+  public placeLimitOrder = async (order: orders.OwnOrder, onUpdate?: (e: PlaceOrderEvent) => void): Promise<PlaceOrderResult> => {
     const stampedOrder = this.stampOwnOrder(order);
-    return this.addOwnOrder(stampedOrder, false, onUpdate, Date.now() + OrderBook.MAX_ADD_OWN_ORDER_ITERATIONS_TIME);
+    if (this.nomatching) {
+      this.addOwnOrder(stampedOrder);
+      onUpdate && onUpdate({ case: PlaceOrderEventCase.RemainingOrder, payload: stampedOrder });
+
+      return Promise.resolve({
```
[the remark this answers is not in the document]
sangaman: I personally think it'd be ok - since anytime this method is called it would still need an `await` or `then` which would end up with a synchronous `PlaceOrderResult` value - but I think it's fine too to keep the method async and drop the `Promise.resolve`. Whichever you prefer.

lib/orderbook/TradingPair.ts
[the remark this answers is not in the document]
sangaman: OK, yeah I'll maybe take another look at this behavior later but it's fine for now.

lib/service/Service.ts
sangaman: Yeah I do think it would be better without the code duplication. I'll make a separate comment for what I have in mind.

lib/orderbook/OrderBook.ts
```diff
@@ -299,42 +321,76 @@ class OrderBook extends EventEmitter {
...
 
+  public executeSwap = async (orderId: string, pairId: string, peerPubKey: string, quantity?: number): Promise<SwapResult> => {
+    const maker = this.removePeerOrder(orderId, pairId, peerPubKey, quantity).order;
+    const taker = this.stampOwnOrder({
+      pairId,
+      localId: '',
+      price: maker.price,
+      isBuy: !maker.isBuy,
+      quantity: quantity || maker.quantity,
+    });
```
sangaman: What I'm thinking here is I'd put the logic that creates `maker` and `taker` back in Service.ts. These are public methods already, so I don't see a problem with that. Then the code below (the try/catch block) can stay in this method, with the method only taking `maker` and `taker` parameters.

Then replace the try/catch starting at line 306 with a call to this method, and in the catch block there you can have the `swapFailures.push` line.

Let me know what you think.

moshababo: That's what I was thinking of doing, but I thought that it would be confusing for `placeOrder` flow that some of the order removal events would be triggered from within `executeSwap`, so I preferred code duplication over that. We can still make `placeOrder` to emit these events, but then `executeSwap` will need another wrapper. 
In addition, `placeOrder` still haven't stabilized, so I didn't want to over-optimize. 

Tell me what you think, and if that's the only thing left, let's merge it and i'll open a separate PR for that.

sangaman: @moshababo To me it's not confusing, it's just that this method would be responsible for calling `swaps.executeSwap` and for emitting the `peerOrder.filled` and `peerOrder.invalidation` events. Then the `placeOrder` method will still be responsible for `swapResults.push`, `swapFailures.push` and the `onUpdate` callback (none of which are relevant when calling `executeSwap` via the RPC).

But yeah it's not a big deal so no need to get stuck on this - if you're still not sure leave as-is and I'll make a separate PR to discuss this specific issue. And I think the only other (minor) thing is the `Promise.resolve` comment, it's the only other unresolved conversation left but it's hidden so you have to expand it in the PR.

lib/orderbook/OrderBook.ts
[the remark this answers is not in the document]
moshababo: I actually thought that it would be an error to do `await` on non-promise, but apparently JS protects against it by wrapping the result in `Promise.resolve`. Anyway, I ended up just removing the `async` keyword.

lib/orderbook/OrderBook.ts
moshababo: Yes, I just thought that having this method emitting the events will be confusing when reading `placeOrder`, which also emits events.
I’ll revise this idea tomorrow, and perhaps open a separate PR. 

Saw your other comment and replied.

Issue #437, opened 2018-09-04
Types refactoring


moshababo:
1) We have two kind of order types: `Order` & `StampedOrder` (and their respective peer/own types). The first is used to define the fields which are needed to create an order, and the later the actual order which is being stored. The later is too verbose, and is used throughout the code base. We should change it to plain `Order`, and change the current `Order` (which is hardly being used) to something like `UnstampedOrder` / `RawOrder` (@sangaman suggested), or maybe `OrderFactory` (i'm suggesting).

2) Move all the files from the types folder into their respective folders, so just like how each folder has an `errors.ts` it could have a `types.ts` (@sangaman suggested)

PR #597, opened 2018-10-23
refactor: create types file for swaps

sangaman:
This commit puts types related to swaps that are used throughout the project in their own file. It also properly derives the `SwapResult` type from `SwapDeal`, of which it is a subset.

I'm using the `types.ts` pattern we discussed several weeks back, eventually I think all the current types files should be moved to `types.ts` files within their respective folders.

### DIFF
--- a/lib/swaps/types.ts
+++ b/lib/swaps/types.ts
@@ -1,0 +1,61 @@
+import { SwapRole, SwapPhase, SwapState } from 'lib/types/enums';
+import { Route } from '../proto/lndrpc_pb';
+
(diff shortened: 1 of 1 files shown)


moshababo:
Can you elaborate on the `types.ts` pattern? just having this file for all types per folder?

sangaman:
Strange, I didn't try to request an additional review from you just now but GitHub says I did.

Anyway, there's an issue for the types refactoring (#437) but the basic idea is that each folder can have its own `types.ts` the same way it has its own `errors.ts`.

Issue #558, opened 2018-10-04
xucli output enhancements [WIP]


sangaman:
This issue is to track enhancements to the output from `xucli` to make it prettier and more user-friendly than the JSON outputs it's currently spitting out

A major point is the `getorders` output, which is currently a mess once you have more than a few orders. We should be displaying only key information and preferably in a table.

Generally speaking we'd want to convert epoch times into local date/times, enum numbers into their corresponding key strings, hide data that is not relevant to the user, and generally make output more concise and readable. Colors would be a bonus.

I figure this can be done across multiple PRs, so if you want to work on a particular command or feature just comment here and reference this issue in the PR.

https://www.npmjs.com/package/cli-table - This package looks promising for when we need to display tabular data, such as with `getorders`.

sangaman:
A suggestion from @offerm 

> My feedback relates to the CLI for buy and sell.
> As a user, I'm expecting to understand what is going on. If I place an order the outcome should be something like
> 'accepted and waiting' (with data)
> 'partially executed' (with data)
> 'fully executed' (with data)
> etc

erkarl:
Going to work on this one.

kilrau:
Check https://sparkswap.com/img/top-fold-terminal.gif @erkarl

kilrau:
https://github.com/ExchangeUnion/xud/issues/620 takes care of the [described](https://github.com/ExchangeUnion/xud/issues/558#issuecomment-428205056) place buy/sell order output.

kilrau:
How about a 'command preview' based on the alphabetically first (on first use of CLI session) or last used command?
![image](https://user-images.githubusercontent.com/19181985/48129236-2a20f500-e289-11e8-8467-0277971355f1.png)

sangaman:
Off the top of my head I don't know how that would be done, but if it's doable then it would be nice for sure.

erkarl:
I'd push this under future improvements/someday/nice to have.

kilrau:
This issue is about going through all CLI outputs and format them properly to be more friendly and readable. Mostly creating tables for `list` and `getinfo`.

rsercano:
I suppose all related issues already have PR. Let me know if there's additional work required

kilrau:
All have PRs or separate issues. Closing here.

PR #600, opened 2018-10-24
feat(cli): show table for getorders

ghost:
Related: https://github.com/ExchangeUnion/xud/issues/558

This PR modifies the getorders cli to return results in a table:
screen shot 2018-10-24 at 16 23 27

With the current implementation all trading pairs are displayed in the same table and can be distinguished by "Pair" column. Should we keep it this way or display a separate table for each pair?

For drawing the table I chose [cli-table3](https://github.com/cli-table/cli-table3) dependency as it's based on cli-table and cli-table2 which are no longer maintained. Additionally cli-table3 has TypeScript support.

I'm thinking we should add additional cli `--raw` option for users who want to receive raw output (perhaps for debugging purposes?).

### DIFF
--- a/bin/xucli
+++ b/bin/xucli
@@ -27,6 +27,12 @@ require('yargs')
       describe: 'Path to the TLS certificate of xud',
       type: 'string',
     },
+    json: {
+      alias: 'j',
+      describe: 'Display output in json format',
+      type: 'boolean',
+      default: false,
+    },
   })
   .commandDir('../dist/cli/commands/', { recurse: true })
   .demandCommand(1, '')
--- a/lib/cli/command.ts
+++ b/lib/cli/command.ts
@@ -40,15 +40,19 @@ interface grpcResponse {
  * @param argv the command line arguments
  * @param callback the callback function to perform a command
  */
-export const callback = (error: Error | null, response: grpcResponse) => {
-  if (error) {
-    console.error(`${error.name}: ${error.message}`);
-  } else {
-    const responseObj = response.toObject();
-    if (Object.keys(responseObj).length === 0) {
-      console.log('success');
+export const callback = (formatOutput?: Function) => {
+  return (error: Error | null, response: grpcResponse) => {
+    if (error) {
+      console.error(`${error.name}: ${error.message}`);
     } else {
-      console.log(JSON.stringify(responseObj, undefined, 2));
+      const responseObj = response.toObject();
+      if (Object.keys(responseObj).length === 0) {
+        console.log('success');
+      } else {
+        formatOutput
+          ? formatOutput(responseObj)
+          : console.log(JSON.stringify(responseObj, undefined, 2));
+      }
     }
-  }
+  };
 };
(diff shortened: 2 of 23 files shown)


sangaman:
Awesome! I pulled down the branch and played with it and it's looking nice. A couple things from a first pass:

- The `Time` column gets cut off for me, ending with `7:26:1…` instead of the full time. I'm thinking we maybe don't need this column tho, as I don't see it as important information just for seeing what's out there. I'm not used to seeing it on most orderbooks on actual exchanges (or if it's there, I don't notice). For me, it'd be nice to have but if it presents issues or takes up too much space we can drop it.

- Rather than have a column for buy/sell and a column for pair, I think it would be nice to have separate tables for each trading pair that is returned and to have all buys on the left side of the tables, and all sells on the right side of the table. Each side would have its own price, quantity, and own order column. 

- I think it would be nice to highlight our own orders a certain color, making it easy to see which orders are "out there" and which are ours. It could even serve as a replacement for the own order column if we don't have space, but then we'd need to make the meaning of the highlight clear to the user.

I really like the idea of a `--raw` option to just output the JSON as we do currently, that can be a global arg for all commands defaulting to `false`.

erkarl:
Without commiting code I'm posting another version for review. Based on the feedback I:
* removed the date column as it was taking up too much space
* removed the ID column since it wasn't really useful in terms of getting an overview of the orderbook
* moved buys to left side of the table & sells to the right
* removed own order column and marked own orders as cyan (I'm not sure if this is the best approach as it might not be apparent for everyone)

screen shot 2018-11-02 at 12 39 14

sangaman:
It feels like there'd be room for the Own Orders column maybe? Then the highlighting will also make sense. I think it looks great otherwise.

erkarl:
I think we can fit in "Own Order" column - the reason I didn't add it was because it felt weird having it "No" for 99% of the cases.

I'll add it for now. We can always remove it in the future when the network's orderbook grows.

erkarl:
Updated with code + added "Own Order" column.

bin/xucli
```diff
@@ -27,6 +27,12 @@ require('yargs')
       describe: 'Path to the TLS certificate of xud',
       type: 'string',
     },
+    raw: {
```
sangaman: For this option, how about calling it `json` with a `-j` flag and description `Display output in json format`. My thinking is that the "raw" responses are serialized data and that the json output is actually formatted. This makes it clear that the output will be in JSON.

sangaman (commented):
Looks great, just a couple of things:

I see that you've colored the "Buy" side red and the "Sell" side green. I'm used to seeing these colors in reverse. Buys are green, sells are red.

Also, about having to print "No" over and over, you could maybe just put an "X" in the column for own orders and leave it empty for peer orders. Or you could only write "Yes" for own orders if "X" is confusing. Just an idea, I'd be fine with how it is currently.

bin/xucli
erkarl: Calling it `json` instead of `raw` feels way better.

erkarl:
@sangaman updated. Let me know if you have reviewed and agree with the changes so I can squash them.

sangaman (approved):
Awesome.

One idea I had, but not for this PR, would be to have a node's alias or node pubkey in the table. I'll open a separate issue to discuss it, but I think this will be a big improvement.

PR #603, opened 2018-10-26
refactor(p2p/orderbook): group peer orders by peerPubKey

erkarl:
This PR is a clone of https://github.com/ExchangeUnion/xud/pull/589 because Travis builds were rejected for the latter.

* rename `this.peerOrders` to `this.peersOrders`.
* change `this.peersOrders` so that instead of having two maps of `orderId` to `StampedPeerOrder` for buys and sells, we now have multiple pairs of buy/sell maps like that nested within a map of `peerPubKeys`.
* remove basic type `orderId: string`
* log generic error message for order invalidation (remove error
handling from `Orderbook.removePeerOrder`

### DIFF
--- a/lib/orderbook/MatchingEngine.ts
+++ b/lib/orderbook/MatchingEngine.ts
@@ -11,9 +11,7 @@ type SplitOrder = {
   remaining: StampedOrder;
 };
 
-type orderId = string;
-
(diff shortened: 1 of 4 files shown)


erkarl:
Didn't work. Closing.

PR #611, opened 2018-10-27
fix(swaps): correctly set taker & maker amounts

### LINKED ISSUE #604: SWAP calculations
@moshababo discovered:

I was looking at the swaps code for something, and there’s something in the currency/amount calculations that I don’t understand.
Considering i’m on pair `LTC/BTC` and the price is `0.01`. we have 2 scenarios:

1)

maker order: buy quantity 5
taker order: sell quantity 5

swap amounts:
maker: 5 LTC
taker: 0.05 BTC

* if the maker is `buy`, then he wants to buy LTC. so why he’s the one sending LTC?

2)

maker order: sell quantity 5
taker order: buy quantity 5

swap amounts
maker: 5 BTC
taker: 0.05 LTC

* same like my previous question for the currency, and the amount also doesn’t seem correct.

--

My calculations are based on:
https://github.com/ExchangeUnion/xud/blob/master/lib/swaps/Swaps.ts#L91
https://github.com/ExchangeUnion/xud/blob/master/lib/swaps/Swaps.ts#L241

sangaman:
This fixes a bug in the `SwapDeal` and `SwapResult` logic for setting values for `takerAmount`, `makerAmount`, `amountReceived`, and `amountSent`. Previously the values could be mistakenly switched.

Fixes #604.

### DIFF
--- a/test/unit/Swaps.spec.ts
+++ b/test/unit/Swaps.spec.ts
@@ -1,0 +1,53 @@
+import { expect } from 'chai';
+import Swaps, { SwapDeal } from '../../lib/swaps/Swaps';
+import Logger, { Level } from '../../lib/Logger';
+import { SwapPhase, SwapState, SwapRole } from '../../lib/types/enums';
+
+const loggers = Logger.createLoggers(Level.Warn);
+
+describe('Swaps', () => {
+  const quantity = 0.01;
+  const price = 0.005;
+
+  /** A swap deal where we are the taker. */
+  const takerDeal: SwapDeal = {
+    quantity,
+    price,
+    role: SwapRole.Taker,
+    phase: SwapPhase.SwapCreated,
+    state: SwapState.Active,
+    peerPubKey: '03029c6a4d80c91da9e40529ec41c93b17cc9d7956b59c7d8334b0318d4a86aef8',
+    orderId: 'f8a85c66-7e73-43cd-9ac4-176ff4cc28a8',
+    localOrderId: '1',
+    proposedQuantity: quantity,
+    pairId: 'LTC/BTC',
+    takerCurrency: 'LTC',
+    makerCurrency: 'BTC',
+    takerAmount: Swaps['SATOSHIS_PER_COIN'] * quantity,
+    makerAmount: Swaps['SATOSHIS_PER_COIN'] * quantity * price,
+    takerCltvDelta: 144,
+    r_hash: '62c8bbef4587cff4286246e63044dc3e454b5693fb5ebd0171b7e58644bfafe2',
+    createTime: 1540716251106,
+  };
+
+  /** A swap deal where we are the maker, mirrored from the taker deal for convenience. */
+  const makerDeal = {
+    ...takerDeal,
+    role: SwapRole.Maker,
+    takerCurrency: takerDeal.makerCurrency,
+    makerCurrency: takerDeal.takerCurrency,
+    takerAmount: takerDeal.makerAmount,
+    makerAmount: takerDeal.takerAmount,
+  };
+
+  it(`should calculate swap amounts`, () => {
+    expect(takerDeal.role).to.equal(SwapRole.Taker);
+    expect(makerDeal.role).to.equal(SwapRole.Maker);
+
+    const { baseCurrencyAmount, quoteCurrencyAmount } = Swaps['calculateSwapAmounts'](takerDeal.quantity!, takerDeal.price);
+    expect(baseCurrencyAmount).to.equal(takerDeal.takerAmount);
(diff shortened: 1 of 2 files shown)


moshababo:
I'll go through this soon, but I think it's important to add unit tests for these sections.

sangaman:
That's a good idea, I will do that.

sangaman:
@moshababo Just added, take a look when you get a chance.

sangaman:
Perhaps it's confusing, but the way it's set up and commented the "taker" amount and currency is what the taker is expecting to receive.

And I just realized the tests are failing, I'd actually fixed that issue locally but apparently forgot to push it. I will do that when I'm home.

moshababo:
Yes, I see now that the maker currency/amount are the ones which are being used in `handleSwapResponse`, hence in the taker side.

test/unit/Swaps.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    createTime: 1540716251106,
+  };
+
+  const makerDeal = {
+    ...takerDeal,
+    role: SwapRole.Maker,
+    takerCurrency: takerDeal.makerCurrency,
+    makerCurrency: takerDeal.takerCurrency,
+    takerAmount: takerDeal.makerAmount,
+    makerAmount: takerDeal.takerAmount,
```
moshababo: why the currency/amount is being assigned to the opposite side (taker on maker, maker on taker)? I don't see how that's aligned with `handleSwapRequest`, which assigns these values as-is (in `acceptDeal`).

test/unit/Swaps.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    orderId: 'f8a85c66-7e73-43cd-9ac4-176ff4cc28a8',
+    localOrderId: '1',
+    proposedQuantity: 0.01,
+    quantity: 0.01,
+    pairId: 'LTC/BTC',
+    price: 0.005,
+    takerCurrency: 'LTC',
+    makerCurrency: 'BTC',
+    takerAmount: 1000000,
+    makerAmount: 5000,
```
moshababo: I would like to see these values being directly tested, although their logic code is currently not testable. Might need to extract it to another static function.

moshababo:
Another thing in regards to the currency/amount: in `handleSwapRequest`, the maker verifies his order availability, and accepts the deal. But I don't see any validation on the deal info. The only validation I see is on LND requests (`validateRequest`), but that's being done on the deal info. It means that the maker is blindly accepting the taker calculations in regards to the deal.

test/unit/Swaps.spec.ts
sangaman: What I was going for here is not exactly the same as the flow that happens within `xud`, where two sides negotiate on the details of a single swap - and where the same node is the "taker" on both sides of the deal. Here I was trying to create two *separate* swaps that are mirrors of each other when it comes to taker/maker and which copy other details for convenience. Let me know if that makes sense, and if you think it's worth adding a comment to the tests to try to clarify that.

test/unit/Swaps.spec.ts
sangaman: Yeah I don't know how I can sensibly break it down further. At the most basic level this is just multiplication. I can break out the `10000000` literal into a `static readonly` property on swaps, and can make a `quantity` variable and make these lines say `quantity * Swaps.SATOSHIS_PER_COIN` and `quantity * Swaps.SATOSHIS_PER_COIN * price`. Let me know if that sounds good to you or if you have a specific suggestion in mind.

sangaman:
@moshababo Let me know what you think of my responses. On the deal side, yes there's still logic missing to validate that the amounts are what we expect, I'm hoping to improve that soon.

test/unit/Swaps.spec.ts
moshababo: Yes that wasn't clear to me. I assumed that it's just the other side, because they usually go together. I think that the fact that we don't have calculations on the maker side is confusing for this matter. We probably need to fix that before continuing with the tests.

test/unit/Swaps.spec.ts
moshababo: I was referring more to the assignment logic (Swaps.ts line 244), but we can also directly assert `calculateSwapAmounts` response with a few test cases.

moshababo:
The latter is a critical security issue, and will also help us with the tests. I'll open a separate issue, and we can continue with the tests after it will be done.

test/unit/Swaps.spec.ts
sangaman: Yup there's a TODO for the calculations on the maker side, it's on my radar.

Issue #613, opened 2018-10-29
Trading issues, part #2


offerm:
Here is what I did:

The market order removed all peer sell order and become a buy order (which is very bad).
See the log below

```
admins-MacBook-Pro-2:xud-simnet admin$ xud-simnet-start
OSX
admins-MacBook-Pro-2:xud-simnet admin$ xucli getorders
{
  "ordersMap": [
    [
      "LTC/BTC",
      {
        "buyOrdersList": [
          {
            "price": 0.86,
            "quantity": 0.003,
            "pairId": "LTC/BTC",
            "id": "0c2802e1-dad8-11e8-ba91-19129e5f0bc9",
            "peerPubKey": "02b66438730d1fcdf4a4ae5d3d73e847a272f160fee2938e132b52cab0a0d9cfc6",
            "localId": "",
            "createdAt": 1540823817942,
            "side": 0,
            "isOwnOrder": false
          },
          {
            "price": 0.87,
            "quantity": 0.0025,
            "pairId": "LTC/BTC",
            "id": "0c285101-dad8-11e8-ba91-19129e5f0bc9",
            "peerPubKey": "02b66438730d1fcdf4a4ae5d3d73e847a272f160fee2938e132b52cab0a0d9cfc6",
            "localId": "",
            "createdAt": 1540823817942,
            "side": 0,
            "isOwnOrder": false
          },
          {
            "price": 0.88,
            "quantity": 0.002,
            "pairId": "LTC/BTC",
            "id": "0c289f22-dad8-11e8-ba91-19129e5f0bc9",
            "peerPubKey": "02b66438730d1fcdf4a4ae5d3d73e847a272f160fee2938e132b52cab0a0d9cfc6",
            "localId": "",
            "createdAt": 1540823817943,
            "side": 0,
            "isOwnOrder": false
          },
          {
            "price": 0.89,
            "quantity": 0.0015,
            "pairId": "LTC/BTC",
            "id": "0c28ed42-dad8-11e8-ba91-19129e5f0bc9",
            "peerPubKey": "02b66438730d1fcdf4a4ae5d3d73e847a272f160fee2938e132b52cab0a0d9cfc6",
            "localId": "",
            "createdAt": 1540823817943,
            "side": 0,
            "isOwnOrder": false
          },
          {
            "price": 0.9,
            "quantity": 0.001,
            "pairId": "LTC/BTC",
            "id": "0c293b61-dad8-11e8-ba91-19129e5f0bc9",
[...]

Issue #612, opened 2018-10-28
Trading issue


offerm:
Guys,

Assume that I'm an exchange that just installed XUD. 

See what I did.
Can you please help me to explain what happened? 

At startup I'm connected to LNDs and to 3 peers and have 11 peer orders. all good.
```
admins-MacBook-Pro-2:~ admin$ xucli getinfo
{
  "version": "1.0.0-alpha.1",
  "nodePubKey": "034606c6a0616397cb5dee9f84cad5186d38bc65712b672ff4fcf1eb46e6e768ea",
  "urisList": [],
  "numPeers": 3,
  "numPairs": 1,
  "orders": {
    "peer": 11,
    "own": 0
  },
  "lndbtc": {
    "error": "",
    "channels": {
      "active": 1,
      "inactive": 0,
      "pending": 0
    },
    "chainsList": [
      "bitcoin"
    ],
    "blockheight": 20386,
    "urisList": [],
    "version": "0.5.0-beta commit=v0.4.2-beta-1247-ge22ae9985b696924d5934004c269c930d3ccba43",
    "alias": "BTC@admins-MacBook-Pro-2.local"
  },
  "lndltc": {
    "error": "",
    "channels": {
      "active": 1,
      "inactive": 0,
      "pending": 0
    },
    "chainsList": [
      "litecoin"
    ],
    "blockheight": 25540,
    "urisList": [],
    "version": "0.5.0-beta commit=v0.4.2-beta-1247-ge22ae9985b696924d5934004c269c930d3ccba43",
    "alias": "LTC@admins-MacBook-Pro-2.local"
  }
}
```
Looking at the orders I see BUY orders and SELL orders.
**The sell orders are not sorted by price**
```
admins-MacBook-Pro-2:~ admin$ xucli getorders
{
  "ordersMap": [
    [
      "LTC/BTC",
      {
        "buyOrdersList": [
          {
            "price": 0.9,
            "quantity": 0.001,
            "pairId": "LTC/BTC",
            "id": "926e2741-dad2-11e8-80b4-af8455b00957",
            "peerPubKey": "02b66438730d1fcdf4a4ae5d3d73e847a272f160fee2938e132b52cab0a0d9cfc6",
            "localId": "",
            "createdAt": 1540746065069,
            "side": 0,
            "isOwnOrder": false
          },
          {
            "price": 0.89,
            "quantity": 0.0015,
            "pairId": "LTC/BTC",
[...]

PR #614, opened 2018-10-29
fix: addOwnOrder iteration call

moshababo:
I saw this issue after checking #612. I'm not sure it's the entire problem, we'll need to have testing for this section soon.


moshababo:
Also fixes #613

PR #618, opened 2018-10-30
proto deprecation fix

moshababo:
Our `protoFixDeprecation` npm script fix was missing.


michael1011:
Every time `npm run proto` is executed `protoFixDeprecation` should be too, shouldn't it? Does someone know how it is possible that there are still files with deprecation warnings in the repository?

moshababo:
I ran `cross-os proto` directly for something, and forgot to run the entire thing. Did we had a plan to run `npm run proto` as part of the build?

michael1011:
No I don't think it should be part of the build.

For what reason did you run `cross-os proto` directly?

moshababo:
I was on a computer without the plugins installed :\

PR #624, opened 2018-10-31
refactor(orderbook): change ORDER_NOT_FOUND msg

sangaman:
This is a minor change to the `ORDER_NOT_FOUND` error message to make it more readable. Instead of `peerPubKey` and `orderId` being separated by only a `/` for peer orders, they are clearly differentiated.

I'd found myself confused when reading the old error message, thinking that it was using `peerPubKey` in place of `orderId`.


erkarl:
LGTM - I also had the same initial reaction when I was debugging this error message.

moshababo:
Another issue is left, where we use `localId` instead of `orderId`. It's in `OrderBook.removeOwnOrderByLocalId`.

sangaman:
@moshababo I created a separate PR for that, #631. I wasn't sure on the best way to treat a case of a local id that we can't find, so comment there if you have any thoughts.

PR #625, opened 2018-10-31
feat(listSwapDeals) - display all completed SwapDeals from database

### LINKED ISSUE #477: new RPC - listSwapDeals
Suggest creating an RPC that extract the list of SwapDeals done. 

We have interesting data there list status, state, errormessae, createtime, executetime, completetime that the educate us and help to debug.

For now, we can show information from memory but in the long run, we should store/extract from the database.

ImmanuelSegol:
closes #477 
This should be merged only after #623.
@sangaman 
Some questions:

- If we end up returning `makerToTakerRoutes` how should we define `Route`, and should I use enums for the `SwapRole`, `SwapPhase`, SwapState` in the `proto` files?

- I was thinking maybe `listSwapDeals` should return two lists one for successfully completed swaps and the second for failed swaps.


sangaman:
With the changes from #623 I think we also need to change the `getSwapDeals` repository method to do a join with the `orders` table. As is, it doesn't return all the fields we need including `price`. I can handle that in a separate PR, and this could then be rebased on top of that.

To answer your questions, I don't think we need to be returning any routes. And maybe instead of returning two lists - one for successful and one for failed - maybe we pass in a parameter to specify whether we want all swaps or only successful ones? Although for now I'd actually probably just lean to keeping it simple and returning a single list, and if we need to enhance the call with features like these we can deal with it later.

kilrau:
* `makerToTakerRoutes`: agree with @sangaman , we don't need any routes for now
* `listSwapDeals`: I'd return one list (successful) as default and add a `all` parameter to include failed ones in this list too. This needs a comprehensive description in `--help`. When using `all` add failed/succeeded info each.

ImmanuelSegol:
@sangaman @kilrau Thanks for the feedback I will make the changes after #665 is merged so there won't be any annoying conflicts.

sangaman:
@moshababo @kilrau @offerm I'm wondering if calling this rpc method something like `ListSwaps` rather than `ListSwapDeals` since the "SwapDeal" term is something we use internally only and may be confusing. I'm also thinking maybe the request should take a flag to either return failed or successful swaps.

proto/xudrpc.proto
```diff
@@ -366,6 +373,57 @@ message ListPeersResponse {
   repeated Peer peers = 1 [json_name = "peers"];
 }
 
+message ListSwapDealsRequest {
+  bool all = 1 [json_name = "all"];
```
sangaman: Rather than an `all` request parameter, I think it'd be better to have a numeric `limit` as you're doing in the ListTrades PR.  If it's 0, we return all swaps, otherwise we return the most recent ones up to the limit. We can also have a `failed` flag to return only failed swaps, otherwise we return only successful swaps. Let me know what you think.

proto/xudrpc.proto
```diff
@@ -366,6 +373,57 @@ message ListPeersResponse {
...
+}
+message ListSwapDealsResponse {
+  repeated Swap swaps = 1 [json_name = "swaps"];
+}
+
+message Swap {
+  SwapRole role = 1 [json_name = "role"];
```
sangaman: Adding every field from the database seems like overkill from an rpc perspective. I think it'd make more sense to have a `oneof` for either a `SwapSuccess` or `SwapFailure` (there are examples of `oneof` elsewhere in the proto file) plus `create_time`.  I'm thinking we may want to add more fields to our `SwapFailure` message (like the failure reason), but it's ok for now. Let me know if you have any questions.

kilrau:
> @moshababo @kilrau @offerm I'm wondering if calling this rpc method something like `ListSwaps` rather than `ListSwapDeals` since the "SwapDeal" term is something we use internally only and may be confusing.

Agree on `ListSwaps`.

> I'm also thinking maybe the request should take a flag to either return failed or successful swaps.

Yep, already [suggested this above](https://github.com/ExchangeUnion/xud/pull/625#issuecomment-437856598). We can add a "failed flag" too. So it would be:
* default, without flag: successful swaps only (I'm convinced most use cases need this)
* "failed" flag: failed swaps only
* "all" flag: successful + failed swaps (indicating "failed/successful" each)

proto/xudrpc.proto
ImmanuelSegol: @sangaman I agree but why would we want to return only successful or failed swap?
Shouldn't `failed` just indicate if we should include failed swaps in the response?

proto/xudrpc.proto
ImmanuelSegol: @sangaman Should the `SwapSuccess` be the same as [this type](https://github.com/ExchangeUnion/xud/blob/d61b07f7c3b2678f08e0dde54e66ef0095853e5a/lib/swaps/types.ts#L59)

sangaman: Yes, although the `SwapDealInstance` type from the database should have all the fields you need to create a `SwapSuccess` gRPC message. You should be able to reuse the `createSwapSuccess` and `createSwapFailure` methods in GrpcService.ts.

sangaman:
In reviewing this, I feel like conceptually having both a `failed` and `all` flag is confusing. `all` is also confusing given the presence of a `limit` parameter - what if I want both failed and successful swaps but only a certain amount of them?

I actually think it would be more clear to simply drop the `all` parameter and go with either failed swaps or successful swaps but not both. Another option is to use an enum paramer that has `SUCCESSFUL`, `FAILED`, and `BOTH` values. I'm even wondering if it might make sense to have `ListSwaps` and `ListFailedSwaps` as they are represented by separate messages, but I don't want to drag this PR out too long so maybe we can keep to a single call for this PR.

Thoughts @kilrau  @ImmanuelSegol ?

kilrau:
> `SUCCESSFUL`, `FAILED`, and `BOTH` value

I'd prefer to keep it in a single call. And agree, now that we all seem to prefer a [`limit` parameter](https://github.com/ExchangeUnion/xud/pull/820#issuecomment-469314764), we should also have it hear and not `all`. So `listswaps` with optional enum `SUCCESSFUL`, `FAILED`, and `BOTH` and optional int `limit` sounds like the way to go.

Default being SUCCESSFUL and all.

sangaman:
That works for me, can you make that change @ImmanuelSegol ? We use a few other enums in the proto definition if you want an example, shouldn't be too hard otherwise.

lib/service/Service.ts
```diff
@@ -316,6 +325,32 @@ class Service extends EventEmitter {
...
+   * Get all completed swap deals.
+   * @returns A list of completed swap deals.
+   */
+  public listSwaps = async (args: { limit: number, status: ListSwapsRequest.Status }): Promise<SwapDealInstance[]> => {
+    const { limit, status } = args;
+
+    let deals = limit > 0 ?
+      await this.swaps.getCompletedDeals(limit) : await this.swaps.getCompletedDeals();
+
+    switch (status) {
```
sangaman: It would be better to pass in the criteria for whether to return completed, failed, or both kinds of deals to the db query above rather than fetching everything and filtering it in node, which will occupy the node thread longer than needed.

lib/cli/commands/listswaps.ts
```diff
@@ -0,0 +1,27 @@
+import { callback, loadXudClient } from '../command';
+import { Arguments } from 'yargs';
+import { ListSwapsRequest } from '../../proto/xudrpc_pb';
+
+export const command = 'listswaps [limit] [failed]';
```
sangaman: `failed` does not match up with `status`. I'm thinking `status` should be renamed anyhow (see my comment below), but however they end up this line ought to have accurate naming.

proto/xudrpc.proto
```diff
@@ -377,6 +384,27 @@ message ListPeersResponse {
   repeated Peer peers = 1 [json_name = "peers"];
 }
 
+message ListSwapsRequest {
+  int32 limit = 1 [json_name = "all"];
+  Status status = 3 [json_name = "status"]; 
+  enum Status {
```
sangaman: `Status` is very generic and vague here. Maybe call this `SwapState` which aligns with the enum used within xud, or perhaps even `RequestedSwapState` to be more descriptive.

sangaman:
In addition to my comments above you should correct the failing tests, here's whats being reported by travis though there may be more. @ImmanuelSegol 

```
TSError: ⨯ Unable to compile TypeScript
lib/grpc/GrpcService.ts (523,49): Conversion of type 'SwapDealInstance' to type 'SwapFailure' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

ImmanuelSegol:
> In addition to my comments above you should correct the failing tests, here's whats being reported by travis though there may be more. @ImmanuelSegol
> 
> ```
> TSError: ⨯ Unable to compile TypeScript
> lib/grpc/GrpcService.ts (523,49): Conversion of type 'SwapDealInstance' to type 'SwapFailure' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
> ```
Yeah, it seems like `deal` doesn't contain all the necessary parameters. Should I fetch them from the DB? @sangaman

kilrau:
Do you still need feedback from @sangaman here?

ImmanuelSegol:
> Do you still need feedback from @sangaman here?

I think this is good as well @sangaman ?

kilrau:
I just gave this a quick test: still JSON format

Since we are beautifying most other calls, we can also put this in a nicer looking table @ImmanuelSegol

rsercano:
Just merged and tried `listswaps` command seems to be working.

lib/grpc/GrpcService.ts
```diff
@@ -508,6 +508,41 @@ class GrpcService {
...
+      const swaps: xudrpc.Swap[] = [];
+      const response = new xudrpc.ListSwapsResponse();
+      listSwapsResponse.forEach((deal) => {
+        const grpcSwap = new xudrpc.Swap();
+        grpcSwap.setCreateTime(deal.createTime);
+        if (deal.failureReason) {
+          const swapFailure = createSwapFailure(deal as SwapFailure);
+          grpcSwap.setSwapFailure(swapFailure);
+        } else {
+          const swapSuccess = new xudrpc.SwapSuccess();
```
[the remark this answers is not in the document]
sangaman: See my original comment, we have an existing `createSwapSuccess` method that we should be able to reuse for this.

lib/grpc/GrpcService.ts
```diff
@@ -134,6 +134,11 @@ const createPlaceOrderEvent = (e: PlaceOrderEvent) => {
   return placeOrderEvent;
 };
 
+const getSwapFailureReason = (reason: SwapFailureReason) => {
```
sangaman: This seems a bit unnecessary to me, elsewhere we do similar ways to convert enums values to their string names without a dedicated function. Any objection to just using `SwapFailureReason[deal.failureReason]` in place of this method? Seems like less code.

lib/cli/commands/listswaps.ts
```diff
@@ -0,0 +1,93 @@
...
+export const command = 'listswaps [limit] [status]';
+
+export const builder = {
+  limit: {
+    description: 'the maximum number of deals to list',
+    type: 'number',
+    default: 0,
+  },
+  status: {
+    description: 'list failed deals',
```
sangaman: This is outdated. Maybe just say `0=all, 1=successful, 2=failed` so people might understand how to use this parameter.

proto/xudrpc.proto
```diff
@@ -442,12 +449,33 @@ message ListPeersResponse {
...
 
 message ListTradesResponse {
   repeated Trade trades = 1 [json_name = "trades"];
 }
 
+message ListSwapsRequest {
+  int32 limit = 1 [json_name = "all"];
+  RequestedSwapState requested_swap_state = 2 [json_name = "requested_swap_state"];
+  enum RequestedSwapState {
+    BOTH = 0;
```
sangaman: I'm thinking that we will have "pending" deals as well. No need to deal with those right now, but calling this option `ALL` instead of `BOTH` might make more sense and cover us if we have more than 2 statuses to choose from later on.

lib/cli/commands/listswaps.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const FAILED_HEADERS = [
+  colors.blue('peer public key'),
+  colors.blue('order id'),
+  colors.blue('trading pair'),
+  colors.blue('quantity'),
+  colors.blue('failure reason'),
+];
+// TODO: create a function that generates these tables
+const SUCCESS_HEADERS = [
```
sangaman: Do we want these to be capitalized like I think we're doing for other commands? Same for the failed table above.

lib/cli/commands/listswaps.ts
```diff
@@ -0,0 +1,93 @@
...
+  colors.blue('local id'),
+  colors.blue('trading pair'),
+  colors.blue('quantity'),
+  colors.blue('price'),
+  colors.blue('amount sent'),
+  colors.blue('amount received'),
+  colors.blue('currency received'),
+  colors.blue('currency send'),
+  colors.blue('role'),
+  colors.blue('rhash'),
```
sangaman: I think `Payment Hash` or maybe `Swap Hash` is more descriptive for users. Thoughts @kilrau ?

lib/cli/commands/listswaps.ts
```diff
@@ -0,0 +1,93 @@
...
+      failedTable.push([
+        fail.peerPubKey,
+        fail.orderId,
+        fail.pairId,
+        fail.quantity,
+        fail.failureReason,
+      ]);
+    } else {
+      const success = swap.swapSuccess!;
+      successTable.push([
```
sangaman: I suspect these are going to be too many columns for a horizontal table in a terminal. I'm thinking we should probably drop:

- preimage
- pairId (assuming we keep currency received and sent)
- orderId (assuming we keep localId)

I'd also consider dropping quantity & price (assuming we keep amounts sent and received).

Thoughts @kilrau ?

lib/cli/commands/listswaps.ts
kilrau: I'd suggest:
`0=all, 1=completed, 2=pending, 3=failed`

lib/cli/commands/listswaps.ts
kilrau: I'd go for `Swap Hash` (since it's a hash for two payments, not one). Let's be consistent about using `Swap Hash` from now on.

lib/cli/commands/listswaps.ts
kilrau: Sounds good.

lib/grpc/GrpcService.ts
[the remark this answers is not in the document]
rsercano: Daniel I feel like I'm missing something, how can I convert `SwapDealInstance` to  `SwapSuccess`, aren't they completely different ?

rsercano:
image

Maybe showing fail or success tables whenever they're requested would be better @kilrau ?
e.g. 
status `all` shows both tables,
status `fail` shows failure table,
status `success` shows success tables.

kilrau:
Almost a full year... Let's get this in in `1.0.0-beta`.

rsercano:
Indeed, so what do you think about what I said ? @kilrau

kilrau:
Sry, managed to overlook that.

Feedback:
* the limit `listswaps <limit>` should be *optional* and only *one* limit. Default is 10.
* as I mentioned [here](https://github.com/ExchangeUnion/xud/pull/625#issuecomment-460562468), default is successful swaps. Flags are `failed` and `all`. `all` divides into two tables as already the case.
* failed swaps should have the same data fields as successful ones
* all tables are always sorted by timestamp
* role shouldn't be `1`, it should be `taker` or `maker`
* `32fd3a...445fde` abbreviate the node key and swap hash if too long. JSON output contains full key.

rsercano:
There were some issues which should be solved and I realized just now, fixed altogether. Now failed & successful swap tables are the same except failure table has failure reason column.
image

sangaman:
Closing this in favor of #1567. `TradeHistory` will be the one-stop call for pulling back information about completed trades/swaps.

PR #626, opened 2018-11-01
feat(orderbook/swaps): release order hold

### LINKED ISSUE #549: Release order holds on failed swaps
This issue is to track functionality that releases a hold on an order when an attempted swap for that order fails or times out. We should probably use the `swap.failed` event for this.

sangaman:
This releases the hold placed on an order when we accept a swap once the swap either completes or fails. Previously, holds would remain on orders post-swap which would make part or all of the order unavailable for future swaps.

Closes #549.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -11,7 +11,7 @@ import Logger from '../Logger';
 import { ms, derivePairId } from '../utils/utils';
 import { Models } from '../db/DB';
 import Swaps from '../swaps/Swaps';
-import { SwapRole, SwapFailureReason } from '../types/enums';
+import { SwapRole, SwapFailureReason, SwapPhase } from '../types/enums';
 import { CurrencyInstance, PairInstance, CurrencyFactory } from '../types/db';
 import {
   Pair, OrderIdentifier, StampedOwnOrder, OrderPortion, OwnOrder, StampedPeerOrder,
@@ -94,11 +94,21 @@ class OrderBook extends EventEmitter {
       this.swaps.on('swap.paid', (swapResult) => {
         if (swapResult.role === SwapRole.Maker) {
           const { orderId, pairId, quantity, peerPubKey } = swapResult;
+
+          // we must remove the amount that was put on hold while the swap was pending for the remaining order
+          this.removeOrderHold(orderId, pairId, quantity);
+
           this.removeOwnOrder(orderId, pairId, quantity, peerPubKey);
           this.emit('ownOrder.swapped', { pairId, quantity, id: orderId });
         }
       });
-      // TODO: bind to other swap events
+      this.swaps.on('swap.failed', (deal) => {
+        if (deal.role === SwapRole.Maker && deal.phase === SwapPhase.SwapAgreed) {
+          // if our order is the maker and the swap failed after it was agreed to but before it was executed
+          // we must release the hold on the order that we set when we agreed to the deal
+          this.removeOrderHold(deal.orderId, deal.pairId, deal.quantity!);
+        }
+      });
     }
   }
 
@@ -345,6 +355,7 @@ class OrderBook extends EventEmitter {
       price: maker.price,
       isBuy: !maker.isBuy,
       quantity: quantity || maker.quantity,
+      hold: 0,
     });
 
     try {
@@ -413,6 +424,18 @@ class OrderBook extends EventEmitter {
     }
 
     this.removeOwnOrder(order.id, order.pairId);
+  }
+
+  private addOrderHold = (orderId: string, pairId: string, holdAmount: number) => {
+    const tp = this.getTradingPair(pairId);
+    tp.addOrderHold(orderId, holdAmount);
+    this.logger.debug(`added hold on ${holdAmount} for order ${orderId}`);
+  }
+
+  private removeOrderHold = (orderId: string, pairId: string, holdAmount: number) => {
+    const tp = this.getTradingPair(pairId);
+    tp.removeOrderHold(orderId, holdAmount);
+    this.logger.debug(`removed hold on ${holdAmount} for order ${orderId}`);
   }
 
   /**
@@ -533,14 +556,14 @@ class OrderBook extends EventEmitter {
 
     const order = this.tryGetOwnOrder(orderId, pairId);
     if (order) {
-      const availableQuantity = order.hold ? order.quantity - order.hold : order.quantity;
+      const availableQuantity = order.quantity - order.hold;
       // TODO: accept the smaller of the proposed quantity and the available quantity
       if (availableQuantity >= proposedQuantity) {
         // put accepted quantity on hold
         // quantityToAccept = Math.min(proposedQuantity, availableQuantity);
         const quantityToAccept = proposedQuantity;
 
-        order.hold = order.hold ? order.hold + quantityToAccept : quantityToAccept;
+        this.addOrderHold(order.id, pairId, quantityToAccept);
 
         // try to accept the deal
         const orderToAccept = {
@@ -551,7 +574,7 @@ class OrderBook extends EventEmitter {
         const dealAccepted = await this.swaps!.acceptDeal(orderToAccept, requestPacket, peer);
         if (!dealAccepted) {
           // release hold amount and reject swap
-          order.hold -= quantityToAccept;
+          this.removeOrderHold(order.id, pairId, quantityToAccept);
         }
       } else {
         peer.sendPacket(new SwapErrorPacket({
--- a/lib/orderbook/TradingPair.ts
+++ b/lib/orderbook/TradingPair.ts
@@ -267,6 +267,20 @@ class TradingPair {
     return maps.buy.get(orderId) || maps.sell.get(orderId);
   }
 
+  public addOrderHold = (orderId: string, holdAmount: number) => {
+    const order = this.getOwnOrder(orderId);
+    assert(holdAmount > 0);
+    assert(order.hold + holdAmount <= order.quantity, 'the amount of an order on hold cannot exceed the available quantity');
+    order.hold += holdAmount;
+  }
+
+  public removeOrderHold = (orderId: string, holdAmount: number) => {
+    const order = this.getOwnOrder(orderId);
+    assert(holdAmount > 0);
+    assert(order.hold >= holdAmount, 'cannot remove more than is currently on hold for an order');
+    order.hold -= holdAmount;
+  }
+
   /**
    * Match an order against its opposite queue. Matched maker orders will be removed from the repository
    * @returns a [[MatchingResult]] with the matches as well as the remaining, unmatched portion of the order
--- a/test/unit/TradingPair.spec.ts
+++ b/test/unit/TradingPair.spec.ts
@@ -17,6 +17,7 @@ const createOwnOrder = (price: number, quantity: number, isBuy: boolean, created
   id: uuidv1(),
   localId: uuidv1(),
   pairId: PAIR_ID,
+  hold: 0,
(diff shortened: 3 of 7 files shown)


lib/orderbook/OrderBook.ts
```diff
@@ -98,7 +98,14 @@ class OrderBook extends EventEmitter {
           this.emit('ownOrder.swapped', { pairId, quantity, id: orderId });
         }
       });
-      // TODO: bind to other swap events
+      this.swaps.on('swap.failed', (deal) => {
+        if (deal.role === SwapRole.Maker && deal.phase === SwapPhase.SwapAgreed) {
+          // if our order is the maker and the swap failed after it was agreed to but before it was executed
+          // we must release the hold on the order that we set when we agreed to the deal
+          const order = this.getOwnOrder(deal.orderId, deal.pairId);
+          order.hold! -= deal.quantity!;
```
moshababo: this breaks the repository encapsulation.

lib/orderbook/OrderBook.ts
```diff
@@ -434,6 +441,12 @@ class OrderBook extends EventEmitter {
         this.pool.broadcastOrderInvalidation(removeResult.order, takerPubKey);
       }
 
+      if (takerPubKey && !removeResult.fullyRemoved) {
+        // if takerPubKey is specified, it means we are removing quantity from the order due to a successful swap
+        // we must also reduce the amount that was put on hold while the swap was pending for the remaining order
+        removeResult.order.hold! -= quantityToRemove!;
```
moshababo: same here

lib/orderbook/OrderBook.ts
sangaman: I don't get what you mean, could you elaborate? Do you mean this line should be in TradingPair.ts?

moshababo: Yes, I means repository in the greater sense, not our db repository class. `TradingPair` now manages the orders in-memory, so all the CRUD operations should be done through its API.

sangaman: Alright that makes sense, thanks for clarifying.

sangaman:
@moshababo I put all the logic that updates the hold amount in TradingPair.ts. it feels a little awkward vs leaving the logic in OrderBook.ts, but if it looks good to you I'll combine the commits and merge.

lib/orderbook/TradingPair.ts
```diff
@@ -267,6 +267,16 @@ class TradingPair {
     return maps.buy.get(orderId) || maps.sell.get(orderId);
   }
 
+  public addOrderHold = (orderOrOrderId: string | StampedOwnOrder, holdAmount: number) => {
```
moshababo: it's indeed don't make much sense to have this method accept an order instance, and perform the same logic as previously `OrderBook` did. I think it should only accept `orderId`, even though you have the order instance sometimes. Otherwise it's still breaking the encapsulation, and reducing `addOrderHold` to a util function. 
The point is not to encourage updates via instance mutations, although they work at the moment. You might get an instance copy for example, and so it won't work, or that other actions need to be applied in the repository (like in `order.quantity = 0`).

lib/orderbook/TradingPair.ts
```diff
@@ -267,6 +267,16 @@ class TradingPair {
...
+    const order = typeof orderOrOrderId === 'string' ? this.getOwnOrder(orderOrOrderId) : orderOrOrderId;
+    order.hold = order.hold ? order.hold + holdAmount : holdAmount;
```
moshababo: what about if `hold` exceeds `quantity`?
you're relying on higher-layer code to protect against that. I'd suggest to encapsulate that in what you're doing here, since it's not a generic update method, but rather specifically for `hold`.

lib/orderbook/TradingPair.ts
```diff
@@ -267,6 +267,16 @@ class TradingPair {
...
+  }
+
+  public removeOrderHold = (orderOrOrderId: string | StampedOwnOrder, holdAmount: number) => {
+    const order = typeof orderOrOrderId === 'string' ? this.getOwnOrder(orderOrOrderId) : orderOrOrderId;
+    order.hold = order.hold ? order.hold + holdAmount : holdAmount;
```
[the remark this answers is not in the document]
sangaman: Thanks for the catch, copied the method and forgot to change the operator.

sangaman:
@moshababo Thanks for the review. I addressed your feedback - one thing I left as is was having separate add and remove order hold methods. It felt a little clumsy to have to multiply parameters by negative one, plus I realized some `assert` statements were in order which made the two methods more distinct from each other. I also went ahead and added unit tests for these methods to make sure the logic was correct.

test/unit/TradingPair.spec.ts
```diff
@@ -339,3 +339,57 @@ describe('MatchingEngine queues and maps integrity', () => {
     expect(queueRemainingOrder).to.be.undefined;
   });
 });
+
+describe('TradingPair.addOrderHold & TradingPair.removeOrderHold', () => {
```
erkarl: Really like this addition...

lib/orderbook/OrderBook.ts
```diff
@@ -98,7 +98,14 @@ class OrderBook extends EventEmitter {
           this.emit('ownOrder.swapped', { pairId, quantity, id: orderId });
         }
       });
-      // TODO: bind to other swap events
+      this.swaps.on('swap.failed', (deal) => {
+        if (deal.role === SwapRole.Maker && deal.phase === SwapPhase.SwapAgreed) {
+          // if our order is the maker and the swap failed after it was agreed to but before it was executed
+          // we must release the hold on the order that we set when we agreed to the deal
+          const tp = this.tradingPairs.get(deal.pairId);
```
moshababo: can use `this.getTradingPair` as well, since you'll get an error anyway if it's missing (but less informative one)

lib/orderbook/OrderBook.ts
moshababo: should use `tp.removeOrderHold`

lib/orderbook/TradingPair.ts
```diff
@@ -267,6 +267,18 @@ class TradingPair {
     return maps.buy.get(orderId) || maps.sell.get(orderId);
   }
 
+  public addOrderHold = (orderId: string, holdAmount: number) => {
+    const order = this.getOwnOrder(orderId);
+    assert((order.hold || 0) + holdAmount <= order.quantity, 'the amount of an order on hold cannot exceed the available quantity');
+    order.hold = order.hold ? order.hold + holdAmount : holdAmount;
```
moshababo: nothing serious, but you're checking twice if `order.hold` exists, and initializing it anyway. Instead you can just initialize it (`order.hold = order.hold || 0`) in the beginning, and avoid the checks.

sangaman: @offerm suggested [here](https://github.com/ExchangeUnion/xud/pull/627#discussion_r231400551) to make hold initialize to 0 to simplify this logic, which I think is a fine idea, so I'll just go with that approach if that's alright with you.

moshababo: Sure.

lib/orderbook/OrderBook.ts
```diff
@@ -434,6 +442,12 @@ class OrderBook extends EventEmitter {
         this.pool.broadcastOrderInvalidation(removeResult.order, takerPubKey);
       }
 
+      if (takerPubKey && !removeResult.fullyRemoved) {
```
offerm: See comment above.
the condition on takerPubKey looks not in place. 
Also, why not to clear to hold amount even if the order is fully removed? Yes, it is redundant but the code will be much clearer.

I also  see no reason why not to hold/release for local matched.

lib/orderbook/OrderBook.ts
```diff
@@ -434,6 +442,12 @@ class OrderBook extends EventEmitter {
         this.pool.broadcastOrderInvalidation(removeResult.order, takerPubKey);
```
offerm: Question: Are we sending invalidation even if !removeResult.fullyRemoved?

lib/orderbook/OrderBook.ts
```diff
@@ -540,7 +554,8 @@ class OrderBook extends EventEmitter {
         // quantityToAccept = Math.min(proposedQuantity, availableQuantity);
         const quantityToAccept = proposedQuantity;
```
offerm: Please remove order.hold ? from the definition of availableQuantity

offerm: I would add an assert to make sure that availableQuantity is positive

lib/orderbook/TradingPair.ts
```diff
@@ -267,6 +267,18 @@ class TradingPair {
     return maps.buy.get(orderId) || maps.sell.get(orderId);
   }
 
+  public addOrderHold = (orderId: string, holdAmount: number) => {
+    const order = this.getOwnOrder(orderId);
```
offerm: assert holdamount is positive

lib/orderbook/TradingPair.ts
```diff
@@ -267,6 +267,18 @@ class TradingPair {
...
+    assert(order.hold + holdAmount <= order.quantity, 'the amount of an order on hold cannot exceed the available quantity');
+    order.hold += holdAmount;
+  }
+
+  public removeOrderHold = (orderId: string, holdAmount: number) => {
+    const order = this.getOwnOrder(orderId);
```
offerm: assert holdamount is positive

offerm:
Looks good. Some small changes please

lib/orderbook/OrderBook.ts
```diff
@@ -98,7 +98,14 @@ class OrderBook extends EventEmitter {
           this.emit('ownOrder.swapped', { pairId, quantity, id: orderId });
```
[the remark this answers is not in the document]
sangaman: Yeah that does seem like a more sensible flow, I'm changing it.

sangaman:
Thanks for the review @offerm, I believe I addressed all your comments.

lib/orderbook/OrderBook.ts
offerm: I don't understand. why to send invalidation of the order is still valid (after partial swap)?

sangaman: It lets peers know that part of the order is no longer available, so they know to reduce the quantity on their end.

offerm: not the best name for the operation.
As a minimum,  I think we should add a comment explaining what this method does.

In addition, while a P2P message was sent to multiple nodes, there is nothing in the logs (senders and receivers) which indicates that a message was sent accepted and processed (not to mention the outcome. Suggest you add this as part of #649 (or a new issue if you like).

erkarl: Good catch @offerm.

It's definitely confusing as of now.

sangaman: I'll add comments for the packet in a separate PR, but we do have that logging [when the packet is received and processed](https://github.com/ExchangeUnion/xud/blob/3868681741bada5ac39166152f14530ee3873f3a/lib/p2p/Pool.ts#L490) as well as logging when an [order is removed or reduced for any reason](https://github.com/ExchangeUnion/xud/blob/3868681741bada5ac39166152f14530ee3873f3a/lib/orderbook/TradingPair.ts#L206). Let's continue this in #649.

kilrau: Need to use verbose mode @offerm

sangaman: Verbose messages should be logged by default, since the default level is `debug` which logs everything except `trace`.

PR #627, opened 2018-11-01
feat(swaps): calculate swap amounts locally

### LINKED ISSUE #615: Swaps security issue: currency/amount info
In `OrderBook.handleSwapRequest`, the maker verifies his order availability, and accepts the deal. Further validation is applied on LND requests (`validateRequests`), based on the deal info. But the deal info was never validated on the maker side. It means that the maker is blindly accepting the taker currency/amount calculation in regards to the deal.

sangaman:
This makes each node involved in a swap determine the swap amounts and currencies (`makerCurrency`, `takerCurrency`, `makerAmount`, and `takerAmount`) independently. If there is any discrepancy in these
values, the swap will fail. These values are no longer sent as part of the `SwapRequestPacket`. Previously, only the taker (who initiates the swap) would determine these values and the maker would blindly accept them.

It also updates the Swap unit tests to align with the new methods and logic used in the `beginSwap` and `acceptDeal` routines.

Closes #615.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -555,37 +555,39 @@ class OrderBook extends EventEmitter {
     const { r_hash, proposedQuantity, orderId, pairId } = requestPacket.body!;
 
     const order = this.tryGetOwnOrder(orderId, pairId);
-    if (order) {
-      const availableQuantity = order.quantity - order.hold;
-      // TODO: accept the smaller of the proposed quantity and the available quantity
-      if (availableQuantity >= proposedQuantity) {
-        // put accepted quantity on hold
-        // quantityToAccept = Math.min(proposedQuantity, availableQuantity);
-        const quantityToAccept = proposedQuantity;
+    if (!order) {
+      peer.sendPacket(new SwapErrorPacket({
+        r_hash,
+        errorMessage: SwapFailureReason[SwapFailureReason.OrderNotFound],
+      }, requestPacket.header.id));
+      return;
+    }
 
-        this.addOrderHold(order.id, pairId, quantityToAccept);
+    const availableQuantity = order.quantity - order.hold;
+    assert(availableQuantity > 0);
+    // TODO: accept the smaller of the proposed quantity and the available quantity
+    if (availableQuantity >= proposedQuantity) {
+      // quantityToAccept = Math.min(proposedQuantity, availableQuantity);
+      /** The quantity of the order that we will accept */
+      const quantity = proposedQuantity;
 
-        // try to accept the deal
-        const orderToAccept = {
-          quantityToAccept,
-          localId: order.localId,
-          price: order.price,
-        };
-        const dealAccepted = await this.swaps!.acceptDeal(orderToAccept, requestPacket, peer);
-        if (!dealAccepted) {
-          // release hold amount and reject swap
-          this.removeOrderHold(order.id, pairId, quantityToAccept);
-        }
-      } else {
-        peer.sendPacket(new SwapErrorPacket({
-          r_hash,
-          errorMessage: SwapFailureReason[SwapFailureReason.OrderUnavailable],
-        }, requestPacket.header.id));
+      this.addOrderHold(order.id, pairId, quantity);
+
+      // try to accept the deal
+      const orderToAccept = {
+        quantity,
+        localId: order.localId,
+        price: order.price,
+        isBuy: order.isBuy,
+      };
+      const dealAccepted = await this.swaps!.acceptDeal(orderToAccept, requestPacket, peer);
+      if (!dealAccepted) {
+        this.removeOrderHold(order.id, pairId, quantity);
       }
     } else {
       peer.sendPacket(new SwapErrorPacket({
         r_hash,
-        errorMessage: SwapFailureReason[SwapFailureReason.OrderNotFound],
+        errorMessage: SwapFailureReason[SwapFailureReason.OrderOnHold],
       }, requestPacket.header.id));
     }
   }
--- a/lib/p2p/packets/types/SwapRequestPacket.ts
+++ b/lib/p2p/packets/types/SwapRequestPacket.ts
@@ -4,10 +4,6 @@ import PacketType from '../PacketType';
 export type SwapRequestPacketBody = {
   proposedQuantity: number;
   pairId: string;
-  takerAmount: number;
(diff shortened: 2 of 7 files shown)


moshababo:
The calculations are now being done on each side, but they are not negotiated. Although the swap will fail due to `validateRequest`, which is the only legit validation, why not to send the values from taker to maker in the swap request packet, to add another pre-swap opportunity for negotiation failure?

sangaman:
My thinking was that the swap amounts are fairly simple and purely deterministic - simple if/else and multiplication, and that explicitly sharing the amounts is redundant p2p traffic that's worth cutting out. I figure that if there's a discrepancy in the calculated amounts, it's either due to intentional misbehavior by one of the parties or a serious bug, but not something that should ever happen under normal operation.

moshababo:
I guess we can keep this out, since an adversary can always send correct values in the agreement, before sending incorrect values in the actual swap. So there's no extra security against that, only against bugs.

lib/orderbook/OrderBook.ts
```diff
@@ -532,37 +532,40 @@ class OrderBook extends EventEmitter {
...
+
+    const availableQuantity = order.hold ? order.quantity - order.hold : order.quantity;
+    // TODO: accept the smaller of the proposed quantity and the available quantity
+    if (availableQuantity >= proposedQuantity) {
+      // put accepted quantity on hold
+      // quantityToAccept = Math.min(proposedQuantity, availableQuantity);
+      /** The quantity of the order that we will accept */
+      const quantity = proposedQuantity;
+
+      order.hold = order.hold ? order.hold + quantity : quantity;
```
offerm: Consider making order.hold not nullable (initialize with zero).

It will make the code clearer 
>order.hold = order.hold + quantity;

lib/orderbook/OrderBook.ts
```diff
@@ -532,37 +532,40 @@ class OrderBook extends EventEmitter {
...
+    if (!order) {
+      peer.sendPacket(new SwapErrorPacket({
+        r_hash,
+        errorMessage: SwapFailureReason[SwapFailureReason.OrderNotFound],
+      }, requestPacket.header.id));
+      return;
+    }
+
+    const availableQuantity = order.hold ? order.quantity - order.hold : order.quantity;
+    // TODO: accept the smaller of the proposed quantity and the available quantity
```
offerm: We used to have partial quantity support in the past, no?

Do we have an open issue/PR for it. This should be `top priority`

lib/orderbook/OrderBook.ts
```diff
@@ -532,37 +532,40 @@ class OrderBook extends EventEmitter {
...
-        }, requestPacket.header.id));
+    if (!order) {
+      peer.sendPacket(new SwapErrorPacket({
+        r_hash,
+        errorMessage: SwapFailureReason[SwapFailureReason.OrderNotFound],
+      }, requestPacket.header.id));
+      return;
+    }
+
+    const availableQuantity = order.hold ? order.quantity - order.hold : order.quantity;
```
offerm: Same comment as below.

lib/orderbook/OrderBook.ts
```diff
@@ -532,37 +532,40 @@ class OrderBook extends EventEmitter {
...
+      const dealAccepted = await this.swaps!.acceptDeal(orderToAccept, requestPacket, peer);
+      if (!dealAccepted) {
+        // release hold amount and reject swap
+        order.hold -= quantity;
       }
     } else {
       peer.sendPacket(new SwapErrorPacket({
         r_hash,
-        errorMessage: SwapFailureReason[SwapFailureReason.OrderNotFound],
+        errorMessage: SwapFailureReason[SwapFailureReason.OrderUnavailable],
```
offerm: `OrderUnavailable ` - this name and the attached description does not reflect the situation properly.

lib/p2p/packets/types/SwapRequestPacket.ts
```diff
@@ -4,10 +4,6 @@ import PacketType from '../PacketType';
 export type SwapRequestPacketBody = {
   proposedQuantity: number;
   pairId: string;
-  takerAmount: number;
-  takerCurrency: string;
-  makerAmount: number;
-  makerCurrency: string;
```
offerm: Removing these fields will also remove them from log (since we print the packet).

This will make debugging much harder. 

If you like to take it out, please include in the log something to replace it. 
Both on the maker and on the taker.

lib/swaps/Swaps.ts
```diff
@@ -258,9 +253,14 @@ class Swaps extends EventEmitter {
...
     const deal: SwapDeal = {
       ...swapRequestBody,
+      takerCurrency,
+      makerCurrency,
+      takerAmount,
+      makerAmount,
       peerPubKey: peer.nodePubKey!,
       localId: taker.localId,
       price: maker.price,
+      isBuy: maker.isBuy,
```
offerm: Do we need this here if we have the orderID?

lib/orderbook/OrderBook.ts
sangaman: No, I had it in #495 but you asked me to take it out, it was never merged. There's a separate PR for it now, #633.

lib/p2p/packets/types/SwapRequestPacket.ts
sangaman: Good point, I will add log statements every time they are calculated.

lib/orderbook/OrderBook.ts
sangaman: OK will do.

lib/orderbook/OrderBook.ts
sangaman: How do you figure? My thinking is that the order is not currently available because it is on hold - although the order exists. Would `OrderOnHold` make more sense?

lib/swaps/Swaps.ts
sangaman: It's not absolutely necessary but it follows the pattern of having details like `price` as part of the deal. I think it makes it easier to deal with the swaps without having to make lookups of orders, by putting the key information on the swap deal itself.

lib/orderbook/OrderBook.ts
sangaman: I'll go ahead and make this `OrderOnHold` because I think it's a more specific indication of why the swap was rejected.

PR #629, opened 2018-11-02
chore: update typedoc dependency

sangaman:
The old version of typedoc was hitting an error that was resolved by updating to latest.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -245,9 +245,9 @@
       }
     },
     "@types/fs-extra": {
-      "version": "5.0.1",
-      "resolved": "https://registry.npmjs.org/@types/fs-extra/-/fs-extra-5.0.1.tgz",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -8274,45 +8281,46 @@
...
-        "marked": "^0.3.17",
+        "lodash": "^4.17.10",
+        "marked": "^0.4.0",
         "minimatch": "^3.0.0",
         "progress": "^2.0.0",
-        "shelljs": "^0.8.1",
+        "shelljs": "^0.8.2",
         "typedoc-default-themes": "^0.5.0",
-        "typescript": "2.7.2"
+        "typescript": "3.1.x"
```
erkarl: With these changes typedoc will be running a newer version of TypeScript. Not a big deal, but ideally I'd have the versions align. Could do the TS upgrade in a separate issue/PR.

sangaman: True, plus I've been thinking about moving to Typescript 3 anyway. I'll do that, thanks.

PR #631, opened 2018-11-02
refactor(orderbook): localId does not exist error

sangaman:
This creates a `LOCAL_ID_DOES_NOT_EXIST` exist error for when `xud` is asked to look up a local id that does not exist. This error is used in place of the `ORDER_NOT_FOUND` error.


moshababo:
Good idea for creating a separate error, but I think that merely referring to "local id" might not be enough. Perhaps `ORDER_LOCAL_ID_NOT_FOUND` or `ORDER_WITH_LOCAL_ID_NOT_FOUND` instead.

sangaman:
@moshababo How about leaving the name as is (in the code, `localId` always refers to an order already so I don't think it's too confusing) but changing the log message to "order with local id ______ does not exist"? That should make the log output that xud users and administrators read more clear.

moshababo:
Sounds good.

PR #633, opened 2018-11-02
Partial swap

### LINKED ISSUE #550: Partial swap acceptance
This issue is to track functionality to allow for swap requests to be partially accepted (for example, when part of an order is on hold) and for the swap amounts to be recalculated when this is the case. Further more, we'll need to not send an `ORDER_INVALIDATION` packet for the the full order when only part of it is successfully swapped.

ImmanuelSegol:
closes #550


lib/swaps/Swaps.ts
```diff
@@ -29,14 +31,17 @@ class Swaps extends EventEmitter {
...
+  private nodes: NodeList;
   /** The number of satoshis in a bitcoin. */
   private static readonly SATOSHIS_PER_COIN = 100000000;
 
   constructor(private logger: Logger, private models: Models, private pool: Pool, private lndBtcClient: LndClient, private lndLtcClient: LndClient) {
     super();
-    this.repository = new SwapRepository(this.models);
+    this.swapRepository = new SwapRepository(this.models);
+    this.p2pRepository = new P2PRepository(this.models);
+    this.nodes = new NodeList(this.p2pRepository);
```
sangaman: We shouldn't need the list of nodes and p2p repository in the swaps module. The list we load from the database on initialization will not stay in sync as we add new nodes.

lib/swaps/Swaps.ts
```diff
@@ -435,15 +440,17 @@ class Swaps extends EventEmitter {
...
       deal.quantity = quantity; // set the accepted quantity for the deal
       if (quantity <= 0) {
-        // TODO: accepted quantity must be a positive number, abort deal and penalize peer
+        this.setDealState(deal, SwapState.Error, 'accepted quantity must be a positive number');
+        await this.nodes.addReputationEvent(peer.nodePubKey!, ReputationEvent.SwapFailure);
+        return;
       } else if (quantity > deal.proposedQuantity) {
-        // TODO: accepted quantity should not be greater than proposed quantity, abort deal and penalize peer
+        this.setDealState(deal, SwapState.Error, 'accepted quantity should not be greater than proposed quantity');
+        await this.nodes.addReputationEvent(peer.nodePubKey!, ReputationEvent.SwapFailure);
```
sangaman: I don't think these reputation events need to be part of this PR, but if we do add them it should be in OrderBook.ts when we handle the `swap.failed` event (that way we don't need to have separate list of nodes in Swaps.ts). A separate PR would be better though just for a clearer commit history, we can penalize based on the reason for the failure. That might benefit from having #671 implemented first.

ImmanuelSegol: Agree this should be in a different pr.

sangaman:
@ImmanuelSegol Needs a rebase too, but I think let's just make these minor changes and then merge before the next release tomorrow.

sangaman:
I'm thinking that we'll also need changes to the matching logic to try to rematch any part of a swap request that *wasn't* accepted, I'll try to handle that as well in the next PR.

PR #634, opened 2018-11-02
refactor(peer) removing banning logic

### LINKED ISSUE #566: Refactor peer penalizing/banning logic
Currently, there's logic in Peer.ts around penalizing and banning peers. When the score exceeds a threshold, it emits a `ban` event, which in turn triggers a manual ban. We should change this so the `Peer` object doesn't track scores or decide when to ban - we have that logic in Pool.ts and NodeList.ts. Instead, it should emit an event when a reputation event occurs, which in turn should trigger calling the `addReputationEvent` method.

ImmanuelSegol:
closes #566

### DIFF
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -13,7 +13,7 @@ const reputationEventWeight = {
   [ReputationEvent.SwapFailure]: -10,
   [ReputationEvent.SwapSuccess]: 1,
   [ReputationEvent.InvalidMessage]: -10,
-  [ReputationEvent.UnknownMessageType]: -20,
+  [ReputationEvent.UnknownPacketType]: -20,
   [ReputationEvent.UnparseableMessage]: -10,
 };
 
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -12,6 +12,9 @@ const reputationEventWeight = {
   [ReputationEvent.PacketTimeout]: -1,
   [ReputationEvent.SwapFailure]: -10,
   [ReputationEvent.SwapSuccess]: 1,
+  [ReputationEvent.InvalidMessage]: -1,
+  [ReputationEvent.UnknownMessageType]: -1,
+  [ReputationEvent.UnparsableMessage]: -1,
 };
 
 // TODO: inform node about getting banned
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -12,9 +12,9 @@ const reputationEventWeight = {
   [ReputationEvent.PacketTimeout]: -1,
   [ReputationEvent.SwapFailure]: -10,
   [ReputationEvent.SwapSuccess]: 1,
-  [ReputationEvent.InvalidMessage]: -1,
-  [ReputationEvent.UnknownMessageType]: -1,
-  [ReputationEvent.UnparsableMessage]: -1,
+  [ReputationEvent.InvalidMessage]: -10,
+  [ReputationEvent.UnknownMessageType]: -20,
+  [ReputationEvent.UnparseableMessage]: -10,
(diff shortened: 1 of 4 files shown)


lib/p2p/NodeList.ts
```diff
@@ -12,6 +12,9 @@ const reputationEventWeight = {
   [ReputationEvent.PacketTimeout]: -1,
   [ReputationEvent.SwapFailure]: -10,
   [ReputationEvent.SwapSuccess]: 1,
+  [ReputationEvent.InvalidMessage]: -1,
+  [ReputationEvent.UnknownMessageType]: -1,
+  [ReputationEvent.UnparsableMessage]: -1,
```
sangaman: I think the old scores made a little more sense (10, 10, and 20 for `UnknownMessageType`). I'd probably go back to those, or maybe make it 5, 5, and 20, because the unknown message type implies a deliberately modified version of the software.

lib/p2p/Peer.ts
```diff
@@ -29,9 +30,9 @@ interface Peer {
   on(event: 'handshake', listener: () => void): this;
   once(event: 'open', listener: () => void): this;
   once(event: 'close', listener: () => void): this;
-  once(event: 'ban', listener: () => void): this;
+  once(event: 'reputationEvent', listener: (event: ReputationEvent) => void): this;
```
sangaman: I think "reputationEvent" is redundant since this is already an event, how about just "reputation" ?

lib/p2p/Peer.ts
```diff
@@ -251,16 +251,8 @@ class Peer extends EventEmitter {
...
-    this.banScore += score;
-
-    if (this.banScore >= 100) { // TODO: make configurable
-      this.logger.debug(`Ban threshold exceeded (${this.nodePubKey})`);
-      this.emit('ban');
-      return true;
-    }
-
-    return false;
+  private increaseBan = (event: ReputationEvent) => {
```
sangaman: I think we can get rid of this method and just use `this.emit('reputationEvent', ...`. Or if you'd prefer to keep it let's rename it to something like `emitReputationEvent`, since `increaseBan` is less descriptive and may even be wrong if we ever emit positive reputation events from here.

lib/types/enums.ts
```diff
@@ -47,6 +47,9 @@ export enum ReputationEvent {
   PacketTimeout = 2,
   SwapFailure = 3,
   SwapSuccess = 4,
+  UnparsableMessage = 5,
```
sangaman: Missing an 'e' in "Unparseable".

lib/p2p/NodeList.ts
ImmanuelSegol: Yeah, I wasn't sure about the penalty value. I'll go back to the original values.

lib/p2p/Peer.ts
ImmanuelSegol: Agree.

lib/types/enums.ts
ImmanuelSegol: Thanks.

lib/types/enums.ts
```diff
@@ -47,6 +47,9 @@ export enum ReputationEvent {
   PacketTimeout = 2,
   SwapFailure = 3,
   SwapSuccess = 4,
+  UnparseableMessage = 5,
+  InvalidMessage = 6,
+  UnknownMessageType = 7,
```
sangaman: One last thing here, I think this should be called `UnknownPacketType` which is more descriptive and matches the name of the parser error.

lib/types/enums.ts
sangaman: Interesting I did not know that, but yeah better to stick to the spelling with the e since that's what we're using elsewhere.

PR #635, opened 2018-11-02
refactor(Pool.ts) able to listen on multiple uris

### LINKED ISSUE #571: Specify interfaces for listening for peers
Currently xud listens on `0.0.0.0` for peers, but we will want to allow users to specify particular addresses/ports to listen on. `listen` in config is currently a boolean that indicates whether we're listening at all, but we should change it to a string array with interfaces to listen on. Any interface without a port should use `listenport`. The default value for `listen` should be `["0.0.0.0"]`, and the option to disable listening altogether should be called `nolisten` and should default to false.

ImmanuelSegol:
close #571


lib/p2p/Pool.ts
```diff
@@ -89,14 +91,25 @@ class Pool extends EventEmitter {
     super();
     this.config = config;
 
-    if (config.listen) {
-      this.listenPort = config.port;
+    if (!config.nolisten) {
+      this.listenPort = config.port || 0;
+      this.interfaces = [];
+
+      /** Makes sure all uris are properlly formatted  */
```
sangaman: properly instead of properlly

ImmanuelSegol:
@sangaman I'm not really sure why the test are failing.

lib/p2p/Pool.ts
```diff
@@ -712,18 +721,14 @@ class Pool extends EventEmitter {
...
-
-        this.server!.removeListener('error', listenErrHandler);
-        resolve();
-      }).on('error', listenErrHandler);
+      this.interfaces!.forEach((uri) => {
+        this.server!.listen(uri.port, uri.host).on('listening', () => {
+          const { address, port } = this.server!.address();
+          this.logger.info(`p2p server listening on ${address}:${port}`);
+          resolve();
+        }).on('error', listenErrHandler);
```
sangaman: It looks like we are adding the `listenErrHandler` once for each interface. You can add it once before starting the loop with `this.server!.on('error', listenErrHandler);`.

lib/p2p/Pool.ts
```diff
@@ -712,18 +721,14 @@ class Pool extends EventEmitter {
...
-        }
-
-        this.server!.removeListener('error', listenErrHandler);
-        resolve();
-      }).on('error', listenErrHandler);
+      this.interfaces!.forEach((uri) => {
+        this.server!.listen(uri.port, uri.host).on('listening', () => {
+          const { address, port } = this.server!.address();
+          this.logger.info(`p2p server listening on ${address}:${port}`);
+          resolve();
```
sangaman: This will resolve the promise after the first interface is being listened to, ideally we should wait for all of them to be listening before considering the method "complete." I think you should keep track of a counter like `listeningCount` which starts at 0, increments every time an interface starts listening, and then each time you increment it check it against `this.interfaces.length` and only if they are equal call `resolve()`.

lib/p2p/Pool.ts
```diff
@@ -712,18 +721,14 @@ class Pool extends EventEmitter {
         reject(err);
       };
 
-      this.server!.listen(this.listenPort || 0, '0.0.0.0').on('listening', () => {
```
sangaman: The comment block for the `listen` method should be updated to specify that multiple interfaces can be used.

ImmanuelSegol:
@kilrau @sangaman fixed the test

kilrau:
Well Travis only fails sporadically if it's 'due to Travis', so please run them until they pass. This is a minimum requirement to get to reviews :)

If Travis fails even though it shouldn't it's worth investing some time and digging up the root cause so this can be fixed permanently.

kilrau:
still needs tests to be fixed

test/p2p/sanity.spec.ts
```diff
@@ -52,8 +53,8 @@ describe('P2P Sanity Tests', () => {
   let unusedPort: number;
 
   before(async () => {
-    nodeOneConfig = createConfig(1, 0);
-    nodeTwoConfig = createConfig(2, 0);
+    nodeOneConfig = createConfig(1, 8855);
```
sangaman: I think we want to keep these ports dynamic for testing purposes, any particular reason this was changed?

test/p2p/networks.spec.ts
```diff
@@ -11,8 +11,8 @@ chai.use(chaiAsPromised);
 describe('P2P Networks Tests', () => {
   function testConnectionFailure(srcNodeNetwork: XuNetwork, destNodeNetwork: XuNetwork) {
     it(`should fail to connect a node from ${srcNodeNetwork} to a node from ${destNodeNetwork}`, async () => {
-      const srcNodeConfig = createConfig(1, 0, false, srcNodeNetwork);
```
sangaman: Same here for keeping the ports dynamic in tests.

lib/Xud.ts
```diff
@@ -62,7 +62,6 @@ class Xud extends EventEmitter {
     const loggers = Logger.createLoggers(this.config.loglevel, this.config.logpath, this.config.instanceid, this.config.logdateformat);
     this.logger = loggers.global;
     this.logger.info('config loaded');
-
```
sangaman: Also can we keep this blank line?

lib/p2p/Pool.ts
```diff
@@ -846,7 +847,7 @@ class Pool extends EventEmitter {
   }
 
   /**
-   * Starts listening for incoming p2p connections on the configured host and port. If `this.listenPort` is 0 or undefined,
+   * Start listening for incoming p2p connections on specified hosts and ports. If `this.listenPort` is 0 or undefined,
```
sangaman: Proper style here is actually "Starts" instead of "Start" for javadoc so I've been trying to stick to that. Also it seems like we're not using `listenPort` any more, can drop it from this method description and maybe from the class altogether?

rsercano: `listenPort` is actually being used in `detectExternalIpAddress` I guess it's not being set correctly but somehow I feel like we need to set one of the ports from interfaces into `this.nodeState.addresses` ?

kilrau: `detectExternalIpAddress` indeed is not working. Did you see why? @rsercano

rsercano: do you mean for this branch @kilrau or totally? :)

rsercano: There's actually a mess in `Pool.ts` right now, since it's been a lot after @ImmanuelSegol 's commits there're lots of changes and I'm trying to figure them out to merge correctly now.

rsercano:
@ImmanuelSegol I need your assistance with this, why do we still need **listenport** in Pool.ts and why there's no assignation of interfaces in `Pool.ts` ? And there's a configuration parameter in p2p with the name `nolisten` which is not being used with the latest commit of yours ?

sangaman:
@rsercano Should we close this PR and start over? If you're unsure how to proceed, no worries, we can juts leave the issue open and get to it later down the road.

kilrau:
> @rsercano Should we close this PR and start over? I

I think that makes a lot of sense @rsercano . If you agree, please close this PR. As you can see the issue is assigned to the `post-1.0.0` project, so not a prio rn.

PR #641, opened 2018-11-06
chore: update typescript dependency

sangaman:
This updates the typescript dependency to be in sync with the version used by typedoc as discussed in #629. The older version of tslint was not compatible with the new version of typescript, so that was updated as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -825,7 +825,7 @@
       "dependencies": {
         "chalk": {
           "version": "1.1.3",
-          "resolved": "https://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
+          "resolved": "http://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -825,7 +825,7 @@
       "dependencies": {
         "chalk": {
           "version": "1.1.3",
-          "resolved": "https://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
+          "resolved": "http://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
```
erkarl: Looks like you're using an old version of npm and it's unnecessarily generating a diff here. On my local machine it's changing it back to https with `npm 6.4.1`.

sangaman: That's odd, because I deliberately updated npm right before making these changes.

```
+ npm@6.4.1
updated 1 package in 17.993s
```

I'll look into it.

sangaman: Checking out this branch again and running `npm install` flips this line back to https. Just some oddity with npm I guess. I'll amend the commit.

PR #642, opened 2018-11-06
Refactor(Swaps.ts) checks if r_hash exists before making swap

### LINKED ISSUE #547: Check that we don't already have an r_hash for swap requests
Currently, we don't have any checks up front when handling a swap request to see if we currently have or ever had a swap for that r_hash. We definitely don't want to allow reuse of preimages or anything like that, so we should detect this and reject the swap.

ImmanuelSegol:
Closes #547

### DIFF
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -293,7 +293,11 @@ class Swaps extends EventEmitter {
     // TODO: max cltv to limit routes
     // TODO: consider the time gap between taking the routes and using them.
     // TODO: multi route support (currently only 1)
-    // TODO: check to make sure we don't already have a deal for the requested r_hash
+
+    if (this.usedHashes.has(requestPacket.body!.r_hash)) {
+      this.sendErrorToPeer(peer, requestPacket.body!.r_hash, 'this r_hash already exsist', requestPacket.header.id);
+      return false;
+    }
     const requestBody = requestPacket.body!;
 
     const takerPubKey = peer.getLndPubKey(requestBody.takerCurrency)!;


lib/swaps/Swaps.ts
```diff
@@ -293,7 +293,11 @@ class Swaps extends EventEmitter {
     // TODO: max cltv to limit routes
     // TODO: consider the time gap between taking the routes and using them.
     // TODO: multi route support (currently only 1)
-    // TODO: check to make sure we don't already have a deal for the requested r_hash
+
+    if (this.usedHashes.has(requestPacket.body!.r_hash)) {
```
erkarl: I'm working on adding integration tests for `Swaps` in https://github.com/ExchangeUnion/xud/pull/650

Added a skipped test for it: https://github.com/ExchangeUnion/xud/pull/650/files#diff-9b4e088bfea3d98a165805245f86a461R79

erkarl:
Didn't test manually, but looks good. 

Nit: s/makeing/making in the commit message

lib/swaps/Swaps.ts
```diff
@@ -293,7 +293,11 @@ class Swaps extends EventEmitter {
...
+      this.sendErrorToPeer(peer, requestPacket.body!.r_hash, 'this r_hash already exsist', requestPacket.header.id);
```
sangaman: typo, "exists" instead of "exists"

sangaman:
Couple other nits for the commit message:

- Since this is adding/changing functionality, it should be a `feat` instead of `refactor` (and start with lower case). Refactoring commits shouldn't change how the code actually functions, and they get skipped over for the changelog.
- A colon after the parentheses, and I think for the scope you can just say `swaps` instead of the actual filename, ultimately that should just be a rough description of what part of the codebase is impacted. It goes in the changelog as well and people reading it don't need to see the filenames changed.
- exsist/exists typo
- Not a big deal but try to keep the first line max 50 characters if you can, it gets cut off beyond that in most places.

I don't at all mind squash merging and writing the commit message myself if you'd prefer, but any commits to master should follow the template.

erkarl:
Also needs a rebase.

lib/swaps/Swaps.ts
ImmanuelSegol: @erkarl Thanks 👍

PR #645, opened 2018-11-06
Refactor(Order proto) added hold value

### LINKED ISSUE #644: Add hold to getorders response
Per recent discussions, we want to have the `hold` property for orders retrieved from the `getorders` rpc call. This would need to be added to the `Order` protobuf message definition and then assigned a value in GrpcService.ts.

ImmanuelSegol:
closes #644

### DIFF
--- a/lib/grpc/GrpcService.ts
+++ b/lib/grpc/GrpcService.ts
@@ -21,7 +21,7 @@ const createOrder = (order: Order) => {
   grpcOrder.setCreatedAt(order.createdAt);
   grpcOrder.setId(order.id);
   if (isOwnOrder(order)) {
-    grpcOrder.setHold(order.hold ? order.hold : 0);
+    grpcOrder.setHold(order.hold);
     grpcOrder.setLocalId((order).localId);
     grpcOrder.setIsOwnOrder(true);
   } else {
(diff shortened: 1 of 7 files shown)


lib/grpc/GrpcService.ts
```diff
@@ -21,6 +21,7 @@ const createOrder = (order: StampedOrder) => {
   grpcOrder.setCreatedAt(order.createdAt);
   grpcOrder.setId(order.id);
   if (isOwnOrder(order)) {
+    grpcOrder.setHold(order.hold ? order.hold : 0);
```
sangaman: `hold` is never undefined in the latest code so you can just set this equal to `order.hold`.

proto/xudrpc.proto
```diff
@@ -387,6 +387,8 @@ message Order {
   OrderSide side = 8 [json_name = "side"];
   // Whether this order is a local own order or a remote peer order.
   bool is_own_order = 9 [json_name = "is_own_order"];
+  // The amount on hold pending swap exectuion.
+  int64 hold = 10 [json_name = "hold"];
```
sangaman: This should be a `double` since it uses decimal values, same as `quantity`.

Issue #240, opened 2018-07-31
Inbound/outbound channel balance check


kilrau:
On `placeOrder`, check over all open payment channels if inbound (Buy) and outbound (sell) balance is enough to execute the order. One hop only. If not, return  'channel balance not enough` event & print.

kilrau:
Channels with balance below trade amount result in "<coin> <incoming/outgoing> channel balance not sufficient". Non-existing channels with "<coin> <incoming/outgoing> channel not existent". "Balance check" is actually an error message and occurs upon match with peer/swap attempt, channel balance is NOT checked immediately after entering an order as channel balances are very dynamic.

Error event via stream, `placeOrder` sync returns immediately. Printed on CLI.

kilrau:
How about this one? @erkarl

erkarl:
@kilrau I'll take a look!

sangaman:
I'm sorry I hadn't looked at this issue in a while, do we really want to *require* a direct channel with sufficient balance for swaps? I figure that as long as we can find a route, we should be good to proceed.

erkarl:
@sangaman agreed. That's what the current implementation is doing, right?

@kilrau do we also really require the channel & channelbalance checks upon placing the order? Upon implementing #650 I got the feeling that it would be redundant. We're verifying that the route with sufficient balance exists before starting the swap anyway.

Having channel & channelbalance checks upon placing the order would protect against orderbook spoofing, but I feel like this is something the reputation system should handle, eventually.

sangaman:
> That's what the current implementation is doing, right?

I think right now we only check for a route when deciding whether to accept a deal but I believe in #650 you're adding the check before we initiate a deal as well which is good I think.

kilrau:
> we really want to require a direct channel with sufficient balance for swaps

We don't require a direct channel, a route is fine - sry if this was not clear. We check if the **balance** of OUR channel with the next peer is sufficient, no matter if the next peer is the final destination or not. We obviously can't check anything after the first hop anyways. @sangaman @kilrau 

> do we also really require the channel & channelbalance checks upon placing the order? Upon implementing #650 I got the feeling that it would be redundant

I'd suggest to do it, but only print a warning/hint as I mentioned [above](https://github.com/ExchangeUnion/xud/issues/240#issuecomment-435883105). Reason is unintended behavior, like match with an internal orders whereas peer orders were targeted.

erkarl:
@kilrau @sangaman 

[...]

PR #650, opened 2018-11-08
feat(swaps): verify route before executing the swap

ghost:
This PR adds an additional check to `verifyExecution` (before we execute a swap) in order to ensure that a valid route to destination exists. Putting this out here for early feedback.

Related issue: https://github.com/ExchangeUnion/xud/issues/240


sangaman (changes requested):
I like breaking out the swap logic into methods, that will definitely make this code more maintainable/testable. I made some comments on the code changes, I haven't looked closely at the tests yet but will get to that tomorrow.

lib/orderbook/OrderBook.ts
```diff
@@ -277,6 +277,9 @@ class OrderBook extends EventEmitter {
         remainingOrder: order,
       });
     }
+    // TODO: Check channel balance to verify that we can swap the order amount.
+    // Case channel not existent -> error!
+    // Case channel with sufficient capacity exists, but not enough local balance -> warning!
```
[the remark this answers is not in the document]
erkarl: This comment was a result of the issue description in: https://github.com/ExchangeUnion/xud/issues/240

That was my reasoning as well - it doesn't make sense to have it in the `OrderBook` for now. The only reason we'd want to have an additional check here is to prevent spoofing orders that can/will never be swapped.

erkarl:
@sangaman thank you for the awesome review. Your comments should now be addressed. Ready for round 2 :).

lib/orderbook/OrderBook.ts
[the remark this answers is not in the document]
sangaman: Ah my mistake I had not looked at that issue in a while. I commented there on whether we should actually require a direct channel, but either way I don't think we'd have it in OrderBook.ts.

lib/swaps/Swaps.ts
```diff
@@ -192,41 +240,33 @@ class Swaps extends EventEmitter {
...
-        return;
-      }
+  public executeSwap = async (maker: PeerOrder, taker: OwnOrder): Promise<SwapResult> => {
+    await this.verifyExecution(maker, taker);
+    const rHash = this.beginSwap(maker, taker);
+    if (!rHash) {
+      throw new Error('cannot execute swap. rHash not found');
+    }
 
+    return await new Promise((resolve, reject) => {
```
sangaman: You don't need the `await` here.

lib/swaps/Swaps.ts
```diff
@@ -192,41 +240,33 @@ class Swaps extends EventEmitter {
...
-      const rHash = this.beginSwap(maker, taker);
-      if (!rHash) {
-        reject();
-        return;
-      }
-
       this.on('swap.paid', onPaid);
       this.on('swap.failed', onFailed);
-    });
+    }) as SwapResult;
```
sangaman: I think it's a little bit more readable if instead of `as SwapResult` here, you did `return new Promise<SwapResult>...` on line 250. This is so far removed from the `return` statement that it's not obvious that they're part of the same line of code. Generally I prefer `as` over `<>` but it took me a second here to realize what was going on.

test/integration/Swap.spec.ts
```diff
@@ -0,0 +1,120 @@
+import chai, { expect } from 'chai';
+import sinon from 'sinon';
+import Pool from '../../lib/p2p/Pool';
+import Peer from '../../lib/p2p/Peer';
+import Swaps from '../../lib/swaps/Swaps';
+import LndClient, { LndClientConfig } from '../../lib/lndclient/LndClient';
+import Logger, { Level } from '../../lib/Logger';
+import DB from '../../lib/db/DB';
+import * as lndrpc from '../../lib/proto/lndrpc_pb';
```
sangaman: lndrpc and LndClientConfig are unused imports.

test/integration/Swap.spec.ts
```diff
@@ -0,0 +1,120 @@
+import chai, { expect } from 'chai';
```
sangaman: I think you need to use this `chai` import by adding the line `chai.use(chaiAsPromised);` if you want to use things like `expect.to.eventually.equal`

test/integration/Swap.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+        amountReceived: 1000,
+        amountSent: 8,
+        rHash: 'd94c22a73d2741ed5cdcf3714f9ab3c8664793b03a54c74a08877726007d67c2',
+        peerPubKey: '020c9a0fb8dac5b91756fb21509aefc4e95b585510c4de6e6311f18348a4723cdd',
+        role: 0,
+      };
+      const swapListenersAdded = sinon.spy(swaps, 'on');
+      const addDealSpy = sinon.spy(swaps, 'addDeal');
+      const swapListenersRemoved = sinon.spy(swaps, 'removeListener');
+      expect(swaps.executeSwap(maker, taker)).to.eventually.equal(swapResult);
```
sangaman: I believe you should be `await`ing this since the assertion is a promise in this case.

test/integration/Swap.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+      await waitForSpy(swapListenersAdded);
+      expect(addDealSpy.calledOnce).to.equal(true);
+      swapResult.rHash = addDealSpy.args[0][0].rHash;
+      swaps.emit('swap.paid', swapResult);
+      expect(waitForSpy(swapListenersRemoved, 'calledTwice')).to.be.fulfilled;
```
sangaman: Same here for `await`

sangaman (commented):
Thanks! The tests look really good and useful, I'm not familiar with sinon and all it can do but it seems like a really handy way to test the swaps module without setting up all the required components. Are you going to implement the rest of the methods as part of this PR?

lib/swaps/Swaps.ts
```diff
@@ -172,18 +172,66 @@ class Swaps extends EventEmitter {
...
+  }
+
+  /**
+   * Checks if route(s) exists given currency, amount and peerPubKey.
+   * @param currency currency/chain to find routes in
+   * @param amount the capacity of the route
+   * @param peerPubKey target node for the route
+   * @returns routes
+   */
+  private checkRoutes =  async (currency: string, amount: number, peerPubKey: string): Promise<lndrpc.Route[]> => {
```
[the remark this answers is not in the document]
erkarl: Good catch - I changed the signature, but didn't change the name.

lib/swaps/Swaps.ts
erkarl: Thanks. It definitely does look better. Will try to use `<>` over `as` from now on.

test/integration/Swap.spec.ts
erkarl: I can't use `await swaps.executeSwap(maker, taker);` here because it would block the execution of the test function until `executeSwap` resolves. It will never happen because `swap.paid` event is not emitted.

`chai-as-promised` allows us to continue executing the test, but still assert `to.eventually.equal` that the promise eventually resolves.

test/integration/Swap.spec.ts
erkarl: It's interesting that it was working as intended even without the import. Regardless, I added the import as per official `chai-as-promised` usage documentation.

offerm:
@erkarl 
1.  agree.
2. found it. Is there a way to reuse the verification code `verifyExecution` also when the maker checks for route?
3. agree. @kilrau please consider if you like to open an issue for this

erkarl:
@offerm 

2. Possibly, but it's not apparent for me right now. I did refactor `acceptDeal` a little by creating `getRoutes` which both `verifyExecution` and `acceptDeal` use. Not sure if it's worth going further in this PR.

lib/swaps/Swaps.ts
sangaman: `as` is definitely fine too, but I think only when it can stay on the same line as the object it's casting.

test/integration/Swap.spec.ts
sangaman: Ok sounds good, thanks for the explanation.

erkarl:
@sangaman your second wave of comments should now be addressed. In addition I added 3 new test cases for `executeSwap`. I'll address the skipped tests for `acceptDeal` in future PRs.

sangaman (approved):
Looks good. Followups can be in future PRs, but I what's done here so far would be a good improvement.

kilrau:
>> Should we include the has route indication with getorders response?
> 3\. agree. @kilrau please consider if you like to open an issue for this

Honestly, I don't think we should care about routes too much at the moment. Exchanges will have direct channels with each other because swaps need to reliable and routing on lightning is far away from that.

General problems I see with this field as part of the `getOrders` response:
* in the moment I get the order book from `getOrders`, `has route` is already outdated.
* when do we update `has route`? Periodically? As part of the `getOrders` call?
* in the end `has route` is only a vague info that might prevent me from doing a trade while actually a route exists. Technically we should remove all orders from the order book where `has route` is false which renders this field obsolete.

What this PR does is enough for now IMHO 👍 

But please open a separate issue if you think differently @offerm

erkarl:
Squashed the commits.

Issue #616, opened 2018-10-30
not connected to peer after disconnect


offerm:
My node was up connected to 3 peers and have peer orders.

wifi down for 3 minutes, wifi back up.

peers are missing (no reconnect).

kilrau:
Maybe add logs @offerm

offerm:
Not really needed.

This is not a bug - we don't [YET] have the code that suppose to do it
(IMHO).

Log only shows the disconnects and that is it.

kilrau:
Hmm shouldn't https://github.com/ExchangeUnion/xud/pull/392 (https://github.com/ExchangeUnion/xud/issues/311) catch it where we defined a connection retry logic for a momentarily unavailable peer?

moshababo:
@kilrau #392 is for the initial connection retries, not after disconnection. We talked before about re-connecting after a disconnection, but decided not to do so.

This specific disconnection happens because peer is stalling ping/pong response. It doesn't make sense to define re-connection from the side which closed the connection. Instead we can check 
network connectivity before closing peers for stalling responses.

But the other side might close the connection as well, and we won't know about it until we're back online. In this case re-connecting make sense. To implement this, we need to define specific cases for reconnecting after peer closed the connection, because in some cases it can be considered as DoS.

To do this, we first need to implement #152.
I expect that we'll get this packet when we'll be back online, but we need to verify it.

sangaman:
Would the idea then be to attempt a reconnection after receiving a `DISCONNECTING` packet? And I'm guessing the packet should specify that the reason for the disconnection is because our node became unresponsive?

offerm:
Kind of inventing the wheel here, no?

See how this started, I had a wifi issue, not a disconnect by a remote peer.

IMHO, if I connected to a peer and disconnected, there should be a simple
retry mechanism that try to connect in a similar way we are doing when LND
disconnected.

Now, if the connection was rejected (reject package) by a peer there is no
need to try an reconnect.

IMHO, peer rejection and peer banning have some time...  wifi problems and
even closing my my laptop going to `sleep` are here already.

sangaman:
I'm not 100% sure but as I understand it the TCP socket will not necessarily close if you lose wifi connectivity temporarily, but your peers will see that you have become unresponsive and disconnect from you. So the actual socket connection would be closed by the peer in this case. Correct me if I'm wrong.

offerm:
TCP socket will close when TCP layers notice it.  If you have a short drop
[...]

PR #658, opened 2018-11-10
feat: disconnecting packet

moshababo:
Implements #152.

I haven't implemented saving the packet to the database on the receiver side. I thought that having it in the logs will be sufficient. If you think otherwise, let me know.

### DIFF
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -3,8 +3,8 @@ import P2PRepository from './P2PRepository';
 import { NodeInstance, NodeFactory } from '../types/db';
 import { Address } from '../types/p2p';
 import addressUtils from '../utils/addressUtils';
-import { UriParts } from '../utils/utils';
 import { ReputationEvent } from '../types/enums';
+import { db } from '../types';
 
 const reputationEventWeight = {
   [ReputationEvent.ManualBan]: Number.NEGATIVE_INFINITY,
@@ -18,8 +18,8 @@ const reputationEventWeight = {
 // TODO: remove reputation events after certain amount of time
 
 interface NodeList {
-  on(event: 'node.ban', listener: (nodePubKey: string) => void): this;
-  emit(event: 'node.ban', nodePubKey: string): boolean;
+  on(event: 'node.ban', listener: (nodePubKey: string, events: db.ReputationEventInstance[]) => void): this;
+  emit(event: 'node.ban', nodePubKey: string, events: db.ReputationEventInstance[]): boolean;
 }
 
 /** Represents a list of nodes for managing network peers activity */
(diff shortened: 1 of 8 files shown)


lib/p2p/packets/types/DisconnectingPacket.ts
```diff
@@ -0,0 +1,31 @@
+import Packet, { PacketDirection } from '../Packet';
+import PacketType from '../PacketType';
+
+export type DisconnectingPacketBody = {
+  reason: DisconnectionReason;
+  payload?: string;
+};
+
+export enum DisconnectionReason {
```
sangaman: I think it would make sense to put this in enums.ts.

lib/p2p/packets/types/DisconnectingPacket.ts
```diff
@@ -0,0 +1,31 @@
...
+  ResponseStalling,
+  IncompatibleProtocolVersion,
+  UnexpectedIdentity,
+  ForbiddenIdentityUpdate,
+  ConnectedToSelf,
+  NotReadyForConnections,
```
sangaman: I think this should be something like "NotAcceptingConnections" - because "not ready" implies that we will be ready in the future, which is not always going to be the case. For now the only time we're using this is when we're in the process of shutting down the p2p pool.

lib/p2p/NodeList.ts
```diff
@@ -141,7 +141,9 @@ class NodeList extends EventEmitter {
         promises.push(this.setBanStatus(node, true));
         this.bannedNodes.add(node.nodePubKey);
         this.nodes.delete(node.nodePubKey);
-        this.emit('node.ban', nodePubKey);
+
+        const events = await this.repository.getReputationEvents(node);
```
sangaman: Sending all reputation events ever seems excessive, I think it would be better to only retrieve something like the last 5 or 10 negative events (no point in sending positive ones). If you don't want to handle that in this PR you can just put a TODO, but it shouldn't be too hard to query a subset of the events.

moshababo:
@sangaman yes, this is a prerequisite for #616.

lib/p2p/NodeList.ts
erkarl: This is marked as resolved, but I can't see @sangaman's comment being addressed. Sorry if I'm missing something.

sangaman: Yes I'm unresolving as it seems like it was overlooked. It's not critical or anything but a TODO would be good so that it's not forgotten.

sangaman:
@moshababo Looks good, can you just add that TODO as requested? Or change it to pull back on the last 5 or 10 events if you're up to it and have the time, either way is fine with me.

moshababo:
@sangaman I already addressed the issue, you can find it in `Pool.ts` line 197.

lib/p2p/NodeList.ts
moshababo: See my comment on the main thread.

sangaman: Oops my mistake, thank you!

PR #659, opened 2018-11-10
feat(OrderBook.ts) persist trades to DB

### LINKED ISSUE #608: Save all completed trades to database
We should store a history of trades made through `xud`. It should use a new table in the database (`trades`) and add a row every time an order is filled. I figure this should not be applicable for `nomatching` mode.

Here's what I'm thinking for columns in this table:

- orderId
- localId
- pairId
- quantity
- price
- matchingOrderId (for the order this was traded against, I'd be open to a different name. there should be a unique constraint for this combined with `orderId`.)
- r_hash (if it was swapped, this can also be used as a foreign key to do a join and find other information about the swap such as the peer)

Once this is in place, I'll create a separate issue to implement a `ListTrades` RPC method.

I also think we might want a `LookupOrder` method which checks if an order is still live or, if not, tries to find record of it in `trades` or `swapdeals`. Also something for a separate issue and PR.

ImmanuelSegol:
Edit: This addresses most of #608 but not the case where a swap is initiated by a remote peer.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -394,9 +394,9 @@ class OrderBook extends EventEmitter {
   private persistTrade = async (quantity: number, makerOrder: orders.PeerOrder | orders.OwnOrder, takerOrder: orders.OwnOrder | orders.PeerOrder) => {
     await this.repository.addOrdersIfNotExists([makerOrder, takerOrder]);
     await this.repository.addTrade({
+      quantity,
       makerOrderId: makerOrder.id,
       takerOrderId: takerOrder.id,
-      quantity,
     });
   }
 
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -316,7 +316,7 @@ class OrderBook extends EventEmitter {
           await this.repository.addOrderIfNotExists(maker);
           const swapResult = await this.swaps.executeSwap(maker, taker);
           this.emit('peerOrder.filled', portion);
-          await this.repository.addOrdersIfNotExists([maker,taker])
+          await this.repository.addOrdersIfNotExists([maker, taker]);
           await this.persistTrade(swapResult.quantity, maker.id, taker.id);
           result.swapResults.push(swapResult);
           onUpdate && onUpdate({ case: PlaceOrderEventCase.SwapResult, payload: swapResult });
@@ -365,7 +365,7 @@ class OrderBook extends EventEmitter {
     try {
       const swapResult = await this.swaps!.executeSwap(maker, taker);
       this.emit('peerOrder.filled', maker);
-      await this.repository.addOrdersIfNotExists([maker,taker]);
+      await this.repository.addOrdersIfNotExists([maker, taker]);
       await this.persistTrade(swapResult.quantity, maker.id, taker.id);
       return swapResult;
     } catch (err) {
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -1,4 +1,4 @@
-import assert, { throws } from 'assert';
+import assert from 'assert';
 import uuidv1 from 'uuid/v1';
 import { EventEmitter } from 'events';
 import OrderBookRepository from './OrderBookRepository';
@@ -17,6 +17,7 @@ import { Pair, OrderIdentifier, OwnOrder, OrderPortion, OwnLimitOrder, PeerOrder
 import { PlaceOrderEvent, PlaceOrderEventCase, PlaceOrderResult } from '../types/orderBook';
 import { SwapRequestPacket, SwapFailedPacket } from '../p2p/packets';
 import { SwapResult } from 'lib/swaps/types';
+import { Order } from 'lib/proto/xudrpc_pb';
 
 interface OrderBook {
   /** Adds a listener to be called when a remote order was added. */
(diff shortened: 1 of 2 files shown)


lib/orderbook/OrderBook.ts
```diff
@@ -387,6 +390,14 @@ class OrderBook extends EventEmitter {
     return true;
   }
 
+  private perssitTrade = async (quantity: number, makerOrderId: string, takerOrderId: string) => {
```
sangaman: typo in "persist"

lib/orderbook/OrderBook.ts
```diff
@@ -1,4 +1,4 @@
-import assert from 'assert';
+import assert, { throws } from 'assert';
```
sangaman: `throws` looks like it's an unused import.

lib/orderbook/OrderBookRepository.ts
```diff
@@ -40,6 +44,29 @@ class OrderbookRepository {
       await this.models.Order.create(order as OrderAttributes);
     }
   }
+
+  /**
+   * Only persist multiple orders if they don't exists.
+   * @param orders array of orders to persist
+   */
+  public addOrdersIfNotExists = async (orders: db.OrderFactory[]) => {
```
sangaman: I'm not sure there's much value in having a batch version of the existing `addOrderIfNotExists` method, mainly since I can't envision a case where we'd ever be trying to add more than 2 orders at a time.

If you do want to keep it, I'd reuse the logic from `addOrderIfNotExists` and call it once for each order in the array, aggregate the promises into a separate array, and then do `Promise.all` on the promises array.

lib/orderbook/OrderBook.ts
```diff
@@ -300,6 +300,8 @@ class OrderBook extends EventEmitter {
         portion.localId = maker.localId;
         result.internalMatches.push(maker);
         this.emit('ownOrder.filled', portion);
+        await this.repository.addOrdersIfNotExists([maker, taker]);
```
sangaman: We should add the orders from within the `persistTrade` method so we don't need to copy this line each time its called.

lib/orderbook/OrderBook.ts
```diff
@@ -314,6 +316,8 @@ class OrderBook extends EventEmitter {
           await this.repository.addOrderIfNotExists(maker);
           const swapResult = await this.swaps.executeSwap(maker, taker);
           this.emit('peerOrder.filled', portion);
+          await this.repository.addOrdersIfNotExists([maker, taker]);
```
sangaman: Same as above.

lib/orderbook/OrderBook.ts
```diff
@@ -361,6 +365,8 @@ class OrderBook extends EventEmitter {
     try {
       const swapResult = await this.swaps!.executeSwap(maker, taker);
       this.emit('peerOrder.filled', maker);
+      await this.repository.addOrdersIfNotExists([maker, taker]);
```
sangaman: Same as above.

lib/orderbook/OrderBook.ts
```diff
@@ -387,6 +391,15 @@ class OrderBook extends EventEmitter {
     return true;
   }
 
+  private persistTrade = async (quantity: number, makerOrder: orders.PeerOrder | orders.OwnOrder, takerOrder: orders.OwnOrder | orders.PeerOrder) => {
+    await this.repository.addOrdersIfNotExists([makerOrder, takerOrder]);
```
erkarl: Looking at the previous comments/changes I also think it would be cleaner to just have:
```
await this.repository.addOrderIfNotExists(makerOrder);
await this.repository.addOrderIfNotExists(takerOrder);
```
instead of `this.repository.addOrdersIfNotExists`.

lib/orderbook/OrderBookRepository.ts
```diff
@@ -40,6 +44,24 @@ class OrderbookRepository {
...
+    const ordersToAdd: Promise<void>[] = [];
+
+    await Bluebird.Promise.each(orders, (order) => {
```
sangaman: I'm not exactly sure what `Bluebird.Promise.each` does, but I think you can just do `orders.forEach` without needing `await`. It might wind up doing the same thing but it's less verbose at least.

lib/orderbook/OrderBook.ts
sangaman: I'd be OK with keeping it, either way I think it would be nice to `await` them in parallel using `Promise.all` (which is currently happening in the `addOrdersIfNotExists` method.

Unless there's another time we'd need to persist multiple orders at once (I can't think of any currently), my preferred approach solution in one line solution and without needing an additional method would be:
```javascript
await Promise.all([this.repository.addOrderIfNotExists(makerOrder), this.repository.addOrderIfNotExists(takerOrder)]);
```

lib/orderbook/OrderBookRepository.ts
```diff
@@ -40,6 +44,24 @@ class OrderbookRepository {
       await this.models.Order.create(order as OrderAttributes);
     }
   }
+
+  /**
+   * Only persist multiple orders if they don't exists.
+   * @param orders array of orders to persist
+   */
+  public addOrdersIfNotExists = async (orders: db.OrderFactory[]) => {
+    const ordersToAdd: Promise<void>[] = [];
```
sangaman: `ordersToAdd` is not  exactly descriptive - we might not be adding anything if the orders already exist for instance. Maybe just call this `orderPromises` or `addOrderPromises`?

PR #663, opened 2018-11-12
refactor(orderbook): reuse executeSwap code

sangaman:
This aims to remove some code duplication and make the `executeSwap` method in OrderBook.ts handle the required logic for when a swap we've initiated fails or succeeds.

### DIFF
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -196,7 +196,17 @@ class Service extends EventEmitter {
     const { orderId, pairId, peerPubKey } = args;
     const quantity = args.quantity > 0 ? args.quantity : undefined; // passing 0 quantity will work fine, but it's prone to bugs
 
-    return this.orderBook.executeSwap(orderId, pairId, peerPubKey, quantity);
+    const maker = this.orderBook.removePeerOrder(orderId, pairId, peerPubKey, quantity).order;
+    const taker = this.orderBook.stampOwnOrder({
+      pairId,
+      localId: '',
+      price: maker.price,
+      isBuy: !maker.isBuy,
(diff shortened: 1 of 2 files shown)


lib/service/Service.ts
```diff
@@ -196,7 +196,17 @@ class Service extends EventEmitter {
     const { orderId, pairId, peerPubKey } = args;
     const quantity = args.quantity > 0 ? args.quantity : undefined; // passing 0 quantity will work fine, but it's prone to bugs
 
-    return this.orderBook.executeSwap(orderId, pairId, peerPubKey, quantity);
+    const maker = this.orderBook.removePeerOrder(orderId, pairId, peerPubKey, quantity).order;
```
erkarl: I don't fully understand why `removePeerOrder` and `stampOwnOrder` was moved to the service level. Can you please elaborate?

sangaman: Basically I tried to make the `executeSwap` method in OrderBook.ts accept only two orders as parameters so it can be called internally as well as from Service.ts. Service.ts gets the orders to pass to `executeSwap` by calling these two methods. If it's not very clear I can always restructure it or add some comments.

erkarl: Fair enough. I think it makes sense this way since `removePeerOrder` and `stampOwnOrder` are public. I actually made `stampOwnOrder` private in #667, but I'll revert the changes.

sangaman:
Rebased to resolve conflicts with latest in master. @moshababo if you get a chance let me know if this change still makes sense to you, I know we'd discussed it earlier as part of a previous PR.

PR #667, opened 2018-11-13
refactor(orderbook): set localId to global id

### LINKED ISSUE #223: set order localId (currently generated separately) = globalId

ghost:
In this PR we:
* modify `Orderbook.stampOwnOrder` to set `localId` to global `id` in case `localId` is not provided
* add integration tests for `Orderbook.stampOwnOrder`

Closes #223.

### DIFF
--- a/test/integration/OrderBook.spec.ts
+++ b/test/integration/OrderBook.spec.ts
@@ -210,6 +210,48 @@ describe('nomatching OrderBook', () => {
     expect(() => orderBook['removeOwnOrder'](remainingOrder!.id, order.pairId, 1)).to.throw;
   });
 
+  describe('stampOwnOrder', () => {
+    const ownOrder = () => {
+      return {
+        pairId: 'LTC/BTC',
+        price: 0.008,
+        quantity: 0.00001,
+        isBuy: false,
+        localId: '',
+        hold: 0,
+      };
+    };
+
+    it('has the same id and localId when localId not provided', async () => {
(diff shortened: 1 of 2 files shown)


test/integration/OrderBook.spec.ts
```diff
@@ -210,6 +210,47 @@ describe('nomatching OrderBook', () => {
...
+        hold: 0,
+      };
+    };
+
+    it('has the same id and localId when localId not provided', async () => {
+      const stampedOrder = orderBook['stampOwnOrder'](ownOrder());
+      expect(stampedOrder.id).to.equal(stampedOrder.localId);
+    });
+
+    it('does not have the same id and localId when localId provided', async () => {
```
sangaman: Rather than simply test that they are different, I think it makes more sense to test that the stamped order's local id is equal to the local id we passed in. We could also have a sanity check in place to make sure they are different, but at that point it would be an issue with how we generate the order id.

sangaman (approved):

test/integration/OrderBook.spec.ts
erkarl: You're right. Updated the spec title and added the assertion.

PR #668, opened 2018-11-13
feat(cli): sort getorders output

### LINKED ISSUE #647: Sort getorders cli output
We should sort the orders in the `getorders` cli output according to price, best bids/asks at the top. When two orders have the same price, the order with the older creation time should be listed first.

ghost:
This PR adds sorting to the CLI `getorders` output:
screen shot 2018-11-13 at 17 50 15

Closes: https://github.com/ExchangeUnion/xud/issues/647


ImmanuelSegol (approved):

sangaman (approved):
:clap::clap::clap:

offerm (approved):
Nice.

Personally I still think that this should be done on the backend since most callers of the RPC would like to get it sorted that way.

Issue #656, opened 2018-11-09
swapResultsList improvements


offerm:
price is missing from the swapResultsList.

Consider also to add the preimage or remove the rhash. 

Should we have role here? isn't it always zero?

```
xud admin$ xucli sell 0.004 LTC/BTC 0.875
{
  "internalMatchesList": [],
  "swapResultsList": [
    {
      "orderId": "d7fa8c31-e42e-11e8-8219-a328c72ea886",
      "localId": "871f9570-e42f-11e8-ba73-5559559634a7",
      "pairId": "LTC/BTC",
      "quantity": 0.0005,
      "rHash": "c3f490a0641aa4d8bd0b456e74d50650fa42182ba4e5ed162304ae1a35ade8c9",
      "amountReceived": 45000,
      "amountSent": 50000,
      "peerPubKey": "03fd337659e99e628d0487e4f87acf93e353db06f754dccc402f2de1b857a319d0",
      "role": 0
    },
    {
      "orderId": "d6a842a1-e42e-11e8-8219-a328c72ea886",
      "localId": "871f9570-e42f-11e8-ba73-5559559634a7",
      "pairId": "LTC/BTC",
      "quantity": 0.001,
      "rHash": "c95ecd85d640879ec1cca3eb447ad13d6bd1651ad2041d2682867af3fc812097",
      "amountReceived": 89000,
      "amountSent": 100000,
      "peerPubKey": "03fd337659e99e628d0487e4f87acf93e353db06f754dccc402f2de1b857a319d0",
      "role": 0
    },
    {
      "orderId": "87a31d51-e42f-11e8-8219-a328c72ea886",
      "localId": "871f9570-e42f-11e8-ba73-5559559634a7",
      "pairId": "LTC/BTC",
      "quantity": 0.001,
      "rHash": "128a1cc547a9b2d7d84cd961e50d6999e7305b3eaa5a8586c0d202a7c925c8cd",
      "amountReceived": 90000,
      "amountSent": 100000,
      "peerPubKey": "03fd337659e99e628d0487e4f87acf93e353db06f754dccc402f2de1b857a319d0",
      "role": 0
    },
    {
      "orderId": "d6a842a1-e42e-11e8-8219-a328c72ea886",
      "localId": "871f9570-e42f-11e8-ba73-5559559634a7",
      "pairId": "LTC/BTC",
      "quantity": 0.0005,
[...]

Issue #609, opened 2018-10-27
Add failed swaps to placeOrder (buy/sell) response


sangaman:
It might be useful to know whether a placed order attempted to swap with any peer orders but failed. Currently `placeOrder` only returns information about successful swaps and internal matches. We could make it also return failed swaps. I figure for each failed swap we'd want to know the `orderId` and maybe `quantity` and `peerPubKey`, maybe the error message for why the swap failed`? Comment if you can think of anything else that might be worth returning.

kilrau:
So instead of only persisting it in the db as per https://github.com/ExchangeUnion/xud/issues/562 you suggest also to return failed swap info in `placeOrder` sync until a swap succeeds, correct?

`orderID`, `pair`, `quantity`, `price`, `peerPubKey`, error message

sangaman:
It would be for both the sync and streaming place order calls, but yeah otherwise that's right. Altho even if a swap never succeeds, we'd still want to return any swaps that failed.

kilrau:
Solves parts of https://github.com/ExchangeUnion/xud/issues/620

kilrau:
Output today:
```
admins-MacBook-Pro-2:xud admin$ xucli buy 0.0025 LTC/BTC 1.105
swapped 0.001 LTC with peer order 62e45ed0-eccb-11e8-a9ee-5d8556a34cdc
remaining 0.0015 LTC entered the order book as 92a14980-eccb-11e8-bb8f-69b162a0c6a5
admins-MacBook-Pro-2:xud admin$ 
```

How it should be:
```
$ xucli buy 0.0025 LTC/BTC 1.105
matched 0.001 LTC with peer order 62e45ed0-eccb-11e8-a9ee-5d8556a34cdc
matched 0.0015 LTC with peer order 87e44ed9-xxxx-xxxx-xxxx-xxxxxxxxxx
swapped 0.001 LTC with peer order 62e45ed0-eccb-11e8-a9ee-5d8556a34cdc
failed to swap 0.0015 LTC with peer order 87e44ed9-xxxx-xxxx-xxxx-xxxxxxxxxx
remaining 0.0015 LTC entered the order book as 92a14980-eccb-11e8-bb8f-69b162a0c6a5
```

PR #669, opened 2018-11-13
feat(cli): format buy & sell output

### LINKED ISSUE #620: Improved buy/sell output for xucli
As per https://github.com/ExchangeUnion/xud/issues/612#issuecomment-434207811 we aim for the following output format:

```
xucli buy 0.0005 LTC/BTC market --stream

market order accepted
matched with peer order : .......
attempting swap: .....
swap failed: .....
peer order removed from book: .....
attempting swap with: .....
swap failed: .....
peer order removed from book: .....
attempting swap with: .....
swap failed: .....
no more matches found
quitting.
```

Successful case:
```
xucli buy 0.0005 LTC/BTC market --stream

market order accepted
matched with peer order : .......
attempting swap: .....
swap succeeded: .....
done.
```

sangaman:
This attempts to make the cli output for the `buy` and `sell` commands more human-readable and succint.

Closes #620.

The missing features that will need to be added in separate PRs are adding `price` to the swap result (#656) and adding an event for failed swaps to the streaming `placeOrder` rpc call (#609).

### DIFF
--- a/lib/cli/commands/getorders.ts
+++ b/lib/cli/commands/getorders.ts
@@ -4,7 +4,7 @@ import { GetOrdersRequest, GetOrdersResponse, Order, OrderSide } from '../../pro
 import Table, { HorizontalTable } from 'cli-table3';
 import colors from 'colors/safe';
 
-type FormatedTradingPairOrders = {
+type FormattedTradingPairOrders = {
   pairId: string,
   orders: string[][],
 };
@@ -39,7 +39,7 @@ const addSide = (orderSide: Order.AsObject[]): string[] => {
 };
 
 export const formatOrders = (orders: GetOrdersResponse.AsObject) => {
-  const formatedOrders: FormatedTradingPairOrders[] = [];
+  const formattedOrders: FormattedTradingPairOrders[] = [];
   orders.ordersMap.forEach((tradingPair) => {
     const buy = sortOrders(tradingPair[1].buyOrdersList, true);
     const sell = sortOrders(tradingPair[1].sellOrdersList, false);
@@ -49,12 +49,12 @@ export const formatOrders = (orders: GetOrdersResponse.AsObject) => {
       .map(() => {
         return addSide(buy).concat(addSide(sell));
       });
-    formatedOrders.push({
+    formattedOrders.push({
       pairId: tradingPair[0],
       orders: tradingPairOrders,
     });
   });
-  return formatedOrders;
+  return formattedOrders;
 };
 
 const createTable = () => {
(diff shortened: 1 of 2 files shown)


kilrau:
Looks much better!

How about:
matched 1 LTC @ price 1 with own order da8d3401-e75a-11e8-972d-a9036be5a943

Why this?
new order 8a22d201-e75e-11e8-972d-a9036be5a943 entered the order book with 5 quantity

don't see a need for this in the buy/sell output. Just matches.

sangaman:
> How about:
> matched 1 LTC @ price 1 with own order da8d3401-e75a-11e8-972d-a9036be5a943

I thought about explicitly stating "price" too but I felt it was a little superfluous. It seems clear enough to me and also I know the "traded [quantity] @ [price]" shorthand is used elsewhere. Still it's close so maybe someone else can weigh in on which way looks better?

> Why this?
> new order 8a22d201-e75e-11e8-972d-a9036be5a943 entered the order book with 5 quantity
> 
> don't see a need for this in the buy/sell output. Just matches.

To me it's an easy way to tell which portion of your order was unmatched and is being added to the orderbook. Also I figure we'd want something like that, otherwise there'd be no output in the case where we place a limit order that doesn't get matched. Others can weigh in on this too tho.

kilrau:
> Edit: Also I realized I'd posted an older example before I'd changed the formatting to use the base currency ("LTC" in this case) instead of "quantity". So if that's what you were suggesting as well I'd already made that change.

👍 that was the main feedback, yes (sry, didn't check the code). Leaving out price is fine.

> > Why this?
> > new order 8a22d201-e75e-11e8-972d-a9036be5a943 entered the order book with 5 quantity
> > don't see a need for this in the buy/sell output. Just matches.
> 
> To me it's an easy way to tell which portion of your order was unmatched and is being added to the orderbook. Also I figure we'd want something like that, otherwise there'd be no output in the case where we place a limit order that doesn't get matched. Others can weigh in on this too tho.

-> Oh then it's just confusing. To me this looks like some new peerOrder came in and the CLI reported about it which doesn't make much sense. Suggest to restructure to sth like this:

`remaining quantity 5 LTC entered order book as new order 8a22d201-e75e-11e8-972d-a9036be5a943`

sangaman:
Ok yeah I see what you mean. Basically I was trying to think about how it would look if there were no matches at all, where I think it's a bit confusing to say "no more matches found." Maybe just `5 unmatched quantity entered order book as 8a22d201-e75e-11e8-972d-a9036be5a943`? @kilrau

kilrau:
I'd prefer to stick with '"remaining" instead of "unmatched" since we use this elsewhere to, but your call. And definitely add "LTC", not only "quantity". Also, if there were no matches found we should say "No matches found".

`No matches found`
`5 LTC remaining quantity entered order book as 8a22d201-e75e-11e8-972d-a9036be5a943`

sangaman:
OK thanks, I'm going to go with @erkarl's suggestion.

PR #672, opened 2018-11-14
perf: use async fs & crypto methods

sangaman:
This replaces all synchronous calls to the NodeJS `fs` and `crypto.randomBytes` methods with asynchronous calls. This improves performance by preventing blocking of the event loop. I was inspired after reading [Don't Block the Event Loop](https://nodejs.org/en/docs/guides/dont-block-the-event-loop/). Most of the benefit is during startup/initialization, but there should also be a small benefit when proposing swaps which should happen quite frequently. 

I also added a test suite for NodeKey.ts, I was primarily motivated by having a quick way to test that the async `crypto.randomBytes` call was working but it would also be good expand in the future.

### DIFF
--- a/lib/utils/fsUtils.ts
+++ b/lib/utils/fsUtils.ts
@@ -1,0 +1,20 @@
+import fs from 'fs';
+import { promisify } from 'util';
+
+// Promisified wrappers for file system methods - only required for support for NodeJS v9 and older
+// NodeJS v10 provides promisified file system methods out of the box
+
+/** A promisified wrapper for the NodeJS `fs.readFile` method. */
+export const readFile = promisify(fs.readFile);
+
+/** A promisified wrapper for the NodeJS `fs.mkdir` method. */
+export const mkdir = promisify(fs.mkdir);
+
+/** A promisified wrapper for the NodeJS `fs.exists` method. */
+export const exists = promisify(fs.exists);
+
+/** A promisified wrapper for the NodeJS `fs.readdir` method. */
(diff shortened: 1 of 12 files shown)


moshababo:
There's definitely no excuse for blocking the thread on Nodejs, since it was created for async I/O, and so all the libs support that. I think we introduced some sync version usages because we needed it in the constructor, but this should be avoided as well.

Looks great!

lib/utils/fsUtils.ts
```diff
@@ -0,0 +1,20 @@
+import fs from 'fs';
+import { promisify } from 'util';
+
+// Promisified wrappers for file system methods - only required for support for NodeJS v9 and older
+// NodeJS v10 provides promisified file system methods out of the box
```
erkarl: We could switch to `>= 10.13.0 LTS` and not have these wrappers.  Is there a good reason to support lower versions?

sangaman: We could, I just wasn't sure whether we should drop support for it yet. For now, I think supporting Node 8 makes sense because it is still "Active" according to the [release chart](https://github.com/nodejs/Release). Then again, by the time xud is production-ready I'd expect that Node 8 will be on its way out. Longer term I think it makes sense to try to support every version that's still being maintained.

Thoughts @moshababo?

moshababo: Yes, i'd say let's not bound ourself yet to support only v10.

erkarl: Sounds good. :).

Issue #652, opened 2018-11-09
DB migration script needed


offerm:
listen ports didn't open.
got this error in log

```
11/9/2018, 8:01:26 AM [GLOBAL] info: config loaded
11/9/2018, 8:01:26 AM [GLOBAL] info: Local nodePubKey is 02db8d5f43940fa1632ad2e2c0761b608e54fcca9c7502f961d17e290e66ad99a6
11/9/2018, 8:01:26 AM [DB] info: connected to database xud.db
11/9/2018, 8:01:26 AM [LND] info: macaroons are disabled for lnd at localhost:10002
11/9/2018, 8:01:26 AM [LND] info: LndClient status: Disconnected
11/9/2018, 8:01:26 AM [LND] info: trying to verify connection to lnd with uri: localhost:10002
11/9/2018, 8:01:26 AM [LND] info: macaroons are disabled for lnd at localhost:10001
11/9/2018, 8:01:26 AM [LND] info: LndClient status: Disconnected
11/9/2018, 8:01:26 AM [LND] info: trying to verify connection to lnd with uri: localhost:10001
11/9/2018, 8:01:26 AM [GLOBAL] error: SequelizeDatabaseError: SQLITE_ERROR: no such column: rHash
    at Query.formatError (/Users/admin/xud-simnet/xud/node_modules/sequelize/lib/dialects/sqlite/query.js:423:16)
    at afterExecute (/Users/admin/xud-simnet/xud/node_modules/sequelize/lib/dialects/sqlite/query.js:119:32)
    at replacement (/Users/admin/xud-simnet/xud/node_modules/sqlite3/lib/trace.js:19:31)
    at Statement.errBack (/Users/admin/xud-simnet/xud/node_modules/sqlite3/lib/sqlite3.js:16:21)
11/9/2018, 8:01:26 AM [LND] info: LndClient status: ConnectionVerified
11/9/2018, 8:01:26 AM [LND] info: LndClient status: ConnectionVerified
```

Impact: all sim-net nodes are down.

erkarl:
Deleting the database `xud.db` worked for me. There was a breaking database change without migration.

offerm:
Even that is a lot of work. I have at least 5 machines where this should be done.
Also, I'm not sure what the impact of this is on the cloud servers. I don't want to configure the sim-net again.

erkarl:
I think we should introduce migrations in the near future to prevent situations like these.

offerm:
[...]

PR #682, opened 2018-11-17
Persist trades from maker order swaps

### LINKED ISSUE #608: Save all completed trades to database

sangaman:
This is a followup to #659, which persisted trades between own orders and a taker own order vs a maker peer order. This saves a trade when one of (the `swap.paid` event) of our orders get filled by a peer taker order.

I needed to make some tweaks to the `tables` db schema, which means a potentially breaking change. However, upgrading existing databases is simple, just need to do `Trade.sync({ alter: true });`. So I'm not sure on the best approach here, whether to create an npm script that does that or whether to put some temporary into xud to do this. Let me know what you think.

I also added test cases for persisting trades that result from a swap, and went ahead with what was discussed previously and moved the helper functions that create order to the test utils methods.

Closes #608.


moshababo:
> I needed to make some tweaks to the tables db schema, which means a potentially breaking change. However, upgrading existing databases is simple, just need to do Trade.sync({ alter: true });. So I'm not sure on the best approach here, whether to create an npm script that does that or whether to put some temporary into xud to do this. Let me know what you think.

What you're suggesting is a simple form of a migration script. I'd say not to introduce it without the proper infrastructure (which anyone could add more migration scripts into), nor to put it temporarily in `xud`. as discussed #652 - we should wait with the migration scripts as much as possible, and in the meanwhile, make it clear that users need to dump their database file.

sangaman:
OK that's my preference as well. I'll probably wait to merge this until the next release is tagged and advise that databases should be wiped.

kilrau:
Can be merged now? @sangaman

sangaman:
When #773 is ready to merge (hopefully soon) I'll resolve the conflicts here and then we can merge.

PR #695, opened 2018-11-20
feat(p2p): reconnect to peer after disconnection

### LINKED ISSUE #616: not connected to peer after disconnect

### LINKED ISSUE #698: connection retries not working
Looks like we have a problem with the connection retries - 
When connecting to a node, we try various addresses. On the first address, the node **expected** public key is being added to `pendingOutgoingConnections`. if connection fails, it should be removed from there, but the current code is doing it only if handshake has completed. If not, we don't have the public key, but we should use the expected one instead.

moshababo:
Closes #616, Closes #698.

If peer got disconnected from us post-handshake due to stalling or without specifying any reason, reconnect to him in case we have a listening address of him.


lib/p2p/Peer.ts
```diff
@@ -187,7 +189,7 @@ class Peer extends EventEmitter {
   /**
    * Close a peer by ensuring the socket is destroyed and terminating all timers.
    */
-  public close = (reason?: DisconnectingPacketBody): void => {
+  public close = (info?: DisconnectingPacketBody): void => {
```
sangaman: `info` is somewhat ambiguous here. Maybe instead of taking a `DisconnectingPacketBody` as a parameter, it can take `reason: DisconnectionReason` and `payload?: string`. I think that's a bit more clear and also more flexible in case we change what fields are used for the `DisconnectingPacketBody` type. If that makes sense to you then I think we'd want a similar method signature for `closePeer` below.

lib/p2p/Pool.ts
```diff
@@ -647,12 +642,26 @@ class Pool extends EventEmitter {
...
         this.peers.remove(peer.nodePubKey);
       }
       this.emit('peer.close', peer);
+
+      // if handshake passed and peer disconnected from us for stalling or without specifying any reason -
+      // reconnect, for that might have been due to a temporary loss in connectivity
+      const legitReconnect =
+        peer.sentDisconnectionReason === undefined &&
+        (peer.recvDisconnectionReason === undefined || peer.recvDisconnectionReason === DisconnectionReason.ResponseStalling);
+      const addresses = peer.inbound ? peer.addresses : [peer.address];
```
sangaman: This looks like we won't attempt to reconnect using the peer's listening addresses if it's an outbound connection. Is that intentional?

If not, why not always use `peer.addresses` but additionally use `peer.address` as the `lastAddress` if the peer is outbound? So like make this line `const lastAddress = peer.inbound ? undefined : peer.address;`, make it `(addresses.length || lastAddress)` in the if statement below, and `const node = { addresses, lastAddress, nodePubKey: peer.nodePubKey };`.

lib/p2p/Pool.ts
```diff
@@ -647,12 +642,26 @@ class Pool extends EventEmitter {
...
+      // cleanup
       if (peer.nodePubKey) {
         this.pendingOutgoingConnections.delete(peer.nodePubKey);
         this.peers.remove(peer.nodePubKey);
       }
       this.emit('peer.close', peer);
+
+      // if handshake passed and peer disconnected from us for stalling or without specifying any reason -
+      // reconnect, for that might have been due to a temporary loss in connectivity
+      const legitReconnect =
```
sangaman: `legitReconnect` seems a bit confusing, how about `shouldReconnect`?

offerm:
Use cases that should work. For all of the A connects to B in advance. 

1. Disconnect wifi/network cable between the A and B
2. B shutdown 
3. B was killed (kill -9)

 result: A should try to reconnect to B until B is back.

currently:
1 - no reconnect at all
2,3 - one time reconnect attempt.

moshababo:
@offerm `2,3` is due to a previous bug in the connection retry logic, which introduced with `pendingOutgoingConnections`. not related to this task, so I opened a separate issue (#698).

In regards to `1` - if a node goes offline, it will perceive the other node as stalling, and will disconnect from it. the other node should reconnect when it will hear about it. 
but if the other node is offline as well, or if it's online but it will perceive the offline node as stalling as well, before it will hear about the disconnection, so neither will try to reconnect. 
Looks like it's the edge case of what defined in #616. We might need to try to reconnect from the disconnecting side as well. @kilrau

offerm:
@moshe,  Not sure why not to fix all related issues together. For 2,3 the fact that you try once does not give any value since the peer is already down. The impact is as if we are doing nothing. IMHO, just fix it.

`1` is not an edge case. I'm not sure how you define "offline". If A and B are connected and A is "offline" B is also "offline" with regards to A.
What I fail to understand is why handling is different is you `hear about disconnection` ot just getting disconnected.  In both cases, if you need to be connected, reconnect.

I also don't understand the definition of `disconnecting side`. There are incoming and outgoing connections and if an outgoing connection dropped we need to reconnect.

moshababo:
Added the connection retries fix here, so you can check it on this branch.

`disconnecting side` is the side which initiated the disconnection by actively closing the socket. You're right that from each node perspective, if one if offline, the other is as well, so both will perceive message stalling. 

The debate here whether to reconnect after we actively closed the socket, or just to relax the stalling penalty, and keep the socket to allow recovery. If we chose the former (disconnect and then reconnect), we just need to remove `Pool.ts` line 657.

moshe:
@offerm you are consistently mentioning the wrong Moshe (:
you should mention moshababo instead of me

moshababo:
> Are we closing the connection beside when we shutdown? what is the most common scenario?

we close the connection if the peer is stalling response to `Ping`.

> If I stop B with xucli shutdown A wouldn't notice that at all and will not try to reconnect (B is still running)

If B had a shutdown, how come he's still running? And if A won't notice that because there's no connection between them, A should disconnect from B because B would be stalling `Ping` response.

> If A is trying to reconnect B after 5,10,20,40,80 seconds and during the wait time B connected to A, the connection will be closed.

The new connection from the retries will be closed if a connection with the public key already exists. Is that a problem? We can add some logic to cut off the retry timeout before attempting a connection, but I don't think it's a big deal.

offerm:
@moshababo:

I have no idea why B is still running after `xucli shutdown`. I opened an issue for this #699 . For some reason the process is still running and the connection to A is still working normally so A don't see B as stalling.

If A should connect to B, A should close current socket and reconnect  B after any of the following:
- B stopped with `xucli shutdown`
- B stopped with Kill
- B stopped with Kill -9
- B stopped for any reason
- B is not responding to  ping
- A has a connectivity problem 
- The IP route between A and B is not working for any reason. 

A should also attempt to connect in a loop (1,5,10,20,40,80 ,etc) if the initial connection to B failed. 

according to my current understanding, the only reason not to do that is if B is banned or if B is not working according to the XUD protocol. Communication failure as part of the handshake should also trigger reconnect.

Please let me know if you don't agree to the above so we can discuss it.
(I will update about the second problem).

lib/p2p/Pool.ts
sangaman: I see what you're saying. It's more like whether this disconnection happened deliberately, or due to temporary network issues. Maybe `unintentionalDisconnect` or `unintentionalClose` might be appropriate. But it's clear enough given the comments so I'm fine with whatever you prefer.

offerm:
If we deploy like its now, the test servers will not connect each other.
I don't think we should merge this before it is working well.
A node that is shutdown will connect to us only if it can do so. If we are
behind Firewall, it will not be able to. It should also need to know how.
This is elementary P2P, why not doing it right?

Take example from lnd please.

moshababo:
@offerm i'm not following the fixes you're requesting. 

From your logs describing the problem, what I can see is that while node A was doing connection retries, node B connected to A. So when A also connected to B, this connection was dropped, because it was duplicated. 

It's what I described here:

> The new connection from the retries will be closed if a connection with the public key already exists. Is that a problem? We can add some logic to cut off the retry timeout before attempting a connection, but I don't think it's a big deal.

If i'm missing something, let me know. But anyway this is not related to reconnecting after disconnection, but to the connection retries. 

So the only problem that I can see now is that if both nodes will experience stalling from their perspective (due to loss of connectivity), they will both close the socket and neither will reconnect. We can fix this easily, but I still not sure that this is the problem you described earlier.

sangaman:
> If we deploy like its now, the test servers will not connect each other.

I'm not following, is that due to something introduced in this PR? If not it shouldn't prevent this from being merged.

> If we are behind Firewall, it will not be able to.

That's why I suggested to reconnect after `Shutdown` if we are not listening for incoming connections. If we are listening for incoming connections, we should be reachable (otherwise we have other problems). But I'd be OK with always reconnecting after `Shutdown` too if that's what you prefer.

> The problem is that the end result is that there is no connection between A and B.  When A try to reconnect with already connected B it will close the connection and A and B will not be connected.

This sounds like a bug if so. Maybe we need to be checking if we're already connected to a peer before we retry a connection. I can open a PR for this for the existing logic, and maybe we can add that to the reconnection logic here as well.

sangaman:
I actually can't find any mechanism by which reconnecting will close an existing connection - and it's not so simple to check before retrying a connection if we already have a `Peer` with a duplicate pubkey because each peer does not know about other peers, but we do check for duplicate pubkeys in `handleOpen` in Pool.ts. So this may require further investigation.

offerm:
Yes, problem created by this PR.

Consider A and B. Both try to connect on startup. This is the situation
with the test servers. Test1 attempts to connect to test2, fail, goes to
sleep until next retry. While sleeping test2 starts and connects to test1.
When test1 wakes up and try to connect test2 rejects and close the
connection. Now they are not connected and no retry takes place.

moshababo:
@offerm this was an old bug with the connection retries. I missed it because I checked that the connection exists, which it did, but the peer was removed from the peer list.

I just pushed a fix for it. 

Let me know if there's something missing for the reconnecting after disconnection feature.

offerm:
looks like the problem is due to **pool.ts:665**. please check

also, I see that you only reconnect if there is no disconnect reason or reason is ResponseStalling. Why not reconnect for `NotAcceptingConnections` and `shutdown` ?
IMHO we should also do it for ALL the other reasons. for example, if A should connect to B and B disconnect with `IncompatibleProtocolVersion`, B should probably be updated. We should keep trying to connect until success. Same with Banned - at some point we will get out of this situation, right? at that point we should reconnect.

What I suggested in the past is that we will reset the reconnect timer only after successful reconnect. This way, if we got Banned we will try to reconnect with a growing gap between each attempt.

Still pending the case when A should connect to B and B should connect to A and B is killed and restarted.

kilrau:
Please update us on the status here @moshababo . We ran into the same issue while testing here on our unstable Chinese connections and it makes testing real difficult.

moshababo:
I didn't fix this yet on purpose so we can isolate this for the other problems. See my comment [here](https://github.com/ExchangeUnion/xud/pull/695#issuecomment-440714647).

@offerm you say you want to reconnect on all cases, but I don't understand why we'll want to have a connect-disconnect loop, which will happen if we're banned or with incompatible version.
We don't have a timer for the reconnecting - it's only for the connection retries, when the target is offline. If you want to add a support for that, we can open a separate issue. 

I removed the line which disables reconnecting from the side which closed the socket. Please try again.

offerm:
Better now. Reconnect after wifi error is working [again]

Also better on server side.

Still not perfect:
- A will not reconnect to B when B is down with `xucli shutdown` 
- If A and B are connected (both can initiate the connection) and you take B down (kill), start B, wait a minute, Kill again ----> A will not attempt reconnect (reason is not cleared).
- The reconnection process will stop after 168 hours for some reason.
- Max time between reconnections should be (IMHO) 5 minutes 

The reason why I think we should always reconnect is to avoid problems like the one above. If we keep it simple it will work better. We should reset the counter only when we made a good connection. So if there is a peer, for which we initiate the connection, and is not properly setup, we will try to connect every 5 minutes. Don't see a problem with this.

I also think we should reconnect only for outgoing peers. No need to reconnect incoming peers.

kilrau:
Ok, this is taking too long. Offer made some good points, I made an "improvement" offshoot: https://github.com/ExchangeUnion/xud/issues/721

But since this improves the situation quite a bit and we need it urgently for our testing, I'd prefer to get it merged soon.

moshababo:
@offerm let's finalize this for this milestone. I returned the check on `sentDisconnectionReason` because it doesn't make to reconnect on **any** case (and it's also failing our tests). I couldn't think on any reasonable case besides `ResponseStalling`, but we can discuss this later. This is not related to `recvDisconnectionReason`, which is a different story.

offerm:
@moshababo which tests scenarios are covered by this PR?

wifi disconnect?
peer killed?
peer shutdown?

moshababo:
> If A and B are connected (both can initiate the connection) and you take B down (kill), start B, wait a minute, Kill again ----> A will not attempt reconnect (reason is not cleared).

Saw the problem now. I'll fix it for #721 because it could depend on other things that we might want to do there.

@sangaman I made some changes to `Pool.spec.ts`.
First was to remove `async` from the main tests wrapper - it was causing some problems with logging the results.
Second was commenting out the last test. It was causing the process to hang because of the reconnecting. Specifying the disconnection reason wasn't efficient, because there was no real connection there. There are a few options of how to fix this, but I wanted to let you review it before i'll change stuff.

offerm:
We can't merge it like this as it will cause the test servers not to be connected and maybe some other effects.

offerm:
I have checked the simnet environment and I'm OK with merge.
We will have problems once xucli shutdown works but we can fix as part of #721

offerm:
@moshababo I'm not sure I follow your fix.

Here is what should work:
A connects to B
B is killed
A tries to reconnect to B
B starts
A connect

What should not work:
A connects to B
A is killed
B tries to reconnect to A

Try to describe the change in this language so it will be easier to understand. 

Let us also know when tests have you done.

moshababo:
> A connects to B
A is killed
B tries to reconnect to A

You said this should not happen, but at the moment B will reconnect to A if A advertised addresses. We didn't limit reconnecting to outbound connections yet.
My fix was for this scenario - if A advertised addresses and B will reconnect, then he will also do connection retries on A first advertised address. I'm not sure that this was the problem you described, but I didn't found any other issue.

> 1/ A connects to B
2/ B is killed
3/ A tries to reconnect to B
4/ B starts
5/ A connect

This works, but mind that after step 4, B might be the one that will connect to A (and not vise versa), in case he had A on his nodes list. So you'll end up with A having B as an inbound.

PR #697, opened 2018-11-21
feat(orderbook): don’t match with quantity on hold

### LINKED ISSUE #666: Matching mode: check "hold" on internal match
As discussed in Slack, this issue is about implementing a check if an `ownOrder` is on hold when matching it with another `ownOrder`.

moshababo:
Closes #666 

If maker own order has quantity on hold, don't match it, but keep the order in the queue for the next match.

### DIFF
--- a/lib/orderbook/TradingPair.ts
+++ b/lib/orderbook/TradingPair.ts
@@ -342,7 +342,9 @@ class TradingPair {
           map.delete(makerOrder.id);
           this.logger.debug(`removed order ${makerOrder.id} while matching order ${takerOrder.id}`);
         } else if (makerAvailableQuantityFullyMatched) {
+          // only an own order can be fully matched for available quantity, but not fully matched in the overall
           assert(isOwnOrder(makerOrder));
+
           assert(queue.poll() === makerOrder);
           queueRemovedOrdersWithHold.push(makerOrder as OwnOrder);
         }
--- a/lib/orderbook/TradingPair.ts
+++ b/lib/orderbook/TradingPair.ts
@@ -292,6 +292,7 @@ class TradingPair {
     let remainingOrder: OwnOrder | undefined = { ...takerOrder };
 
     const queue = takerOrder.isBuy ? this.queues!.sell : this.queues!.buy;
+    const queueRemovedOrdersWithHold: OwnOrder[] = [];
     const getMatchingQuantity = (remainingOrder: OwnOrder, oppositeOrder: Order) => takerOrder.isBuy
       ? TradingPair.getMatchingQuantity(remainingOrder, oppositeOrder)
       : TradingPair.getMatchingQuantity(oppositeOrder, remainingOrder);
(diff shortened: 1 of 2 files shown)


lib/orderbook/TradingPair.ts
```diff
@@ -300,13 +301,18 @@ class TradingPair {
     while (remainingOrder && !queue.isEmpty()) {
       // get the best available maker order from the top of the queue
       const makerOrder = queue.peek()!;
-      const matchingQuantity = getMatchingQuantity(remainingOrder, makerOrder);
+      const makerAvailableQuantityOrder = isOwnOrder(makerOrder)
+        ? { ...makerOrder, quantity: makerOrder.quantity - makerOrder.hold, hold: 0 }
+        : { ...makerOrder };
```
sangaman: Can we just set this equal to `makerOrder` instead of a copy of `makerOrder`? It seems like shouldn't be an issue, and it should make this slightly more performant.

lib/orderbook/TradingPair.ts
```diff
@@ -317,27 +323,36 @@ class TradingPair {
           const matchedMakerOrder = TradingPair.splitOrderByQuantity(makerOrder, matchingQuantity);
           this.logger.debug(`reduced order ${makerOrder.id} by ${matchingQuantity} quantity while matching order ${takerOrder.id}`);
           matches.push({ maker: matchedMakerOrder, taker: remainingOrder });
-        } else if (makerFullyMatched) {
+        } else if (makerFullyMatched || makerAvailableQuantityFullyMatched) {
```
sangaman: I think this can just check `makerAvailableQuantityFullyMatched` because I believe it will be true any time `makerFullyMatched` is true.

lib/orderbook/TradingPair.ts
```diff
@@ -317,27 +323,36 @@ class TradingPair {
...
         }
+
         if (makerFullyMatched) {
           // maker order is fully matched, so remove it from the queue and map
           assert(queue.poll() === makerOrder);
           const map = this.getOrderMap(makerOrder)!;
           map.delete(makerOrder.id);
           this.logger.debug(`removed order ${makerOrder.id} while matching order ${takerOrder.id}`);
+        } else if (makerAvailableQuantityFullyMatched) {
+          assert(isOwnOrder(makerOrder));
```
sangaman: What's the point of this assert? Isn't it possible that the maker order is a peer order, but still is fully matched?

lib/orderbook/TradingPair.ts
moshababo: I started with it to be consistent with having an immutable object on both cases, but in the end we're only mutating `makerOrder` and not `makerAvailableQuantityOrder`, so it can indeed be removed.

lib/orderbook/TradingPair.ts
moshababo: only ownOrder can be `makerAvailableQuantityFullyMatched` but not `makerFullyMatched`. So it's just for checking our own code.

sangaman: Oh right, I overlooked that this `else if` block will be skipped if `makerFullyMatched` is true. You could maybe add a comment here explaining the assert, but otherwise this makes sense.

PR #709, opened 2018-11-27
refactor: change getOrders to ListOrders

### LINKED ISSUE #681: gRPC+CLI renaming "list"/"get" cleanup
This issue is about renaming `list`/`get` calls in a consistent manner. I know we all have our feelings about this and especially all the renaming. Brought that up because I believe it's still fixable now. Please leave comments.

Motivation: Let's analyze it from a new dev/CLI user perspective: it looks inconsistent how we use `get` and `list` e.g. `getOrders` and `listTrades`. Especially on CLI usage one tends to pick the wrong one.

Option 1: All `list` -> `get`. E.g. `listPeers`, `listPairs` to `getPeers`, `getPairs`. I'd go for `get` since it's the more generic form and will always make sense for a dev implementing the gRPC. IMO the CLI just represents data it gets from the gRPC/xud in the most suitable form. That might be a list, a table, a JSON or a circle.
Good: consistent
Bad: differs from `lnd`, renaming effort

Option 2: The `lnd` way: `get` only for info calls (`getInfo`), rest is consistently `list`.
Good: follows `lnd`
Bad: not consistent

ImmanuelSegol:
closes #681


lib/cli/commands/listorders.ts
```diff
@@ -84,7 +84,7 @@ const displayOrdersTable = (tradingPair: FormattedTradingPairOrders) => {
   console.log(table.toString());
 };
 
-const displayTables = (orders: GetOrdersResponse.AsObject) => {
+const displayTables = (orders: ListOrdersResponse.AsObject) => {
   formatOrders(orders).forEach(displayOrdersTable);
 };
```
sangaman: The command below should say `listorders`

sangaman:
@ImmanuelSegol One more thing, when I pull down this branch and run `npm run proto` I get a number of changes to the xudrpc type declarations. Could you try running `npm install` (and maybe `npm rebuild` for good measure) and then `npm run proto` again and see if you get those changes on your end as well? If so, check those in and then this should be good to merge I think.

sangaman:
@ImmanuelSegol Could you take a look at this conflict as well? You should only have to run the proto scripts again I believe, including for the p2p proto.

sangaman:
@ImmanuelSegol Let me know if you can handle the conflict soon so we can merge before the next release, I'll handle it myself if you're busy.

PR #710, opened 2018-11-28
feat(orderbook/swaps): accept partial swaps

### LINKED ISSUE #550: Partial swap acceptance

sangaman:
This commit adds logic that will:

1. Accept a swap request for a reduced amount when the full amount is not available due to part of the order being on hold.
2. Retry the portion that was not accepted when matching a newly placed order and having a swap request only partially accepted.

Closes #550.

I started trying to write a test case for this but it was too complex, this would be a good case to cover once the simulation tests are up and running.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -289,8 +289,8 @@ class OrderBook extends EventEmitter {
       remainingOrder: matchingResult.remainingOrder,
     };
 
-    // instantiate a container for failed swaps, for retry purposes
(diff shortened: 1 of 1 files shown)


sangaman:
I'll fix conflicts, do some quick manual tests, and then merge.

Issue #341, opened 2018-08-14
Sign & verify `nodePubKey` on handshake


kilrau:
https://github.com/ExchangeUnion/xud/issues/185 exchanges an unsigned `nodePubKey` for now, this task is about signing the `nodePubKey` packet on handshake (sender) and verifying it (receiver). If verification fails, disconnect and print according error message.

sangaman:
This is related to #158, but I do think that virtually any approach should include signing and verifying the handshake so this will be a good first step.

kilrau:
This is (like) a prerequisite for https://github.com/ExchangeUnion/xud/issues/158

kilrau:
Moved to next milestone, not focus for now

kilrau:
As discussed in the weekly call, our goal is full encryption. Hence signing and verifying the handshake is a first step, post-handshake encryption covers both - authenticity & encryption of each packet.

kilrau:
Yes, go ahead we definitely need to sign in addition to encrypting @moshababo

Thinking more about this, we might extend this issue (can do a separate issue) to sign *all* p2p packets to prevent man in the middle attacks which somehow manage to hijack a socket. What do you think? @moshababo @sangaman @offerm 

For [encryption](https://github.com/ExchangeUnion/xud/issues/158), I vouch for signing first and encrypt the whole packet after.

moshababo:
> For encryption, I vouch for signing first and encrypt the whole packet after.

Signing is in the messaging protocol level, so yes it should be done before the encryption (which is part of the wire protocol). 
I thought of implementing the encryption similarly to [BIP 151](https://github.com/bitcoin/bips/blob/master/bip-0151.mediawiki). Let me know what you think.

> Thinking more about this, we might extend this issue (can do a separate issue) to sign all p2p packets to prevent man in the middle attacks which somehow manage to hijack a socket. What do you think? @moshababo @sangaman @offerm

I'm not sure it's necessary if both sides are authenticated and the channel is encrypted. I'd say it's needed only if we plan to relay messages, but not otherwise. Tell me if i'm missing something.

In regards to how to perform the actual authentication, so both sides identity will be proven:
The proof should be non re-usable, so it should be applied on session-specific data. The handshake data doesn't contain at the moment anything about the target peer. 

[...]

PR #712, opened 2018-11-28
feat(p2p): protobuf serialization/validation

### LINKED ISSUE #159: Packets serialization
We're currently using JSON string format for all packets serialization. We might want to introduce a pre-defined structure serialization at some point.

moshababo:
Closes #159.

Still some more cleanup & testing left to do, but it's basically done. 

Main points:
* p2p serialization changed from JSON to protobuf.
* type and length are written to the buffer before the actual packet. parser doesn't rely on the delimiter string anymore. I didn't thought this approach is reasonable anymore, when we don't have a string-based serialization.
* strong validation is applied on the packets. This could have been done in a higher-layer, but I thought that enforcing it on the packet level would be the better.  Tests were added.

Open:
* packet hash is still applied to the JSON representation. We might want to change that.
* I haven't optimized the packet classes yet, just went for the simplest solution (static methods on the derived classes). Seems to be working fine.
* need to improve the parser error handling


proto/xudp2p.proto
```diff
@@ -0,0 +1,116 @@
+syntax = "proto3";
+
+import "annotations.proto";
```
sangaman: Is this line actually needed? I thought it was just for the http metadata, which it doesn't seem like we are using here.

proto/xudp2p.proto
```diff
@@ -0,0 +1,116 @@
...
+
+package xudp2p;
+
+message Header {
+    string id = 1 [json_name = "id"];
+    string reqId = 2 [json_name = "req_id"];
```
sangaman: I noticed that in the `validate` methods, we are checking to make sure that `reqId` and `hash` are undefined for packets that don't use them. It seems like it would be better to only use those fields on the message definitions for packets that actually use them, therefore we don't need to explicitly validate in our code. So instead of having a `Header` with all three fields like we did in the prior JSON implementation, we just put these fields directly on messages that actually use them. For example, `GetNodesPacket` would have `id` as its only field. But `NodesPacket` would have `id`, `reqId`, and `hash` along with the actual nodes.

lib/p2p/packets/PacketType.ts
```diff
@@ -1,3 +1,21 @@
+enum PacketType {
+  Hello = 0,
```
sangaman: I think it would make sense to explicitly number each packet instead of just the first. That way if we ever delete a packet type, the numbering of the other packets won't change.

lib/p2p/packets/PacketType.ts
```diff
@@ -14,5 +32,6 @@ enum PacketType {
   SwapComplete = 'SWAP_COMPLETE',
   SwapError = 'SWAP_ERROR',
 }
+*/
```
sangaman: Can delete the old enum too rather than just comment out.

sangaman: I figure we can delete the old enum too rather than just comment it out.

lib/p2p/packets/types/DisconnectingPacket.ts
```diff
@@ -15,6 +18,46 @@ class DisconnectingPacket extends Packet<DisconnectingPacketBody> {
...
     return PacketDirection.Unilateral;
   }
+
+  public static deserialize(binary: Uint8Array): DisconnectingPacket | undefined {
+    const msg = pb.DisconnectingPacket.deserializeBinary(binary).toObject();
+    return DisconnectingPacket.validate(msg) ? DisconnectingPacket.convert(msg) : undefined;
+  }
+
+  private static validate(msg: pb.DisconnectingPacket.AsObject): boolean {
+    return !!(msg.header
```
sangaman: For these, I don't think we need the `!!` when we're using the `&&` operators already. It doesn't hurt to have it, but it seems a bit cleaner without.

lib/p2p/packets/types/DisconnectingPacket.ts
```diff
@@ -15,6 +18,46 @@ class DisconnectingPacket extends Packet<DisconnectingPacketBody> {
   public get direction() {
     return PacketDirection.Unilateral;
   }
+
+  public static deserialize(binary: Uint8Array): DisconnectingPacket | undefined {
```
sangaman: The `deserialize` and `validate` methods aren't using the arrow methods we use elsewhere. It doesn't really matter but it would be nice to keep them consistent.

lib/p2p/packets/types/DisconnectingPacket.ts
```diff
@@ -15,6 +18,46 @@ class DisconnectingPacket extends Packet<DisconnectingPacketBody> {
...
+    return !!(msg.header
+      && msg.header.id
+      && msg.header.hash
+      && !msg.header.reqid
+      && msg.reason
+    );
+  }
+
+  private static convert = (msg: pb.DisconnectingPacket.AsObject): DisconnectingPacket => {
+    return new DisconnectingPacket({
```
sangaman: What do you think about using the native protobuf objects for the packets? Maybe for a future PR? We could keep these Packet classes as wrapper objects, but have them hold a pb object instead of the header & body JSON objects.

lib/p2p/packets/types/DisconnectingPacket.ts
```diff
@@ -15,6 +18,46 @@ class DisconnectingPacket extends Packet<DisconnectingPacketBody> {
...
+      header: {
+        id: msg.header!.id,
+        hash: msg.header!.hash,
+      },
+      body: removeUndefinedProps({
```
sangaman: What's the point of using `removeUndefinedProps` here for example? Isn't it fine if `payload` is undefined? I haven't examined the other usages of it but I'm not sure if it's needed.

proto/xudp2p.proto
moshababo: Sounds fine, just the it won't give us a clear distinction between the body and the header. Also, when we'll want to hash the body, we need to make sure we don't get confused about which fields we want to put in. 

If it sounds ok to you, i'll do the change.

proto/xudp2p.proto
moshababo: Yes, forgot to remove this one.

lib/p2p/packets/types/DisconnectingPacket.ts
moshababo: It's needed because otherwise the return type won't be a boolean.

lib/p2p/packets/types/DisconnectingPacket.ts
moshababo: It's basically the same problem we previously had - an object with an undefined prop isn't the same as an object without that prop. Because of that, the objects deep equal assertions will fail.
You can remove the `removeUndefinedProps` and see yourself how that tests are failing.

It's not a big deal, and we can relax the test assertions,  but I think it's still better to remove the prop entirely.

moshababo:
The main missing thing for this PR is your [comment](https://github.com/ExchangeUnion/xud/pull/712#discussion_r238552276) on removing the generic `Header` type, and putting its fields inside the packet.

I think that doing so is fine, but i'm trying to solve this problem together with https://github.com/ExchangeUnion/xud/issues/157 - so we need to settle on how we're hashing the packet body (which is relevent also for #341).

The current approach of hashing the JSON representation isn't optimal, because `JSON.stringify` doesn't guarantee deterministic output. There are a few solutions for that by ordering the fields, but it's still problematic in an open network that theoretically can have multiple clients, in that we don't clearly define the protocol.

Another option would be to define the fields order and type, and to serialize to binary ourselves. I think that in our case it will be an overkill.

Another option is to work with protobuf. We can construct the body fields and then use `serializeBinary()`, and in the absence of other fields from the header, the output won't be affected.  protobuf adds the field type and number before the raw value - I don't think it's a problem, but it means that in practice the protocol will be defined by protobuf. In that case, we can construct the protobuf message, populate the body fields, and then serialize it for the hash/signature. It might be a bit confusing to work like that, and the fields numbers won't start from 1, so in addition we can create separate message type for each body.

proto/xudp2p.proto
sangaman: Those are good points. Another possibility would be defining different Header messages such as `Header` & `ResponseHeader` and using those where appropriate. But I'm thinking this way might be good after all, so go with whatever you think is best.

lib/p2p/packets/types/DisconnectingPacket.ts
sangaman: Yes actually you are right, my mistake.

lib/p2p/packets/types/DisconnectingPacket.ts
sangaman: Alright that's fine, just a thought.

sangaman:
Sorry for the delay on getting back to you on this. It looks good to me but one of the tests is failing and I haven't dug into why yet, maybe you can take a look. You can go with what you think is best for how the header is structured.

I think the final option of serializing the message with only the body fields and then hashing the serialized bytes makes the most sense. We're already baking protobuf into the protocol, so I don't see any problem with the field types and numbers being part of the serialized data. I'm assuming that the serialized data will be deterministic?

moshababo:
In regards to the serialization for hashing/signing - I looked into into what I suggested some more, and I was wrong in assuming that [protobuf serialization](https://gist.github.com/kchristidis/39c8b310fd9da43d515c4394c3cd9510) is deterministic.
So I think we should stick with `JSON.stringify` for the sake of simplicity, but add a wrapper for a deterministic output, similar to [this](https://www.npmjs.com/package/json-stringify-deterministic).

moshababo:
Let me know if you're good with it, and i'll finish this PR.

sangaman:
I'm good with it, we'll deal with the hashing in a separate PR. I'm thinking another approach for that might be to construct the JSON object ourselves (rather than `toObject`) and strictly follow ordering the properties alphabetically, that way `JSON.stringify` should give us the same output each time. We can continue discussing that separately tho.

moshababo:
I removed the `Header` protobuf message type as you suggested. Will deal with the hashing/signing in a separate PR. 

Also added some more error handling. If packet validation failed, the invalid protobuf object will be logged.

I'll leave this open until Sunday and then will merge it.

kilrau:
Did you do some realworld (SIMNET) testing with this? @moshababo

moshababo:
@kilrau no.. we'll need to upgrade all the instances for this.

moshababo:
I'm merging this, to not further delay the dev pipeline. 
If issues will arise due to the strict validation, we'll fix it.

PR #716, opened 2018-11-30
feat(p2p/orderbook): drop orders for dropped pairs

### LINKED ISSUE #599: Remove peer orders for dropped trading pairs
As discussed in #585, when a peer drops support for a trading pair in a handshake update, we should drop all orders for that peer and trading pair locally.

sangaman:
This adds logic to drop a peer's orders for a given trading pair when we receive an updated handshake state that does not include a trading pair that was supported according to the previous handshake. Without this, we may keep orders in our order book for a peer that no longer supports (and can swap) those orders.

Closes #599.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -89,6 +89,7 @@ class OrderBook extends EventEmitter {
       this.pool.on('packet.getOrders', this.sendOrders);
       this.pool.on('packet.swapRequest', this.handleSwapRequest);
       this.pool.on('peer.close', this.removePeerOrders);
+      this.pool.on('peer.pairDropped', this.removePeerPair);
     }
   }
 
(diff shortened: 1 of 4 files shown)


lib/p2p/Peer.ts
```diff
@@ -577,6 +581,7 @@ class Peer extends EventEmitter {
       return;
     }
 
+    const oldHandshakeState = this.handshakeState;
```
moshababo: perhaps `prevHandshakeState`.

lib/p2p/Pool.ts
```diff
@@ -633,6 +637,18 @@ class Pool extends EventEmitter {
       await this.handlePacket(peer, packet);
     });
 
+    peer.on('handshake', (oldHandshakeState) => {
+      // drop all orders for trading pairs that are no longer supported
+      if (oldHandshakeState) {
+        oldHandshakeState.pairs.forEach((pairId) => {
+          if (!peer.pairs!.includes(pairId)) {
+            // a trading pair was in the old handshake state but not in the updated one
```
moshababo: This doesn't need anything from `Pool`, so we can handle it entirely in `Peer`.

lib/p2p/Pool.ts
```diff
@@ -633,6 +637,18 @@ class Pool extends EventEmitter {
...
+            this.emit('peer.dropPair', peer.nodePubKey!, pairId);
```
moshababo: `dropPair` sounds like a command, not an event. perhaps `pairDropped` / `pairRemoved` / `pairRemoval`.

sangaman: I was trying to follow the imperative style of the `close` event for example, but I also prefer `pairDropped` so I'll go with that.

lib/p2p/Pool.ts
sangaman: I'm emitting `peer.dropPair` from `Pool` because other components are listening to `Pool` and not `Peer`. But if you're suggesting to move the logic to check which pairs have been dropped or added into `Peer` - in that case I suppose I'd have it emit specific events like `pairAdded` and `pairRemoved` rather than `handshake`?

moshababo: Yes, and then propagate it from `Pool`.

lib/p2p/Pool.ts
moshababo: I'm not sure it's the same, because `close` doesn't imply that we should close something. So maybe it's the equivalent of `pairDrop`, but you're the english expert here ;)

sangaman:
@moshababo I addressed your feedback, take another look when you get a chance.

lib/orderbook/OrderBook.ts
```diff
@@ -541,6 +542,11 @@ class OrderBook extends EventEmitter {
     this.logger.debug(`removed all orders for peer ${peer.nodePubKey}`);
   }
 
+  private removePeerPair = (peerPubKey: string, pairId: string) => {
+    const tp = this.getTradingPair(pairId);
+    tp.removePeerOrders(peerPubKey);
```
moshababo: shouldn't this be similar to how `removePeerOrders` is being handled, and so to emit `peerOrder.invalidation` for every removed order?

sangaman: Yes you are right, good catch. I'll make a reusable method for this so it can be called in the future to make sure this event gets emitted.

PR #717, opened 2018-11-30
fix(rpc): close streaming calls on shutdown

### LINKED ISSUE #699: xucli 'shutdown' doesn't shut down xud process
xucli shutdown does not bring the process down and does not close the active peer connections.

sangaman:
Fixes #699.

This tracks all active streaming gRPC calls and closes them when the gRPC server is shut down. Previously, active streaming calls would prevent `xud` from shutting down gracefully.

I've tested this and it works well for me. I also confirmed that the `cancelled` event is fired when any streaming call is ended by the client.

### DIFF
--- a/lib/grpc/GrpcServer.ts
+++ b/lib/grpc/GrpcServer.ts
@@ -12,6 +12,7 @@ import { exists, readFile, writeFile } from '../utils/fsUtils';
 
 class GrpcServer {
   private server: Server;
+  private grpcService: GrpcService;
 
   constructor(private logger: Logger, service: Service) {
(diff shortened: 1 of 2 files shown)


sangaman:
@offerm I added a line to clear the streams after they've ended. And I think I see what you mean about incoming peers, I'll handle that in a separate PR though.

offerm:
@sangaman sure. This PR however does not close  #699 yet.

We need this fix to check #616 

Thanks

sangaman:
I'll open a separate issue for incoming peers interfering with the shutdown process.

PR #718, opened 2018-12-02
fix(p2p): terminate inbound peers on shutdown

sangaman:
This tracks inbound peer connections so that they can be closed when `xud` is gracefully shutdown. Previously, pending inbound peers could prevent `xud` from shutting down gracefully.

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -70,7 +70,9 @@ class Pool extends EventEmitter {
   /** The local handshake data to be sent to newly connected peers. */
   public handshakeData!: HandshakeState;
   /** A map of pub keys to nodes for which we have pending outgoing connections. */
-  private pendingOutgoingConnections = new Map<string, Peer>();
+  private pendingOutboundPeers = new Map<string, Peer>();
+  /** A set of peers for which we have pending incoming connections. */
+  private pendingInboundPeers = new Set<Peer>();
   /** A collection of known nodes on the XU network. */
   private nodes: NodeList;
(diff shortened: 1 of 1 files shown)


lib/p2p/Pool.ts
```diff
@@ -72,6 +72,8 @@ class Pool extends EventEmitter {
   public handshakeData!: HandshakeState;
   /** A map of pub keys to nodes for which we have pending outgoing connections. */
   private pendingOutgoingConnections = new Map<string, Peer>();
+  /** A set of peers for which we have pending incoming connections. */
+  private pendingIncomingConnections = new Set<Peer>();
```
moshababo: In this case the connection isn't pending, because it was already established.

To make this more clear, maybe we should rename these 2 to `pendingOutboundPeers` / `pendingInboundPeers`.

sangaman:
> It looks like this would fix the problem only if the shutdown cmd came in the short interval between the inbound connection and the handshake completion, because otherwise the peer/conn would be on the active peers list.
> Is that really the case, or am I missing something?

Yes, just from looking at the code that's the only gap I could see where an inbound connection could interfere with the shutdown process (we already close active peers), but according to @offerm something else is in play because the issue happens even . So this PR might be fixing some edge case that we haven't encountered yet. I'm looking into it and might update the description for this PR and open a different one depending on what I find.

sangaman:
@moshababo I renamed the properties, good suggestion. This doesn't appear to solve #799 after all, so I updated the commit and PR descriptions. Still a useful fix though I think.

PR #720, opened 2018-12-02
feat: graceful shutdown on SIGINT

sangaman:
This intercepts SIGINT and begins the graceful shutdown process.


lib/Xud.ts
```diff
@@ -170,6 +175,10 @@ class Xud extends EventEmitter {
   }
 
   private shutdown = async () => {
+    if (this.shuttingDown) {
```
moshababo: We can add "Already shutting down..." logging here.

PR #730, opened 2018-12-05
feat(subscribeorders.ts): SubscribeAddedOrders should have `existing` true by default 

### LINKED ISSUE #728: SubscribeAddedOrders should have `existing` true by default

ImmanuelSegol:
Closes #728


ImmanuelSegol:
@sangaman  How should I decide what orders should be returned?

sangaman:
I just don't think we should be returning a subset of orders, and that's not the problem highlighted in the issue, so I think we should stick to returning all orders up front when `existing` is true.

sangaman:
@ImmanuelSegol Will you be able to fix this typo soon? I think it makes sense to merge this before the next release and obviously the change is very minor, but if you're busy I can handle it.

Issue #733, opened 2018-12-06
remove local_id from `SwapResult` in nomatching mode


kilrau:
Currently nomatching mode (here a `SwapResult`) still maintains a `local_id` which doesn't make sense to me. There should be only one (global) `order_id`.

```
order_id: "631145c0-f901-11e8-b518-5fd626b71b77"
local_id: "e6fe35e0-f911-11e8-b9ca-27ecace2dafc"
pair_id: "LTC/BTC"
quantity: 0.001
r_hash: "aad8fae44138922fc52b1fca949c7f75294fa5f8345418634d23ba4d961b5b9f"
amount_received: 100000
amount_sent: 110000
peer_pub_key: "03fd337659e99e628d0487e4f87acf93e353db06f754dccc402f2de1b857a319d0"
currency_received: "LTC"
currency_sent: "BTC"
```

sangaman:
Sorry I'd missed this earlier, conceptually I don't follow though. Why wouldn't we want a local id for nomatching mode? What if an order has an id on the local exchange? Wouldn't we want a way to track that id in `xud` and associate it with the global order id?

kilrau:
> Wouldn't we want a way to track that id in xud and associate it with the global order id?

Yep, that's exactly what we want. BUT `local_id` is not an order id the exchange system assigned to an order; it's another uuid `xud` generated for this order. @sangaman 

To visualize what we have:
`exchange_order_id` + `local_id` + (global) `order_id`

and what we want:
`exchange_order_id` <-> (global) `order_id` maintained within `xud`

sangaman:
In that taker image on the right, the `local_id` is the randomly generate global order id - since no local id was specified the local and global ids are the same.

I can see how that may be a little confusing, but what if the exchange does want to associate a taker order with an actual order on their exchange? Imagine I'm an exchange in no matching mode, I get order 123 and match it with an order in xud, now I call `ExecuteSwap` and I want to tell `xud` that the local id of the order is 123, so it gets saved in the database and can be searched/retrieved later.

Maybe the confusing part is that the `order_id` in the swap result refers to the global id of the maker order, whereas `local_id` refers to the local id of the local order, whether it's the maker or the taker. It's definitely not clear based on the comments in the proto file, so I'm thinking at a minimum we should clarify the comments (I can handle that) and maybe rename fields to be more explicit.

kilrau:
[...]

PR #741, opened 2018-12-09
refactor: remove local_id in nomatching mode

ImmanuelSegol:
Closes: #733 
Please fix the error in the commit message before merging.

@kilrau I hope this takes care of the issue.


sangaman:
I asked some questions in the issue, conceptually I don't understand why we wouldn't want `localid` for nomatching mode.

proto/xudrpc.proto
```diff
@@ -540,28 +540,28 @@ message SwapSuccess {
...
-  Role role = 9 [json_name = "role"];
+  Role role = 8 [json_name = "role"];
   // The ticker symbol of the currency received.
-  string currency_received = 10 [json_name = "currency_received"];
+  string currency_received = 9 [json_name = "currency_received"];
   // The ticker symbol of the currency sent.
-  string currency_sent = 11 [json_name = "currency_sent"];
+  string currency_sent = 10 [json_name = "currency_sent"];
+  // The trading pair that the swap is for.
+  string pair_id = 11 [json_name = "pair_id"];
```
sangaman: It's hard to tell what all is going on here. Could you change the field numbers/ordering back to how it was before?

sangaman:
@ImmanuelSegol Could you take another look at this? I don't actually see any changes except for in the generated swagger file. The key thing here is to not set a value for `local_id` in `SwapSuccess` messages where our role is the taker.

PR #743, opened 2018-12-10
fix(swap): allow one extrac block to be created during swap

offerm:


lib/swaps/Swaps.ts
```diff
@@ -569,7 +569,8 @@ class Swaps extends EventEmitter {
       return false;
     }
 
-    if (cltvDelta > resolveRequest.getTimeout() - resolveRequest.getHeightNow()) {
+    // allow 1 additional one block to be created during the swap
+    if (cltvDelta - 1 > resolveRequest.getTimeout() - resolveRequest.getHeightNow()) {
       this.logger.error(`got timeout ${resolveRequest.getTimeout()} at height ${resolveRequest.getHeightNow()}`);
       this.logger.error(`cltvDelta is ${resolveRequest.getTimeout() - resolveRequest.getHeightNow()} expected delta of ${cltvDelta}`);
```
sangaman: I figure this should say `expected delta of at least ${cltvDelta -1}`.

offerm: I don't think we should change this. If we do it may create confusion with the configuration.

offerm:
I will continue thinking about this. Not simple. 
Still, 2 new blocks (same net) during the swap is very unlikely. 

@sangaman Lets accept it as is since I need this for the trading tests.

Note that I added ceil in the ctlv calculation to avoid fraction (have seen this in action).

sangaman:
Alright sounds good for now, thanks.

Issue #402, opened 2018-08-27
Share newly connected peers dynamically at runtime


kilrau:
This issue is to discuss 2 opposing approaches in how to share new peers (reminder: peer is a node I'm currently successfully connected to):
a) Manually send `GET_NODES` packet to learn about all peers of my peer (concern: nobody will ever do that, it should be automated). Also it's potentially a huge load every time since it's ALL peers.
b) My peer shares newly connected peers dynamically at run time every time it successfully connects to one (concern: spammy?)

I would define our overall goal as: "All xuds automatically find each other and connect to each other." Just like BitTorrents DHT gets bootstrapped from some hard coded seeds (or a swarm) and the rest happens *automatically*. Hence, I think we should combine a) and b).
a) happens on startup
b) during run time

for b) we can add additional peer meta data (like trading pairs) to only get peers of interest.

offerm:
@kilrau If the goal is : "All xuds automatically find each other and connect to each other." we should assume that all nodes are already connected to all connectable nodes. `TBD: is this scalable model`

So when a new node  (X) joins the graph it is doing so by connecting to one of the existing nodes (Y). Y should already be connected to most other nodes and should share their addresses to X. X will connect to these nodes itself so all other nodes become connected with X without the need for anything like a and b being done.

sangaman:
> a) Manually send GET_NODES packet to learn about all peers of my peer (concern: nobody will ever do that, it should be automated). Also it's potentially a huge load every time since it's ALL peers.
> b) My peer shares newly connected peers dynamically at run time every time it successfully connects to one (concern: spammy?)

Manually querying for new nodes is the type of thing that would have to happen quite rarely, and while it can potentially be a big packet it's already something we do every time start up and I don't think it's a problem. My thinking is that, for nodes that stay online for long stretches of time, manually querying for new nodes is a way to simulate that start-up querying for nodes but without actually shutting down the server. But maybe we don't even need either of these options. We already do GET_NODES each time we start up, and if we're online, new nodes will hear about us. 

>  TBD: is this scalable model

Not exactly, I think this is fine for a network on the order of dozens or hundreds of nodes. Much beyond that I'm not sure, and at some point we just exhaust the number of ports. That would definitely be a phase 2 sort of concern.

moshababo:
I think we have total 4 different approaches: 

1. Manually query `GET_NODES` from peers

I don't think it's relevant. everything should be done automatically.

[...]

PR #746, opened 2018-12-11
Sending GET_NODES scheduled #402

rsercano:
Hopefully this does the trick for #402 , please let me know if that's not the right way to do it.


lib/Config.ts
```diff
@@ -63,6 +63,7 @@ class Config {
     this.p2p = {
       listen: true,
       discover: true,
+      discoverInSeconds: 60 * 60, // 1 hour
```
sangaman: I think an hour is pretty short, I was thinking more like a day would make sense? Or maybe 12 hours? I think every hour can get pretty noisy and most of the time will not get us any new info.

Also, I think we can safely measure this in minutes instead of seconds, I doubt we need second precision. It's a little easier to calculate that way.

And for the name, I'm struggling to come up with something I really like but maybe `rediscoverTimer` or `rediscoverMinutes` might be a little clearer?

lib/p2p/Peer.ts
```diff
@@ -50,6 +50,8 @@ class Peer extends EventEmitter {
   public sentDisconnectionReason?: DisconnectionReason;
   public expectedNodePubKey?: string;
   public active = false; // added to peer list
+  /** Timer to schedularly call getNodes #402. */
```
sangaman: `schedularly` sounds like it should be a word, but I confirmed that it's not. How about just `Timer to periodically call GetNodes`?

lib/p2p/Pool.ts
```diff
@@ -579,8 +582,10 @@ class Pool extends EventEmitter {
     // request peer's orders
     peer.sendPacket(new packets.GetOrdersPacket({ pairIds: this.handshakeData.pairs }));
     if (this.config.discover) {
-      // request peer's known nodes only if p2p.discover option is true
-      peer.sendPacket(new packets.GetNodesPacket());
+      peer.getNodesTimer = setInterval(() => {
```
sangaman: It might be nice to be able to set the configured value to `0` to disable the scheduled GetNodes calls, if so here's where we'd want to make that check.

lib/Config.ts
rsercano: Makes sense, but since discover's word meaning already something like; find something new, do we really want to rename these ?

lib/p2p/Peer.ts
rsercano: haha, my bad of course it's not.

lib/p2p/Pool.ts
rsercano: good point

lib/Config.ts
sangaman: Well it's a new config option, I don't meant to rename any of the existing options. Here  I think that "rediscover" might make more sense because we're telling `xud` to continuously attempt to discover new nodes, e.g. repeating discovery, instead of only doing so on first connection.

kilrau: Most people won't know that `xud` discovers once in the beginning, so discover is fine with me. Rediscover raises more ? I'm afraid.

sangaman: We'll leave the existing option as `discover` for sure, but then the new option is tricky. I'd be ok with `discoverminutes` I suppose, it should definitely be all lower case tho to be consistent with the other config options.

Either way, this is an advanced option and not at all something people should have to mess with unless they have a good reason to, and definitely not if they don't know how the `discover` option works.

lib/p2p/Pool.ts
```diff
@@ -580,7 +583,16 @@ class Pool extends EventEmitter {
     peer.sendPacket(new packets.GetOrdersPacket({ pairIds: this.handshakeData.pairs }));
     if (this.config.discover) {
       // request peer's known nodes only if p2p.discover option is true
-      peer.sendPacket(new packets.GetNodesPacket());
+      if (this.config.discoverminutes === 0) {
+        // timer is disabled
+        peer.sendPacket(new packets.GetNodesPacket());
+        peer.getNodesTimer = undefined; // defensive programming
+      }else {
```
moshababo: should add white space between `}` and `else`

lib/p2p/Pool.ts
```diff
@@ -580,7 +583,16 @@ class Pool extends EventEmitter {
...
+        // timer is enabled
+        peer.getNodesTimer = setInterval(() => {
```
moshababo: what about moving this in `Peer.open`, with all the rest of the intervals? (sendPing, checkTimeout)

moshababo: looks like that if timer is enabled, there won't be an initial discovery - only after the interval.

lib/p2p/Peer.ts
```diff
@@ -50,6 +50,8 @@ class Peer extends EventEmitter {
   public sentDisconnectionReason?: DisconnectionReason;
   public expectedNodePubKey?: string;
   public active = false; // added to peer list
+  /** Timer to periodically call getNodes #402 */
+  public getNodesTimer?: NodeJS.Timer;
```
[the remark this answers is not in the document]
rsercano: done

lib/p2p/Pool.ts
rsercano: I don't know why tslint didn't warn me, but done :)

lib/p2p/Pool.ts
rsercano: done

lib/p2p/Peer.ts
```diff
@@ -190,6 +192,21 @@ class Peer extends EventEmitter {
 
     // TODO: Check that the peer's version is compatible with ours
 
+    // request peer's known nodes only if p2p.discover option is true
+    if (this.discover) {
+      if (this.discoverminutes === 0) {
+        // timer is disabled
+        this.sendPacket(new packets.GetNodesPacket());
```
moshababo: you can take this out of the `if` clause - it's executed in both cases.
so there should only one `if` statement - checking whether `this.discoverminutes` is truthy.

lib/p2p/Peer.ts
```diff
@@ -190,6 +192,21 @@ class Peer extends EventEmitter {
...
+        this.discoverTimer = undefined; // defensive programming
+      } else {
+        // timer is enabled
+        this.sendPacket(new packets.GetNodesPacket()); // initial execution
+        this.discoverTimer = setInterval(() => {
```
moshababo: perhaps follow the `pingTimer` one-liner, by creating a non-anonymous method `sendGetNodes`.

lib/p2p/Peer.ts
```diff
@@ -121,8 +123,8 @@ class Peer extends EventEmitter {
   /**
    * Creates a Peer from an inbound socket connection.
    */
-  public static fromInbound = (socket: Socket, logger: Logger): Peer => {
-    const peer = new Peer(logger, addressUtils.fromSocket(socket));
+  public static fromInbound = (socket: Socket, logger: Logger, discover: boolean, discoverminutes: number): Peer => {
```
moshababo: perhaps pass along the entire Pool config object. will be easier to follow when it's being read from (`this.discover` vs `this.config.discover`), and easier to add more fields later. 
It might require to move the `PoolConfig` type to a different location. If it's getting complicated you can leave it as is for now.

lib/p2p/Peer.ts
rsercano: right

lib/p2p/Peer.ts
rsercano: sure

lib/p2p/Peer.ts
rsercano: passing config into it from `Pool` I'm not the one to decide where to put config type, therefore leaving it as is, but changing signature of Peer constructor to accept PoolConfig

rsercano:
Please ignore this PR, followup is here: #770

Issue #722, opened 2018-12-04
Revoke connection retries


moshababo:
When attempting to connect to an unreachable node, connection retries procedure might trigger, which will last a long time. During that time, it's not possible to connect to a different address of the same `nodePubKey`.

We should add support for cancelling the connection retries, perhaps by re-adding the `disconnect` cmd, which was removed, and apply it on the connection retries as well.

kilrau:
How this happened:
we setup 2 environments with 2 different xud's and 2 different `nodePubKeys` but which (for some reason) had the same IP. Connecting xud1 one to xud2 succeeded once. Then we noticed the IP issue, changed IP of xud 2 and tried connecting again. Not possible because `xud` threw an "Already exisitng connection attempt". Problem: "Already existing connection attempt checks only `nodePubKey`, but not not IP:Port.
Solution01: make connection attempt check the full URI, including IP & Port
Solution02: Implement a "disconnect" funtion which closes socket to peer, remove peer from nodes list

Now thinking a bit about it, actually Solution01 sounds better (&easier). Opinions? @moshababo @sangaman @offerm

moshababo:
Implementing **Solution01** isn't straightforward - we structured many features on the premise having one-to-one relationship between a connection and a `nodePubKey`.
Also, it might not even solve the problem - the connection retries wouldn't trigger if "Already existing connection" error is thrown. They probably started when you weren't connected, so implementing **Solution01** might not be relevant to the problem.

I'd suggest to go with **Solution02** which sounds useful non-related to this issue, since the connection retries procedure is very long.

sangaman:
How about when we manually try to connect to a given node pub key, we cancel any pending connections for that node pub key and then start the manual attempt. That should solve the problem, right? I think it would be easier/cleaner to implement.

moshababo:
@sangaman i'd say that having implicit side effects is less optimal, but since it's just about cancelling the connection retries (and not an actual connection), it should be fine.

PR #750, opened 2018-12-17
feat(p2p): revoke connection retries

moshababo:
Solves #722 

* retries are revoked even if the same IP is used. I thought that it might be useful for triggering an immediate connection attempt, when the retries delay become very big.
* ~~I added the option of having connection retries on manual connection as well, mostly for the automatic tests. I also added it to the gRPC because I thought it might be useful - let me know if you think otherwise.~~

### DIFF
--- a/test/p2p/sanity.spec.ts
+++ b/test/p2p/sanity.spec.ts
@@ -112,13 +112,13 @@ describe('P2P Sanity Tests', () => {
       retryConnecting: true,
     });
 
-    setTimeout(() => {
+    setImmediate(() => {
       expect(nodeOne.service.connect({
         nodeUri: toUri({ port, nodePubKey, host: 'localhost' }),
         retryConnecting: false,
       })).to.be.rejectedWith(`could not connect to peer at localhost:${unusedPort}`);
       done();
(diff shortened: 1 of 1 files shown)


lib/cli/commands/connect.ts
```diff
@@ -11,10 +11,16 @@ export const builder = {
     description: 'uri of peer in [nodePubKey]@[host]:[port]',
     type: 'string',
   },
+  retry_connecting: {
+    description: 'whether to apply multiple connection retry attempts',
```
sangaman: I'm not that sure this is a useful feature, wouldn't rpc call potentially take a very long time?

moshababo: Yes, it will be blocked until connection will succeed. 
Our options is either to propagate updates (same ones from the logs), or to remove this feature.

sangaman: Gotcha. I'd lean towards removing this, maybe we can discuss in gitter and see what others think.

moshababo: I removed this feature from cli/grpc.

PR #751, opened 2018-12-17
feat(p2p): verify data integrity

moshababo:
Solves #157.

As discussed [here](https://github.com/ExchangeUnion/xud/pull/712#issuecomment-445180853), the packet body serialization for the hash function is done with `JSON.stringify`. Lack of consistency in the serialization was already found, so I used `json-stable-stringify` to fix it. Without it the tests would fail.

### DIFF
--- a/lib/p2p/packets/Packet.ts
+++ b/lib/p2p/packets/Packet.ts
@@ -2,6 +2,7 @@ import PacketType from './PacketType';
 import CryptoJS from 'crypto-js';
 import MD5 from 'crypto-js/md5';
 import uuidv1 from 'uuid/v1';
+import stringify from 'json-stable-stringify';
 
 type PacketHeader = {
   /** An identifer for the packet which must be unique for a given socket. */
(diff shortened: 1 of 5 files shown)


lib/p2p/packets/Packet.ts
```diff
@@ -81,12 +82,31 @@ abstract class Packet<T = any> implements PacketInterface {
...
         this.body = bodyOrPacket;
-        this.header.hash = MD5(JSON.stringify(bodyOrPacket)).toString(CryptoJS.enc.Base64);
+        this.header.hash = this.hash(bodyOrPacket);
       }
     }
   }
   public abstract serialize(): Uint8Array;
 
+  private hash(value: any): string {
+    return MD5(stringify(value)).toString(CryptoJS.enc.Base64);
```
sangaman: It might be nice to have some sanity tests for this method.

moshababo: You mean to test against static values (input/output)?

moshababo: I'll add some testing in the upcoming #756. Merging this now due to dependency in #756.

PR #754, opened 2018-12-18
feat(rpc): remove peerpubkey from executeswap

### LINKED ISSUE #694: ExecuteSwap -> remove peerPubKey
This call shouldn't require any peerPubKey since exchanges order gets mapped with XU `oderID` which has all info including the peerPubKey. Remove, not optional.

sangaman:
This call removes the `peer_pub_key` field from the `ExecuteSwap` request. This is not truly required information and it is inconvenient to have to provide, particularly from the cli. Finding which peer an order belongs to is not much of a performance concern.  It is constant time in regards to the number of orders and linear in regards to the number of peers.

Closes #694.

### DIFF
--- a/lib/orderbook/TradingPair.ts
+++ b/lib/orderbook/TradingPair.ts
@@ -137,7 +137,7 @@ class TradingPair {
   /**
    * Remove all orders given a peer pubKey.
    */
-  public removePeerOrders = (peerPubKey: string): PeerOrder[] => {
+  public removePeerOrders = (peerPubKey?: string): PeerOrder[] => {
     // if incoming peerPubKey is undefined or empty, don't even try to find it in order queues
     if (!peerPubKey) return [];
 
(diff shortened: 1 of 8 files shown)


kilrau:
Nice!

sangaman:
I'm actually second guessing my own work here, maybe we should leave this in and specify in the comments that it is optional? While manually inputting the `peerpubkey` is a pain, it's simple to do from code, and it will save us a bit of work within `xud`.

moshababo:
Keeping it optional in the gRPC and not in the CLI could be fine IMO.

sangaman:
@moshababo I went ahead and put it back on the gRPC call only, but it's still gone from the cli.

lib/orderbook/TradingPair.ts
```diff
@@ -161,7 +161,20 @@ class TradingPair {
    * @returns the removed order or order portion, otherwise undefined if the order wasn't found
    */
   public removePeerOrder = (orderId: string, peerPubKey: string, quantityToRemove?: number): { order: PeerOrder, fullyRemoved: boolean} => {
```
moshababo: i'd say to keep `peerPubKey` optional both here and in `OrderBook.removePeerOrder`, same as we're doing with the quantity. Otherwise it's implicit.

Issue #158, opened 2018-06-27
Signing & encrypting


moshababo:
Decide which encryption schemes we want to apply.

* ECDSA (sign on own public key)
* Encrypt on peers public keys
* AES (via encryption handshaking) / SSL

kilrau:
`xud` enables two levels of signing/encrypting:
1. Sign only: Sign all order packets, peer verifies authenticity with my public key
2. Sign & encrypt: Sign & encrypt order to verify authenticity and encrypt orders with clients `nodePubKey` to ensure order info can only be decrypted by white-listed peers

Notes:
*1.* is the **default** and signs all orders with the XUD's private key. This ensures that the `xud`s `nodePubKey` carried in the order's invoice is the actual pubKey of the `xud` sending the invoice. We decided to place the `nodePubKey` directly in the invoice description field for a faster processing, since this avoids a payment hash vs. `nodePubKey` lookup for each incoming new order. This mode prevents `xud`'s pretending to be someone else and knowing the actual issuer of an order is important for e.g. the incentive & punishing mechanism to work. This mode connects to all available peers and is suitable for exchanges located in jurisdictions without a strict legal framework for digital assets. 

To be discussed @sangaman @moshababo : it could be enough to just verify the `nodePubKey` on handshake (https://github.com/ExchangeUnion/xud/issues/341) and not sign and verify every packet. This would increase processing speed.

*2.* is optional, can be enabled in the config. Additionally to everything *1.* does, `xud` encrypts orders with the public key of each white-listed peer. This mode is more resource intensive and requires maintaining a white-list of peers. It is suitable for exchanges located in jurisdictions with a strict legal framework and ensures that order information can only be received and processed by white-listed trading partners, e.g. other licensed exchanges.

This issue is about implementing *Mode 2*.

kilrau:
Not priority for next milestone, moved to next

moshababo:
As discussed in #341, 

1: there’s no need to sign all p2p messages while we’re in fully-connected network mode. Signing is necessary only for the communication session authentication. 

2: we will encrypt all communication via shared secret key (ECDH/ECDHE) per session. This will make encrypting on white-list peers public key unnecessary, since only the session peer will be able to decrypt it.

moshababo:
I'm working on this together with #341.

PR #756, opened 2018-12-19
feat(p2p): encryption & authentication

moshababo:
Solves #158 & #341 

* interactive handshake: `Hello` packet changed to `SessionInit`/`SessionAck`.
* `SessionInit`/`SessionAck` interaction is done from both sides, and yield a different secret key (ECDHE) for each side of the communication (following [BIP151](https://github.com/bitcoin/bips/blob/master/bip-0151.mediawiki)). on each side, when `SessionAck` is received, all following messages are expected to be encrypted. 
* added `aes-256-cbc` symmetric encryption using the secret key (wire protocol)
* added ECDSA authentication
* added crypto primitives tests
* `Hello` is no longer used for updating the handshake state. the term handshake is now used only to describe the initial encryption / authentication interaction. In this interaction each party also send its initial `NodeState`, similar to the previously named `HandshakeState`. `NodeStateUpdate` packet was added to support state changes (removed/added pairs, etc.)
* added `Framer` class - to handle the wire protocol serialization/deserialization (on top of the packets serialization).
* added `Network` class - to contain all network-specific data - currently just the network magic value.
* terminology/structure: packets are used in the application layer. when sent over the wire, a header is being added, to receive a wire msg. if the wire msg is encrypted, another header is being added. 
* packet `hash` changed to `checksum` and was moved to the wire msg header - so it's no longer exist on the packet js object.


moshababo:
@sangaman you can review now - a bit more work is left, which i'll return to on Monday.

lib/p2p/packets/Packet.ts
```diff
@@ -82,46 +78,26 @@ abstract class Packet<T = any> implements PacketInterface {
...
-    const type = Buffer.alloc(1);
-    type.writeUInt8(this.type, 0, true);
-
-    const size = Buffer.allocUnsafe(4);
-    size.writeUInt32LE(msg.length, 0, true);
+    return Buffer.from(this.serialize().buffer as ArrayBuffer);
+  }
 
-    return Buffer.concat([type, size, Buffer.from(msg.buffer as ArrayBuffer)]);
+  public checksum(): Buffer {
```
sangaman: Also, for stylistic purposes we're not using the arrow function format for this and several other methods in this class, so it'd be worth changing. I think `checksum` might actually make the most sense as a getter.

test/unit/Parser.spec.ts
```diff
@@ -269,7 +357,8 @@ describe('Parser', () => {
       parser.feed(Buffer.alloc(0));
     });
 
-    it(`should try parse just the metadata as a packet`, (done) => {
+    /*
+    it(`should not try parse just the header as a packet`, (done) => {
```
sangaman: Should we just drop this test case?

lib/p2p/Peer.ts
```diff
@@ -174,25 +189,26 @@ class Peer extends EventEmitter {
    * @param nodePubKey the expected nodePubKey of the node we are opening a connection with
    * @param retryConnecting whether to retry to connect upon failure
    */
-  public open = async (handshakeData: HandshakeState, nodePubKey?: string, retryConnecting = false): Promise<void> => {
+  public open = async (nodeState: NodeState, nodeKey: NodeKey, expectedNodePubKey?: string, retryConnecting = false): Promise<void> => {
```
sangaman: For consistency (and improved clarity) we can call `nodeState` `ownNodeState` as you're done elsewhere.

lib/p2p/Peer.ts
```diff
@@ -410,7 +437,7 @@ class Peer extends EventEmitter {
    * Waits for a packet to be received from peer.
    * @returns A promise that is resolved once the packet is received or rejects on timeout.
    */
-  private wait = (packetId: string, timeout?: number) => {
+  private wait = (packetId: string, timeout?: number): any => {
```
sangaman: I'd prefer avoiding `any`, I figure this can at least return `Promise<Packet>` and then when calling this method we can define the return type more specifically like:

```typescript
const inSessionAck: Packet<SessionAckPacketBody> = await this.wait(outSessionInit.header.id, Peer.RESPONSE_TIMEOUT);
```

lib/p2p/Peer.ts
```diff
@@ -511,7 +538,16 @@ class Peer extends EventEmitter {
       this.close();
     });
 
-    this.socket!.on('data', this.parser.feed);
+    this.socket!.on('data', (data) => {
+      this.lastRecv = Date.now();
+      const dataStr = data.toString();
+      if (this.nodePubKey !== undefined) {
+        this.logger.trace(`Received data from ${this.nodePubKey}: ${dataStr}`);
+      } else {
```
sangaman: I'm not sure if there's value in logging the raw traffic, even on a trace level. I figure it will be incomprehensible and that it will quickly flood logs/consoles when there's network activity.

lib/p2p/Peer.ts
```diff
@@ -604,69 +644,126 @@ class Peer extends EventEmitter {
...
+    const body = packet.body!;
+    const { sign, ...bodyWithoutSign } = body;
+    const { nodePubKey } = body.nodeState; // the peer pubkey
+    const { peerPubKey } = body; // our own pubkey
 
-    this.sendPacket(packet);
+    // verify that the msg was signed the for us
+    if (peerPubKey !== nodeKey.nodePubKey) {
+      this.close(DisconnectionReason.AuthFailureInvalidTarget);
+      throw errors.AUTH_FAILURE_INVALID_TARGET(nodePubKey, peerPubKey);
```
sangaman: Should we penalize a peer for this?

lib/p2p/Peer.ts
```diff
@@ -604,69 +644,126 @@ class Peer extends EventEmitter {
...
-  private initHello = async (handshakeData: HandshakeState) => {
-    const packet = new packets.HelloPacket(handshakeData);
+  private authenticate = (packet: packets.SessionInitPacket, nodeKey: NodeKey) => {
+    const body = packet.body!;
+    const { sign, ...bodyWithoutSign } = body;
+    const { nodePubKey } = body.nodeState; // the peer pubkey
+    const { peerPubKey } = body; // our own pubkey
 
-    this.sendPacket(packet);
+    // verify that the msg was signed the for us
```
sangaman: Can drop "the" in "was signed the for us", and maybe say "was intended for us" since it's possible it wasn't signed at all

lib/p2p/Framer.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * Frame a packet with a header to be used as a wire msg
+   */
+  public frame = (packet: Packet, encryptionKey?: Buffer): Buffer => {
+    const payload = packet.toRaw();
+    const msg = Buffer.allocUnsafe(Framer.MSG_HEADER_LENGTH + payload.length);
+
+    // network magic value
+    msg.writeUInt32LE(this.network.magic, 0, true);
```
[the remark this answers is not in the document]
moshababo: Yes, but I thought it will make things simpler so that encryption would just be a wrapper on the msg.

lib/p2p/Framer.ts
```diff
@@ -0,0 +1,155 @@
...
+  public unframe = (data: Buffer, encryptionKey?: Buffer): WireMsg => {
+    let msg;
+    if (!encryptionKey) {
+      msg = data;
+    } else {
+      const length = data.readUInt32LE(0, true);
+      const ciphertext = data.slice(4);
+
+      if (length !== ciphertext.length) {
+        throw new Error('invalid ciphertext length');
```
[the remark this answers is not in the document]
moshababo: Yes, it was something I didn't got to yet. Did it in the last commit (`error handling`)

lib/p2p/packets/Packet.ts
moshababo: I remember having issues with the inheritance after using arrow functions, but I just checked and it's not the case. I'll change it (and `checksum` to a getter).

test/unit/Parser.spec.ts
moshababo: It was fixed in the last commit.

lib/p2p/Peer.ts
moshababo: Good call, will change it. 

I'm afraid we're also not verifying that the response packet is of the expected type. This could lead to runtime errors. i'll open an issue for it.

lib/p2p/Peer.ts
moshababo: Sorry, I added it for myself and forgot removing.

lib/p2p/Peer.ts
moshababo: Yes, i'll add it.

lib/p2p/packets/Packet.ts
moshababo: I changed `checksum` to a property, since it's to be calculated only in the constructor.

moshababo: In regards to `checksum` - I'd rather not have a getter which is calculating stuff. Making it a property isn't good either, because it's supposed to be only for the wire protocol. So I ended up leaving it as a function.

moshababo:
the p2p sanity tests are failing due to late activation of encryption mode on the parser. I'm going to try fix it tomorrow (feel free to look into it). Wanted to finalize other things before. 

Still open:
* fix `verifyReachability` (can be done in a follow up PR) 
* sha256 -> md5
* reconsider the design of two-sided SessionInit/SessionAck (been busy in implementing recently, need to re-think about this).

lib/p2p/Framer.ts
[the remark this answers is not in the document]
sangaman: Gotcha, I think it'd be preferable not to include it when it's not needed. Same with the checksum I'm thinking as well, I don't think it's needed when we're encrypting the message because any corrupted bits will distort the entire message anyway. I was going to suggest leaving it as an optimization for later, but it'd be kind of a hassle to do later because it's a breaking change.

lib/p2p/packets/Packet.ts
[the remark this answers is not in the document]
sangaman: OK thanks that works for me.

lib/p2p/Pool.ts
```diff
@@ -122,7 +126,7 @@ class Pool extends EventEmitter {
       this.logger.error('Unexpected error connecting to known peers on startup', reason);
     });
 
-    this.verifyReachability();
+    // this.verifyReachability();
```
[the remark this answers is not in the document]
sangaman: Alright no problem.

lib/p2p/Peer.ts
sangaman: Right, we're only checking that the packet id matches what we expect. This crossed my mind as well and I agree it would be good to handle the case when the packet we receive doesn't have the type we expect.

sangaman:
I like the session init/ack handshake, why would you reconsider it?

lib/p2p/Peer.ts
moshababo: I opened #788

moshababo:
I managed to find the problem with the late activation of in-encryption - it was that `Peer.wait` was relying on promise resolve, and so it wasn't synchronous. It was solved by introducing callbacks to `PendingResponseEntry`, in addition to the promises, when synchronous resolution is needed.

moshababo:
I found the problem with the late activation of in-encryption - it was that `Peer.wait` was relying on promise resolve, and so it wasn't synchronous. It was solved by introducing callbacks to `PendingResponseEntry`, in addition to the promises, to be used when synchronous resolution is needed.

lib/p2p/Framer.ts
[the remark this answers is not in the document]
moshababo: Sounds fine, i'm gonna work on it now.

moshababo:
@sangaman conflicts resolved, waiting for your review now.

moshababo:
Yes, but nothing which includes swaps - just basic communication.

sangaman:
Cool, I have previously set up two instances of `xud-simnet` nodes so I'm going to try updating them to this branch and testing a swap between them for good measure, then we can merge.

sangaman:
OK, I tested this branch with a swap and it worked flawlessly. Great work @moshababo.

kilrau:
Very nice @moshababo

PR #757, opened 2018-12-20
Version Compatibility Check #154

### LINKED ISSUE #154: Advertise version
Advertise `xud` `version` on handshaking.
When received, verify that it exceeds `MIN_VERSION`. If not, disconnect from peer.

rsercano:
Hello,

hopefully that resolves #154


rsercano:
Actually this PR includes the commits from #746, I forgot to create a new branch before #746 and that messed up a bit, I can close this one or the other one depending on your request.

lib/p2p/Peer.ts
```diff
@@ -173,19 +177,23 @@ class Peer extends EventEmitter {
...
 
-    if (this.nodePubKey === handshakeData.nodePubKey) {
+    if (this.nodePubKey === handshakeState.nodePubKey) {
       this.close(DisconnectionReason.ConnectedToSelf);
       throw errors.ATTEMPTED_CONNECTION_TO_SELF;
     }
 
-    // TODO: Check that the peer's version is compatible with ours
+    // Check version compatibility
+    if (this.version !== handshakeState.version) {
```
moshababo: not every version update will be non-compatible. We need to define `MIN_VERSION`, and verify the compatibility by converting the string to numbers (or something similar to that).

kilrau: As Daniel said in https://github.com/ExchangeUnion/xud/issues/154#issuecomment-436832861  we need a separate field indicating a min compatible version. We will set this manually with every release. The `MIN_VERSION` will be = current version if the current version introduces a significant breaking change e.g. on p2p level. I don't see any other way apart from doing it manually on release tag.

So instead of `if (this.version !== handshakeData.version)` something like `if (this.minversion !== handshakeData.version)` @rsercano

kilrau: and take care we cannot forget to set `MIN_VERSION` when tagging. Any idea about this? @sangaman

moshababo: I’d suggest to define `minCompatibleVersion` field in `package.json` and read it from there in `Xud.ts`, same as with `version`. 
We just need to upgrade it when we're upgrading `version`, before a non-compatible release.

Both the min version and the peer version strings should be parsed, and get evaluated so that the peer version needs to be higher or equal to the min version.

@kilian are we planning to go from v1.0.0-alpha.x to v1.0.0?

moshababo: When we'll get to v1.0.0, we want to increment the first number for incompatible changes release, instead of the min version field?

rsercano: @moshababo actually that was my entry point for the my first question; 
`would it be sufficient to check major version numbers ?`

kilrau: As a general rule per [semver](https://semver.org/), once we are at v1.0.0 we'll increment MAJOR in vMAJOR.MINOR.PATCH for every breaking change that effects the majority of users. Backwards-compatible or breaking changes for a small subset of users might go via MINOR, we'll use our judgement on a case by case base.

Nevertheless this is just for the sake of readability, your check shouldn't care about this and just go for the `minCompatibleVersion` field @rsercano 

Btw, I got zero feeling for staying in low version numbers as long as possible (like bitcoin) a lot of major software products nowadays are in double digits which is perfectly fine. But of course we should reduce breaking changes to a minimum once we are at v1.0.0.

rsercano: Alright, I've implemented semantic version check which has a logic as below;

- 1.0.0 < 2.0.0 < 2.1.0 < 2.1.1
- 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0

lib/p2p/Peer.ts
```diff
@@ -11,6 +11,9 @@ import { Packet, PacketDirection, PacketType } from './packets';
 import { HandshakeState, Address, NodeConnectionInfo } from '../types/p2p';
 import errors from './errors';
 import addressUtils from '../utils/addressUtils';
+import { compare } from 'semver';
+
+const minCompatibleVersion: string = require('../../package.json').minCompatibleVersion;
```
moshababo: this should be imported together with `version` in `Xud.ts`, and get passed in a similar way to `Pool`.

bin/xud
```diff
@@ -109,10 +109,14 @@ const { argv } = require('yargs')
       default: undefined,
     },
     'p2p.discover': {
-      describe: 'Automatically discover and connect to new nodes',
+      describe: 'Automatically discover and connect to new nodes with discoverminutes interval',
```
moshababo: would you be able to separate this PR commits from #746 ones?

lib/p2p/errors.ts
```diff
@@ -31,6 +31,10 @@ const errors = {
     message: `node at ${address} sent pub key ${nodePubKey}, expected ${expectedNodePubKey}`,
     code: errorCodes.UNEXPECTED_NODE_PUB_KEY,
   }),
+  INCOMPATIBLE_VERSION: (address: string, minCompatibleVersion: string, incomingPeerVersion: string | undefined) => ({
```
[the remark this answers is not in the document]
rsercano: sure

lib/p2p/Peer.ts
```diff
@@ -188,7 +193,27 @@ class Peer extends EventEmitter {
       throw errors.ATTEMPTED_CONNECTION_TO_SELF;
     }
 
-    // TODO: Check that the peer's version is compatible with ours
+    // Check version compatibility
+    // dev.note: compare returns 0 if v1 == v2, or 1 if v1 is greater, or -1 if v2 is greater.
+    if (compare(handshakeData.version, minCompatibleVersion) === -1) {
```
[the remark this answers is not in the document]
rsercano: correct, my bad.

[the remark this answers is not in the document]
rsercano: was thinking about it, will add some.

lib/p2p/Peer.ts
```diff
@@ -11,6 +11,9 @@ import { Packet, PacketDirection, PacketType } from './packets';
 import { HandshakeState, Address, NodeConnectionInfo } from '../types/p2p';
 import errors from './errors';
 import addressUtils from '../utils/addressUtils';
+import { compare } from 'semver';
```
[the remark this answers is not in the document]
rsercano: right

rsercano:
Please ignore this, #772 is the real one.

lib/p2p/Peer.ts
kilrau: Looks right to me.

PR #759, opened 2018-12-21
Fix peer stalling

### LINKED ISSUE #758: Peers disconnect
@Moshe @sangaman Looks like the latest master is faulty. After connecting nodes, they are disconnected very fast (15 seconds) while claiming for  `stalling`. This is urgent as xud-simnet is basically not working now

Can be seen between xud1 and xud3.   or any other peer running the latest master and connected to these two. Note that xud2 is using an old master and has no issue.

sangaman:
This includes several fixes I uncovered while trying to figure out what was causing #758, as well as logging enhancements to help debug p2p issues. Each commit has its own description, but the key one is the last one:

> This fixes a bug where the hash calculated for outgoing Order and Orders packets wouldn't match the calculated hash for the packet received by the peer due to the order objects being hashed having extraneous properties that are not transmitted to the peer, namely `hold` and `initialQuantity`.

I've tested the fix between two peers using this branch and they don't stall and disconnect from each other any more, but it would be helpful if we could try this on the seed nodes before merging.

Fixes #758.

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -303,7 +303,10 @@ class Peer extends EventEmitter {
         }
         this.socket!.removeListener('error', onError);
         this.socket!.removeListener('connect', onConnect);
-        this.retryConnectionTimer = undefined;
+        if (this.retryConnectionTimer) {
+          clearTimeout(this.retryConnectionTimer);
+          this.retryConnectionTimer = undefined;
+        }
       };
 
       const onConnect = () => {
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -388,6 +388,7 @@ class Peer extends EventEmitter {
     for (const [packetId, entry] of this.responseMap) {
       if (now > entry.timeout) {
         this.emitError(`Peer is stalling (${packetId})`);
+        entry.reject('response timed out');
         this.close(DisconnectionReason.ResponseStalling, packetId);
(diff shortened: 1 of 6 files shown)


offerm:
@sangaman 
Can't find time for code review. 
Deployed to XUD test servers and it looks like the PR is solving the problem.

For now, the XUD test servers are using this branch.

lib/p2p/Peer.ts
```diff
@@ -496,15 +487,15 @@ class Peer extends EventEmitter {
 
       switch (err.type) {
         case ParserErrorType.InvalidPacket:
-          this.logger.warn(`parser: invalid peer packet: ${err.payload}`);
+          this.logger.error(`parser: invalid peer packet: ${err.payload}`);
           this.emit('reputation', ReputationEvent.InvalidPacket);
           break;
         case ParserErrorType.UnknownPacketType:
-          this.logger.warn(`parser: unknown peer packet type: ${err.payload}`);
+          this.logger.error(`parser: unknown peer packet type: ${err.payload}`);
```
moshababo: i'd say that `error` is for problems in the local node, while here it's most likely detecting a problem in the peer node.

sangaman: OK, that makes sense. I'll change it back.

sangaman:
@moshababo Yes exactly, that's the key line. Removing those two properties from what gets hashed.

PR #761, opened 2018-12-21
refactor(proto): follow proto3 style for p2p proto

sangaman:
This applies underscore_snake_case to all field names in the p2p proto definition as per the proto3 Style Guide. As a result, the generated code follows camelCase variable naming which is consistent with the rest of the code base.


sangaman:
Going to resolve conflicts and then merge, thanks.

Issue #573, opened 2018-10-09
Mark lnd as Disconnected when calls to it fail


sangaman:
We ought to handle the case where calls to lnd fail, whether it be a single failure or some number of multiple consecutive failures. The reason for this is that when lnd goes down for any reason, we lose the ability to perform any swaps for that lnd's currency. If we continue to place orders and attempt swaps for any trading pairs involving that currency, we will cause many swap failures and peers may ban us. Handling an lnd instance that goes down may involve canceling all our orders for any trading pairs affected by that instance, and not accepting new orders until we reestablish connectivity.

We may also consider periodically calling `getInfo` on lnd to ensure that we get a timely response and that the state is nomral, `synced_to_chain` is true, and so on.

sangaman:
Yes, we probably need further discussion on the afterthought of periodically pinging `lnd` to make sure it's working, but the first part of disabling the LndClient if calls to `lnd` fail due to connectivity errors or anything we don't expect to recover from automatically can be worked.

kilrau:
Please prepare the code to include pings to raiden and other swap nodes, not only LND. @ImmanuelSegol

sangaman:
I'd recommend opening a PR just for the lnd side of  things first.

kilrau:
Agree, let's do that @ImmanuelSegol

kilrau:
use `SubscribeInvoices` streaming call to detect if lnd is down

PR #762, opened 2018-12-23
feat: listen to subscribeInvoices to verify lnd connection

ImmanuelSegol:
#573

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -1,4 +1,4 @@
-import grpc, { ChannelCredentials } from 'grpc';
+import grpc, { ChannelCredentials, ClientReadableStream } from 'grpc';
 import Logger from '../Logger';
 import BaseClient, { ClientStatus } from '../BaseClient';
 import errors from './errors';
@@ -53,7 +53,7 @@ class LndClient extends BaseClient {
   private credentials!: ChannelCredentials;
   private reconnectionTimer?: NodeJS.Timer;
   private identityPubKey?: string;
-
+  private invoiceSubscription!: ClientReadableStream<lndrpc.InvoiceSubscription>;
   /** Time in milliseconds between attempts to recheck connectivity to lnd if it is lost. */
   private static readonly RECONNECT_TIMER = 5000;
 
@@ -66,7 +66,6 @@ class LndClient extends BaseClient {
    */
   constructor(private config: LndClientConfig, logger: Logger) {
     super(logger);
-
     this.cltvDelta = config.cltvdelta || 0;
   }
 
@@ -177,7 +176,6 @@ class LndClient extends BaseClient {
     if (!this.isConnected()) {
       this.logger.info(`trying to verify connection to lnd with uri: ${this.uri}`);
       this.lightning = new LightningClient(this.uri, this.credentials);
-
       try {
         const getInfoResponse = await this.getInfo();
         if (getInfoResponse.getSyncedToChain()) {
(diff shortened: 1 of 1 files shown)


lib/lndclient/LndClient.ts
```diff
@@ -290,6 +289,21 @@ class LndClient extends BaseClient {
...
+   */
+  private subscribeInvoices = (): void => {
+    if (this.invoiceSubscription) {
+      this.invoiceSubscription.cancel();
+    }
+
+    this.invoiceSubscription = this.lightning.subscribeInvoices(new lndrpc.InvoiceSubscription(), this.meta)
+    .on('error', (error) => {
+      this.setStatus(ClientStatus.Disconnected);
+      this.logger.error(`lnd has been disconnected at ${this.uri}, error: ${error}`);
```
sangaman: We want to start attempt to reconnect after this I believe, like so `this.reconnectionTimer = setTimeout(this.verifyConnection, LndClient.RECONNECT_TIMER);`

lib/lndclient/LndClient.ts
```diff
@@ -66,7 +66,6 @@ class LndClient extends BaseClient {
    */
   constructor(private config: LndClientConfig, logger: Logger) {
     super(logger);
-
```
sangaman: We have blank lines after all calls to `super` in the codebase so we should keep this consistent.

lib/lndclient/LndClient.ts
```diff
@@ -177,7 +176,6 @@ class LndClient extends BaseClient {
     if (!this.isConnected()) {
       this.logger.info(`trying to verify connection to lnd with uri: ${this.uri}`);
       this.lightning = new LightningClient(this.uri, this.credentials);
-
```
sangaman: Personally I think this blank line improves readability so I'd prefer to keep it.

lib/lndclient/LndClient.ts
```diff
@@ -53,7 +53,7 @@ class LndClient extends BaseClient {
   private credentials!: ChannelCredentials;
   private reconnectionTimer?: NodeJS.Timer;
   private identityPubKey?: string;
-
```
sangaman: Here too I'd prefer to keep the blank line separating `private` from `private static` fields.

lib/lndclient/LndClient.ts
```diff
@@ -290,6 +289,21 @@ class LndClient extends BaseClient {
...
+   * Subscribe to invoices.
+   */
+  private subscribeInvoices = (): void => {
+    if (this.invoiceSubscription) {
+      this.invoiceSubscription.cancel();
+    }
+
+    this.invoiceSubscription = this.lightning.subscribeInvoices(new lndrpc.InvoiceSubscription(), this.meta)
+    .on('error', (error) => {
+      this.setStatus(ClientStatus.Disconnected);
```
sangaman: I think it would make sense to set `this.invoiceSubscription` to `undefined` here, so we don't try to close a stream that has already been canceled.

lib/lndclient/LndClient.ts
```diff
@@ -53,7 +53,7 @@ class LndClient extends BaseClient {
   private credentials!: ChannelCredentials;
   private reconnectionTimer?: NodeJS.Timer;
   private identityPubKey?: string;
-
+  private invoiceSubscription!: ClientReadableStream<lndrpc.InvoiceSubscription>;
```
sangaman: I don't think we need the `!` operator here, but rather the nullable `?` operator to indicate that sometimes we won't have an active invoice subscription. You already have null checks where this is being used, which is good, so I don't think you'd need to change anything else.

PR #770, opened 2018-12-25
Sending GET_NODES scheduled #402

### LINKED ISSUE #402: Share newly connected peers dynamically at runtime

rsercano:
resolves #402, please ignore #746


moshababo:
You can move the `PoolConfig` type to `lib/types/p2p.ts` to avoid the circular dependency.

rsercano:
Fixed !

PR #773, opened 2018-12-25
Swap failureReason field for SwapDeal db model and SwapFailedPacket

sangaman:
This is a followup to #753 which includes breaking changes. Here's an overview of what it changes:

- Adds the `failureReason` field to the `SwapDeal` database model as well as the `SwapFailed` packet. It removes the `PeerFailedSwap` failure reason, and assigns swaps the failure reason specified by the peer in cases when the peer fails the swap. It adds a `PaymentHashReuse` reason which was missing to cover all current swap failure reasons.
- Tweaks the Parser tests to verify parsed packets with the
stable `stringify` method we use when hashing packets. This was done to
work around issues where the parsed packet did not deeply equal the
original packet because of the presence of an undefined `failureReason`.
- Renames the `errorReason` swap deal property to `errorMessage` to help differentiate it from `failureReason` and describe it as a field for details about why a swap failed or what caused an error. It also makes it optional on the `SwapFailed` packet to accommodate cases where there are no relevant details to include besides the `failureReason`.

### DIFF
--- a/lib/p2p/packets/types/SwapFailedPacket.ts
+++ b/lib/p2p/packets/types/SwapFailedPacket.ts
@@ -1,13 +1,12 @@
 import Packet, { PacketDirection } from '../Packet';
 import PacketType from '../PacketType';
 import * as pb from '../../../proto/xudp2p_pb';
-import SwapCompletePacket from './SwapCompletePacket';
 import { removeUndefinedProps } from '../../../utils/utils';
 
 // TODO: proper error handling
 export type SwapFailedPacketBody = {
   rHash: string;
-  errorMessage: string;
+  errorMessage?: string;
 };
 
 class SwapFailedPacket extends Packet<SwapFailedPacketBody> {
@@ -42,10 +41,10 @@ class SwapFailedPacket extends Packet<SwapFailedPacketBody> {
         hash: obj.hash,
         reqId: obj.reqId || undefined,
       }),
-      body: {
+      body: removeUndefinedProps({
         rHash: obj.rHash,
-        errorMessage: obj.errorMessage,
-      },
+        errorMessage: obj.errorMessage || undefined,
+      }),
     });
   }
 
@@ -55,7 +54,9 @@ class SwapFailedPacket extends Packet<SwapFailedPacketBody> {
     msg.setHash(this.header.hash!);
     msg.setReqId(this.header.reqId!);
     msg.setRHash(this.body!.rHash);
-    msg.setErrorMessage(this.body!.errorMessage);
+    if (this.body!.errorMessage) {
+      msg.setErrorMessage(this.body!.errorMessage!);
+    }
 
     return msg.serializeBinary();
   }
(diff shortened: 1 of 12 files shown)


lib/p2p/packets/types/SwapFailedPacket.ts
```diff
@@ -44,7 +45,8 @@ class SwapFailedPacket extends Packet<SwapFailedPacketBody> {
       }),
       body: {
```
moshababo: you can wrap this with `removeUndefinedProps` and undo the parser testing workaround, no?

lib/p2p/packets/types/SwapFailedPacket.ts
```diff
@@ -55,7 +57,10 @@ class SwapFailedPacket extends Packet<SwapFailedPacketBody> {
     msg.setHash(this.header.hash!);
     msg.setReqId(this.header.reqId!);
     msg.setRHash(this.body!.rHash);
-    msg.setErrorMessage(this.body!.errorMessage);
+    if (this.body!.errorMessage) {
```
moshababo: this case looks redundant, and to produce the same thing without it. If you wanted to have it anyway, you can leave it in.

proto/xudp2p.proto
```diff
@@ -118,4 +118,5 @@ message SwapFailedPacket {
     string hash = 3 [json_name = "hash"];
     string r_hash = 4 [json_name = "r_hash"];
     string error_message = 5 [json_name = "error_message"];
+    uint32 failure_reason = 6;
```
moshababo: `json_name` is missing

sangaman: I can add it in, but I wasn't sure if it makes a difference here. Like I don't know if anyone is going to be looking at these in json format, since it's all machine to machine. I'd followed this in the main rpc proto definition to be consistent with lnd, not sure if it also makes sense here.

lib/p2p/packets/types/SwapFailedPacket.ts
sangaman: Isn't it needed to cover the case of `errorMessage` being undefined?

lib/p2p/packets/types/SwapFailedPacket.ts
sangaman: Got it, let me try that.

sangaman:
Yes I think it'd be good to have in the next release due to the breaking changes. I will test this soon between two simnet nodes, merge it if all is good, and then do more testing against master before we tag a release.

lib/p2p/packets/types/SwapFailedPacket.ts
moshababo: AFAIK if it's undefined or an empty string then it won't affect the pb serialization anyway.

proto/xudp2p.proto
moshababo: I see, so let's decide about that soon but until then i'd rather have consistency, otherwise it might suggest that something is unique about this particular field.

sangaman: Ok, I'll make it consistent for now but revisit this later.

lib/p2p/packets/types/SwapFailedPacket.ts
sangaman: I think you're right but my only concern is that the method signature for `setErrorMessage` expects a `string` parameter, not `string | undefined`, so I'd prefer to avoid passing `undefined` just to be safe.

Issue #767, opened 2018-12-24
remove order is not being send to peers


offerm:
My node (A) is connected to xud1,2,3.

listorders shows 2 x sell 2.5@0.0079 (one order from xud1 and one order with xud2).

Using CLI: 
```
xucli buy 2.5 LTC/BTC 0.0079
```
I swap with one of the nodes (xud1 in this case):
```
swapped 2.5 LTC with peer order 29dff850-07c0-11e9-a5e9-eba6222927d0
```
Post swap the sell order is replaced by the bot on xud1 and my node shows again 2 sell orders.

checking `xucli getorders` on xud3 shows 3 x  sell 2.5@0.0079 (one from xud3 and TWO from xud1).

The log does not show any remove order from xud1.

Very easy to reproduce.

sangaman:
Looks like this is a similar bug/cause as #758, where superfluous fields are being hashed by the sender for the Order Invalidation packet and then failing validation on the receiver's end, thus the orders are not being invalidated by the receiver.

offerm:
Not solved. 
Now, not only that the executed order is not being removed from peers the counter order is also added with a price of zero.

<img width="898" alt="screen shot 2018-12-26 at 8 14 40" src="https://user-images.githubusercontent.com/9712754/50434568-5ac6f780-08e6-11e9-9d88-b2b0ed590c7c.png">

sangaman:
Looks like this is the key change that needed to happen in addition to my fix above, OrderInvalidationPackets were being converted into the wrong type of packet. https://github.com/ExchangeUnion/xud/pull/774/commits/27e289bff72ca0c137430e843d01082bfa7ebd3d

Thanks offer for finding this.

PR #774, opened 2018-12-26
fix(p2p): order invalidation packet

moshababo:
* fix typing mismatch in order invalidation conversion return type
* fix the tests to be able to find such problems
* adding data integrity errors logging/penalization (taken from #756)

Still need to check if it actually solves #767 or that we have another problem.

### DIFF
--- a/test/unit/Parser.spec.ts
+++ b/test/unit/Parser.spec.ts
@@ -36,8 +36,9 @@ describe('Parser', () => {
     return new Promise((resolve, reject) => {
       wait(packets.length)
         .then((parsedPackets) => {
-          for (let i = 0; i <= packets.length; i += 1) {
+          for (let i = 0; i < packets.length; i += 1) {
             expect(packets[i]).to.deep.equal(parsedPackets[i]);
+            expect(packets[i].type).to.equal(parsedPackets[i].type);
           }
           resolve();
         })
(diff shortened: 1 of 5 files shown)


test/unit/Parser.spec.ts
```diff
@@ -36,8 +36,9 @@ describe('Parser', () => {
     return new Promise((resolve, reject) => {
       wait(packets.length)
         .then((parsedPackets) => {
-          for (let i = 0; i <= packets.length; i += 1) {
+          for (let i = 0; i < packets.length; i += 1) {
             expect(packets[i]).to.deep.equal(parsedPackets[i]);
+            expect(packets[i].type).to.equal(parsedPackets[i].type);
```
sangaman: Isn't this redundant? The deep equal check above should also check `type`.

moshababo: Without it the test would have passed (while having the bug) - haven't looked much into it, but probably because `type` is a getter and so a function, not a plain field.

moshababo:
I was thinking to squash the commits, but i'll see if I can easily rename the commits although they were pushed, and merge without squash.

test/unit/Parser.spec.ts
sangaman: Ohhh that's right, that makes sense. That's fine.

sangaman:
Squash merging is fine too, however you think is best but just to keep the template format. Thanks!

PR #775, opened 2018-12-26
Added date formatting for logger via config param #686

### LINKED ISSUE #686: Enhance log to include milliseconds
Enhance log output to inlcude milliseconds in each printing.

Today we only should the time up to the seconds.

rsercano:
resolves #686, please let me know for any changes that you wish

### DIFF
--- a/lib/utils/utils.ts
+++ b/lib/utils/utils.ts
@@ -4,6 +4,7 @@ import { assert } from 'chai';
 import { Pair } from '../types/orders';
 import crypto from 'crypto';
 import { promisify } from 'util';
+import moment from 'moment';
 
 export type UriParts = {
   nodePubKey: string;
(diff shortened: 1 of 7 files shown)


rsercano:
@kilrau set default date format as LND !

lib/utils/utils.ts
```diff
@@ -75,9 +76,9 @@ export const isEmptyObject = (val: any): boolean => {
   return isObject(val) && Object.keys(val).length === 0;
 };
 
-/** Get the current date in the LocaleString format.
+/** Get the current date in the given dateFormat, if not provided formats with `DD/MM/YYYY HH:mm:ss.SSS`.
  */
-export const getTsString = (): string => (new Date()).toLocaleString();
+export const getTsString = (dateFormat?: string): string => moment().format(dateFormat ? dateFormat : 'YYYY-MM-DD hh:mm:ss.sss');
```
moshababo: the default format is a bit different from the one written in the comment.

also, you can write it as `dateFormat || 'YYYY...'` instead of the ternary operator.

rsercano: Good point !

moshababo:
@sangaman take a look if you get to it, if not i'll merge on Sunday.

Issue #622, opened 2018-10-31
Limit Order (IOC)


offerm:
EDIT by @kilrau : we want IOC only for now: https://github.com/ExchangeUnion/xud/issues/622#issuecomment-557157547

I suggest to support FOK & IOC with limit  orders.

Use case:

with getorders I see a peer order that I wish to take.
since I never use market orders :-) I issued a limit order targeting the peer order.

The matching with the peer order failed for some reason (peer order does not exists anymore, swap failed, etc), As a result my take order entered the book and broadcasted to all my peers. This was not my intention.
FOK - fill the order or kill it.
IOC - fill immediate or cancel 

reference - http://optionstradingbeginner.blogspot.com/2009/01/fill-or-kill-fok-immediate-or-cancel.html

kilrau:
... which we could take and simulate market orders if we wanted/needed to. Other thoughts? @sangaman @moshababo

moshababo:
Sounds good to me.

sangaman:
This would be simple enough to do, it's basically a limit order (with a price) where we set `discardRemaining` to true. It can be an optional flag on the `placeOrder` request.

kilrau:
Let me know what you think @offerm @moshababo @sangaman 

**Summary** 

`placeOrder` with flag `fok` (Limit Fill-or-Kill, make sure `fok` flag doesn't work with Market):
search order book for match of total quantity (no partial matching allowed)
    if match found & swap successful
        remove matched order as usual
    else cancel order and return discarded remaining order

`placeOrder` with flag `ioc` (Limit Immediate-or-Cancel, make sure `ioc` flag doesn't work with Market):
search order book for match (partial matching allowed)
    for match found & swap successful
        remove matched order as usual
    `discardRemaining` quantity, nothing enters the order book

CLI UX Example FOK:
```
$ xucli buy 0.0001 LTC/BTC 1.1 --fok
{
  "internalMatchesList": [],
  "swapResultsList": [],
  "remainingOrder": {
    "price": 1.1,
    "quantity": 0.0001,
    "pairId": "LTC/BTC",
    "id": "0298b691-e26f-11e8-bc86-4938fbce63e9",
    "peerPubKey": "",
    "localId": "0298b690-e26f-11e8-bc86-4938fbce63e9",
    "createdAt": 1541582704761,
    "side": 0,
    "isOwnOrder": true,
    "discarded": true
  }
}
```
CLI example IOC:
```
[...]

PR #777, opened 2019-01-03
New Order Types FOK/IOC #622

rsercano:
Hello guys,

this is an initial commit for #622, so it's not finished yet. Even though I don't feel like I'm lost yet, but feel like gotta lose soon, so please let me know what you think.

I've changed `buy` and `sell` commands a bit, now I guess they look more meaningful, instead of controlling `price` arg to decide now we accept another arg for it, so it can be `lmt`, `ioc`, `fok`, `mkt` which's clearer I guess. 

It would be great if you can give me some directions from here, just maybe shorten to understand further, thanks in advance !

p.s. I'm aware tests are failing, just wanted some directions from you guys.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -58,6 +58,7 @@
   
     - [AddCurrencyRequest.SwapClient](#xudrpc.AddCurrencyRequest.SwapClient)
     - [OrderSide](#xudrpc.OrderSide)
+    - [OrderType](#xudrpc.OrderType)
     - [SwapResult.Role](#xudrpc.SwapResult.Role)
   
   
@@ -610,6 +611,7 @@
 | pair_id | [string](#string) |  | The trading pair that the order is for. |
 | order_id | [string](#string) |  | The local id to assign to the order. |
 | side | [OrderSide](#xudrpc.OrderSide) |  | Whether the order is a Buy or Sell. |
+| type | [OrderType](#xudrpc.OrderType) |  | Whether the order type is a IOC or FOK. |
 
 
 
@@ -860,6 +862,20 @@
 | ---- | ------ | ----------- |
 | BUY | 0 |  |
 | SELL | 1 |  |
+
+
+
+<a name="xudrpc.OrderType"></a>
+
+### OrderType
+
+
+| Name | Number | Description |
+| ---- | ------ | ----------- |
+| IOC | 0 |  |
+| FOK | 1 |  |
+| LIMIT | 2 |  |
+| MARKET | 3 |  |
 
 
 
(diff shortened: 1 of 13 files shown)


sangaman:
For the command line, I feel like the `fok`/`ioc` orders would probably not be as common. I'd be hesitant to change the regular way to add limit orders. Maybe add `--fok` and `--ioc` flags that can be set manually? I'm actually thinking it might make sense to have `[stream]` be a flag like `--stream` in the same way as well.

Fill-or-kill will be complicated to implement properly I think, but you can definitely go for it. If you want to just do `ioc` first that might be easier.

sangaman:
One thing that we'd need to change for `fok` is allowing multiple swap deals with the same `rHash`. Right now it's basically the key by which we identify swap deals. It'd require db changes as well as changes to several places in the code. We could maybe make swaps unique according to the pair of `rHash` and the maker `orderId`.

kilrau:
> For the command line, I feel like the `fok`/`ioc` orders would probably not be as common. I'd be hesitant to change the regular way to add limit orders. Maybe add `--fok` and `--ioc` flags that can be set manually?

Yep, separate flags with the regular limit order commands NOT changing (a bit messy on gitter, check chats from Jan 04)

offerm:
Why we need multiple swap deals having the same hash?

sangaman:
> Why we need multiple swap deals having the same hash?

@offerm I think I was mistaken, after thinking about this some more it shouldn't be necessary. But I figure we would need to make sure that we have HTLC's to us for all portions of the `fok` order before releasing the preimages, which is still a significant change.

sangaman:
@kilrau I figure we would need to find all matches and get all swaps accepted and up to the point where the HTLCs are set up but right before we release the preimage. We can't execute any trades before we get to that point, and if we can't get there we have to discard the trade entirely.

It would also be tricky because we'd need to hold off on executing the internal match trades immediately as we always do now. I'm actually not sure how to go about this part, because if we find 2 matches, one with an internal order and one with an external order, we can't execute the internal order until the swap for the external order is accepted and set up. But since setting up the swap is not instant, in the meantime it's possible that the internal order gets canceled or filled by a different order.

In other words I think fill-or-kill orders introduce a great deal of complexity and probably shouldn't be a priority.

kilrau:
Agree, let's move [this](https://github.com/ExchangeUnion/xud/issues/622) to `alpha.11`

kilrau:
According to @rsercano ([comment](https://github.com/ExchangeUnion/xud/issues/622#issuecomment-457147216)) close this PR?

Issue #687, opened 2018-11-18
streamOrders - wait for XUD


offerm:
Enhance streamorders CLI in the following ways:

a. allow it to start before `xud`. `streamorders` should wait for `xud` to start and not issue an error. It is OK to issue a warn/info for waiting for XUD. 

b. print a line upon connection

c. upon termination of `xud`  or any other RPC error - reconnect to `xud` (wait if `xud` is not available.

d. upon connection to xud - print existing orders.

With these changes it will be easier to track xud internals.

PR #780, opened 2019-01-07
improved streamOrders #687

rsercano:
Hello again,

I was trying to work on new order types but since there're lots of struggles there I wanted to focus something else, as requested by @offerm I've implemented all the following changes;

- [x] a. allow it to start before xud. streamorders should wait for xud to start and not issue an error. It is OK to issue a warn/info for waiting for XUD.
- [x] b. print a line upon connection 
- [x] c. upon termination of xud or any other RPC error - reconnect to xud (wait if xud is not available. 
- [x] d. upon connection to xud - print existing orders.

**In a technical overview**;  I've implemented those by calling `GetInfo` to ensure connection and if there's no `xud` node running then I'm trying to connect in `3 seconds` (let me know if you want to adjust time) This seemed to be easiest way to me to implement such logic. Btw. error messages are not being printed every 3 seconds it's only being printed at first.

### DIFF
--- a/lib/cli/commands/subscribeorders.ts
+++ b/lib/cli/commands/subscribeorders.ts
@@ -15,6 +15,22 @@ export const builder = {
 };
 
 export const handler = (argv: Arguments) => {
+  ensureConnection(argv, true);
+};
+
+const ensureConnection = (argv: Arguments, printError?: boolean) => {
+  loadXudClient(argv).getInfo(new xudrpc.GetInfoRequest(), (error: Error | null) => {
+    if (error) {
+      if (printError) console.error(`${error.name}: ${error.message}`);
+      setTimeout(ensureConnection.bind(undefined, argv), 3000);
+    } else {
+      console.log('Successfully connected, subscribing for orders');
+      subscribeOrders(argv);
+    }
+  });
+};
+
+const subscribeOrders = (argv: Arguments) =>  {
   const addedOrdersRequest = new xudrpc.SubscribeAddedOrdersRequest();
   addedOrdersRequest.setExisting(argv.existing);
   const addedOrdersSubscription = loadXudClient(argv).subscribeAddedOrders(addedOrdersRequest);
@@ -22,13 +38,28 @@ export const handler = (argv: Arguments) => {
     console.log(`Order added: ${JSON.stringify(order.toObject())}`);
   });
 
+  // adding end, close, error events only once,
+  // since they'll be thrown for three of subscriptions in the corresponding cases, catching once is enough.
+  addedOrdersSubscription.on('end', ensureConnection.bind(undefined, argv, true));
+  addedOrdersSubscription.on('close', ensureConnection.bind(undefined, argv, true));
+  addedOrdersSubscription.on('error', (err: Error) => {
+    console.log(`Unexpected error occured: ${JSON.stringify(err)}, retrying to connect`);
+    ensureConnection(argv);
+  });
+
   const removedOrdersSubscription = loadXudClient(argv).subscribeRemovedOrders(new xudrpc.SubscribeRemovedOrdersRequest());
   removedOrdersSubscription.on('data', (orderRemoval: xudrpc.OrderRemoval) => {
     console.log(`Order removed: ${JSON.stringify(orderRemoval.toObject())}`);
   });
 
+  // prevent exiting and do nothing, it's already cached above.
+  removedOrdersSubscription.on('error', () => {});
+
   const swapsSubscription = loadXudClient(argv).subscribeSwaps(new xudrpc.SubscribeSwapsRequest());
   swapsSubscription.on('data', (swapResult: xudrpc.SwapResult) => {
     console.log(`Order swapped: ${JSON.stringify(swapResult.toObject())}`);
   });
+
+  // prevent exiting and do nothing, it's already cached above.
+  swapsSubscription.on('error', () => {});
 };


lib/cli/commands/subscribeorders.ts
```diff
@@ -15,20 +15,51 @@ export const builder = {
...
   addedOrdersRequest.setExisting(argv.existing);
   const addedOrdersSubscription = loadXudClient(argv).subscribeAddedOrders(addedOrdersRequest);
   addedOrdersSubscription.on('data', (order: xudrpc.Order) => {
     console.log(`Order added: ${JSON.stringify(order.toObject())}`);
   });
 
+  // adding end, close, error events only once,
+  // since they'll be thrown for three of subscriptions in the corresponding cases, catching once is enough.
+  addedOrdersSubscription.on('end', ensureConnection.bind(undefined, argv, true));
+  addedOrdersSubscription.on('close', ensureConnection.bind(undefined, argv, true));
```
moshababo: perhaps log here that xud went down, before trying to reconnect

lib/cli/commands/subscribeorders.ts
```diff
@@ -15,20 +15,51 @@ export const builder = {
 };
 
 export const handler = (argv: Arguments) => {
+  ensureConnection(argv, true);
+};
+
+const ensureConnection = (argv: Arguments, printError?: boolean) => {
+  loadXudClient(argv).getInfo(new xudrpc.GetInfoRequest(), (error: Error | null) => {
```
moshababo: perhaps you can use `loadXudClient(argv).waitForReady` instead

lib/cli/commands/subscribeorders.ts
```diff
@@ -15,20 +15,51 @@ export const builder = {
...
+  // prevent exiting and do nothing, it's already cached above.
+  removedOrdersSubscription.on('error', () => {});
+
   const swapsSubscription = loadXudClient(argv).subscribeSwaps(new xudrpc.SubscribeSwapsRequest());
   swapsSubscription.on('data', (swapResult: xudrpc.SwapResult) => {
     console.log(`Order swapped: ${JSON.stringify(swapResult.toObject())}`);
   });
+
+  // prevent exiting and do nothing, it's already cached above.
+  swapsSubscription.on('error', () => {});
```
moshababo: why do we need this line (and also line 56)? client will throw error if not subscribing to it?

rsercano: No actually these lines are required in case of streaming errors, if we don't catch them like that they'll stop the reading stream

lib/cli/commands/subscribeorders.ts
rsercano: Will check that

lib/cli/commands/subscribeorders.ts
rsercano: Of course

lib/cli/commands/subscribeorders.ts
```diff
@@ -15,20 +15,56 @@ export const builder = {
...
+};
+
+const reconnect = (argv: Arguments) => {
+  console.log('Stream is closed unexpectedly, trying to reconnect');
```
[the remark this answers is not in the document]
rsercano: Trust me I'm an amateur in English, so I'll follow your lead.

lib/cli/commands/subscribeorders.ts
sangaman: I think if there's an error on the stream, it will end anyway (because the server is ending it, nothing the client can do) and adding a listener here won't change it. If we want to reconnect, I think we'd need to do so by making the rpc calls again, which you're already doing. I'm not sure this "catches" anything, but in case I'm wrong could you explain some more? Is there information on this online anywhere or did you run into needing this in testing?

lib/cli/commands/subscribeorders.ts
```diff
@@ -15,20 +15,56 @@ export const builder = {
 };
 
 export const handler = (argv: Arguments) => {
+  ensureConnection(argv, true);
+};
+
+const ensureConnection = (argv: Arguments, printError?: boolean) => {
+  loadXudClient(argv).waitForReady(Number.POSITIVE_INFINITY, (error: Error | null) => {
```
sangaman: Do we need to reload the XudClient every time? Or can we just load it once, maintain the reference to it,  and call `waitForReady` as needed.

lib/cli/commands/subscribeorders.ts
```diff
@@ -15,20 +15,56 @@ export const builder = {
...
   addedOrdersRequest.setExisting(argv.existing);
   const addedOrdersSubscription = loadXudClient(argv).subscribeAddedOrders(addedOrdersRequest);
   addedOrdersSubscription.on('data', (order: xudrpc.Order) => {
     console.log(`Order added: ${JSON.stringify(order.toObject())}`);
   });
 
+  // adding end, close, error events only once,
+  // since they'll be thrown for three of subscriptions in the corresponding cases, catching once is enough.
+  addedOrdersSubscription.on('end', reconnect.bind(undefined, argv));
+  addedOrdersSubscription.on('close', reconnect.bind(undefined, argv));
```
sangaman: Based on [this](https://grpc.io/docs/tutorials/basic/node.html#streaming-rpcs) I think we only need to listen to the `end` event and not `close`.

lib/cli/commands/subscribeorders.ts
```diff
@@ -15,20 +15,56 @@ export const builder = {
...
 
+  // prevent exiting and do nothing, it's already cached above.
+  removedOrdersSubscription.on('error', () => {});
+
   const swapsSubscription = loadXudClient(argv).subscribeSwaps(new xudrpc.SubscribeSwapsRequest());
   swapsSubscription.on('data', (swapResult: xudrpc.SwapResult) => {
     console.log(`Order swapped: ${JSON.stringify(swapResult.toObject())}`);
   });
+
+  // prevent exiting and do nothing, it's already cached above.
```
sangaman: If we keep these comments, change to "caught" instead of "cached."

lib/cli/commands/subscribeorders.ts
```diff
@@ -15,20 +15,56 @@ export const builder = {
...
   addedOrdersSubscription.on('data', (order: xudrpc.Order) => {
     console.log(`Order added: ${JSON.stringify(order.toObject())}`);
   });
 
+  // adding end, close, error events only once,
+  // since they'll be thrown for three of subscriptions in the corresponding cases, catching once is enough.
+  addedOrdersSubscription.on('end', reconnect.bind(undefined, argv));
+  addedOrdersSubscription.on('close', reconnect.bind(undefined, argv));
+  addedOrdersSubscription.on('error', (err: Error) => {
+    console.log(`Unexpected error occured: ${JSON.stringify(err)}, retrying to connect`);
```
sangaman: Minor nit but this can say "trying to reconnect" as well as you've phrased it elsewhere, I think that sounds a little better.

lib/cli/commands/subscribeorders.ts
```diff
@@ -15,20 +15,56 @@ export const builder = {
...
   swapsSubscription.on('data', (swapResult: xudrpc.SwapResult) => {
     console.log(`Order swapped: ${JSON.stringify(swapResult.toObject())}`);
   });
+
+  // prevent exiting and do nothing, it's already cached above.
+  swapsSubscription.on('error', () => {});
+};
+
+const reconnect = (argv: Arguments) => {
+  console.log('Stream has closed unexpectedly, trying to reconnect');
```
sangaman: I think we can just say "stream has closed" without "unexpectedly" because the error is unexpected, whereas this block is for when `xud` has been deliberately shutdown or stopped.

lib/cli/commands/subscribeorders.ts
rsercano: good catch

lib/cli/commands/subscribeorders.ts
rsercano: call me paranoid, just wanted to be safe with this one, but removing `close`, it's okay.

lib/cli/commands/subscribeorders.ts
rsercano: typo, sorry.

lib/cli/commands/subscribeorders.ts
rsercano: sure thing

lib/cli/commands/subscribeorders.ts
rsercano: sure

lib/cli/commands/subscribeorders.ts
```diff
@@ -18,8 +18,11 @@ export const handler = (argv: Arguments) => {
   ensureConnection(argv, true);
 };
 
+let xud: any;
```
sangaman: Minor nit, but this can be `XudClient` instead of `any`.

rsercano: sure thing

rsercano:
If you may squash and merge I would be happier, don't want to take risk to break commit history of master :)

PR #783, opened 2019-01-08
 feat: parse uri without port

sangaman:
This allows for node connection uris without a specified port to be parsed and assigned a default port of 8885. This allows the port to be left off in the rpc `Connect` call when connecting to the default port.


test/unit/utils.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { expect } from 'chai';
+import { parseUri, getUri, UriParts } from '../../lib/utils/utils';
```
moshababo: maybe it's worth extracting these uri util functions to a `uriUtils` file, and have this test file name accordingly.

sangaman: That would be fine with me, I'll make the change.

Issue #688, opened 2018-11-18
streamOrders - missing swap information


offerm:
if `streamorders` is running and my node acts as the taker, swap information is not printed.

Fix that.

sangaman:
I don't think it's guessing, the results for swaps you initiate are returned when you initiate a swap via `PlaceOrder` or `ExecuteSwap` - and the description of the `SubscribeSwaps` call starts with `Subscribes to completed swaps that are initiated by a remote peer.`

I still am not really seeing the need for this, but if we do go with it I'd rather see this as an optional flag that can be set on the `SubscribeSwaps` call to subscribe to taker swaps as well. The reason being is that if I put myself in the shoes of a developer for an exchange, I wouldn't have any use for the duplicate notification of a successful swap coming from `SubscribeSwaps` when I'm already handling it as part of the response to `PlaceOrder` or `ExecuteSwap`.

kilrau:
Doesn't look like we have consensus on this one yet. Other opinions? @moshababo @erkarl @ImmanuelSegol @michael1011

kilrau:
@sangaman

PR #784, opened 2019-01-16
feat(rpc): optionally subscribe to taker swaps

### LINKED ISSUE #698: connection retries not working

sangaman:
This adds an optional `include_taker` flag to the `SubscribeSwaps` rpc call to include taker swaps resulting from `ExecuteSwap` or `PlaceOrder`. When this flag is enabled, swap results will be transmitted both in the response of the `ExecuteSwap` or `PlaceOrder` call as well as via the `SubscribeSwaps` stream.

Closes #698.


moshababo:
> Closes #698.

Did you mean a different one?

sangaman:
> Did you mean a different one?

Oops yes! It's #688. Will fix.

PR #785, opened 2019-01-16
fix(rpc): subscribeorders http endpoints

sangaman:
This corrects overlapping http endpoint naming for the `SubscribeAddedOrders` and `SubscribeRemovedOrders` rpc calls.

### DIFF
--- a/lib/proto/xudrpc.swagger.json
+++ b/lib/proto/xudrpc.swagger.json
@@ -466,7 +466,34 @@
         ]
       }
     },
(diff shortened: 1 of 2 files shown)


moshababo:
Does the HTTP endpoints support the streaming in any way?

sangaman:
I believe so (LND appears to support it at least) although I can't say I've tested it or understand fully how it works. Still, good to not have the name used twice.

PR #786, opened 2019-01-16
feat(swaps): timeout stalled swaps

### LINKED ISSUE #653: Held amount not released if taker doesn't follow the swap protocol
One a swap agreed between the taker and the maker, the maker is holding the agreed amount for the swap.

In case of a swap error, the amount is being release #637 and available for trading.

However, if the taker is not starting the swap and is not sending a swap error  the amount will be kept on hold by the maker until a restart of the maker.

sangaman:
This adds timeouts to all swap deals to fail them after a certain amount of time. Current time limits are set to 10 seconds for a swap to be accepted and 30 seconds for a swap to be completed once accepted. Previously, swaps that stalled out due to an unresponsive peer or swap client would last indefinitely and the corresponding order would remain on hold.

Closes #653.

### DIFF
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -29,10 +29,16 @@ interface Swaps {
 class Swaps extends EventEmitter {
   /** A map between payment hashes and swap deals. */
   private deals = new Map<string, SwapDeal>();
+  /** A map between payment hashes and timeouts for swaps. */
+  private timeouts = new Map<string, number>();
   private usedHashes = new Set<string>();
   private repository: SwapRepository;
   /** The number of satoshis in a bitcoin. */
   private static readonly SATOSHIS_PER_COIN = 100000000;
+  /** The maximum time in milliseconds we will wait for a swap to be accepted before failing it. */
+  private static readonly SWAP_ACCEPT_TIMEOUT = 10000;
+  /** The maximum time in milliseconds we will wait for a swap to be completed before failing it. */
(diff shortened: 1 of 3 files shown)


lib/swaps/Swaps.ts
```diff
@@ -454,10 +467,19 @@ class Swaps extends EventEmitter {
...
+      return;
+    }
+    if (deal.phase !== SwapPhase.SwapRequested) {
+      this.logger.warn(`received swap accepted for deal that is not in SwapRequested phase: ${rHash}`);
+      // TODO: penalize peer
       return;
     }
 
+    clearTimeout(this.timeouts.get(rHash));
+    this.timeouts.set(rHash, setTimeout(this.handleSwapTimeout, Swaps.SWAP_COMPLETE_TIMEOUT, rHash));
```
moshababo: perhaps add a comment to describe these 2 lines

lib/swaps/Swaps.ts
```diff
@@ -636,6 +658,12 @@ class Swaps extends EventEmitter {
 
   }
 
+  private handleSwapTimeout = (rHash: string) => {
+    const deal = this.getDeal(rHash)!;
+    this.timeouts.delete(rHash);
+    this.failDeal(deal, SwapFailureReason.Timeout);
```
moshababo: perhaps create a different failure reason to differ between acceptance/completion timeout

sangaman:
@moshababo Thanks, you can take a look at the changes when you get a chance.

moshababo:
Looks good 👍

PR #791, opened 2019-01-19
feat: listtrades grpc command #667

### LINKED ISSUE #677: new RPC - ListTrades
We should have an RPC method to return all trades from the database.

Should be implemented after #608 is merged.

ImmanuelSegol:
Closes #677


lib/orderbook/OrderBookRepository.ts
```diff
@@ -51,6 +51,14 @@ class OrderbookRepository {
   public addTrade = (trade: db.TradeFactory) => {
     return this.models.Trade.create(trade);
   }
+
+  public getTrades = (): Bluebird<db.TradeInstance[]> => {
+    return this.models.Trade.findAll();
+  }
+
+  public getTradesLimit = (limit: number): Bluebird<db.TradeInstance[]> => {
+    return this.models.Trade.findAll({ limit });
```
sangaman: I'm not sure this is returning the most recent trades by default, we may want to order results by `createdAt`.

lib/orderbook/OrderBook.ts
```diff
@@ -132,6 +132,19 @@ class OrderBook extends EventEmitter {
     });
   }
 
+  /**
+   * Get all trades or a limit number of trades.
```
sangaman: "limited" instead of "limit"

lib/cli/commands/listtrades.ts
```diff
@@ -0,0 +1,21 @@
+import { Arguments } from 'yargs';
+import { callback, loadXudClient } from '../command';
+import { ListTradesRequest } from '../../proto/xudrpc_pb';
+
+export const command = 'listtrades [all]';
+
+export const describe = 'list all trades';
+
+export const builder = {
+  existing: {
```
sangaman: This should say "all" rather than "existing"

However I think it might actually be better to allow the caller to specify a number of results, rather than all or a fixed number. You can just make this a number called `limit` as you've done elsewhere, set a default value of 15 from the cli, and make it so a value of 0  (which is the value if it's unspecified through the grpc call) returns all trades.

proto/xudrpc.proto
```diff
@@ -396,6 +410,17 @@ message Order {
   double hold = 10 [json_name = "hold"];
 }
 
+message Trade {
+  /** The order id of the maker order involved in this trade. */
+  string makerOrderId = 1 [json_name = "maker_order_id"];
+  // The order id of the taker order involved in this trade, if applicable.
+  string takerOrderId = 2 [json_name = "taker_order_id"];
+  // The rHash of the swap that filled this trade, if applicable.
+  string rHash = 3 [json_name = "r_hash"];
```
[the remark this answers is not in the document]
ImmanuelSegol: wow can believe I messed that up :p

proto/xudrpc.proto
```diff
@@ -410,6 +424,21 @@ message Order {
...
+  string maker_order_id = 1 [json_name = "maker_order_id"];
+  // The order id of the taker order involved in this trade, if applicable.
+  string taker_order_id = 2 [json_name = "taker_order_id"];
+  // The payment hash of the swap that filled this trade, if applicable.
+  string r_hash = 3 [json_name = "r_hash"];
+  // The quantity transacted in this trade.
+  int64 quantity = 4 [json_name = "quantity"];
+  // pair id of trade
+  string pair_id = 5 [json_name = "pair_id"];
+  // local id of trade
```
sangaman: The trade itself doesn't have a local id, so this field can probably be dropped.

proto/xudrpc.proto
```diff
@@ -410,6 +424,21 @@ message Order {
   double hold = 10 [json_name = "hold"];
 }
 
+message Trade {
+  // The order id of the maker order involved in this trade.
+  string maker_order_id = 1 [json_name = "maker_order_id"];
```
sangaman: I think it would be best to return the `Order` messages here, which means we need to fetch the orders from the database for each trade.

sangaman:
You should be able to get the maker and taker `OrderInstance` objects from a `TradeInstance` with the methods defined in #802.

proto/xudrpc.proto
ImmanuelSegol: I don't think I understood.

sangaman: Basically, I don't think the order id is particularly helpful to the caller of the API. They would want to see actual information about the orders involved in the trade. Since the database is relational the `trades` table only contains references to the actual orders, so for this API call I think we should actually fetch the order information. So instead of `string maker_order_id` it should be `Order maker_order`.

See the methods defined in #802, you should be able to call those to get the order data you need. Comment here or ping me on gitter if you have more questions.

lib/grpc/GrpcService.ts
```diff
@@ -431,6 +432,32 @@ class GrpcService {
     }
   }
 
+  /**
+   * See [[Service.listTrades]]
+   */
+  public listTrades: grpc.handleUnaryCall<xudrpc.ListTradesRequest, xudrpc.ListTradesResponse> = async (call, callback) => {
+    try {
+      const listTrades = await this.service.listTrades(call.request.toObject());
```
sangaman: This can just be `trades`. Calling it `listTrades` makes it sounds like a function rather than a collection of trades.

proto/xudrpc.proto
```diff
@@ -410,6 +425,19 @@ message Order {
   double hold = 10 [json_name = "hold"];
 }
 
+message Trade {
+  // The order id of the maker order involved in this trade.
```
sangaman: This comment can just say "The maker order involved in this trade" - same for the taker order below.

proto/xudrpc.proto
```diff
@@ -410,6 +425,19 @@ message Order {
...
+  Order maker_order = 1 [json_name = "maker_order"];
+  // The order id of the taker order involved in this trade, if applicable.
+  string taker_order_id = 2 [json_name = "taker_order_id"];
```
sangaman: This should also be an `Order` message like the maker order above, not just the id.

proto/xudrpc.proto
```diff
@@ -410,6 +425,19 @@ message Order {
...
+  // The payment hash of the swap that filled this trade, if applicable.
+  string r_hash = 3 [json_name = "r_hash"];
+  // The quantity transacted in this trade.
+  int64 quantity = 4 [json_name = "quantity"];
+  // pair id of trade
```
sangaman: "The trading pair for this trade"

lib/grpc/GrpcService.ts
```diff
@@ -431,6 +432,32 @@ class GrpcService {
...
+      const response = new xudrpc.ListTradesResponse();
+      const trades: xudrpc.Trade[] = [];
+      listTrades.forEach(async (trade: TradeInstance) => {
+        const grpcTrade = new xudrpc.Trade();
+        const order = grpcTrade.getMakerOrder();
```
sangaman: This should be `const makerOrder = await trade.getMakerOrder();` - otherwise we are just getting the empty `Order` message from the grpc response we are trying to construct.

I put together this method you can use to convert an `OrderInstance` into a grpc order:

```
      const getGrpcOrderFromOrderInstance = (order: OrderInstance) => {
        const grpcOrder = new xudrpc.Order();
        grpcOrder.setCreatedAt(order.createdAt);
        grpcOrder.setId(order.id);
        grpcOrder.setIsOwnOrder(order.nodeId === undefined)
        if (order.localId) {
          grpcOrder.setLocalId(order.localId);
        }
        grpcOrder.setPairId(order.pairId);
        // TODO: set peer pub key if order.nodeId has a value
        if (order.price) {
          grpcOrder.setPrice(order.price);
        }
        grpcOrder.setSide(order.isBuy ? xudrpc.OrderSide.BUY : xudrpc.OrderSide.SELL);
        return grpcOrder;
      };
```

So you can use `const grpcMakerOrder = getGrpcOrderFromOrderInstance(makerOrder!);` and `grpcTrade.setMakerOrder(grpcMakerOrder);` - and eventually do the same for the taker order as well.

lib/grpc/GrpcService.ts
```diff
@@ -431,6 +432,32 @@ class GrpcService {
...
+        const deal = await trade.getSwapDeal();
```
sangaman: It looks like they only thing we're using the `deal` for is the pair id, which we're getting anyway from the orders. We can probably drop this line and if you make my changes above instead use `grpcTrade.setPairId(makerOrder!.pairId);` below.

lib/orderbook/OrderBook.ts
```diff
@@ -102,7 +102,8 @@ class OrderBook extends EventEmitter {
           // we must remove the amount that was put on hold while the swap was pending for the remaining order
           this.removeOrderHold(orderId, pairId, quantity);
 
-          await this.persistTrade(swapSuccess.quantity, this.getOwnOrder(swapSuccess.orderId, swapSuccess.pairId), undefined, swapSuccess.rHash);
+          await this.persistTrade(swapSuccess.quantity,
+          this.getOwnOrder(swapSuccess.orderId, swapSuccess.pairId), undefined, swapSuccess.rHash);
```
sangaman: I'd prefer keeping this on one line, but if it's going to be on two lines I think the second line should be indented to be more readable.

lib/orderbook/OrderBook.ts
```diff
@@ -131,6 +132,14 @@ class OrderBook extends EventEmitter {
     });
   }
 
+  /**
+   * Get all trades or a limited number of trades.
```
sangaman: I think it would be good to add "from the database" to the end of the description of the method.

lib/orderbook/OrderBookRepository.ts
```diff
@@ -50,6 +50,14 @@ class OrderbookRepository {
   public addTrade = (trade: db.TradeFactory) => {
     return this.models.Trade.create(trade);
   }
+
+  public getTradesLimit = (limit?: number): Bluebird<db.TradeInstance[]> => {
```
sangaman: Can we just call this `getTrades`? I think that makes more sense since the limit is optional.

lib/grpc/GrpcService.ts
```diff
@@ -434,6 +455,32 @@ class GrpcService {
...
+      const response = new xudrpc.ListTradesResponse();
+      const tradesList: xudrpc.Trade[] = [];
+      trades.forEach(async (trade: TradeInstance) => {
+        const grpcTrade = new xudrpc.Trade();
+        const makerOrder = await trade.getMakerOrder();
+        const takerOrder = await trade.getTakerOrder();
+        grpcTrade.setQuantity(trade.quantity);
+        grpcTrade.setRHash(trade.rHash ? trade.rHash : '');
+        grpcTrade.setMakerOrder(getGrpcOrderFromOrderInstance(makerOrder!));
+        grpcTrade.setTakerOrder(getGrpcOrderFromOrderInstance(takerOrder!));
```
sangaman: It's possible we won't know of the taker order (if, for example, a peer filled one of our orders via a swap) so instead of using the `!` operator we should put this line within an `if (takerOrder)` statement.

kilrau:
Good to merge? @sangaman

kilrau:
Push @sangaman

ImmanuelSegol:
> Push @sangaman

This has been good to push for a long time now @sangaman ?

kilrau:
It is because of raiden, see https://github.com/ExchangeUnion/xud/issues/1011, thanks for the hint @reliveyy 

Can't reproduce it on master, could you give this a rebase? @ImmanuelSegol

kilrau:
Also, I couldn't get the `listtrades` call to work:
```
~/xud/bin$ ./xucli listtrades
Error: 2 UNKNOWN: SQLITE_ERROR: no such column: Trade.time
```

I dropped the db and xud recreated it. No matter if xud.db, xud-testnet.db or xud-simnet.db - all throw the same error that `Trade.time` colum doesn't exist.

kilrau:
if the output is json atm, please also put it in a table to make it look decent @ImmanuelSegol

lib/proto/lndinvoices_pb.js
```diff
@@ -653,7 +655,7 @@ proto.invoicesrpc.AddHoldInvoiceRequest.prototype.getExpiry = function() {
 
 /** @param {number} value */
 proto.invoicesrpc.AddHoldInvoiceRequest.prototype.setExpiry = function(value) {
-  jspb.Message.setField(this, 5, value);
+  jspb.Message.setProto3IntField(this, 5, value);
```
sangaman: Can you check your protoc compiler version? I believe it should be 3.4. There shouldn't be these changes to this file. Ensure you have `grpc-tools` 1.6.0 installed in the xud directory and try regenerating protos, these changes should go away.

ImmanuelSegol: I do have those versions.
For some reason, this still happens.
I will look into it.

rsercano:
Fixed async bug, merged master seems to be working

lib/proto/xudrpc.swagger.json
```diff
@@ -21,7 +21,7 @@
         "operationId": "AddCurrency",
         "responses": {
           "200": {
-            "description": "",
+            "description": "A successful response.",
```
sangaman: Can you try regenerating these with the version of grpc-swagger I specified in a separate PR? I can find it for you again if you need.

lib/proto/annotations_grpc_pb.js
```diff
@@ -1 +1 @@
-// GENERATED CODE -- NO SERVICES IN PROTO
+// GENERATED CODE -- NO SERVICES IN PROTO
```
sangaman: Not sure why we're getting these changes, see if regenerating and committing removes this if not I'll take a closer look.

lib/service/Service.ts
```diff
@@ -349,6 +349,18 @@ class Service {
     return this.pool.listPeers();
   }
 
+  /**
+   * Get the list of trades.
```
sangaman: "Gets" instead of "Get".

proto/xudrpc.proto
```diff
@@ -396,6 +410,17 @@ message Order {
   double hold = 10 [json_name = "hold"];
 }
 
+message Trade {
```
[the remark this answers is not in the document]
sangaman: I'm thinking we should still have the executed price for the trade here.

lib/orderbook/OrderBook.ts
```diff
@@ -199,6 +199,14 @@ class OrderBook extends EventEmitter {
     return currencyInstance ? currencyInstance.toJSON() : undefined;
   }
 
+  /**
+   * Get all trades or a limited number of trades from the database.
```
sangaman: "Gets"

lib/cli/commands/listtrades.ts
```diff
@@ -0,0 +1,45 @@
+import { Arguments } from 'yargs';
+import { callback, loadXudClient } from '../command';
+import { ListTradesRequest, ListTradesResponse, Trade } from '../../proto/xudrpc_pb';
+import Table, { HorizontalTable } from 'cli-table3';
+import colors from 'colors/safe';
+
+const HEADERS = [
+  colors.blue('rhash'),
```
sangaman: I don't think we really need the payment hash on the prettified, user-friendly table. Thoughts @kilrau?

If we do keep it, I'd probably call this "Payment hash" or maybe "Swap hash".

lib/cli/commands/listtrades.ts
```diff
@@ -0,0 +1,45 @@
...
+  colors.blue('trade quantity'),
```
sangaman: These should be capitalized (such as "Trade quantity") to be consistent with the tables in other calls.

lib/cli/commands/listtrades.ts
```diff
@@ -0,0 +1,45 @@
...
+  colors.blue('order type'),
+];
+
+const displayTrades = (trades: ListTradesResponse.AsObject) => {
+  const table = new Table({ head: HEADERS }) as HorizontalTable;
+  trades.tradesList.forEach((trade: Trade.AsObject) => {
+    const type = trade.makerOrder ? 'maker' : 'taker';
+    table.push([
+      trade.rHash,
+      trade.quantity,
```
sangaman: We should convert this from satoshis to whole units using the `satsToCoinsStr` utils method.

lib/cli/commands/listtrades.ts
```diff
@@ -0,0 +1,45 @@
...
+  colors.blue('trade quantity'),
+  colors.blue('trading pair'),
+  colors.blue('order type'),
+];
+
+const displayTrades = (trades: ListTradesResponse.AsObject) => {
+  const table = new Table({ head: HEADERS }) as HorizontalTable;
+  trades.tradesList.forEach((trade: Trade.AsObject) => {
+    const type = trade.makerOrder ? 'maker' : 'taker';
+    table.push([
```
sangaman: I think the price is a key piece of information to display, we should add it here.

lib/proto/xudrpc.swagger.json
rsercano: got it thanks :)

lib/cli/commands/listtrades.ts
rsercano: removed it for now, please let me know if you want to put it back

rsercano:
Implemented @sangaman, thanks a lot.

PR #800, opened 2019-02-03
Execute swaps in parallel when matching orders

### LINKED ISSUE #654: Parallel taking of orders & retry after failures
Consider the case when a taker order (market or limit)  is matched with multiple orders. As of today, we execute the first match, wait for result, execute the next match etc.

Why all these matches not executed in parallel?

This is not just performance improvement issue. When the first swap fails the orderbook may be completely different  compared to what it was when we planned the swaps. It might be that we can swap the user order with a peer order having a better price.

Maybe we should  perform all swaps in parallel and have a seperate retry for each swap that failed?
Maybe we should extend swap-request to include multiple swaps?

sangaman:
This begins all swaps in parallel when placing/matching an order and retries failed swaps immediately. Any unmatched quantity is tracked and added to the orderbook (for limit orders) after matching and swaps are complete. Previously, swaps would execute one at a time and any failed quantity would wait to be retried until all first-attempt swaps completed. This should result in significantly faster order execution for orders that match with multiple peer orders and therefore require multiple swaps.

Closes #654.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -15,7 +15,7 @@ import { CurrencyInstance, PairInstance, CurrencyFactory } from '../db/types';
 import { Pair, OrderIdentifier, OwnOrder, OrderPortion, OwnLimitOrder, PeerOrder, Order, PlaceOrderEvent,
   PlaceOrderEventType, PlaceOrderResult, OutgoingOrder, OwnMarketOrder, isOwnOrder, IncomingOrder } from './types';
 import { SwapRequestPacket, SwapFailedPacket } from '../p2p/packets';
-import { SwapResult, SwapDeal } from 'lib/swaps/types';
+import { SwapSuccess, SwapDeal } from 'lib/swaps/types';
 import Bluebird from 'bluebird';
 
 interface OrderBook {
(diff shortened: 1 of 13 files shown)


lib/orderbook/OrderBook.ts
```diff
@@ -289,25 +289,37 @@ class OrderBook extends EventEmitter {
     // perform match. maker orders will be removed from the repository
     const tp = this.getTradingPair(order.pairId);
     const matchingResult = tp.match(order);
```
moshababo: An empty line after this one would help. Also IMO we need now to add very verbose documentation for this method (`placeOrder`).

lib/orderbook/OrderBook.ts
```diff
@@ -289,25 +289,37 @@ class OrderBook extends EventEmitter {
...
-      swapResults: [],
-      remainingOrder: matchingResult.remainingOrder,
+    /** Any portion of the placed order that could not be swapped or matched internally. */
+    let { remainingOrder } = matchingResult;
+    /** Local orders that matched with the placed order. */
+    const internalMatches: OwnOrder[] = [];
+    /** Successful swaps performed for the placed order. */
+    const swapResults: SwapResult[] = [];
+
+    const retryFailedSwap = async (failedSwapQuantity: number) => {
```
moshababo: Unlike the previous version, it looks that you're calling this function for every failed swap, and not for the accumulative failed quantity (of all the failed swaps). Is that on purpose?

sangaman: Yeah it is, just to minimize the total amount of time expected for the order matching routine to complete. If we start retrying a failed swap immediately, it potentially eliminates a wait for other swaps to complete.

lib/orderbook/OrderBook.ts
sangaman: I added this line and some comments in a separate commit, let me know if you think it's sufficient or if you have anything else in mind.

lib/orderbook/OrderBook.ts
```diff
@@ -278,45 +289,68 @@ class OrderBook extends EventEmitter {
...
-    const result: PlaceOrderResult = {
-      internalMatches: [],
-      swapResults: [],
-      remainingOrder: matchingResult.remainingOrder,
+    /** Any portion of the placed order that could not be swapped or matched internally. */
+    let { remainingOrder } = matchingResult;
+    /** Local orders that matched with the placed order. */
+    const internalMatches: OwnOrder[] = [];
+    /** Successful swaps performed for the placed order. */
+    const swapResults: SwapResult[] = [];
```
moshababo: This should be changed to `swapSuccesses` (also in `PlaceOrderResult`).

lib/orderbook/OrderBook.ts
moshababo: Looks great, thanks.

lib/orderbook/OrderBook.ts
sangaman: I thought about doing this but I wasn't sure about it, but if you think it should be changed I'll go ahead. I may also change the name of the `SwapResult` type.

sangaman: I went ahead and renamed all references I could find to make it consistent.

PR #803, opened 2019-02-05
Partial order removal

### LINKED ISSUE #714: Remove part of an order
Currently we can only remove an entire order via the rpc layer. I think it would probably be useful to be able to remove only part of an order by specifying a quantity on the `RemoveOrderRequest`.

sangaman:
This replaces #796. Closes #714.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -479,51 +479,94 @@ class OrderBook extends EventEmitter {
    * is not supported or if the order to cancel could not be found.
    * @returns any quantity of the order that was on hold and could not be immediately removed.
    */
-  public removeOwnOrderByLocalId = (localId: string) => {
+  public removeOwnOrderByLocalId = (localId: string, quantityToRemove?: number) => {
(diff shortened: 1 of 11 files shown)


lib/orderbook/OrderBook.ts
```diff
@@ -479,51 +479,61 @@ class OrderBook extends EventEmitter {
...
+      // quantity to be removed can't be higher than order's quantity.
+      throw errors.QUANTITY_DOES_NOT_MATCH(remainingQuantityToRemove, order.quantity);
+    }
+
+    const removableQuantity = order.quantity - order.hold;
+    if (remainingQuantityToRemove <= removableQuantity) {
+      this.removeOwnOrder(orderIdentifier.id, orderIdentifier.pairId, quantityToRemove);
+      remainingQuantityToRemove = 0;
+    } else {
+      // we can't remove the entire amount because of a hold on the order
```
moshababo: perhaps put `failedHandler` / `paidHandler` / `cleanup` inside of this clause, since they are only relevant on this case.

PR #806, opened 2019-02-12
feat(swaps): define behavior for swap errors

### LINKED ISSUE #661: Define & implement behavior for different kind of swap errors
This issue is about a swap failure after the deal was agreed with the peer and the amount was held by the peer. Providing that the taker is acting reasonably fast once a deal agreed, the only "error" has to do with actual swap configuration and routing.

In such a case, if we get an error from the executeswap and we decide to remove the part of the peer order we tried to swap, we should ask why we are not removing the other part and the other orders.

For example, if we get "unknowPaymentHash" it is clear that the maker is not setup properly. If so, why to remove just that part that we tried to swap? why not to remove all orders and disconnect from the peer until next restart or next connect rpc call?

I think we need to consider this error by error and decide what we should do about it. If the swap fails on noroute from the peer to us, we may be able to swap a smaller amount so removing the current amount may not be the right handling.

sangaman:
Closes #661.

This adds logic to handle swap failures according to the reason for the failure. Certain failures suggest a persistent, underlying issue with swaps for that peer and trading pair, and in this case we drop all orders for that peer and trading pair to avoid subsequent failures. We also penalize the peer for swaps that fail due to potential misbehavior or due to a timeout. Timeouts result in delays that degrade the trading experience moreso than regular swap failures.

### DIFF
--- a/lib/constants/enums.ts
+++ b/lib/constants/enums.ts
@@ -51,7 +51,7 @@ export enum SwapPhase {
   SendingAmount = 3,
   /** We have received the agreed amount of the swap, and the preimage is now known to both sides. */
   AmountReceived = 4,
-  /* The swap has been formally completed and both sides have confirmed they've received payment. */
+  /** The swap has been formally completed and both sides have confirmed they've received payment. */
   SwapCompleted = 5,
 }
 
@@ -65,10 +65,15 @@ export enum ReputationEvent {
   ManualBan = 0,
   ManualUnban = 1,
   PacketTimeout = 2,
-  SwapFailure = 3,
-  SwapSuccess = 4,
-  WireProtocolErr = 5,
-  InvalidAuth = 6,
+  WireProtocolErr = 3,
+  InvalidAuth = 4,
+  SwapSuccess = 5,
+  /** When a swap is accepted and is attempted to be executed but fails. */
+  SwapFailure = 6,
+  /** When a swap is accepted and then fails due to exceeding time limits. */
+  SwapTimeout = 7,
+  /** When a swap fails due to unexpected or possibly malicious behavior. */
+  SwapMisbehavior = 8,
 }
 
 export enum SwapFailureReason {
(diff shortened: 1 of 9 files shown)


lib/constants/enums.ts
```diff
@@ -65,10 +65,15 @@ export enum ReputationEvent {
...
   PacketTimeout = 2,
+  /** When a swap is accepted and is attempted to be executed but fails. */
   SwapFailure = 3,
   SwapSuccess = 4,
   WireProtocolErr = 5,
   InvalidAuth = 6,
+  /** When a swap is accepted and then fails due to exceeding time limits. */
+  SwapTimeout = 7,
+  /** When a swap fails due to unexpected or possibly malicious behavior. */
+  SwapMisbehavior = 8,
```
moshababo: Perhaps group the swap types togeather (I think it's still fine to do breaking changes), and explain a bit the difference between `SwapFailure` and `SwapTimeout`/`SwapMisbehavior`, since they also indicate a failure.

lib/p2p/Peer.ts
```diff
@@ -339,6 +339,21 @@ class Peer extends EventEmitter {
...
+  /**
+   * Manually drops support for a trading pair with this peer.
+   */
+  public dropPair = (pairId: string) => {
+    if (!this.nodeState) {
+      throw new Error('cannot drop a trading pair before handshake is complete');
+    }
+    const index = this.nodeState.pairs.indexOf(pairId);
+    if (index >= 0) {
+      this.nodeState.pairs.slice(index, 1);
```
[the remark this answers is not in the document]
sangaman: Yes good point... I should probably track the deactivated pairs separately for this reason.

lib/constants/enums.ts
sangaman: I grouped them together, but I'm not sure exactly what more to say about the differences between the errors without getting too verbose or specific. I tried outlining the general differences with the existing comments, let me know if you have anything in mind.

Issue #807, opened 2019-02-13
Improve swaps error handling


moshababo:
* Add `SwapFailures` to `PlaceOrderResult`. Otherwise we're just getting the final results, but have no idea about swap attempts which failed. (@sangaman @kilrau please verify that we want this)

* Propagate meaningful errors from `Swaps.executeSwap` (taker side), and log them once in the higher layer (`OrderBook.PlaceOrder`). Same for the maker side. 

* Calling `Swaps.getRoutes` and it's underlying `lnd.queryRoutes` throws exception on failure (when there's no channel, channel isn't active, channel local balance isn't sufficient, etc.). At the moment it's not handled properly and `lnd`s internal errors are returned ("target not found", "unable to find a path to destination").

kilrau:
> * Add `SwapFailures` to `PlaceOrderResult`. Otherwise we're just getting the final results, but have no idea about swap attempts which failed. (@sangaman @kilrau please verify that we want this)

Not sure. Is it inconsistent to have swap failures in the async call, but not in the sync version? I'd rather just write these in the db (please confirm that we already do this @moshababo ). Waiting for @sangaman opinion

sangaman:
For the first bullet point, I'm not really sure whether it's needed. Originally we were OK with the current behavior of only returning the failures for the asynchronous call, is there a new reason why we would change this? I personally don't have much of a preference but I wouldn't want to change something unless there's a reason.

For the other two bullet points I think is improved slightly by #806, but I've realized myself that we could have neater exceptions from the `executeSwap` method.

kilrau:
Thx Daniel! Swap failures are already recorded in the db, correct?

sangaman:
Yes correct, the only ones we don't record are the ones proposed to us but which be reject (like if the order is no longer available). We save all requested swaps and all accepted swaps whether the fail or succed.

kilrau:
Can you take out the first bullet point then if you agree? @moshababo

moshababo:
I thought it would be good to have it for the sake of UX and consistency with the async version. If i'm using `PlaceOrderSync`, I don't get this info, and having it saved in the DB doesn't help me to understand from the response why my order didn't trigger a swap.

moshababo:
[...]

PR #814, opened 2019-02-27
feat(rpc): add failed swaps to PlaceOrderSync

sangaman:
This adds a list of failed swaps to the `PlaceOrderSync` response to match the behavior of the asynchronous `PlaceOrder` call which returns information about failed swaps.

Solves part of #807.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -247,6 +247,7 @@ class OrderBook extends EventEmitter {
       return {
         internalMatches: [],
         swapSuccesses: [],
+        swapFailures: [],
         remainingOrder: stampedOrder,
       };
     }
@@ -292,6 +293,7 @@ class OrderBook extends EventEmitter {
       return {
         internalMatches: [],
         swapSuccesses: [],
+        swapFailures: [],
         remainingOrder: order,
       };
     }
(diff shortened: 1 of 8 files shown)


lib/orderbook/types.ts
```diff
@@ -15,6 +15,7 @@ export type PlaceOrderResult = {
   internalMatches: OwnOrder[];
   swapSuccesses: SwapSuccess[];
   remainingOrder?: OwnOrder;
+  swapFailures: PeerOrder[];
```
moshababo: Minor: should jump this one line up, next to `swapSuccesses`. also in other sections in this PR.

lib/orderbook/OrderBook.ts
```diff
@@ -371,6 +375,7 @@ class OrderBook extends EventEmitter {
           onUpdate && onUpdate({ type: PlaceOrderEventType.SwapSuccess, payload: swapSuccess });
         } catch (err) {
           this.logger.warn(`swap for ${portion.quantity} failed during order matching, will repeat matching routine for failed swap quantity`);
+          swapFailures.push(maker);
```
moshababo: We should add the swap reason here (and in `onUpdate` callback as well), and not to return merely the taker order. Did you plan to keep this for a different PR?

kilrau: Same question, see our discussion on discord.

lib/orderbook/types.ts
sangaman: OK I was actually thinking about doing the same, so I'll go ahead with that since you had the same thought.

lib/orderbook/OrderBook.ts
sangaman: Yes I was, but I can add it to this PR in a separate commit so it completely closes #807.

sangaman: Actually I'll open a separate PR because it's not a very simple change and I think it'd be better to merge this then discuss more changes.

PR #816, opened 2019-02-27
 feat(rpc): add failureReason to SwapFailure msg

### LINKED ISSUE #807: Improve swaps error handling

sangaman:
Closes #807.

This specifies the `SwapFailureReason` for each `SwapFailure` message sent in gRPC responses. It refactors the `Swaps.executeSwap` method to return a `SwapFailureReason` when a swap fails rather than throwing an exception which may not always describe why a swap failed. Exceptions are now thrown only for unexpected, unrecoverable errors. It also logs the failure reason in `OrderBook.ts` when a swap we initiated fails.


moshababo:
After executing the test scenario which is expected to have a swap failure, i'm getting this log:
````
01/03/2019 13:47:30.126 [RPC] debug: received call /xudrpc.Xud/PlaceOrderSync
01/03/2019 13:47:30.129 [ORDERBOOK] debug: matched with peer 0258a9425c67eea1f0234c409862037c9be5fd5f2ab1b3648985b7ad32c12aab08, executing swap on taker 8ea865f0-3c28-11e9-be77-6914090f1a2e and maker 8ea50a90-3c28-11e9-a041-5340f82a6e0d for 0.1
01/03/2019 13:47:30.158 [RPC] error: 13 INTERNAL: Failed to parse server response
````
And `PlaceOrderSync` response is returned with the same logged error (`13 INTERNAL: Failed to parse server response`).
I'm not sure what's causing this, but I verified that the client `xudrpc.proto` and the auto-generated pb file are updated.

**Comment by sangaman** [comment]
@moshababo In researching that error, it looks like it is generated by the client when the server returns a message in an unrecognized format. xud is only the grpc client when it comes to calls to lnd, which makes me think this may have to do with a discrepancy between lnd. In looking at the latest lnd rpc proto definition nothing jumps out as what might have changed (mainly the routes and payment messages).

Can you check the proto file your lnd is using? Is it this one https://github.com/ExchangeUnion/lnd/blob/resolver-cmd%2Bsimnet-ltcd/lnrpc/rpc.proto ?

I'll help dig into this more tomorrow my time.

**Comment by moshababo** [comment]
Yes, i'm using the `ExchangeUnion/lnd/resolver-cmd+simnet-ltcd` branch, which its proto isn't different from our standard `ExchangeUnion/lnd/resolver+simnet-ltcd` branch.
 
Looking at `xud`, it's lnd proto file is different from these two branches proto. Although I don't see something that could affect the swap procedure calls, it might be worthwhile to update `xud/proto/lndrpc.proto` and the auto-generated files. 

Let me know if you have more ideas about this problem, or you want me too look into it.

**Comment by sangaman** [comment]
I was able to reproduce your error message and spent hours trying to figure out what's wrong but am stuck, so going to come back to this tomorrow night with a set of fresh eyes. Do you think we can merge this PR anyway if the changes look good and I'll continue to debug the problem we're having separately? @moshababo .

**Comment by moshababo** [comment]
NP, if so i'll take closer look in the code soon. I was going for the easy way of waiting for it to be working first. 

If you have problems with reproducing this error, let me know.

I believe you didn't address the 3rd bullet from #807. Is that correct?

**Comment by sangaman** [comment]
> I believe you didn't address the 3rd bullet from #807. Is that correct?

Yes I think not, I can handle that separately.

File: lib/orderbook/OrderBook.ts [hunk 1]
```diff
@@ -357,26 +357,32 @@ class OrderBook extends EventEmitter {
...
           } else {
             this.logger.info(`match executed on taker ${taker.id} and maker ${maker.id} for ${maker.quantity} with peer ${maker.peerPubKey}`);
           }
-          swapSuccesses.push(swapSuccess);
-          onUpdate && onUpdate({ type: PlaceOrderEventType.SwapSuccess, payload: swapSuccess });
-        } catch (err) {
+          swapSuccesses.push(swapResult);
+          onUpdate && onUpdate({ type: PlaceOrderEventType.SwapSuccess, payload: swapResult });
+        } else {
           this.logger.warn(`swap for ${portion.quantity} failed during order matching, will repeat matching routine for failed swap quantity`);
```
moshababo: You can log here the swap failure reason as well (same as you did in `OrderBook.executeSwap`.

lib/swaps/Swaps.ts
```diff
@@ -243,16 +243,21 @@ class Swaps extends EventEmitter {
...
    */
-  public executeSwap = async (maker: PeerOrder, taker: OwnOrder): Promise<SwapSuccess> => {
+  public executeSwap = async (maker: PeerOrder, taker: OwnOrder): Promise<SwapSuccess | SwapFailureReason> => {
     await this.verifyExecution(maker, taker);
-    const rHash = await this.beginSwap(maker, taker);
-    if (!rHash) {
-      throw new Error('cannot execute swap. rHash not found');
+    const beginResult = await this.beginSwap(maker, taker);
+    let rHash: string;
+    if (typeof beginResult === 'string') {
```
moshababo: Optional: create another type called `SwapBeginSuccess` & `isSwapBeginSuccess` method, to make it easier to follow and understand why it's differed from `SwapSuccess`.

lib/swaps/Swaps.ts
```diff
@@ -243,16 +243,21 @@ class Swaps extends EventEmitter {
...
+      // we were able to request the swap and got its rHash value
+      rHash = beginResult;
+    } else {
+      // we could not begin the swap, return the failure reason
+      return beginResult;
     }
 
-    return new Promise<SwapSuccess>((resolve, reject) => {
+    return new Promise<SwapSuccess | SwapFailureReason>((resolve) => {
```
moshababo: Why not use the failure case as a promise reject, so you wont need to deal with the Success/Failure value inspection?

lib/swaps/Swaps.ts
```diff
@@ -243,16 +243,21 @@ class Swaps extends EventEmitter {
    * A promise wrapper for a swap procedure
    * @param maker the remote maker order we are filling
    * @param taker our local taker order
-   * @returns A promise that is resolved once the swap is completed, or rejects otherwise
+   * @returns A promise that resolves to a [[SwapSuccess]] once the swap is completed, or a [[SwapFailureReason]] if it fails
    */
-  public executeSwap = async (maker: PeerOrder, taker: OwnOrder): Promise<SwapSuccess> => {
+  public executeSwap = async (maker: PeerOrder, taker: OwnOrder): Promise<SwapSuccess | SwapFailureReason> => {
     await this.verifyExecution(maker, taker);
```
moshababo: This won't return the swap failure.

lib/swaps/Swaps.ts
```diff
@@ -336,7 +341,7 @@ class Swaps extends EventEmitter {
     if (errMsg) {
       this.logger.error(errMsg);
       this.failDeal(deal, SwapFailureReason.SwapClientNotSetup, errMsg);
-      return;
+      return SwapFailureReason.SwapClientNotSetup;
```
moshababo: Lines 496, 500, 550: all these swap failures won't be returned to the linked `executeSwap`. To achieve this more work is necessary to link between the async events.
Same for more failures in `Swap.js`, which some of them can be linked to the taker swap call.

lib/orderbook/OrderBook.ts
sangaman: Wouldn't that be repetitive? Since this is calling `OrderBook.executeSwap` I figure it would log the same error twice and both from the `ORDERBOOK` logger. I wanted to have the error logging within `executeSwap` so that we get similar logging in `nomatching` mode as well. If you still see a reason to have it here as well let me know and I'll add it.

lib/swaps/Swaps.ts
sangaman: The type of `SwapBeginSuccess` would just be a string right? That might be a little confusing as well as `isSwapBeginSuccess` would return true for any string. How about either:

1. I create an `isRHash` method that checks if the supplied parameter is a hex string of the correct length and use this here for line 252.
2. I create an `isSwapFailure` type check method and check for the failure first, `else` set the rHash value.

lib/swaps/Swaps.ts
sangaman: I thought about this, but I was unsure about relying on the parameter type in the `catch` blocks we'd need to handle rejections. I'm pretty sure that by default the compiler always treats the error in a `catch` block as `any`. I could explicitly define it as a `SwapFailureReason` though. I was just concerned about what would happen should any other type of exception get thrown for whatever reason.

If you think that's not a big problem though, I can definitely restructure these to throw/reject `SwapFailureReason` - this could also solve the need for anything like a `isSwapBeginSuccess` above because `beginSwap` could also throw the failure reason.

lib/swaps/Swaps.ts
sangaman: I'm not sure I follow, in these cases, calling `failDeal` will in turn set the failure reason on the deal and emit the `swap.failed` event. `executeSwap` is listening for that event and will return the failure reason if it occurs. Am I missing something that would cause this not to work?

sangaman:
@moshababo I addressed the issue with `getRoutes` and `verifyExecution`. Still have a couple questions about your other comments to clarify the best approach for returning or throwing the failure reasons.

lib/orderbook/OrderBook.ts
moshababo: I didn't notice it's calling `OrderBook.executeSwap`. Forget about it!

lib/swaps/Swaps.ts
moshababo: Now that I understand the limitation, i'd say you should go with whatever you think is better.

You can perhaps move the inspection to the thrown value, so the method interface will be straightforward. Perhaps by defining and throwing `SwapFailure`, an object with one property, `reason`, unlike `SwapFailureReason` which is a plain number.

lib/swaps/Swaps.ts
moshababo: Both are fine, but I think in this case rejecting the promise on failure will be the simplest.

lib/swaps/Swaps.ts
moshababo: Indeed, I got confused and didn't notice that the events are already connected. Forget about it!

lib/swaps/Swaps.ts
```diff
@@ -214,44 +214,67 @@ class Swaps extends EventEmitter {
...
-    const routes = (await client.queryRoutes(req)).getRoutesList();
-    this.logger.debug(`got ${routes.length} routes to destination: ${routes}`);
-    return routes;
+    try {
+      const routes = (await client.queryRoutes(req)).getRoutesList();
+      this.logger.debug(`got ${routes.length} routes to destination: ${routes}`);
+      return routes;
+    } catch (err) {
+      this.logger.debug(`error calling getRoutes for ${currency} to ${pubKey}: ${JSON.stringify(err)}`);
+      if (typeof err.message === 'string' && err.message.includes('unable to find a path to destination')) {
```
moshababo: You're relying on `unable to find a path to destination`, but I encountered also `target not found` (might be others as well).

moshababo:
Ok, so this is what we're getting now:

````
14/03/2019 18:28:14.592 [RPC] debug: received call /xudrpc.Xud/PlaceOrderSync
14/03/2019 18:28:14.594 [ORDERBOOK] debug: matched with peer 0303e66f52d10ad3971c775248e419f2f2440c9cab3f3135f7c5f3ca4fe4a30506, executing swap on taker ee1c8800-4686-11e9-9292-3d747b64fba9 and maker ee197ac0-4686-11e9-8aab-8b2197f9e90a for 0.1
14/03/2019 18:28:14.617 [SWAPS] debug: error calling getRoutes for LTC to 0249f07abe1c49e76fa40a303dd65b0ae60005b4a986aff4941978d86da76f8f07: {"code":13,"details":"Failed to parse server response"}
14/03/2019 18:28:14.617 [ORDERBOOK] error: swap between orders ee197ac0-4686-11e9-8aab-8b2197f9e90a & ee1c8800-4686-11e9-9292-3d747b64fba9 failed due to UnexpectedLndError
14/03/2019 18:28:14.618 [ORDERBOOK] warn: swap for 0.1 failed during order matching, will repeat matching routine for failed swap quantity
14/03/2019 18:28:14.618 [ORDERBOOK] debug: repeating matching routine for ee1c8800-4686-11e9-9292-3d747b64fba9 for failed quantity of 0.1
14/03/2019 18:28:14.619 [ORDERBOOK] debug: order added: {"pairId":"LTC/BTC","price":10,"quantity":0.1,"isBuy":false,"localId":"taker_order_id","hold":0,"id":"ee1c8800-4686-11e9-9292-3d747b64fba9","initialQuantity":0.1,"createdAt":1552588094592}
````

Looks much better, and can see that the `Failed to parse server response` error is associated with `queryRoutes` call to LND-LTC.

lib/swaps/Swaps.ts
sangaman: Thanks, I'll add that case as well. If we encounter others we can add those.

lib/swaps/Swaps.ts
sangaman: OK I'm going to refactor this by rejecting/throwing the failure reason and declaring the type in the `catch` blocks.

sangaman:
Thanks, I handled your feedback and did some testing and I think this is good to merge. I still have to dig deeper into that server response error but I'll address that separately.

PR #818, opened 2019-03-01
refactor: conditionally log init connections

sangaman:
This only logs messages that we've started and completed connecting to known nodes on startup when we have known, unbanned nodes to connect to.

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -118,10 +118,14 @@ class Pool extends EventEmitter {
     this.bindNodeList();
 
     this.nodes.load().then(() => {
-      this.logger.info('Connecting to known / previously connected peers');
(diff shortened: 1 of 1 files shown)


lib/p2p/Pool.ts
```diff
@@ -118,10 +118,14 @@ class Pool extends EventEmitter {
     this.bindNodeList();
 
     this.nodes.load().then(() => {
```
moshababo: Maybe refactor this to use async/await. It will also help making the code simpler with the new conditions.

sangaman: I would prefer to as well, but this was deliberate as per https://github.com/ExchangeUnion/xud/pull/389 so that this `init` method can resolve before all the connections to peers resolve.

PR #820, opened 2019-03-02
Listorders all

### LINKED ISSUE #748: ListOrders `all` flag
If `all` is `true` we return all existing orders, else we return a select number of orders.
We should decide how to select orders to display if -`all=false`:
1) Return by date.
2) Return the 10 best (closest to mid-market price) orders.
3) Some other factor.

@kilrau @sangaman @moshababo Opinions?

ImmanuelSegol:
Closes #748 
When `all=false` we return only the top 10 orders sorted by best market price.


sangaman:
If we do go with a max results parameter, I think we should call it `limit` mainly to match the terminology we use elsewhere but also because it's shorter, but just clarify with a comment that it's the max number of orders to return for each side of the order book.

kilrau:
> Looks good for the most part. One thing I wish I'd thought of sooner is that I think it would make more sense to have an optional `max_results` rather than an `all` field. If it's zero (the default value) we return all orders. This way a caller who wants more than the hardcoded 10 orders can fetch exactly how many orders they want. It should be a simple change from what you have here. Does that make sense @kilrau ?

`max_results` parameter is definitely good idea. My initial motivation for this was to prevent the platform/user from accidentally getting spammed with ALL orders whereas they are only interested in the top 10, 20 to e.g. show these in the UI. Thinking about this again it might not be that crazy important.

Could you clarify your suggestion?
`listorders` = 10 orders (hardcoded)
`listorders 0` = all orders
`listorders 20` = twenty orders

or

`listorders` = all orders
`listorders 20` = twenty orders

?

Honestly, the latter might just be the easiest to understand and ok with me.

sangaman:
I'm thinking the latter, zero (which is the default value if you don't specify a value for the parameter) returns all, specifying a limit returns that many.

kilrau:
> I'm thinking the latter, zero (which is the default value if you don't specify a value for the parameter) returns all, specifying a limit returns that many.

Yep, exactly. Good with you too? And sorry again for changing our minds on that ;)

lib/cli/commands/listorders.ts
```diff
@@ -88,20 +77,32 @@ const displayTables = (orders: ListOrdersResponse.AsObject) => {
...
     type: 'string',
   },
+  include_own_orders: {
+    describe: 'should include own orders',
+    type: 'boolean',
+    default: true,
+  },
+  limit: {
+    describe: 'return all orders',
+    type: 'boolean',
```
sangaman: `number` instead of `boolean`

lib/cli/commands/listorders.ts
```diff
@@ -88,20 +77,32 @@ const displayTables = (orders: ListOrdersResponse.AsObject) => {
...
+    default: false,
```
sangaman: default should be 0

lib/cli/commands/listorders.ts
```diff
@@ -88,20 +77,32 @@ const displayTables = (orders: ListOrdersResponse.AsObject) => {
...
+    describe: 'pair id for which to retrieve orders.',
     type: 'string',
   },
+  include_own_orders: {
+    describe: 'should include own orders',
+    type: 'boolean',
+    default: true,
+  },
+  limit: {
+    describe: 'return all orders',
```
sangaman: max number of orders to return

proto/xudrpc.proto
```diff
@@ -342,6 +342,8 @@ message ListOrdersRequest {
   string pair_id = 1 [json_name = "pair_id"];
   // Whether own orders should be included in result or not.
   bool include_own_orders = 2 [json_name = "include_own_orders"];
+  // Whether to return all orders
```
sangaman: The maximum number of orders to return from each side of the order book.

test/integration/Service.spec.ts
```diff
@@ -79,6 +79,7 @@ describe('API Service', () => {
     const args = {
       pairId,
       includeOwnOrders: true,
+      all: false,
```
sangaman: drop this line

lib/service/Service.ts
```diff
@@ -269,12 +270,15 @@ class Service extends EventEmitter {
   /**
    * Get a map between pair ids and its orders from the order book.
    */
-  public listOrders = (args: { pairId: string, includeOwnOrders: boolean }): Map<string, OrderSidesArrays<any>> => {
-    const { pairId, includeOwnOrders } = args;
+  public listOrders = (args: { pairId: string, includeOwnOrders: boolean, limit: boolean }): Map<string, OrderSidesArrays<any>> => {
```
sangaman: limit should be a number, not a boolean

lib/service/Service.ts
```diff
@@ -283,6 +287,11 @@ class Service extends EventEmitter {
         orders.sell = [...orders.sell, ...ownOrders.sell];
       }
 
+      if (!limit) {
+        // get 10 best orders
```
sangaman: The logic here per what kilian and I discussed should be `if (limit > 0)` then `slice(0, limit)`

lib/utils/utils.ts
```diff
@@ -155,3 +155,14 @@ export const removeUndefinedProps = (obj: any) => {
 
   return obj;
 };
+
+export const sortOrders = (orders: PeerOrder[], isBuy: boolean) => {
```
sangaman: Minor note but this should probably be `orders: Order[]` since we are not only interested in sorting Peer Orders.

ImmanuelSegol:
@sangaman I think there is a bug in the test, could you have a look.

lib/service/Service.ts
```diff
@@ -263,20 +264,37 @@ class Service extends EventEmitter {
...
         orders.buy = [...orders.buy, ...ownOrders.buy];
         orders.sell = [...orders.sell, ...ownOrders.sell];
       }
 
+      // sort all orders
+      orders.buy = sortOrders(orders.buy, true);
+      orders.sell = sortOrders(orders.sell, false);
+
+      if ((limit > 0)) {
+        // get 10 best peerOrdersorders
```
sangaman: Fix or remove this comment.

lib/cli/commands/listorders.ts
```diff
@@ -89,20 +78,32 @@ const displayTables = (orders: ListOrdersResponse.AsObject) => {
...
 };
 
-export const command = 'listorders [pair_id]';
+export const command = 'listorders [pair_id] [include_own_orders] [limit]';
 
 export const describe = 'list orders from the order book';
 
 export const builder = {
   pair_id: {
+    describe: 'pair id for which to retrieve orders.',
```
sangaman: I think it's a bit more descriptive to say "trading pair for which to retrieve orders" since "pair id"  is already in the name.

lib/cli/commands/listorders.ts
```diff
@@ -89,20 +78,32 @@ const displayTables = (orders: ListOrdersResponse.AsObject) => {
...
     type: 'string',
   },
+  include_own_orders: {
+    describe: 'should include own orders',
```
sangaman: "whether to include own orders" assuming that still fits on one line in the help output.

lib/service/Service.ts
```diff
@@ -7,8 +7,9 @@ import { EventEmitter } from 'events';
 import errors from './errors';
 import { SwapClient, OrderSide, SwapRole } from '../constants/enums';
 import { parseUri, toUri, UriParts } from '../utils/uriUtils';
+import { sortOrders } from '../utils/utils';
 import * as lndrpc from '../proto/lndrpc_pb';
-import { Pair, Order, OrderPortion, PlaceOrderEvent } from '../orderbook/types';
+import { Pair, Order, OrderPortion, PlaceOrderEvent, PeerOrder } from '../orderbook/types';
```
sangaman: Is `PeerOrder` actually being used here? It looks like an unused import.

lib/utils/utils.ts
```diff
@@ -1,6 +1,6 @@
 import http from 'http';
 import p2pErrors from '../p2p/errors';
-import { Pair } from '../orderbook/types';
+import { Pair, PeerOrder, Order } from '../orderbook/types';
```
sangaman: Also doesn't look like we're using `PeerOrder` here.

test/unit/Command.spec.ts
```diff
@@ -104,17 +104,6 @@ describe('Command.listorders.formatOrders', () => {
     const output = formatOrders(jsonOrders);
     expect(output.length).to.equal(1);
     expect(output[0].pairId).to.equal('LTC/BTC');
-    const expectedFirstRow = [
```
sangaman: This test doesn't do very much after these lines are removed. I'm thinking it might be nice to have some tests to verify the sorting and the limit parameter in the simulation tests, but that can be a separate PR.

PR #824, opened 2019-03-05
perf(p2p): use async randomBytes for packet framer

sangaman:
This replaces the synchronous usage of `crypto.randomBytes` with the asynchronous version when framing packets. The synchronous version can block the thread until enough entropy is collected. Since this is done every time an encrypted packet is sent, the delay may become noticeable.

This required rewriting synchronous code as asynchronous in several methods and tests which makes up the bulk of the code changes.

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -215,29 +215,29 @@ class Peer extends EventEmitter {
     await this.handshake(ownNodeState, nodeKey);
 
     if (this.expectedNodePubKey && this.nodePubKey !== this.expectedNodePubKey) {
-      this.close(DisconnectionReason.UnexpectedIdentity);
+      await this.close(DisconnectionReason.UnexpectedIdentity);
       throw errors.UNEXPECTED_NODE_PUB_KEY(this.nodePubKey!, this.expectedNodePubKey, addressUtils.toString(this.address));
     }
 
     if (this.nodePubKey === ownNodeState.nodePubKey) {
-      this.close(DisconnectionReason.ConnectedToSelf);
+      await this.close(DisconnectionReason.ConnectedToSelf);
       throw errors.ATTEMPTED_CONNECTION_TO_SELF;
     }
 
     // Check if version is semantic, and higher than minCompatibleVersion.
     if (!semver.valid(this.version)) {
-      this.close(DisconnectionReason.MalformedVersion);
+      await this.close(DisconnectionReason.MalformedVersion);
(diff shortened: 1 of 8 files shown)


lib/p2p/Peer.ts
```diff
@@ -748,35 +748,32 @@ class Peer extends EventEmitter {
...
       const sessionInit = await this.waitSessionInit();
-      this.ackSession(sessionInit, nodeKey);
+      return this.ackSession(sessionInit, nodeKey);
     } else {
       // inbound handshake
       const sessionInit = await this.waitSessionInit();
-      this.ackSession(sessionInit, nodeKey);
-      await this.initSession(ownNodeState, nodeKey, sessionInit.body!.nodeState.nodePubKey);
+      await this.ackSession(sessionInit, nodeKey);
+      return this.initSession(ownNodeState, nodeKey, sessionInit.body!.nodeState.nodePubKey);
```
moshababo: It might be simpler to do `await` instead of `return`, since that's how this function was structured. Unless you'll do `return` on the first `if`, and remove the `else`. But in this case, I think if/else makes more sense. 

Also, in many places in this PR you're doing `return` instead of `await`. But if the returned value is void, it might be less confusing to simply do `await`.

sangaman: I definitely agree in this example, and in regards to other places where I'm returning the `Promise<void>` explicitly rather than `await`, I did it mainly because it required fewer lines changed because I didn't have to make the functions `async`. But you may be right that the code is clearer without any `return` statements in those methods that aren't intended to return anything. Functionally they are the same, but if you still think it's a good idea to use `await` instead of `return` let me know and I'll change them elsewhere.

kilrau:
I understand this similar to replacing `/dev/random` with `/dev/urandom`. We definitely don't want to wait for entropy when encrypting packets. Good catch!!

lib/p2p/Peer.ts
moshababo: If it's just about adding `async` to the function, i'd say it's preferred over using `return`, which I found confusing.

sangaman:
> I understand this similar to replacing `/dev/random` with `/dev/urandom`. We definitely don't want to wait for entropy when encrypting packets. Good catch!!

I've read that `crypto.randomBytes` is actually calling `urandom` behind the scenes on linux boxes either way. The main thing here is that should we have to wait any amount of time at all to get back cryptographically secure random numbers, we don't block the thread/event loop.

PR #825, opened 2019-03-06
[WIP] fix(p2p): xucli connect to a peer which banned me shows success

### LINKED ISSUE #735: xucli connect to a peer which banned me shows success (but connect failed)
vagrant@xud-simnet-1:~$ xucli connect 030b20e28e42b95677587c7b2d4a7f2aeaf886491ef70648191ce57d55012e4b58@192.168.33.11:8885
success

vagrant@xud-simnet-2:~$ xucli connect 037c514ee25ebbfc10726442325bdf3210b6ee80a9b3618aa96788a06b01696535@192.168.33.10:8885
Error: 9 FAILED_PRECONDITION: could not connect to node 037c514ee25ebbfc10726442325bdf3210b6ee80a9b3618aa96788a06b01696535 because it is banned

reliveyy:
Fixes #735 
Implementation 1: Move peer banned checking from Pool#handleOpen to Peer#ackSession
Implementation 2: Let "xucli connect" success/fail until getting peer.active/close events

### DIFF
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -170,7 +170,32 @@ class Service extends EventEmitter {
     argChecks.HAS_HOST({ host });
     argChecks.VALID_PORT({ port });
 
+    const peerActive = new Promise((resolve) => {
+      this.pool.once('peer.active', (peerPubKey) => {
+        if (peerPubKey === nodePubKey) {
+          resolve();
+        }
+      });
+    });
+
+    const peerClose: Promise<string> = new Promise((resolve) => {
+      this.pool.once('peer.close', (peerPubKey, rejectionMsg) => {
+        if (peerPubKey === nodePubKey) {
+          resolve(rejectionMsg);
+        }
+      });
+    });
+
+    const timeout: Promise<void> = new Promise((resolve) => {
+      setTimeout(resolve, 1000);
+    });
+
     await this.pool.addOutbound({ host, port }, nodePubKey, retryConnecting, true);
+    await peerActive;
+    const rejectionMsg = await Promise.race([peerClose, timeout]);
+    if (rejectionMsg !== undefined) {
+      throw errors.CONNECTION_CLOSED(rejectionMsg);
+    }
   }
 
   /*
--- a/lib/service/errors.ts
+++ b/lib/service/errors.ts
@@ -4,6 +4,7 @@ const codesPrefix = errorCodesPrefix.SERVICE;
 const errorCodes = {
   INVALID_ARGUMENT: codesPrefix.concat('.1'),
   NOMATCHING_MODE_IS_REQUIRED: codesPrefix.concat('.2'),
+  CONNECTION_CLOSED: codesPrefix.concat('.3'),
 };
 
 const errors = {
@@ -14,6 +15,10 @@ const errors = {
   NOMATCHING_MODE_IS_REQUIRED: () => ({
     message: 'nomatching mode is required',
     code: errorCodes.NOMATCHING_MODE_IS_REQUIRED,
+  }),
+  CONNECTION_CLOSED: (rejectionMsg: string) => ({
+    message: 'connection closed: ' + rejectionMsg,
+    code: errorCodes.CONNECTION_CLOSED,
   }),
 };
 
--- a/lib/grpc/GrpcService.ts
+++ b/lib/grpc/GrpcService.ts
@@ -156,6 +156,7 @@ class GrpcService {
       case orderErrorCodes.CURRENCY_CANNOT_BE_REMOVED:
       case orderErrorCodes.MARKET_ORDERS_NOT_ALLOWED:
       case serviceErrorCodes.NOMATCHING_MODE_IS_REQUIRED:
+      case serviceErrorCodes.CONNECTION_CLOSED:
         code = status.FAILED_PRECONDITION;
         break;
       case lndErrorCodes.LND_IS_UNAVAILABLE:
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -40,11 +40,11 @@ interface Peer {
   on(event: 'nodeStateUpdate', listener: () => void): this;
   on(event: 'reputation', listener: (event: ReputationEvent) => void): this;
   once(event: 'open', listener: () => void): this;
-  once(event: 'close', listener: () => void): this;
+  once(event: 'close', listener: (rejectionMsg: string) => void): this;
   emit(event: 'connect'): boolean;
   emit(event: 'reputation', reputationEvent: ReputationEvent): boolean;
   emit(event: 'open'): boolean;
-  emit(event: 'close'): boolean;
+  emit(event: 'close', rejectionMsg: string): boolean;
   emit(event: 'error', err: Error): boolean;
   emit(event: 'packet', packet: Packet): boolean;
   emit(event: 'pairDropped', pair: string): boolean;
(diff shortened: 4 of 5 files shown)


moshababo:
@reliveyy looks good to begin with, although I would prefer to implement all `Pool.handleOpen` validations via callback provided to `Peer`, because otherwise `Peer.ackSession` flow will continue, while the peer is at closed state. So in this case, using an event emitted isn't the most appropriate IMO.

Also, @sangaman suggested a different approach [here](https://github.com/ExchangeUnion/xud/issues/735#issuecomment-470401348).

To finalize this PR, make sure we're applying all the validations, and not letting them to be run twice.

reliveyy:
@moshababo Tried @sangaman's suggestions. I hard coded 1 second timeout for `peer.close` event and pass `rejectionMsg` through `close` and `peer.close` events to reveal the true reason behind the peer connection close.

lib/service/Service.ts
```diff
@@ -170,7 +170,32 @@ class Service extends EventEmitter {
...
+    const peerClose: Promise<string> = new Promise((resolve) => {
+      this.pool.once('peer.close', (peerPubKey, rejectionMsg) => {
+        if (peerPubKey === nodePubKey) {
+          resolve(rejectionMsg);
+        }
+      });
+    });
+
+    const timeout: Promise<void> = new Promise((resolve) => {
+      setTimeout(resolve, 1000);
```
sangaman: One second seems a little bit short for me. This should be enough most of the time but given a few round trips needed for the handshake and the possibility of intermittent latency, maybe we could do 3 or 5 seconds?

lib/service/Service.ts
```diff
@@ -170,7 +170,32 @@ class Service extends EventEmitter {
...
+    });
+
     await this.pool.addOutbound({ host, port }, nodePubKey, retryConnecting, true);
+    await peerActive;
```
sangaman: If we never get the `peer.active` event, will we proceed past this line? I suspect not although I may be reading this wrong.

I was thinking you could combine the `peerActive` and `peerClose` promises into a single promise that resolves if it gets the active event and rejects if it gets the closed event. You could also put the timeout in there and have it reject if the timeout is reached.

lib/service/Service.ts
```diff
@@ -170,7 +170,32 @@ class Service extends EventEmitter {
...
+    const rejectionMsg = await Promise.race([peerClose, timeout]);
+    if (rejectionMsg !== undefined) {
```
sangaman: Wouldn't rejectionMsg be `undefined` if it times out? This could possibly be solved as well if you go with a single promise which rejects on timeout as I suggested above.

lib/service/errors.ts
```diff
@@ -15,6 +16,10 @@ const errors = {
     message: 'nomatching mode is required',
     code: errorCodes.NOMATCHING_MODE_IS_REQUIRED,
   }),
+  CONNECTION_CLOSED: (rejectionMsg: string) => ({
```
sangaman: Having an error for this is a good idea, but we should also add it to the switch statement in the `getGrpcError` method in GrpcService.ts. I figure this should probably go in the list of cases for `FAILED_PRECONDITION`.

lib/service/Service.ts
reliveyy: @sangaman If the delay is above 2 seconds then it will trigger p2p santity tests' failure

sangaman: Good point, maybe let's just make it 1.5 seconds for now? We can revisit this timeout limit later if it's a problem.

lib/service/Service.ts
reliveyy: @sangaman `peer.active` will always be triggered first. Although the peer which is banned will receive `peer.close` later. So put active and close in one Promise will always get resolved.

lib/service/Service.ts
reliveyy: When I put `setTimeout` into the `new Promise`, it will be executed immediately. And if addOutbound or peerActive blocks then timeout reject will be triggered first. And that timeout is  not the time waiting close packet but the time waiting addOutbound, peerActive, peerClose all together I think.

lib/service/errors.ts
reliveyy: Good point!

reliveyy:
Here is what I think about the pros and cons between two implementations:

### Impl. 1: Move peer banned checking from Pool#handleOpen to Peer#ackSession

Cons:
* Expose peer `ackSession` lifecycle event. (Maybe you don't want to expose such kind of events)
* Move some peer checking logic from Pool#handleOpen to a new peer `ackSession` event callback

Pros:
* Don't need to care about the `xucli connect` timeout for waiting the close packet. Because banned rejections will be received in ackSession packet before it.

### Impl. 2: Let "xucli connect" success/fail until getting peer.active/close events

Cons:
* It's hard to determine the timeout length for the close packet waiting. And it will add some useless waiting period in unit tests.
* Need to pass through `rejectionMsg` from Peer to Pool then to Service

Pros:
* Don't need to change to much about Peer and Pool's peer checking logic

IMO `xucli connect` should not wait too long and the errors it can feedback to the user is some **FAST** rejection errors like peer banned checks. And the waiting timeout for the close packet depends on some unpredictable factors like networking. Also we cannot guarantee the close packet will coming soon in some other situations in the feature. So do we need to increase the timeout for those cases?

So my suggestion is categorizing peer connecting rejections into two groups: fast and slow. For the fast rejections we should put checking logic into the ackSession and for slow rejection we can put checking logic everywhere. And `xucli connect` response only shows the fast rejections, meanwhile, the slow rejections should be printed in the xud log. @moshababo @sangaman @kilrau

lib/service/Service.ts
sangaman: Why would it be executed immediately? I realize that the time to complete the handshake may actually take longer than our timeout limit, but that's going to be an issue regardless of where the timeout is in this method.

lib/service/Service.ts
sangaman: No I don't think `peer.active` is triggered if the peer is banned or if we don't complete the connection for any reason, we have logic right before where we emit `peer.active` to make these checks. So I still think we won't be able to proceed past this line when a peer is banned.

sangaman:
I actually don't think we'll need to have useless waiting for the regression tests, we will emit `peer.close` immediately after the handshake is complete for a banned peer.

The main problem I'm thinking we'd have with checking banned logic within `Peer.ts` is that the peer object is not aware of which nodes are banned, only `Pool` does. So I think lets stick to this approach, at least for now.

lib/service/Service.ts
reliveyy: @sangaman We got `peer.active` on the banned peer because the ban-assertion on it only checks it's own ban list. But our case is it is banned by another peer. So the fact is the other peer's ban-assertion works and `peer.active` will not be emitted. But on our side it will bypass the peer ban-assertion and emit `peer.active`

reliveyy:
> I actually don't think we'll need to have useless waiting for the regression tests, we will emit `peer.close` immediately after the handshake is complete for a banned peer.

@sangaman The useless waiting I think is not in the banned case but the normal case. Because in the normal case, we will wait until timeout to ensure the connection is not closed by the other peer.

sangaman:
I see what you mean. But I'd also rather not have to depend on waiting an arbitrary amount of time to determine whether a connection was successful or not.

Maybe then the best approach really is to check if a node is banned before we respond with `SessionAck`. It would require some async/event-driven communication between Peer and Pool but it's possible. Does that seem sensible to you?

sangaman:
I think this can probably be closed given the underlying issue seems to be solved per #848.

kilrau:
Could you test #848 and close this issue here if it's fixed by #848? @reliveyy Don't forget to leave a review :)

reliveyy:
Since https://github.com/ExchangeUnion/xud/issues/735 has been resolved by #848, so this PR.

Issue #826, opened 2019-03-07
Add minimum viable RaidenClient REST API integration


erkarl:
Offshoot from https://github.com/ExchangeUnion/xud/issues/797

## Base Client requirements
- [x] getInfo (for general sanity checks before attempting the swap)
  - [x] isConnected
  - [x] getAddress
- [x] sendPayment (for swaps)
- [x] getChannels (for xucli channelbalance)

### Everything else
- open channel
- close channel
- depositing to channel
- join token network
- bootstrap token network
- register token
- check if token registered

Everything else classifies as nice to have functionality since we do the same manual channel management for Lightning implementations as well. Ultimately, xud should abstract channel management away from the end user, but it is not required for ERC20 <-> BTC/LTC swaps.

kilrau:
Added feedback directly above.

PR #828, opened 2019-03-12
feat(raiden): add minimum viable RaidenClient REST API integration 

ghost:
https://github.com/ExchangeUnion/xud/issues/826

In this PR we introduce we add updated support for Raiden REST API:
- basic connection verification and retry mechanism
- `xucli getinfo` correctly displays raiden status/address
- `xucli channelbalance` correctly displays WETH balance after adding it with `xucli addcurrency`
- `RaidenClient.sendPayment` - supports providing custom secret & secret hash (does not work with [v0.100.2](https://github.com/raiden-network/raiden/releases/tag/v0.100.2) binaries, compile latest Raiden master from source)
- removed [legacy tokenSwap](https://github.com/raiden-network/raiden/commit/345a89621b2c8895229395f595f05e8e948706fa) endpoint

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -51,18 +51,14 @@ class LndClient extends BaseClient {
   private meta!: grpc.Metadata;
   private uri!: string;
   private credentials!: ChannelCredentials;
-  private reconnectionTimer?: NodeJS.Timer;
   private identityPubKey?: string;
   private invoiceSubscription?: ClientReadableStream<lndrpc.InvoiceSubscription>;
-
-  /** Time in milliseconds between attempts to recheck connectivity to lnd if it is lost. */
-  private static readonly RECONNECT_TIMER = 5000;
 
   /** Time in milliseconds between attempts to recheck if lnd's backend chain is in sync. */
   private static readonly RECHECK_SYNC_TIMER = 30000;
 
   /**
-   * Create an lnd client.
+   * Creates an lnd client.
    * @param config the lnd configuration
    */
   constructor(private config: LndClientConfig, logger: Logger) {
@@ -167,7 +163,7 @@ class LndClient extends BaseClient {
   }
 
   /**
-   * Verify that the lnd gRPC service can be reached by attempting a `getInfo` call.
+   * Verifies that the lnd gRPC service can be reached by attempting a `getInfo` call.
    * If successful, subscribe to invoice events and store the lnd identity pubkey.
    * If not, set a timer to attempt to reach the service again in 5 seconds.
    */
@@ -212,7 +208,7 @@ class LndClient extends BaseClient {
   }
 
   /**
-   * Return general information concerning the lightning node including it’s identity pubkey, alias, the chains it
+   * Returns general information concerning the lightning node including it’s identity pubkey, alias, the chains it
    * is connected to, and information concerning the number of open+pending channels.
    */
   public getInfo = (): Promise<lndrpc.GetInfoResponse> => {
@@ -220,14 +216,14 @@ class LndClient extends BaseClient {
   }
 
   /**
-   * Send a payment through the Lightning Network.
+   * Sends a payment through the Lightning Network.
    */
   public sendPaymentSync = (request: lndrpc.SendRequest): Promise<lndrpc.SendResponse> => {
     return this.unaryCall<lndrpc.SendRequest, lndrpc.SendResponse>('sendPaymentSync', request);
   }
 
   /**
-   * Get a new address for the internal lnd wallet.
+   * Gets a new address for the internal lnd wallet.
    */
   public newAddress = (addressType: lndrpc.NewAddressRequest.AddressType): Promise<lndrpc.NewAddressResponse> => {
     const request = new lndrpc.NewAddressRequest();
@@ -236,21 +232,24 @@ class LndClient extends BaseClient {
   }
 
   /**
-   * Return the total of unspent outputs for the internal lnd wallet.
+   * Returns the total of unspent outputs for the internal lnd wallet.
    */
   public walletBalance = (): Promise<lndrpc.WalletBalanceResponse> => {
     return this.unaryCall<lndrpc.WalletBalanceRequest, lndrpc.WalletBalanceResponse>('walletBalance', new lndrpc.WalletBalanceRequest());
   }
 
   /**
-   * Return the total balance available across all channels.
+   * Returns the total balance available across all channels.
    */
-  public channelBalance = (): Promise<lndrpc.ChannelBalanceResponse> => {
-    return this.unaryCall<lndrpc.ChannelBalanceRequest, lndrpc.ChannelBalanceResponse>('channelBalance', new lndrpc.ChannelBalanceRequest());
+  public channelBalance = async (): Promise<lndrpc.ChannelBalanceResponse.AsObject> => {
+    const channelBalanceResponse = await this.unaryCall<lndrpc.ChannelBalanceRequest, lndrpc.ChannelBalanceResponse>(
+      'channelBalance', new lndrpc.ChannelBalanceRequest(),
+    );
+    return channelBalanceResponse.toObject();
   }
 
   /**
-   * Connect to another lnd node.
+   * Connects to another lnd node.
    */
   public connectPeer = (pubkey: string, host: string, port: number): Promise<lndrpc.ConnectPeerResponse> => {
     const request = new lndrpc.ConnectPeerRequest();
@@ -262,7 +261,7 @@ class LndClient extends BaseClient {
   }
 
   /**
-   * Open a channel with a connected lnd node.
+   * Opens a channel with a connected lnd node.
    */
   public openChannel = (node_pubkey_string: string, local_funding_amount: number): Promise<lndrpc.ChannelPoint> => {
     const request = new lndrpc.OpenChannelRequest;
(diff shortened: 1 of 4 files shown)


lib/raidenclient/RaidenClient.ts
```diff
@@ -309,6 +281,31 @@ class RaidenClient extends BaseClient {
     const res = await this.sendRequest(endpoint, 'PATCH', { state: 'settled' });
   }
 
+  /**
+   * Send a payment.
+   * @param target_address recipient of the payment
+   * @param token_address contract address of the token
+   * @param amount
+   * @param secret optional; provide your own secret
```
offerm: We don't need the secret for XUD functionality.
I suggest to remove it.

lib/raidenclient/errors.ts
```diff
@@ -35,6 +35,10 @@ const errors = {
     message: 'unexpected error during raiden request',
     code: errorCodes.UNEXPECTED,
   },
+  MISSING_SECRET: {
+    message: 'providing only the secret hash is currently not supported - please provide a secret',
+    code: errorCodes.UNEXPECTED,
+  },
```
offerm: You can remove it.

When running raiden, you can use  the exteranl-swap-resolver  from https://github.com/offerm/raiden.git.

lib/raidenclient/RaidenClient.ts
```diff
@@ -125,19 +108,35 @@ class RaidenClient extends BaseClient {
...
+      this.reconnectionTimer = setTimeout(this.verifyConnection, RaidenClient.RECONNECT_TIMER);
     }
   }
 
   public getRaidenInfo = async (): Promise<RaidenInfo> => {
     let channels: number | undefined;
     let address: string | undefined;
     let error: string | undefined;
-    const version = 'v0.3.0'; // Hardcoded for now until they expose it to their API;
+    const version = 'v0.100.2'; // Hardcoded for now until they expose it to their API;
```
offerm: Why do we need this version?

lib/lndclient/LndClient.ts
```diff
@@ -245,8 +245,11 @@ class LndClient extends BaseClient {
   /**
    * Return the total balance available across all channels.
    */
-  public channelBalance = (): Promise<lndrpc.ChannelBalanceResponse> => {
-    return this.unaryCall<lndrpc.ChannelBalanceRequest, lndrpc.ChannelBalanceResponse>('channelBalance', new lndrpc.ChannelBalanceRequest());
+  public channelBalance = async (): Promise<lndrpc.ChannelBalanceResponse.AsObject> => {
```
offerm: Why going back to use AsObject? is there a reason for it?

lib/raidenclient/RaidenClient.ts
erkarl: You're right, we don't need this - it's currently just for display purposes when querying for `xucli getinfo`. At the moment Raiden client does not expose its version via the API - I imagine they will in the future.

lib/lndclient/LndClient.ts
sangaman: I think that's the right idea but then maybe let's define this method signature on the `BaseClient` class, then override it here. I also think that rather than using the `AsObject` type we can define an internal type that mirrors it, just so that the Raiden client isn't relying on type definitions from lnd.

lib/raidenclient/RaidenClient.ts
```diff
@@ -99,8 +80,10 @@ class RaidenClient extends BaseClient {
   public address?: string;
   private port: number;
   private host: string;
-  /** Map of token swap identifiers to order ids */
-  private swapIdOrderMap = new Map<number, Order>();
+  private reconnectionTimer?: NodeJS.Timer;
+
+  /** Time in milliseconds between attempts to recheck connectivity to Raiden. */
+  private static readonly RECONNECT_TIMER = 5000;
```
sangaman: This and the `reconnectionTimer` can also be moved to `BaseClient`, I think they'll be used the same way.

lib/raidenclient/RaidenClient.ts
sangaman: Maybe lets just leave version blank until they expose it? I wouldn't want to have to keep manually updating this line.

lib/raidenclient/RaidenClient.ts
```diff
@@ -166,15 +165,11 @@ class RaidenClient extends BaseClient {
    */
   private sendRequest = (endpoint: string, method: string, payload?: object): Promise<http.IncomingMessage> => {
     return new Promise((resolve, reject) => {
-      if (!this.isConnected()) {
```
sangaman: We may still want to have a check here to verify that we are connected, right? Even though it wouldn't necessarily be disabled.

lib/raidenclient/RaidenClient.ts
```diff
@@ -309,6 +281,24 @@ class RaidenClient extends BaseClient {
     const res = await this.sendRequest(endpoint, 'PATCH', { state: 'settled' });
   }
 
+  /**
+   * Send a payment.
```
sangaman: Minor note but I've been trying to follow the present-tense for method comment blocks like "Sends a payment" rather than "Send a payment."

lib/service/Service.ts
```diff
@@ -124,21 +124,24 @@ class Service extends EventEmitter {
     const { currency } = args;
     const balances = new Map<string, { balance: number, pendingOpenBalance: number }>();
     const getBalance = async (currency: string) => {
-      let cmdLnd: LndClient;
+      let client: LndClient | RaidenClient;
```
sangaman: Here it would be good to use `BaseClient`

lib/service/Service.ts
```diff
@@ -124,21 +124,24 @@ class Service extends EventEmitter {
...
         case 'LTC':
-          cmdLnd = this.lndLtcClient;
+          client = this.lndLtcClient;
+          break;
+        case 'WETH':
+          client = this.raidenClient;
           break;
         default:
-          // TODO: throw an error here indicating that lnd is disabled for this currency
+          // TODO: throw an error here indicating that client is disabled for this currency
```
sangaman: "the client"

lib/raidenclient/RaidenClient.ts
```diff
@@ -309,6 +281,24 @@ class RaidenClient extends BaseClient {
...
+   * @param target_address recipient of the payment
+   * @param token_address contract address of the token
+   * @param amount
+   * @param secret_hash optional; provide your own secret hash
+   */
+  public sendPayment = async (token_address: string, target_address: string, amount: number, secret_hash?: string): Promise<{}> => {
```
sangaman: One more thought, we'll want a common `sendPayment` interface for both lnd and raiden. Right now lnd has `sendPaymentSync`, and right now I'm looking at making that private and exposing a public `sendPayment` method with generic parameters that's defined on the base class. If you want to structure this similarly, maybe you could name this `tokenPayment` and make it private, then (possibly in a future PR) implement the public `sendPayment` method.

lib/raidenclient/RaidenClient.ts
erkarl: Good idea - done.

lib/raidenclient/RaidenClient.ts
erkarl: I'll address this in a future PR as I already have some logic implemented for BaseClient timers (for channelbalance sanity checks/caching).

lib/raidenclient/RaidenClient.ts
erkarl: I thought about keeping it, but the `verifyConnection` also uses `sendRequest` internally - making it fail because the client is not connected. I keep pondering whether this is the correct place to do this check. Analogically, `lndClient` doesn't have such safeguards.

lib/lndclient/LndClient.ts
erkarl: Done. `BaseClient` now has the method signature.

lib/raidenclient/RaidenClient.ts
sangaman: OK that makes sense, I think I may have been confusing this method with `sendPayment` actually. This isn't the place for the check, thanks.

lib/raidenclient/RaidenClient.ts
sangaman: I'll open a PR with a BaseClient `sendPayment` interface soon, right now I'm thinking of having it just accept a `SwapDeal` parameter and then have the logic to extract the necessary fields from it within.

erkarl: 👍 - I rename the current implementation to `tokenPayment` and made it private.

lib/service/Service.ts
```diff
@@ -120,15 +121,16 @@ class Service extends EventEmitter {
     const { currency } = args;
     const balances = new Map<string, { balance: number, pendingOpenBalance: number }>();
     const getBalance = async (currency: string) => {
-      const cmdLnd = this.lndClients[currency.toUpperCase()];
+      const client = this.lndClients[currency.toUpperCase()]
+        // TODO: support all registered tokens for Raiden
+        || (currency.toUpperCase() === 'WETH' && this.raidenClient);
 
-      if (!cmdLnd) {
-        // TODO: throw an error here indicating that lnd is disabled for this currency
```
sangaman: We might want to keep this todo, but maybe change the message to "indicating that there is no swap client for this currency". I feel like an error is appropriate if we're asked for the channel balance of a currency we don't support.

sangaman (approved):
Looks good, I haven't tested this myself but if you've done some smoke testing on your end then you can go ahead and merge. Thanks.

erkarl:
Going to merge this as I've [been building](https://github.com/ExchangeUnion/xud/pull/833) on top of this branch for several days now - haven't noticed problems.

PR #836, opened 2019-03-17
fix(p2p): fix connection timeout unhandled error

sangaman:
This fixes a bug where xud could crash due to an unhandled error event when a socket connection attempt times out due to an unreachable peer. Previously we manually set a timeout upon attempting a connection which would call the `onError` function, which in turn called `cleanup` to remove all listeners. The socket is not destroyed in cases where we automatically retry connecting, but the `bind` function that sets new listeners isn't called until the delay between automatic retries is up. Since the socket would still be live, when the `ETIMEDOUT` internal error is raised there is no listener and it crashes the process.

This removes the manual connection timeout and instead relies on the internal `ETIMEDOUT` error. I have tested this and failed connection attempts still time out and get retried as expected, but without risk of the process crashing.

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -66,7 +66,6 @@ class Peer extends EventEmitter {
   private closed = false;
   /** Timer to retry connection to peer after the previous attempt failed. */
   private retryConnectionTimer?: NodeJS.Timer;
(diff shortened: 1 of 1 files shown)


sangaman:
Going to merge to see if this clears up issues getting in the way of other testing.

PR #840, opened 2019-03-18
feat(rpc): add timestamps

### LINKED ISSUE #685: streamOrders - missing timestamp
CLI streamorders is missing the timestamp for each event. 
Best if the time stamp will be the event timestamp and not the stream's data timestamp.
It would be great if we have a similar time stamp as the log so we can easily find the event in the log.
Time stamp should include milliseconds.

reliveyy:
* Order: executed_at
* OrderRemoval: removed_at
* SwapFailure: fail_time

Resolves #685

### DIFF
--- a/lib/grpc/GrpcService.ts
+++ b/lib/grpc/GrpcService.ts
@@ -610,6 +610,7 @@ class GrpcService {
       orderRemoval.setQuantity(order.quantity);
       orderRemoval.setLocalId(order.localId || '');
       orderRemoval.setIsOwnOrder(order.localId !== undefined);
+      orderRemoval.setRemovedAt(Date.now());
       call.write(orderRemoval);
     });
     this.addStream(call);
--- a/proto/xudrpc.proto
+++ b/proto/xudrpc.proto
@@ -306,6 +306,8 @@ message SwapFailure {
   string peer_pub_key = 4 [json_name = "peer_pub_key"];
   // The reason why the swap failed.
   string failure_reason = 5 [json_name = "failure_reason"];
+  // The epoch time in milliseconds when the swap failed.
+  int64 fail_time = 6 [json_name = "fail_time"];
 }
 
 message GetInfoRequest {}
@@ -422,6 +424,8 @@ message OrderRemoval {
   string local_id = 4 [json_name = "local_id"];
   // Whether the order being removed is a local own order or a remote peer order.
   bool is_own_order = 5 [json_name = "is_own_order"];
+  // The epoch time in milliseconds when the order was removed from the order book.
+  int64 removed_at = 6 [json_name = "removed_at"];
 }
 
 message Orders {
@@ -443,7 +447,7 @@ message Peer {
   string address = 1 [json_name = "address"];
   // The node pub key to uniquely identify this peer.
   string node_pub_key = 2 [json_name = "node_pub_key"];
-  // A map of ticker symbols to lnd pub keys for this peer 
+  // A map of ticker symbols to lnd pub keys for this peer.
   map<string, string> lnd_pub_keys = 3 [json_name = "lnd_pub_keys"];
   // Indicates whether this peer was connected inbound.
   bool inbound = 4 [json_name = "inbound"];
@@ -570,6 +574,8 @@ message SwapSuccess {
   string r_preimage = 14 [json_name = "r_preimage"];
   // The price used for the swap.
   double price = 15 [json_name = "price"];
+  // The epoch time in milliseconds when the swap was completed.
+  int64 complete_time = 16 [json_name = "complete_time"];
 }
 
 message UnbanRequest {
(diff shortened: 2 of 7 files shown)


kilrau:
Could somebody review this so that @reliveyy can continue? @erkarl @sangaman @moshababo

proto/xudrpc.proto
```diff
@@ -306,6 +306,8 @@ message SwapFailure {
   string peer_pub_key = 4 [json_name = "peer_pub_key"];
   // The reason why the swap failed.
   string failure_reason = 5 [json_name = "failure_reason"];
+  // The fail time for the swap.
```
sangaman: I might say "The epoch time in milliseconds when the swap failed" to be a little more clear.

proto/xudrpc.proto
```diff
@@ -409,6 +411,8 @@ message Order {
   bool is_own_order = 9 [json_name = "is_own_order"];
   // The amount on hold pending swap exectuion.
   double hold = 10 [json_name = "hold"];
+  // The execution time for the Order.
```
sangaman: "The epoch time in milliseconds when the order was executed."

proto/xudrpc.proto
```diff
@@ -422,6 +426,8 @@ message OrderRemoval {
   string local_id = 4 [json_name = "local_id"];
   // Whether the order being removed is a local own order or a remote peer order.
   bool is_own_order = 5 [json_name = "is_own_order"];
+  // The removal time for the Order.
```
sangaman: "The epoch time in milliseconds when the order was removed from the order book."

proto/xudrpc.proto
```diff
@@ -443,7 +449,7 @@ message Peer {
   string address = 1 [json_name = "address"];
   // The node pub key to uniquely identify this peer.
   string node_pub_key = 2 [json_name = "node_pub_key"];
-  // A map of ticker symbols to lnd pub keys for this peer 
+  // A map of ticker symbols to lnd pub keys for this peer
```
sangaman: Good catch on the whitespace, but while we're changing this line we might as well add a period at the end to make it consistent with the other comments.

proto/xudrpc.proto
```diff
@@ -568,6 +574,8 @@ message SwapSuccess {
   string r_preimage = 14 [json_name = "r_preimage"];
   // The price used for the swap.
   int64 price = 15 [json_name = "r_preimage"];
+  // The completion time for the swap.
```
sangaman: "The epoch time in milliseconds when the swap was completed."

lib/grpc/GrpcService.ts
```diff
@@ -32,6 +32,7 @@ const createOrder = (order: Order) => {
   grpcOrder.setPrice(order.price);
   grpcOrder.setQuantity(order.quantity);
   grpcOrder.setSide(order.isBuy ? xudrpc.OrderSide.BUY : xudrpc.OrderSide.SELL);
+  grpcOrder.setExecutedAt(Date.now());
```
sangaman: We don't want to always set this to the current date, because then even standing, unexecuted orders (like those we get back from `ListOrders`) will show that they've been executed.

I'm actually thinking we should drop the `executed_at` proto field for now because it's not actually an event from the perspective of the API right now. Also it's a little complicated given that orders can be executed partially at various different times.

lib/grpc/GrpcService.ts
```diff
@@ -66,6 +67,7 @@ const createSwapFailure = (swapFailure: SwapFailure) => {
   grpcSwapFailure.setPeerPubKey(swapFailure.peerPubKey);
   grpcSwapFailure.setQuantity(swapFailure.quantity);
   grpcSwapFailure.setFailureReason(SwapFailureReason[swapFailure.failureReason]);
+  grpcSwapFailure.setFailTime(Date.now());
```
sangaman: This won't give us accurate times when we use the synchronous place order calls. I'm thinking we should add a `failTime` property to the `SwapDeal` type (and probably for the db as well, although that can be a separate PR) and have the `SwapFailure` type inherit it, then assign it a value of `Date.now()` when we create the `SwapFailure` in https://github.com/ExchangeUnion/xud/blob/master/lib/orderbook/OrderBook.ts#L388. Then we can use that value here.

reliveyy: I will create a new PR for it. @sangaman

lib/grpc/GrpcService.ts
reliveyy: Agree.

lib/grpc/GrpcService.ts
sangaman: Can you assign this a value in this PR at least in the rpc layer? I was thinking of a separate PR only for the db option. Or if you think that's too much, let's remove `fail_time` altogether from this PR, I just don't think it makes sense to add it to the proto file but not assign a value to it.

reliveyy: Sorry. Forgot to remove that gRPC proto field.

PR #841, opened 2019-03-19
fix(p2p): handle packets received during handshake

### LINKED ISSUE #839: seed nodes keep disconnecting
the connection to the seed nodes is not stable.

log attached

[xud.log](https://github.com/ExchangeUnion/xud/files/2976279/xud.log)

sangaman:
This commit addresses an edge case bug whereby one peer finishes the handshake before the other and begins sending regular packets. It is possible for the other peer to attempt to process these regular packets while it is executing asynchronous code to complete the handshake, which results in the regular packets being discarded as "unsolicited."

This makes two changes:

1. Moves the logic to asynchronously send a `GetNodesPacket` until after we mark the peer as `opened`.
2. Add an `initializing` flag that is set to true once we receive a `SessionAckPacket`. This is an indicator that any subsequent packets we receive should not be acted upon until the handshake is complete. If
we receive packets while `opened` is false but `initializing` is true, we wait to see which of the `open` or `close` events fire first to either accept or discard a packet, respectively.

Fixes #839.

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -60,6 +60,8 @@ class Peer extends EventEmitter {
   public active = false; // added to peer list
   /** Timer to periodically call getNodes #402 */
   public discoverTimer?: NodeJS.Timer;
+  /** Whether we have received a [[SessionAckPacket]] and are in the process of completing the handshake to open the connection. */
+  private initializing = false;
   private opened = false;
   private socket?: Socket;
   private parser: Parser;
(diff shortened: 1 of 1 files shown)


sangaman:
I've been testing this branch quite a bit and it's working fine.

offerm:
Did you consider to mark the connection open before sending the ack?

sangaman:
Not exactly, although I'm thinking that would work too and be simpler. The only thing is for that approach, we'd need to make all our checks as to whether the peer is valid (not banned, node pub key matches the advertised one, not ourselves, etc) before responding with the ack, which is already something being discussed as part of #825. We can talk about the best way forward on the call.

sangaman:
I took a totally different approach that I think makes some good structural improvements to the code and solves the same issue. Since it's very different from this PR, I'm closing this one and opening a separate one shortly.

PR #843, opened 2019-03-19
Deprecation fix

sangaman:
This reverts commit a352e29.

This is a temporary measure until an underlying bug with the `grpc-tools` dependency is fixed and we can upgrade to a newer version without issue.


kilrau:
Works!

PR #844, opened 2019-03-19
fix(raiden): cleanup reconnectTimer on shutdown

ghost:
This commit fixes a bug where the Raiden reconnect timer would not cleanup when triggering a graceful shutdown.

In addition, I have already fixed the issue in my other feature branch where I moved the cleanup logic into the BaseClient.

### DIFF
--- a/lib/BaseClient.ts
+++ b/lib/BaseClient.ts
@@ -45,6 +45,14 @@ abstract class BaseClient extends EventEmitter {
   public isDisconnected(): boolean {
     return this.status === ClientStatus.Disconnected;
   }
+  /** Ends all connections, subscriptions, and timers for for this client. */
+  public close() {
+    if (this.reconnectionTimer) {
+      clearTimeout(this.reconnectionTimer);
+    }
+    this.closeSpecific();
+  }
+  protected abstract closeSpecific(): void;
 }
 
 export default BaseClient;
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -340,14 +340,10 @@ class LndClient extends BaseClient {
       });
   }
 
-  /** Ends all subscriptions and reconnection attempts. */
-  public close = () => {
+  /** Lnd client specific cleanup. */
+  protected closeSpecific = () => {
     if (this.invoiceSubscription) {
       this.invoiceSubscription.cancel();
-    }
-
-    if (this.reconnectionTimer) {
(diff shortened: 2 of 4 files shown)


sangaman (approved):

sangaman:
Looks good. I think it'd be nice (but not necessary) to have the reconnection timer cleanup logic in the base client since. Do you want to just merge this for now and then refactor it in your next branch? I figure you could handle that in this PR if it's not any trouble.

erkarl:
Good idea. This will make the other PR smaller and easier to review. I've moved the reconnect cleanup logic into the abstract `BaseClient`.

lib/BaseClient.ts
```diff
@@ -45,6 +45,13 @@ abstract class BaseClient extends EventEmitter {
   public isDisconnected(): boolean {
     return this.status === ClientStatus.Disconnected;
   }
+  public async close() {
+    if (this.reconnectionTimer) {
+      clearTimeout(this.reconnectionTimer);
+    }
+    await this.stop();
+  }
+  protected abstract stop(): Promise<void>;
```
sangaman: Why have this return a promise if we're not doing anything asynchronous with it?

lib/lndclient/LndClient.ts
```diff
@@ -340,15 +340,12 @@ class LndClient extends BaseClient {
       });
   }
 
-  /** Ends all subscriptions and reconnection attempts. */
-  public close = () => {
+  /** Ends all subscriptions attempts. */
```
sangaman: Maybe have this comment say something along the lines of "Lnd specific cleanup" (like you've done for Raiden) and move this general description to the base `close` method. Also "subscriptions attempts" looks like a typo here, maybe best to say something like "Ends all connections, subscriptions, and timers for for this client."

lib/BaseClient.ts
sangaman: Another thought I have here is that "close" and "stop" might get confused with each other, maybe it'd be more descriptive to call this method "closeSpecific" or something like that?

erkarl: > Why have this return a promise if we're not doing anything asynchronous with it?
I was assuming we're going to have async cleanup in the future, but you're right - it's not necessary right now. Removed it.

> maybe it'd be more descriptive to call this method "closeSpecific"
`closeSpecific` sounds great. Changed.

sangaman (approved):
Looks good, thanks! Just squash the commits then merge.

PR #845, opened 2019-03-19
fix(p2p): open peer before sending session ack

### LINKED ISSUE #839: seed nodes keep disconnecting

sangaman:
This commit addresses an edge case bug whereby one peer finishes the handshake before the other and begins sending regular packets. It is possible for the other peer to attempt to process these regular packets while it is executing asynchronous code to complete the handshake, which results in the regular packets being discarded as "unsolicited."
 
The key change is marking the peer as `opened` after receiving and authenticating the `SessionInitPacket`, as well as validating the peer (not banned, not ourselves, not an incompatible version, etc...), but before we send the `SessionAckPacket`. This ensures that we are ready to accept packets before the peer is ready to send packets.
 
This involves restructuring the open flow, splitting it into two methods (`beginOpen` and `completeOpen`). The first establishes the socket connection and begins the handshake. The second marks the peer as `opened` and compelts the handshake. In between we validate the peer, and afterwards we perform the routine for newly opened peers including sharing and asking for open orders.
 
Fixes #839.

### DIFF
--- a/lib/grpc/GrpcService.ts
+++ b/lib/grpc/GrpcService.ts
@@ -164,6 +164,9 @@ class GrpcService {
       case p2pErrorCodes.COULD_NOT_CONNECT:
         code = status.UNAVAILABLE;
         break;
+      case p2pErrorCodes.POOL_CLOSED:
+        code = status.ABORTED;
+        break;
     }
 
     // return a grpc error with the code if we've assigned one, otherwise pass along the caught error as UNKNOWN
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -4,7 +4,6 @@ import { EventEmitter } from 'events';
 import crypto from 'crypto';
 import secp256k1 from 'secp256k1';
 import stringify from 'json-stable-stringify';
-import semver from 'semver';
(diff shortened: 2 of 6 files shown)


sangaman:
I've done a good amount of testing with xud-simnet with this branch and it's looking good, both between two nodes using this branch and with the seed nodes which are still on master.

kilrau:
Gave this some quick tests on simnet and connections & swaps work as intended. Could you give this your final go @offerm ? You are the one who could reliably reproduce these handshake asymmetries.

offerm:
Will do later today

sangaman:
It will still be possible to see the bug in #839 coming from nodes that are not on this branch which includes the seed nodes, because they discard packets on their side.

sangaman:
Also I believe this will actually solve #735 without any further changes, although I'll need to test that to make sure.

offerm:
Suggest you merge it and I will check after you do. Im checking against the
seed nodes anyway

kilrau:
> Also I believe this will actually solve #735 as well without any further changes, although I'll need to test that to make sure.

if you are on, please check that @reliveyy

PR #847, opened 2019-03-20
refactor: remove local_id from SwapSuccess

### LINKED ISSUE #733: remove local_id from `SwapResult` in nomatching mode

ImmanuelSegol:
closes #733 
Totally removes local_id according to:

> decision from the call: simply remove local_id from SwapResult response (both matching and nomatching mode) @ImmanuelSegol


lib/proto/google/api/http_pb.js
```diff
@@ -153,15 +153,15 @@ proto.google.api.Http.serializeBinaryToWriter = function(message, writer) {
 
 /**
  * repeated HttpRule rules = 1;
- * @return {!Array.<!proto.google.api.HttpRule>}
+ * @return {!Array<!proto.google.api.HttpRule>}
```
erkarl: I'm curious as to why we're getting diffs like this? @sangaman, is this related to the grpc-tools version changes?

sangaman: Yes I think so, @ImmanuelSegol could you run `npm install` and then `npm run proto` again?

ImmanuelSegol:
@kilrau @sangaman Also here the failed test are due to Travis, locally they work just fine.

kilrau:
Restarted Travis build

rsercano:
There's a test case in the GO tests which checks local id, that's why tests are failing; `order_matching_and_swap`

sangaman:
@rsercano Do you need help modifying the go simulation test not to check the local id? It shoulnd't be too hard but if you're not sure where to start I can point you in the right direction. Thanks.

rsercano:
Commented it out @sangaman, thanks a lot.

test/simulation/actions.go
```diff
@@ -228,7 +228,7 @@ func (a *actions) placeOrderAndSwap(srcNode, destNode *xudtest.HarnessNode,
 
 	// Verify the swap events info.
 	a.assert.Equal(eMaker.swap.OrderId, eTaker.swap.OrderId)
-	a.assert.NotEqual(eMaker.swap.LocalId, eTaker.swap.LocalId)
+	//a.assert.NotEqual(eMaker.swap.LocalId, eTaker.swap.LocalId)
```
michael1011: Why commenting it our instead of removing this line entirely?

rsercano: I can remove it if you say so guys, can't imagine why would we require it @sangaman

sangaman:
@kilrau I know the original idea behind this PR first came up many months ago, but I feel I was confused then as to its motive and am still confused now when reviewing this again - maybe it's just that we've made other changes that make this irrelevent.

If we give our orders local ids, we would want them on the `SwapSuccess` message that we get back from `SubscribeSwaps` and even from `PlaceOrder`. The *only* case where it clearly doesn't make sense to me to have a local id is in `nomatching` mode when calling `ExecuteSwap`. If that's the case, we can just ensure the local id is blank on the `ExecuteSwap` response. Does that make sense? If so I think it makes sense to close this PR and handle that in another one (I'm thinking it would be just a couple of lines).

kilrau:
> Does that make sense? If so I think it makes sense to close this PR and handle that in another one (I'm thinking it would be just a couple of lines).

Probably. I'm closing this now and we'll look at it again once `nomatching` mode is a topic again. This has been dragged way too long.

PR #850, opened 2019-03-21
feat(client): add maximum capacity checks to client

ghost:
This PR builds upon https://github.com/ExchangeUnion/xud/pull/844.

In this PR we're adding a `maximumOutboundCapacity` property to every client and update it every 60 seconds by querying the `channelBalance`. This is required to ultimately add sanity checks before `Orderbook.placeOrder`.

### DIFF
--- a/lib/BaseClient.ts
+++ b/lib/BaseClient.ts
@@ -21,11 +21,15 @@ type ChannelBalance = {
  */
 abstract class BaseClient extends EventEmitter {
   public abstract readonly cltvDelta: number;
+  public maximumOutboundCapacity = 0;
   protected status: ClientStatus = ClientStatus.NotInitialized;
   protected reconnectionTimer?: NodeJS.Timer;
 
   /** Time in milliseconds between attempts to recheck connectivity to the client. */
   protected static readonly RECONNECT_TIMER = 5000;
+  private updateCapacityTimer?: NodeJS.Timer;
+  /** Time in milliseconds between updating the maximum outbound capacity */
+  private CAPACITY_REFRESH_INTERVAL = 60000;
 
   constructor(protected logger: Logger) {
     super();
@@ -38,6 +42,23 @@ abstract class BaseClient extends EventEmitter {
   protected setStatus(status: ClientStatus): void {
     this.logger.info(`${this.constructor.name} status: ${ClientStatus[status]}`);
     this.status = status;
+    this.checkTimers();
+  }
+  private checkTimers() {
+    if (this.status === ClientStatus.ConnectionVerified) {
+      this.updateCapacityTimer = setInterval(async () => {
+        try {
+          this.maximumOutboundCapacity = (await this.channelBalance()).balance;
+        } catch (e) {
+          // TODO: Mark client as disconnected
+          this.logger.error(`failed to fetch channelbalance from client: ${e}`);
+        }
+      }, this.CAPACITY_REFRESH_INTERVAL);
+    } else {
+      if (this.updateCapacityTimer) {
+        clearInterval(this.updateCapacityTimer);
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


lib/BaseClient.ts
```diff
@@ -50,6 +66,9 @@ abstract class BaseClient extends EventEmitter {
     if (this.reconnectionTimer) {
       clearTimeout(this.reconnectionTimer);
     }
+    if (this.updateCapacityTimer) {
+      clearTimeout(this.updateCapacityTimer);
```
sangaman: Same here for `clearInterval`

lib/BaseClient.ts
```diff
@@ -35,6 +39,18 @@ abstract class BaseClient extends EventEmitter {
   protected setStatus(status: ClientStatus): void {
     this.logger.info(`${this.constructor.name} status: ${ClientStatus[status]}`);
     this.status = status;
+    this.checkTimers();
+  }
+  private checkTimers() {
+    if (this.status === ClientStatus.ConnectionVerified) {
+      this.updateCapacityTimer = setInterval(async () => {
+        this.maximumOutboundCapacity = (await this.channelBalance()).balance;
```
sangaman: I think we would want some error handling here, right? In case the call fails, I'm thinking we should catch the error, log it,  and mark the client as disconnected. As is I think an exception could crash the process.

If you want to put the part where it marks the client as disconnected off for another PR, that's fine, because when I'm looking at it we'd need to reenable the reconnect timer as well and right now that's done manually only in LndClient. Ideally there should be a base method that we call to mark the client disconnected and enable the reconnection timer, and Lnd & Raiden can have their own logic for how to check for reconnection. It's not trivial and a little outside the scope of this PR.

sangaman (changes requested):
Nice, made some comments below. The next step would then be to check the available outbound balance before adding orders to the order book, right?

lib/BaseClient.ts
```diff
@@ -18,11 +18,15 @@ type ChannelBalance = {
...
 abstract class BaseClient extends EventEmitter {
+  public maximumOutboundCapacity = 0;
   protected status: ClientStatus = ClientStatus.NotInitialized;
   protected reconnectionTimer?: NodeJS.Timer;
 
   /** Time in milliseconds between attempts to recheck connectivity to the client. */
   protected static readonly RECONNECT_TIMER = 5000;
+  private updateCapacityTimer?: NodeJS.Timer;
+  /** Time in milliseconds to update the maximum outbound capacity cache */
+  private CAPACITY_CACHE_TIMEOUT = 60000;
```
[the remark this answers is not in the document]
erkarl: Renamed to `CAPACITY_REFRESH_INTERVAL`.

lib/BaseClient.ts
```diff
@@ -35,6 +39,18 @@ abstract class BaseClient extends EventEmitter {
...
+    this.checkTimers();
+  }
+  private checkTimers() {
+    if (this.status === ClientStatus.ConnectionVerified) {
+      this.updateCapacityTimer = setInterval(async () => {
+        this.maximumOutboundCapacity = (await this.channelBalance()).balance;
+      }, this.CAPACITY_CACHE_TIMEOUT);
+    } else {
+      if (this.updateCapacityTimer) {
+        clearTimeout(this.updateCapacityTimer);
```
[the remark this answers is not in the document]
erkarl: Nice catch. Interestingly enough it was working as intended 😲

lib/BaseClient.ts
erkarl: It was indeed crashing the process. Added error handling and logging + TODO for marking the client as disconnected.

erkarl:
Rebased, squashed the commits and applied the feedback.

sangaman (approved):

erkarl:
Rebased again - @sangaman

PR #851, opened 2019-03-22
refactor(swaps): remove lnd references from swaps

sangaman:
This commit restructures the Swaps module to remove references to LndClient and instead use the BaseClient which can be either lnd or Raiden. This is in anticipation of interoperability between lnd and Raiden.

### DIFF
--- a/lib/BaseClient.ts
+++ b/lib/BaseClient.ts
@@ -1,5 +1,7 @@
 import Logger from './Logger';
 import { EventEmitter } from 'events';
+import { SwapDeal } from './swaps/types';
+import { Route } from './proto/lndrpc_pb';
 
 enum ClientStatus {
   NotInitialized,
@@ -15,9 +17,10 @@ type ChannelBalance = {
 };
 
 /**
- * A base class to represent a client for an external service such as LND or Raiden.
+ * A base class to represent an external swap client such as lnd or Raiden.
  */
 abstract class BaseClient extends EventEmitter {
+  public abstract readonly cltvDelta: number;
   protected status: ClientStatus = ClientStatus.NotInitialized;
   protected reconnectionTimer?: NodeJS.Timer;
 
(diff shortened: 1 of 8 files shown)


lib/BaseClient.ts
```diff
@@ -36,6 +39,26 @@ abstract class BaseClient extends EventEmitter {
...
+   */
+  public abstract async sendPayment(deal: SwapDeal): Promise<string>;
+
+  /**
+   * Gets routes for the given currency, amount and peerPubKey.
+   * @param amount the capacity of the route
+   * @param destination target node for the route
+   * @returns routes
+   */
+  public abstract async getRoutes(amount: number, destination: string): Promise<Route[]>;
```
erkarl: In the case of raiden there's no `getRoutes` API, but if it were to be implemented it would also require either `contract_address` or `currency` in addition to `amount` and `destination`. For now it's fine, though.

lib/swaps/Swaps.ts
```diff
@@ -172,44 +176,8 @@ class Swaps extends EventEmitter {
...
-        return [];
-      } else {
-        throw err;
-      }
-    }
-  }
-
-  /**
-   * Checks if a swap for two given orders can be executed.
+   * Checks if a swap for two given orders can be executed by ensuring both swap clients are acftive
```
erkarl: %s/acftive/active

lib/swaps/Swaps.ts
```diff
@@ -223,12 +191,23 @@ class Swaps extends EventEmitter {
     // generalize the logic below to work for when we are the maker as well
     const { makerCurrency } = Swaps.deriveCurrencies(maker.pairId, maker.isBuy);
     const { makerAmount } = Swaps.calculateSwapAmounts(taker.quantity, maker.price, maker.isBuy);
-    let routes: lndrpc.Route[];
+
+    const swapClient = this.swapClients[makerCurrency];
+    if (!swapClient) throw new Error('swap client not found');
+
+    const peer = this.pool.getPeer(maker.peerPubKey);
+    const destination = peer.getLndPubKey(makerCurrency);
```
erkarl: What do you think about calling this `peer.getPubKey`?

lib/swaps/Swaps.ts
```diff
@@ -491,50 +476,37 @@ class Swaps extends EventEmitter {
...
     } catch (err) {
-      this.logger.error(`Got exception from sendPaymentSync: ${JSON.stringify(request.toObject())}`, err.message);
       this.failDeal(deal, SwapFailureReason.SendPaymentFailure, err.message);
       await this.sendErrorToPeer(peer, rHash, err.message);
     }
   }
 
   /**
-   * Verifies that the request from LND is valid. Checks the received amount vs
+   * Verifies that the rewsolve request is valid. Checks the received amount vs
```
erkarl: %s/rewsolve/resolve

lib/swaps/Swaps.ts
sangaman: I was thinking that this code will need to change, but raiden does not have a pub key. I was thinking of maybe having something like `getIdentifier()` but I wasn't sure so I put it aside for now.

sangaman:
Fixed the typos, thanks for catching them. As you pointed out there are still a few areas that will need to be refactored, but I think it makes sense to break it into several PRs and leave those for later.

lib/swaps/Swaps.ts
erkarl: `getIdentifier` sounds good, but I suggest to leave it out of this PR.

PR #856, opened 2019-03-27
feat(database): enable WETH trading pair and currency by default

ghost:
In this PR we're modifying the initial database bootstrapping code
to include `WETH` currency by default along with `WETH/BTC` trading
pair.

As a default Rinkeby testnet contract address is used (until we get
support for network specific database/schema).

### DIFF
--- a/lib/db/DB.ts
+++ b/lib/db/DB.ts
@@ -89,15 +89,20 @@ class DB {
       promises.push(Currency.bulkCreate(<db.CurrencyAttributes[]>[
         { id: 'BTC', swapClient: SwapClients.Lnd, decimalPlaces: 8 },
         { id: 'LTC', swapClient: SwapClients.Lnd, decimalPlaces: 8 },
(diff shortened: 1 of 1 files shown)


sangaman (approved):

sangaman:
Super minor note, in the past I've written the commit message as `feat(db)` instead of writing out database makes the commit message a bit shorter. Change if you so please.

erkarl:
Updated the commit message.

erkarl:
Going to close this because after discussing it with @offerm and @sangaman we felt like `xud-simnet` script would be better place to add this. The contract address for simnet is subject to change and we don't want to have a hardcoded value in the xud internals.

kilrau:
Makes sense, but maybe prepare the testnet and mainnet defaults? (I guess it makes sense to wait for https://github.com/ExchangeUnion/xud/pull/827)

If lnd is enabled = LTC/BTC
If lnd+raiden is enabled = LTC/BTC, WETH/BTC, LTC/WETH

erkarl:
Created: https://github.com/ExchangeUnion/xud/issues/872

PR #858, opened 2019-03-27
feat(raiden): detect raiden address change

sangaman:
This detects if our raiden address changes between calls to `getAddress`. If so it updates peers of our new raiden address.

### DIFF
--- a/lib/raidenclient/RaidenClient.ts
+++ b/lib/raidenclient/RaidenClient.ts
@@ -2,7 +2,6 @@ import http from 'http';
 import Logger from '../Logger';
 import BaseClient, { ClientStatus, ChannelBalance } from '../BaseClient';
 import errors from './errors';
-import { Order } from '../orderbook/types';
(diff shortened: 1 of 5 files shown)


lib/raidenclient/RaidenClient.ts
```diff
@@ -70,8 +70,8 @@ type ChannelEvent = {
 };
 
 interface RaidenClient {
-  on(event: 'swap', listener: (order: Order) => void): this;
-  emit(event: 'swap', order: Order): boolean;
+  on(event: 'connectionVerified', listener: (newAddress?: string) => void): this;
```
erkarl: I'm wondering if these should also be moved to `BaseClient`?

sangaman: Yeah I was on the fence but in thinking about it some more it makes sense, I can call the parameter `newIdentifier` to be generic. I'll make that change.

PR #861, opened 2019-03-29
fix(raiden): add getRoutes and getHeight compatibility

ghost:
In this commit we're modifying `RaidenClient.getRoutes` and `getHeight` to be compatibile with the current `Swaps` implementation. Since raiden's API does not support `getRoutes` nor `getHeight` we're returning the minimum viable mock implementation to continue with the swap.

### DIFF
--- a/lib/raidenclient/RaidenClient.ts
+++ b/lib/raidenclient/RaidenClient.ts
@@ -136,11 +136,14 @@ class RaidenClient extends BaseClient {
   }
 
   public getRoutes =  async (_amount: number, _destination: string) => {
-    return []; // stub placeholder, query routes not currently implemented in raiden
+    // stub placeholder, query routes not currently implemented in raiden
+    return [{
+      getTotalTimeLock: () => 1,
+    }];
   }
 
   public getHeight = async () => {
-    return 0; // raiden's API does not tell us the height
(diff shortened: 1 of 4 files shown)


kilrau:
Could you make this a topic in our channel with the raiden team? Basically: are api calls for these planned?

lib/raidenclient/RaidenClient.ts
```diff
@@ -135,11 +135,16 @@ class RaidenClient extends BaseClient {
   }
 
   public getRoutes =  async (_amount: number, _destination: string) => {
-    return []; // stub placeholder, query routes not currently implemented in raiden
+    // stub placeholder, query routes not currently implemented in raiden
+    return Promise.resolve([
```
sangaman: Since it's `async` you don't need to wrap the array in `Promise.resolve()`, and we're just returning plain synchronous values from other `async` methods in the codebase. It doesn't make any real difference but might as well make it consistent.

sangaman (approved):
Thanks, this is an elegant solution to the unimplemented raiden methods.

lib/raidenclient/RaidenClient.ts
erkarl: Changed. Old habits die hard :-).

PR #862, opened 2019-03-30
feat(p2p): verify response packet type

### LINKED ISSUE #788: Verify response packet type
Currently we're not verifying the response packet type for each request - only that its `reqId` is matching the request packet `Id`. This can lead to runtime exceptions and node crash.

moshababo:
Closes #788.


lib/p2p/packets/types/SwapRequestPacket.ts
```diff
@@ -11,14 +11,18 @@ export type SwapRequestPacketBody = {
 };
 
 class SwapRequestPacket extends Packet<SwapRequestPacketBody> {
-  public get type() {
+  public get type(): PacketType {
```
sangaman: Fwiw the restriction to return a `PacketType` already exists because it's declared on the base `Packet` class, so it's not necessary to explicitly state it here for the compiler to enforce it. But if you think it's more readable this way it's fine to leave these changes in place that's fine.

moshababo:
@sangaman It's due to `SwapRequestPacket`, which can either get an acceptance or a failure packet as the response. So it's just about referring to the type of the response.

I could have done everything as an array (which can have 0, 1 or 2 items), but I preferred to make it more explicit.

lib/p2p/packets/types/SwapRequestPacket.ts
moshababo: It was necessary to be declared for `responseType`, so I preferred the style to be unified.

Issue #675, opened 2018-11-15
Sanity checks before adding to orderbook


erkarl:
Started as a result of #240 discussion.

> To the extent that this is a valuable safeguard, I think it makes sense to employ it on a broader scope and not just via the cli. Maybe xud should make these sanity checks internally - it's not in our interest to advertise orders that can't be swapped. For example, we shouldn't broadcast orders if we don't have channels open with some minimum capacity. We already do some checks on initialization like verifying lnd is in sync with the chain. I believe checking our incoming capacity would be a bit more complicated - calling listchannels and then iterating over each one - but still doable. Getting this done right wouldn't be simple but it seems worthwhile, and once that's in place we could return a grpc error for the placeorder calls indicating that the order can't be swapped without needing to make an extra call like channelbalance.

### High level specs
* before `placeOrder` add `orderSanityChecks`

`orderSanityChecks` should verify that:
* inbound capacity to execute the order exists
* outbound capacity to execute the order exists

if either of the above checks fails - ERROR!

sangaman:
We could also do this on *peer* orders potentially, so that we check if capacity exists before we add it to our orderbook. In both cases, I think it makes sense to save the capacity in memory (cached in other words) , and only recalculate it after it gets old, maybe 10 minutes? I'm thinking that doing this every time we place an order will introduce too much of a delay - we really want the order placement and settlement to be as fast as possible.

In fact, maybe we shouldn't tie these checks into the `placeOrder` flow per se, but rather calculate our capacities on a scheduled basis. Every 5-10 minutes we update the balances in memory, and on new orders we just compare the amount to the balance we've stored in memory. That should add almost no delay to the routine.

Also one extra detail, the [lightning spec](https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md) recommends that 1% of the balance of each channel be held in reserve on each side. So we probably want to make sure that our channel balance is at least 101% of the  order amount.

kilrau:
Cache channel balances in memory and then check this value on `placeOrder` sounds good! Since this is an independent job (not affecting the `placeOrder` performance) I think we can target 1 min updates (unless there is a reason not to @sangaman @erkarl ), looking a bit into the future where channel balances change very fast when trading volume goes in one direction.

Since we want to support AMPs in future, let's prepare aggregating balances of multiple channels. Also wouldn't distinguish between peers for now, meaning let the check succeed as long as one channel to one peer with sufficient capacity exists.
[...]

PR #863, opened 2019-03-31
feat(orderbook): add sanity checks to placeOrder

ghost:
In this PR we're going to use the `maximumOutboundCapacity` from BaseClient to make sure that when placing the order sufficient balance exists. https://github.com/ExchangeUnion/xud/issues/675

There's 1 test case currently failing within [Service.spec](https://github.com/ExchangeUnion/xud/blob/master/test/integration/Service.spec.ts#L73) because the code now assumes reachable and funded swap client, but the test setup does not enable them. I'm open to suggestions on how to move forward with modifications within Service.spec. Should we enable lnds within test setup config and assume channels exist and are funded?

### DIFF
--- a/lib/Config.ts
+++ b/lib/Config.ts
@@ -27,6 +27,11 @@ class Config {
   public dbpath: string;
   /** Whether matching will be disabled */
   public nomatching: boolean;
+  /** Whether to disable sanity checks that verify that the orders can possibly
+   * be swapped before adding them to the order book, can be enabled for
+   * testing & debugging purposes.
+   */
+  public nosanitychecks: boolean;
 
   constructor() {
     const platform = os.platform();
@@ -60,6 +65,7 @@ class Config {
     this.logdateformat = 'DD/MM/YYYY HH:mm:ss.SSS';
     this.network = this.getDefaultNetwork();
     this.dbpath = this.getDefaultDbPath();
+    this.nosanitychecks = false;
 
     this.p2p = {
       listen: true,
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -68,7 +68,8 @@ class OrderBook extends EventEmitter {
     return Array.from(this.pairs.keys());
   }
 
-  constructor(private logger: Logger, models: Models, public nomatching = false, private pool?: Pool, private swaps?: Swaps) {
+  constructor(private logger: Logger, models: Models, public nomatching = false,
+    private pool?: Pool, private swaps?: Swaps, private nosanitychecks = false) {
     super();
 
     this.repository = new OrderBookRepository(logger, models);
@@ -299,6 +300,19 @@ class OrderBook extends EventEmitter {
         swapFailures: [],
         remainingOrder: order,
       };
+    }
+
+    if (!this.nosanitychecks) {
+      // check if sufficient outbound channel capacity exists
+      const { makerAmount } = Swaps.calculateSwapAmounts(order.quantity, order.price, order.isBuy, order.pairId);
+      const { makerCurrency } = Swaps.deriveCurrencies(order.pairId, order.isBuy);
+      const swapClient = this.swaps && this.swaps.swapClients.get(makerCurrency);
+      if (!swapClient) {
+        throw errors.SWAP_CLIENT_NOT_FOUND(makerCurrency);
+      }
+      if (makerAmount > swapClient.maximumOutboundCapacity) {
+        throw errors.INSUFFICIENT_OUTBOUND_BALANCE(makerCurrency, makerAmount);
+      }
     }
 
     // perform matching routine. maker orders that are matched will be removed from the order book.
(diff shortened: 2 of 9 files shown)


kilrau:
Missing `maximumInboundCapacity`?

lib/orderbook/OrderBook.ts
```diff
@@ -298,6 +298,17 @@ class OrderBook extends EventEmitter {
       };
     }
 
+    // check if sufficient outbound channel capacity exists
+    const { makerAmount } = Swaps.calculateSwapAmounts(order.quantity, order.price, order.isBuy);
+    const { makerCurrency } = Swaps.deriveCurrencies(order.pairId, order.isBuy);
+    const swapClient = this.swaps && this.swaps.swapClients[makerCurrency];
```
sangaman: For getting the tests to pass, how about wrapping this block of code in an `if (this.swaps)` check and dropping it from this line?

sangaman (commented):
Is the plan to put in similar checks before we add peer orders to the order book in a separate PR?

erkarl:
> Missing maximumInboundCapacity?

Do we need to include the check for inbound capacity if the other peer is already also checking its outbound (our inbound) capacity?

sangaman:
> Does it make sense include the check for inbound capacity if the other peer is already also checking its outbound (our inbound) capacity?

I'd be ok with leaving it up to the peer to check their side of the available capacity, particularly because LND doesn't give us a simple way of fetching the inbound capactiy (we'd have to list all channels and calculate it ourselves).

kilrau:
> Does it make sense include the check for inbound capacity if the other peer is already also checking its outbound (our inbound) capacity?

Generally speaking: `don't trust, verify` , but

> LND doesn't give us a simple way of fetching the inbound capactiy (we'd have to list all channels and calculate it ourselves).

doing that regularly is a bit over the top, agree.

kilrau:
> Is the plan to put in similar checks before we add peer orders to the order book in a separate PR?

Well here's the thing: `ChannelBalance` only matters in the moment of a match, when an actual swap is triggered. And a match only occurs upon `OwnOrder` entering the order book, never when a `PeerOrder` enters the order book. So letting an "out of channel bounds" `PeerOrder` enter our local order book is totally fine, checks happen when `OwnOrder` comes in and if "within channel bounds" it can still fill this over-sized `PeerOrder` partially.

Unless I'm missing something, I believe it's the right thing to NOT do this check on `PeerOrders`.

lib/orderbook/OrderBook.ts
erkarl: This won't work because the service spec is firing up a real instance of xud (with config defined in the test setup). It will have `this.swaps`.

erkarl: I've updated this PR/commit to make sanity checks configurable from xud.conf (`sanitychecks = true`). This allows us to disable those checks in the service spec.

Alternative option is to remove the offending tests from service spec and add them to xud-simulation... 🤔

kilrau:
[Above](https://github.com/ExchangeUnion/xud/pull/863#issuecomment-479892229) I said channel capacity checks upon order entering shouldn't be done on incoming peer orders, just on `OwnOrder`s since incoming matching `OwnOrder`s directly trigger a swap and only my `OwnOrder` size matters for the channel capacity - I can fill oversized `PeerOrder`s partially. Does that make sense? @erkarl 

Also as @erkarl pointed out before, the `getRoutes()` check on maker side when accepting a deal covers the capacity check back to the taker. So I think we don't have to do another `getRoutes()` check from taker to maker. The sanity check cover it. 

Bottom line I think we can safely replace [this in-code comment](https://github.com/ExchangeUnion/xud/blob/master/lib/swaps/Swaps.ts#L387) in this PR with an explainer that the sanity checks already cover the local amount of the channel.

erkarl:
@kilrau 
In general I agree with the logic above, but having outbound balance does not always mean that routes exist (assuming non-direct channels). Therefore, I'd leave the TODO in place as a future improvement to be made.

@sangaman - rebased and ready for review again.

sangaman:
I think that TODO comment can be removed with #822, as we'll be verifying that a route exists before accepting *any* orders from a peer as part of the sanity swap.

lib/Config.ts
```diff
@@ -27,6 +27,8 @@ class Config {
   public dbpath: string;
   /** Whether matching will be disabled */
   public nomatching: boolean;
+  /** Whether to perform sanity checks */
+  public sanitychecks: boolean;
```
sangaman: "sanity checks" is a little ambiguous but maybe I'll use this same option to disable the sanity swaps in which case it's a good thing it's not too specific. Maybe just add a some detail to the comment like "Whether to check that orders can possibly be swapped before adding them to the order book, can be disabled for testing & debugging purposes."

Another consideration is that if we ever pass a value for this option from the command line (which I think could happen if I reuse it for disabling #822) then we might want to call it `nosanitychecks` so that it can be set with `--nosanitychecks` from the command line.

sangaman (commented):
Looks good but can you look into why the simulation tests are failing? It's not immediately obvious to me, but it's failing due to insufficient balance.

lib/Config.ts
erkarl: Good idea. I've renamed the config option to `nosanitychecks` and updated the comment.

erkarl:
@sangaman - build is passing now.

sangaman (approved):
Awesome, thank you.

PR #867, opened 2019-04-03
style: clean up imports

sangaman:
Remove unused imports and replace absolute module paths with relative.


sangaman:
Yeah that's probably a good idea, we have a couple of unused variables still but at this point I think we can just comment them out or get rid of them.

PR #852, opened 2019-03-23
test: xud-simulation migration 

moshababo:
Closes #765.

This PR is migrating `xud-simulation` under `xud/test/simulation`.

I changed `.travis` to have `language: go` instead of `language: node_js`. Otherwise `go1.7.4` is being shipped, which isn't sufficient for Go Modules.

The new `.travis` file is a bit messy now, combining 2 environments. Feel free to suggest improvements.

Also, please execute it on your local machine by running `npm run test:sim` (`go1.11.x` installation is required).

PR #869, opened 2019-04-03
fix: use integer satoshi quantities

sangaman:
This commit changes all `quantity` values to be represented by integer satoshi amounts (10 ^ -8 of a full unit) rather than fractional/floating full units. This applies both to the internal representations as well as
the gRPC API definition. This is done to prevent any rounding errors when adding or subtracting fractional units.

Output and arguments for the cli remain unchanged and still deal with full units.

Fixes#740.

BREAKING CHANGE: Database and p2p field type changes


kilrau:
> LGTM - `SATOSHIS_PER_COIN` has a weird ring to it, but I don't have a better name to suggest.

Agree, especially if we use it as generic smallest unit for all coins & tokens. Seems we are not the only ones having troubles to find a sexy name for "[smallest unit of a currency](https://money.stackexchange.com/questions/85562/generic-name-for-the-smallest-unit-of-)" :/

sangaman:
Yeah, I'm trying to use "satoshi" in that instance to indicate 10^-8. I don't actually want the smallest unit in that context, since we're trying to cap precision at 8 decimal places anyhow.

I'm rebasing this PR onto #852 now that it's merged.

sangaman:
@moshababo I needed to regenerate the go protobuf file for this PR, and I made a few changes to the `gen_protos.sh` script in a separate commit (explained briefly in the commit message). If you have a minute could you check to make sure it makes sense to you?

PR #874, opened 2019-04-08
feat(raiden): add sendPayment support

### LINKED ISSUE #797: [Overview] Prepare XUD for ERC20<>Lightning swap
In parallel to extending Raiden with swap support we can enhance XUD to make sure we are ready.

Need to take care for the following:
- Make gRPC work for Raiden (https://github.com/ExchangeUnion/xud/issues/792)
- Staring Raiden - need to have account with tokens and ETH,  keystore file , EndpointRegistry address, SecretRegistry address, TokenNetworkRegistry address, and eth-rpc-endpoint. I can share a startup script when needed.   
- Create a channel with a peer `PUT /api/(version)/channels`
- implement XUD to Raiden connectivity (like we do with LND). Make sure we have a connection and we are ready to trade. Need configuration support. 
- Create an XUD currency for the raiden Token. Should include the Token address
- Get to the point where Swap agreed and the taker initiates the  swap by using `POST /api/(version)/payments/(token_address)/(target_address)` 

Suggest to gave a separate issue for each step once we start the implementation.

ghost:
In this commit we're adding support for `RaidenClient.sendPayment` that
enables us to do a **trusted** swap.

closes https://github.com/ExchangeUnion/xud/issues/797

### DIFF
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -4,6 +4,7 @@ import Logger from './Logger';
 import Config from './Config';
 import DB from './db/DB';
 import OrderBook from './orderbook/OrderBook';
+import BaseClient from './BaseClient';
 import LndClient from './lndclient/LndClient';
 import RaidenClient from './raidenclient/RaidenClient';
 import GrpcServer from './grpc/GrpcServer';
@@ -74,25 +75,30 @@ class Xud extends EventEmitter {
       this.db = new DB(loggers.db, this.config.dbpath);
       await this.db.init(this.config.network, this.config.initdb);
 
+      const swapClients = new Map<string, BaseClient>();
       // setup LND clients and initialize
       for (const currency in this.config.lnd) {
         const lndConfig = this.config.lnd[currency]!;
         if (!lndConfig.disable) {
           const lndClient = new LndClient(lndConfig, currency, loggers.lnd);
           this.lndClients[currency] = lndClient;
+          swapClients.set(currency, lndClient);
           initPromises.push(lndClient.init());
         }
       }
 
       // setup raiden client and connect if configured
-      this.raidenClient = new RaidenClient(this.config.raiden, loggers.raiden);
+      this.raidenClient = new RaidenClient(this.config.raiden, loggers.raiden, this.db.models);
       if (!this.raidenClient.isDisabled()) {
-        initPromises.push(this.raidenClient.init());
+        await this.raidenClient.init();
+        for (const currency of this.raidenClient.tokenAddresses.keys()) {
+          swapClients.set(currency, this.raidenClient);
+        }
       }
 
       this.pool = new Pool(this.config.p2p, this.config.network, loggers.p2p, this.db.models);
 
-      this.swaps = new Swaps(loggers.swaps, this.db.models, this.pool, this.lndClients);
+      this.swaps = new Swaps(loggers.swaps, this.db.models, this.pool, swapClients);
       initPromises.push(this.swaps.init());
 
       this.orderBook = new OrderBook(loggers.orderbook, this.db.models, this.config.nomatching, this.pool, this.swaps);
--- a/lib/raidenclient/RaidenClient.ts
+++ b/lib/raidenclient/RaidenClient.ts
@@ -3,7 +3,10 @@ import Logger from '../Logger';
 import BaseClient, { ClientStatus, ChannelBalance } from '../BaseClient';
 import errors from './errors';
 import { SwapDeal } from '../swaps/types';
-import { SwapClient } from '../constants/enums';
+import { SwapClient, SwapState, SwapRole } from '../constants/enums';
+import assert from 'assert';
+import OrderBookRepository from '../orderbook/OrderBookRepository';
+import { Models } from '../db/DB';
 
 /**
  * A utility function to parse the payload from an http response.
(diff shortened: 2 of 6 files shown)


lib/Xud.ts
```diff
@@ -74,25 +75,30 @@ class Xud extends EventEmitter {
       this.db = new DB(loggers.db, this.config.dbpath);
       await this.db.init(this.config.initdb);
 
+      const swapClients: { [currency: string]: BaseClient | undefined } = {};
```
sangaman: I'm thinking it actually might have made sense to do this earlier, but I'm thinking it might be cleaner to make this a `Map<string, BaseClient>` type. We're mapping `currency` to other values elsewhere in the codebase. But this change could also happen in a separate PR.

lib/raidenclient/RaidenClient.ts
```diff
@@ -75,19 +78,23 @@ class RaidenClient extends BaseClient {
   public readonly type = SwapClient.Raiden;
   public readonly cltvDelta: number = 0;
   public address?: string;
+  public currencies: { [key: string]: string } = {};
```
sangaman: This can be a `Map` type as well, and maybe it would make more sense to call this `tokenAddresses` since that's what the values are?

lib/raidenclient/RaidenClient.ts
```diff
@@ -101,11 +108,27 @@ class RaidenClient extends BaseClient {
...
+    try {
+      const currencies = await this.repository.getCurrencies();
+      const raidenCurrencies = currencies.filter((currency) => {
+        const tokenAddress = currency.getDataValue('tokenAddress');
+        if (tokenAddress) {
+          this.currencies[currency.getDataValue('id')] = tokenAddress;
+        }
+      });
+    } catch (e) {
+      this.logger.error(`failed to set currencies for Raiden: ${e}`);
```
sangaman: If you make this `this.logger.error('failed to set currencies for Raiden', e);` the `error` method will log the exception message and stack trace if applicable.

lib/raidenclient/RaidenClient.ts
```diff
@@ -134,7 +157,24 @@ class RaidenClient extends BaseClient {
...
 
-  public sendPayment = async (_deal: SwapDeal): Promise<string> => {
+  public sendPayment = async (deal: SwapDeal): Promise<string> => {
+    assert(deal.state === SwapState.Active);
+    assert(deal.destination);
+    let amount = 0;
+    let tokenAddress = '';
+    if (deal.role === SwapRole.Maker) {
+      // we are the maker paying the taker
+      amount = deal.makerAmount;
```
sangaman: I think this should be `takerAmount` as that's the amount the taker is expecting to receive.

lib/raidenclient/RaidenClient.ts
```diff
@@ -134,7 +157,24 @@ class RaidenClient extends BaseClient {
...
+      tokenAddress = this.currencies[deal.takerCurrency];
+    } else {
+      // we are the taker paying the maker
+      amount = deal.takerAmount;
```
sangaman: Should be `makerAmount`.

lib/raidenclient/RaidenClient.ts
```diff
@@ -134,7 +157,24 @@ class RaidenClient extends BaseClient {
...
+      tokenAddress = this.currencies[deal.makerCurrency];
+    }
+    // TODO: Secret hash. Depending on sha256 <-> keccak256 problem:
+    // https://github.com/ExchangeUnion/xud/issues/870
+    const tokenPaymentResponse = await this.tokenPayment(tokenAddress, deal.destination!, amount);
+    console.log('tokenPaymentResponse is', tokenPaymentResponse);
```
sangaman: Replace this with a `debug` statement? And `JSON.stringify` the response.

sangaman (commented):
Thanks, made some comments below.

lib/raidenclient/RaidenClient.ts
```diff
@@ -68,26 +71,34 @@ type ChannelEvent = {
...
+  secret: string;
+};
+
 /**
  * A class representing a client to interact with raiden.
  */
 class RaidenClient extends BaseClient {
   public readonly type = SwapClient.Raiden;
-  public readonly cltvDelta: number = 0;
+  public readonly cltvDelta: number = 1;
```
erkarl: The reason for this change is that the `SwapRequestPacket` will not validate otherwise:
https://github.com/ExchangeUnion/xud/blob/master/lib/p2p/packets/types/SwapRequestPacket.ts#L37

lib/swaps/Swaps.ts
```diff
@@ -426,7 +433,7 @@ class Swaps extends EventEmitter {
     }
 
     const responseBody: packets.SwapAcceptedPacketBody = {
-      makerCltvDelta: deal.makerCltvDelta || 0,
+      makerCltvDelta: deal.makerCltvDelta || 1,
```
erkarl: The reason for this change is that the `SwapAcceptedPacket` will not validate otherwise:
https://github.com/ExchangeUnion/xud/blob/master/lib/p2p/packets/types/SwapAcceptedPacket.ts#L36

ghost (commented):
@sangaman feedback applied. I've got the PR to the closest as I can to a real swap, but without being able to provide a custom sha256 hash for Raiden payments the swaps will not succeed. 

I chose not to apply temporary code to support a "trusted" swap as there were too many places that required changes. However, it's still possible to test Raiden payments via xud:

1. Setup Raiden for xud A & B
2. Fund both Raiden accounts with `ETH` & `WETH`
3. Setup direct channels between A & B
4. (optional) test payment without xud
```
curl -i -X POST http://localhost:5001/api/v1/payments/0xc778417E063141139Fce010982780140Aa0cD5Ab/<receiverAddress> -H 'Content-Type: application/json' --data-raw '{"amount": 1234, "identifier": <randomInteger>}'
```
5. Add WETH currency to A & B - `./bin/xucli addcurrency WETH Raiden 18 0xc778417E063141139Fce010982780140Aa0cD5Ab`
6. Add WETH pair to A & B - `./bin/xucli addpair WETH BTC`
7. Execute buy WETH order on B
8. Execute sell WETH order on A

sangaman (approved):
Awesome, thank you. Looks like about as close as we can get it without the hashes matching up.

PR #876, opened 2019-04-10
fix(p2p): NodeStateUpdatePacket process crash

ghost:
This commit fixes a bug where the process crashed while trying to send the `NodeStateUpdatePacket`.

Steps to reproduce the bug in current master:
1. Start xud with lnd/raiden closed
2. Connect to a peer so that we start sending out NodeStateUpdatePackets
3. Start lnd/raiden
4. Error! `NodeStateUpdatePacket.serialize` crashes.

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -165,12 +165,12 @@ class Pool extends EventEmitter {
    */
   public updateNodeState = (nodeStateUpdate: NodeStateUpdate) => {
     this.nodeState = { ...this.nodeState, ...nodeStateUpdate };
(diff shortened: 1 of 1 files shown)


sangaman (approved):
This definitely looks like it would fix the crash. But it does so by having us send the entire node state on every update. I'm wondering if it might be better to fix this longer term by changing the `serialize` method in NodeStateUpdatePacket.ts. It uses the `!` operator on `pairs`, `raidenAddress`, and `addresses`. And in looking at the `handleNodeStateUpdate` method that might need some rework too.

But I think we should go ahead and merge this for now since it prevents a crash, and can maybe revisit this later.

PR #880, opened 2019-04-11
fix(orderbook): rename SwapClients to SwapClient

ghost:
This commit fixes an accidental bug related to SwapClient enum import where during the rebase Orderbook.ts import was not renamed.


sangaman (approved):
Thanks for catching.

PR #883, opened 2019-04-12
feat(p2p): blacklist and whitelist

### LINKED ISSUE #179: Strict/Permissive p2p modes
The default behavior is that a node will create/accept connection to any peer which is not on the black list. We previously discussed about having a strict mode, which unlike the permissive one, will block all peers by default, except these who are explicitly mentioned in a white list.

### LINKED ISSUE #458: Ban by IP address
Offshoot of #185. We want to be able to ban by IP address when a certain IP address spams us with invalid messages or repeatedly fails to complete the handshake. Since IP addresses change, we need not persist this ban permanently. Rather, we can ban for the duration of an xud session or perhaps set a timer to lift the ban after several days.

reliveyy:
This commit add a feature to block incoming connections according to the blacklist or whitelist IP addresses.

The following topics needs to be discussed

- [ ] Persistence for such blacklist or whitelist
- [ ] Add IP addresses into blacklist or whitelist dynamically through RPC calls
- [ ] Enhance `xucli ban <pubkey>` with `xucli ban <ip>` and persist that IP address in blacklist.

Resolve #458
Resolve #179

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -74,6 +74,10 @@ class Pool extends EventEmitter {
   private listenPort?: number;
   /** This node's listening external socket addresses to advertise to peers. */
   private addresses: Address[] = [];
+  /** This node's P2P blacklist IP addresses */
+  private blacklist: string[] = [];
+  /** This node's P2P whitelist IP addresses */
+  private whitelist: string[] = [];
   /** Points to config comes during construction. */
   private config: PoolConfig;
   private repository: P2PRepository;
@@ -92,6 +96,12 @@ class Pool extends EventEmitter {
       config.addresses.forEach((addressString) => {
         const address = addressUtils.fromString(addressString, config.port);
         this.addresses.push(address);
+      });
+      config.blacklist.forEach((addressString) => {
+        this.blacklist.push(addressString);
+      });
+      config.whitelist.forEach((addressString) => {
+        this.whitelist.push(addressString);
       });
     }
   }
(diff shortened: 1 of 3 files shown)


reliveyy:
@kilrau This PR dose not extend original `xucli ban` command. And the logic in this PR is blacklist first then whitelist. If whitelist is empty then bypass it.

lib/p2p/types.ts
```diff
@@ -49,4 +49,15 @@ export type PoolConfig = {
    * It will be advertised with peers for them to try to connect to the server in the future.
    */
   addresses: string[];
+
+  /**
+   * An array of IP addresses or host names which can be used as a whitelist to limit connecting peers.
+   * Is will bypass all peers when it is empty.
```
sangaman: "It" instead of "Is". Also I think it's a bit more clear to say "It will allow all IP addresses when it is empty."

lib/p2p/types.ts
```diff
@@ -49,4 +49,15 @@ export type PoolConfig = {
...
+   */
+  whitelist: string[];
+
+  /**
+   * An array of IP addresses or host names which can be used as a blacklist to limit connecting peers.
+   */
+  blacklist: string[];
```
sangaman: Conceptually, I wonder whether we need a configurable blacklist of IP addresses within xud.  For one thing, I believe it would be much more efficient to block IPs before they even get to this point, banning them within the OS rather than within our app. When I wrote #458, I was thinking of banning IPs dynamically while xud is running if we get a lot of bad requests from them.

But in general, I don't any sort of fixed list will be all that effective since attackers can easily change IP addresses. My thinking is we should drop this config option altogether. Thoughts @kilrau ?

lib/p2p/Pool.ts
```diff
@@ -93,6 +97,12 @@ class Pool extends EventEmitter {
         const address = addressUtils.fromString(addressString, config.port);
         this.addresses.push(address);
       });
+      config.blacklist.forEach((addressString) => {
+        this.blacklist.push(addressString);
+      });
+      config.whitelist.forEach((addressString) => {
+        this.whitelist.push(addressString);
```
sangaman: No need to loop through and `push` every element into a new array, you can just do `this.whitelist = config.whitelist;`.

sangaman:
I agree that we should not simultaneously have a blacklist and whitelist active, but I think actually it doesn't make sense to have a blacklist period. See my comment above.

kilrau:
>  I think actually it doesn't make sense to have a blacklist

Agree - I believe we can drop the blacklist altogether. I don't see a great use case for a manual ip/domain based blacklist. Please make this PR about a white-list only.

> #458, I was thinking of banning IPs dynamically while xud is running if we get a lot of bad requests from them.

Agree too. This should be handled automatically in the background at run-time, please take this out of this PR.

kilrau:
As discussed with @reliveyy , there will be a major revision of the PR after agreeing on [these basics](https://github.com/ExchangeUnion/xud/issues/179#issuecomment-484058527).

Issue #684, opened 2018-11-18
What exactly to do when lnd or raiden is disconnected


kilrau:
Prerequisite: https://github.com/ExchangeUnion/xud/issues/573

This issue is about answering the question: What should we do if we `xud` identifies via #573  that it is not properly connected to LND?

Current opinions:
a) disconnect from all peers
Good: easy
Bad: doesn't differentiate between e.g. LND and Raiden
b) Throw an error through gRPC and revoke all trading pairs with swap protocol LND. Also disable all peerOrders with swap protocol LND. Same with Raiden.
Good: considers different swap clients (LND, Raiden)
Bad: Complicated

sangaman:
We shouldn't disconnect from peers, as losing connectivity to lnd only means that one currency is impacted, so I think the better route is to temporarily disable all affected trading pairs. That would update the handshake with our peers to let them know we aren't supporting that pair (which in turn would make them drop all our orders for that pair) and would fail any rpc calls to place an order (or execute a swap) for that trading pair. We would also want to drop orders for that trading pair locally.

kilrau:
* don't block `xud` start when lnd or raiden fails to connect
* how to denounce a pair / protocol
* how to reannounce a pari / protocol

sangaman:
I verified that we actually do connect to peers on startup even without a connection to lnd, however we are advertising the trading pair that we don't have an lnd connection for and that needs to change.

sangaman:
> how to denounce a pair / protocol
> how to reannounce a pari / protocol

We already have the ability to add and remove trading pairs (see `addPair` and `removePair` in OrderBook.ts`) so we just need to call these methods, this part should be simple actually.

kilrau:
Great, anything that still needs to be defined?

kilrau:
Good to go? @sangaman

kilrau:
Dependency: https://github.com/ExchangeUnion/xud/pull/762

kilrau:
Depends on https://github.com/ExchangeUnion/xud/issues/884#issuecomment-482553162

sangaman:
We have logic that more or less covers b) I think. What did you have in mind for a gRPC error?

Right now when we encounter any lnd error (including an unexpected end to our invoice subscription) we will mark that lnd as disconnected and reject any swaps involving that currency as well as stop ourselves from requesting any swaps involving that currency. That logic hasn't been tested very much I don't think.

[...]

PR #888, opened 2019-04-15
docs: simulation tests add swap client disabled test

kilrau:


test/simulation/README.md
```diff
@@ -38,7 +38,7 @@ $ GO111MODULE=on go test -v
 - [ ] Peer disconnection should trigger orders removal to all his orders.
 
 ### Node State
-- [ ] Removed pair should trigger order removal for all the pair's orders, and removed from connected peers' order books.
+- [ ] Removed pair or swap client (lnd/raiden) unavailable should trigger order removal for all the pair's orders, and removed from connected peers' order books.
```
erkarl: Do we want to trigger order removals immediately when swap client becomes unavailable? @sangaman suggested to have a short timeout period before we do remove orders: https://github.com/ExchangeUnion/xud/issues/684#issuecomment-483110342

kilrau: No, we don't i agree on the timeout period, see https://github.com/ExchangeUnion/xud/issues/684#issuecomment-483209152. Clarified that.

test/simulation/README.md
```diff
@@ -38,7 +38,7 @@ $ GO111MODULE=on go test -v
 - [ ] Peer disconnection should trigger orders removal to all his orders.
 
 ### Node State
-- [ ] Removed pair should trigger order removal for all the pair's orders, and removed from connected peers' order books.
+- [ ] Removed pair or swap client (lnd/raiden) disconnected (after timeout elapsed) should trigger order removal for all the pair's orders, and removed from connected peers' order books.
```
sangaman: Per discussion in #684, if we do use a timeout of ~30s before deactivating the pair that would be kind of a long delay to build into automated tests - plus we'd have to be careful it doesn't exceed the time limits for each test. What do you think @moshababo, do you know of a good way to handle a delay within a test?

moshababo: We should have the timeout length configurable, and so for the test node instance we'll do it as 1-2 sec, and it should be fine.

erkarl: I like the idea - `debug.swapClientDeactivationDelay`?

kilrau: Sounds good!

sangaman: Hm maybe make it an environment variable? If it's not something that should actually be configurable, and we'd avoid having a "debug" section in config.

erkarl: I'm good with both options. The benefit of having a debug section in the config would be automatic "documentation" when running `npm run config`. On the other hand it could be confusing for regular users.

sangaman: Right, I'd be thinking that it would be confusing for regular users. I'm thinking of this as something that we wouldn't want to have users fiddling with, only something that we want to be able to change for simulation test purposes.

kilrau:
Included in https://github.com/ExchangeUnion/xud/pull/905

Issue #40, opened 2018-04-24
Add wallet rpc methods and integrations


sangaman:
Allow for creation of new deposit addresses and for withdrawals through xud.

sangaman:
Partly solved by #94, still need to add appropriate rpc calls.

kilrau:
I think we are ready to do this now. Volunteers? @ImmanuelSegol @erkarl

sangaman:
We have the code to create a new deposit address in lnd, still need a wrapper method to send coins in lnd but that is easy. Then we'd need wrapper rpc methods for these which also shouldn't be hard. I think this should be low priority, but maybe we should discuss whether we'd want `xud` to be the sole point of interaction that manages all the `lnd` instances entirely, or whether it's ok to let some calls be made directly on `lnd`.

ImmanuelSegol:
> I think we are ready to do this now. Volunteers? @ImmanuelSegol @erkarl

I would be happy to take this @kilrau

kilrau:
I'd be pro `xud` being the sole point of interaction to reach a certain abstraction.
* only expose calls that matter, cleaner
* enables a much nicer CLI experience (compare https://sparkswap.com/docs/getting-started/first-use)
* enables us to switch/add support of different implementations like c-lightning

kilrau:
You got it @ImmanuelSegol

ImmanuelSegol:
@kilrau  @sangaman Which methods should I implement?

kilrau:
Hold-off on this. I would like to finalize testing bitcoind and litecoind first and then list calls here.

ImmanuelSegol:
@kilrau Could I get a better description of which methods exactly I should implement?

kilrau:
Yes, of course. Relevant for the wallet usage are [LND](https://api.lightning.community)'s:

https://api.lightning.community/#genseed
https://api.lightning.community/#initwallet
https://api.lightning.community/#unlockwallet
https://api.lightning.community/#changepassword
https://api.lightning.community/#walletbalance
https://api.lightning.community/#gettransactions
https://api.lightning.community/#transactiondetails
https://api.lightning.community/#estimatefee
https://api.lightning.community/#listunspent
https://api.lightning.community/#sendcoins
https://api.lightning.community/#sendmany
https://api.lightning.community/#newaddress
https://api.lightning.community/#subscribetransactions

[...]

Issue #246, opened 2018-08-01
Include raiden into docker


kilrau:

sangaman:
I think it's worth discussing whether we should necessarily include lnd and raiden in the docker image. Since we have the option to disable either of those, . I don't we might not always need both. I don't know if it's reasonable to have multiple docker images or the option to exclude certain components from the image for people who only want to run one or the other, but I thought I'd bring it up.

kilrau:
Definitely worth discussing:
how about a docker 'setup' asking which coins/tokens the person wants to swap with a small hint what that means (download, disk space) based on this different docker images are chosen.

kilrau:
Can you lead this discussion @itsbalamurali

kilrau:
@reliveyy

reliveyy:
I think xud dockerfile should only cares about xud itself. And we provide a docker-compose file or a k8s yaml file to describe the setup process of the whole system instead.

kilrau:
That sounds about right! @reliveyy 

We can expect to use official [docker lnd](https://github.com/lightningnetwork/lnd/tree/master/docker) and [raiden](https://github.com/raiden-network/raiden/tree/master/docker) in future.

reliveyy:
https://github.com/reliveyy/xud-docker A quick PoC here (some services may not running properly and  I will fix theme later). If you have any suggestions please fire an issue in this PoC project and If the dockerization enhancement to xud is acceptable I will issue a PR to `xud` project

kilrau:
I learned quite a bit from the sprkswap's docker flow today. This is the level of experience we want. Added more details to the description above. Please check and let me know what you think @reliveyy @sangaman @erkarl

kilrau:
> I think it's worth discussing whether we should necessarily include lnd and raiden in the docker image. Since we have the option to disable either of those, . I don't we might not always need both. I don't know if it's reasonable to have multiple docker images or the option to exclude certain components from the image for people who only want to run one or the other, but I thought I'd bring it up.

Btw: That's taken care of by a) user choosing supported coins/tokens and also b) #891 @sangaman .

PR #889, opened 2019-04-15
Lnd wallet api wrapper 

### LINKED ISSUE #912: `xud` master password
Needs some research if this could work like this @sangaman :

* let's btclnd create memonic, use it for xud and inject it in all other non-preexisting nodes
* encrypts xud's private key and all other nodes non-preexisting nodes private key with `xud` master password
* lnd macaroons, full-node rpc passwords
* pre-existing LND/raiden/fullnode's: user has to enter passwords and xud stores and encrypts these pws with master password. Needs great explainer on what is happening here and why this is secure.
* as just discussed, this should be a xud core function exposed as gRPC and CLI call

ImmanuelSegol:
EDIT (by @kilrau ): closes #912


sangaman:
I'm a bit unclear as to the scope of this PR, could you elaborate a bit before I review? Also the tests are currently failing due to compilation errors, could you clear those up?

ImmanuelSegol:
> I'm a bit unclear as to the scope of this PR, could you elaborate a bit before I review? Also, the tests are currently failing due to compilation errors, could you clear those up?

It's not ready for review yet, @kilrau stated in issue #40 that we should create wrapper commands for specific LND commands.

sangaman:
Ah OK I missed the WIP in the PR title. When you get a chance it'd still be good to get an overview of what's being changed and added to have context

kilrau:
Any updates here? We would need these calls for the docker setup soonish @ImmanuelSegol

ImmanuelSegol:
> Any updates here? We would need these calls for the docker setup soonish @ImmanuelSegol

Will complete this today as well.

> Any updates here? We would need these calls for the docker setup soonish @ImmanuelSegol
@kilrau Yeah test it to make sure everything works well.

kilrau:
I think it should be clear from [the original issue](https://github.com/ExchangeUnion/xud/issues/40) that we decided to NOT integrate all `lnd` client calls, but only the ones that matter.

What might have not been clear: integrating a call shouldn't be 1-1, because that would indeed require the user to learn `lncli`, which doesn't make wrapping very meaningful in the first place. What makes it meaningful: we can simplify lnd client calls a lot and consequently use `xucli` syntax.

I suggest as a next step for you to go over the [relevant calls I listed](https://github.com/ExchangeUnion/xud/issues/40#issuecomment-482053134) and propose a syntax and info the call should have with an example each. @reliveyy @sangaman and I will then go and check to make sure this meets the use case, because @sangaman was perfectly right - they are very use case specific; mainly #912. Let me give you an example what I mean by 'simplify': the user should never have to provide a peer's LND URI, all should be automatically mapped internally by providing `xud` `NodePubKey` only.

So please go ahead and list the calls with samples here and we'll help you to finalize these. Not all calls [I listed](https://github.com/ExchangeUnion/xud/issues/40#issuecomment-482053134) are needed immediately, so we can leave them out for now. To make it simple, I closed #40 because it was too generic and we make this issue about wrapping calls for the #912 use case. All other calls we'll need after, we play it by ear - as soon as we have a use case, we'll add it.

kilrau:
Also since it's targeting #912 now, this PR should include wrappers for `geth` with the same functionality. Maybe change the title to something more generic like "Layer1/2 client api wrapper"

kilrau:
Here an update list of lnd calls that should be exposed in a simplified manner via xud's grpc:

https://api.lightning.community/#genseed
https://api.lightning.community/#initwallet
https://api.lightning.community/#unlockwallet
https://api.lightning.community/#walletbalance
https://api.lightning.community/#gettransactions
https://api.lightning.community/#estimatefee
https://api.lightning.community/#sendcoins
https://api.lightning.community/#newaddress
https://api.lightning.community/#subscribechannelbackups (for #246 )

Anything missing for #912 @sangaman @erkarl ?

kilrau:
aight, so close this PR and include in new and existing issues? @erkarl @ImmanuelSegol

PR #894, opened 2019-04-16
test: fix placeOrderAndBroadcast race condition

sangaman:
This attempts to fix a potential race condition that occasionally caused the simulation tests to fail. In `placeOrderAndBroadcast` we call `SubscribeAddedOrders` on node A, then immediately place an order on node B, which immediately broadcasts the order to node A. It's possible that the order subscription is not active on node A by the time it hears about the order since there's no latency between nodes.

This fix calls `ListPeers` after `SubscribeAddedOrders` to ensure that the two nodes are still connected and moves the `getOrdersCount` call to after `SubscribeAddedOrders`. This in turn creates a slight delay between subscribing for added orders and adding the order, resolving the race.

### DIFF
--- a/test/simulation/actions.go
+++ b/test/simulation/actions.go
@@ -70,12 +70,24 @@ func (*actions) connect(ctx context.Context, assert *require.Assertions, srcNode
 
 func (*actions) placeOrderAndBroadcast(ctx context.Context, assert *require.Assertions, srcNode, destNode *xudtest.HarnessNode,
 	req *xudrpc.PlaceOrderRequest) *xudrpc.Order {
-	// 	Fetch nodes current order book state.
+	// Subscribe to added orders on destNode
+	destNodeAddedOrderChan := subscribeAddedOrders(ctx, destNode)
+
+	// Fetch nodes current order book state.
 	prevSrcNodeCount, prevDestNodeCount, err := getOrdersCount(ctx, srcNode, destNode)
 	assert.NoError(err)
 
-	// Subscribe to added orders on destNode
-	destNodeAddedOrderChan := subscribeAddedOrders(ctx, destNode)
+	// Ensure that destNode is connected to srcNode
+	resListPeers, err := destNode.Client.ListPeers(context.Background(), &xudrpc.ListPeersRequest{})
+	assert.NoError(err)
+	hasPeer := false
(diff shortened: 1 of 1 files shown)


test/simulation/actions.go
```diff
@@ -70,12 +70,18 @@ func (*actions) connect(ctx context.Context, assert *require.Assertions, srcNode
...
 	// 	Fetch nodes current order book state.
 	prevSrcNodeCount, prevDestNodeCount, err := getOrdersCount(ctx, srcNode, destNode)
 	assert.NoError(err)
 
-	// Subscribe to added orders on destNode
-	destNodeAddedOrderChan := subscribeAddedOrders(ctx, destNode)
+	// Ensure that destNode is connected to srcNode
+	resListPeers, err := destNode.Client.ListPeers(context.Background(), &xudrpc.ListPeersRequest{})
+	assert.NoError(err)
+	assert.Len(resListPeers.Peers, 1)
```
moshababo: Node might have more peers connected. Just need to check that this peer is connected, not that it's the only one.

sangaman: Good point.

sangaman:
> Did you also try to replace listPeers with a timeout/sleep to test how long the latency is?

I haven't tried testing that, I would guess that the latency is virtually nothing, but that the subscribe orders and place order call are happening almost simultaneously from the tests but being handled asynchronously by xud and that the order is guaranteed. I wouldn't be surprised if a millisecond is enough to wait between calls, because the starting the subscription isn't computationally expensive at all.

I'll probably merge this after handling moshe's comment above so that I can have reliable tests while working on a separate branch.

sangaman:
A test just failed due to a similar timeout error in a different place (this time right after subscribing to removed orders), so I'm going to fix that as well and scan the rest of the tests if there's anywhere else where we set up an order subscription that may be vulnerable to the same issue.

```
    --- FAIL: TestExchangeUnionDaemon/order_broadcast_and_invalidation (5.00s)
        require.go:794: 
            	Error Trace:	actions.go:151
            	            				xud_test.go:143
            	            				xud_test.go:180
            	            				xud_test.go:405
            	            	timeout reached before event was received
            	Test:       	TestExchangeUnionDaemon/order_broadcast_and_invalidation
```

PR #904, opened 2019-04-25
refactor(rpc): combine added & removed order calls

### LINKED ISSUE #903: Combine `SubscribeAddedOrders` and `SubscribeRemovedOrders` to ensure correct order

sangaman:
This combines the `SubscribeAddedOrders` and `SubscribeRemovedOrders` rpc calls into a single `SubscribeOrders` calls. This is done to ensure that updates to the order book do not get transmitted out of order. gRPC guarantees message ordering within an individual call: https://grpc.io/docs/guides/concepts.html#server-streaming-rpc. However, with separate calls an edge case exists where a client receives an order removal message before the order added message for that same order.

Closes #903.


sangaman:
Just a note, the tests are failing because I worked on the simulation tests using Go 1.12, which is currently the minimum version required by lnd. I'm thinking it would make sense to use 1.12 for our simulation tests as well, particularly in anticipation of using lnd 0.6 with hold invoices (which will not worker with less than 1.12). I'll put that in a separate PR though so it can be merged sooner, then rebase this PR.

sangaman:
Tests passing now with Go 1.12.

PR #906, opened 2019-04-25
test: use go 1.12 for simulation tests

sangaman:
This updates the simulation tests to use Go 1.12 in anticipation of using lnd 0.6, for which the minimum Go version is 1.12.


kilrau:
Alex Bosworth reported some 30x-100x performance loss with go 1.12 on Mac. If that is also the case for you, you might just want to continue developing on 1.11 @moshababo

PR #913, opened 2019-04-26
Listorderhistory

### LINKED ISSUE #689: New CLI - ListOrderHistory
I see a new for a new cli `getorder` which accept orderId as a mandatory parameter.

The CLI should list all events related to the order since entered until fully removed from the orderbook.

The output of the CLI should allow us to understand what had happend with the order similar to this:
https://github.com/ExchangeUnion/xud/issues/638#issuecomment-436665439

#677 and #477 don't provide a good answer to the need.

ImmanuelSegol:
closes #689 

@kilrau @sangaman 
We currently don't store the `Order status` in the DB, but we do emit events for all operations on orders in the order book.

Should we persist the order status ? it doesn't have to necessarily be in the DB.

We need this for `Order status: filled/cancelled/open`.


kilrau:
Can't you get the order status by checking `trades` along these lines:
trades covering full quantity = `filled`
trades covering partial quantity = `partially filled`
no trade = either `open` or `cancelled`. Cancelled should have an entry in the log. Open should have nothing.

Other ideas? @sangaman

ImmanuelSegol:
> Can't you get the order status by checking `trades` along these lines:
> trades covering full quantity = `filled`
> trades covering partial quantity = `partially filled`
> no trade = either `open` or `cancelled`. Cancelled should have an entry in the log. Open should have nothing.
> 
> Other ideas? @sangaman

@kilrau @sangaman Would it not be better to just save the status if the order using these event listeners 
[here](https://github.com/ExchangeUnion/xud/blob/ffd7dc8b42911ff9cdc69eee62355475795936e8/lib/orderbook/OrderBook.ts#L21)

sangaman:
@ImmanuelSegol Let's discuss this when I (finally) make time to go over the codebase with you. I think @kilrau's approach is generally correct. I definitely don't think we need to persist status in the db. Orders do have an `initialQuantity` field, fwiw, you can check this to determine whether an order is partially filled.

kilrau:
I think it makes sense to schedule a 1-on-1 with @sangaman @ImmanuelSegol

ImmanuelSegol:
@kilrau @sangaman 
I add the ability to the order status by checking the trades.

The last dilemma that I have has to do with looking throw the log for canceled orders, it seems like a bad idea. I think we should approach it from another angle.

kilrau:
> The last dilemma that I have has to do with looking throw the log for canceled orders, it seems like a bad idea. I think we should approach it from another angle.

Ok. Please propose sth!

ImmanuelSegol:
@sangaman review, please?

kilrau:
gave this a quick test by adding an order:
```
~/xud/bin$ ./xucli buy 1 ltc/btc 0.1
no matches found
remaining 1 LTC entered the order book as 5b856430-8948-11e9-a70b-db3f228855e8
```
but
```
~/xud/bin$ ./xucli listorderhistory 5b856430-8948-11e9-a70b-db3f228855e8
Error: 2 UNKNOWN: pairId provided dosent exsist
```

also pls fix the 'doesn't exist' typo

lib/constants/enums.ts
```diff
@@ -14,6 +14,13 @@ export enum OrderSide {
   Sell,
 }
 
+export enum OrderStatus {
+  FILLED  = 0,
+  PARTIALLYFILLED = 1,
```
sangaman: This should be `CAPITALS_WITH_UNDERSCORES` style

proto/xudrpc.proto
```diff
@@ -405,6 +413,30 @@ message ListPeersResponse {
...
+}
+
+message ListOrderHistoryResponse {
+  repeated Trade trades = 1 [json_name = "trades"];
+  repeated SwapSuccess swaps = 2 [json_name = "swaps"];
+  OrderStatus status = 3 [json_name = "status"];
+
+  enum OrderStatus {
+    FILLED  = 0;
+    PARTIALLYFILLED = 1;
```
sangaman: enums should be `CAPITALS_WITH_UNDERSCORES` according to the [proto3 style guide](https://developers.google.com/protocol-buffers/docs/style).

proto/xudrpc.proto
```diff
@@ -405,6 +413,30 @@ message ListPeersResponse {
...
+    OPEN = 2;
+    CANCELLED = 3;
+  };
+}
+
+message Trade {
```
sangaman: I think it would be nice to have the price in this message as well, so that someone can quickly see the price of the trade without having to look up the orders.

lib/service/errors.ts
```diff
@@ -25,6 +26,10 @@ const errors = {
     message: 'a pending call is ongoing that conflicts with this call',
     code: errorCodes.PENDING_CALL_CONFLICT,
   },
+  PAIRID_NON_EXISTENT: () => ({
+    message: 'pairId provided dosent exsist',
```
sangaman: "doesn't exist"

proto/xudrpc.proto
```diff
@@ -405,6 +413,30 @@ message ListPeersResponse {
   repeated Peer peers = 1 [json_name = "peers"];
 }
 
+message ListOrderHistoryRequest {
+  string order_id = 1 [json_name = "order_id"];
```
sangaman: It seems a bit weird for me to use this call to also look up single orders, that feels like it should be a separate rpc call. Instead maybe it can accept a `limit` like we do elsewhere, or maybe a parameter to only get orders that are less than X days old (can be a future PR though). Thoughts @kilrau?

kilrau: Without checking the code: ListOrderHistory is supposed to accept ONE `order_id` and lists all events like

Order status: filled/cancelled/open
Swapdeals:
Trades:
Log entries:

Nothing multiple orders..

sangaman: Ah ok my mistake, I'm trying to track too many things at once. I will review again with that in mind.

proto/xudrpc.proto
ImmanuelSegol: I messaged you privately about this.

rsercano:
Merged master, checked a bit but as far as I understood there's an issue as @kilrau already mentioned I'll add it as a test case after I find out what causes it.

rsercano:
As per our talk with @kilrau this should be capable of returning in-memory orders too, I'm on it.

rsercano:
I've added `ownOrders` which are in-memory but not sure about `peerorders` ? Should I check `peerorders` too ?

p.s. Also added two new test cases

kilrau:
> Should I check `peerorders` too ?

No, only `ownOrders`.

rsercano:
Great then I assume that's done @kilrau

lib/proto/xudrpc.swagger.json
```diff
@@ -21,7 +21,7 @@
         "operationId": "AddCurrency",
         "responses": {
           "200": {
-            "description": "",
+            "description": "A successful response.",
```
sangaman: Can you try regenerating protos to see if it gets rid of some of these unintended changes?

proto/xudrpc.proto
```diff
@@ -163,6 +163,14 @@ service Xud {
     };
   }
 
+  /* List all events related to the order. */
+  rpc ListOrderHistory(ListOrderHistoryRequest) returns (ListOrderHistoryResponse) {
+    option (google.api.http) = {
+      post: "/v1/listorderhistory"
```
sangaman: This should probably be a `get` and just `/v1/orderhistory`

rsercano:
Implemented @sangaman, thanks :)

sangaman:
@rsercano Now that #791 is merged, can you take care of these conflicts and I'll review again? Thanks.

rsercano:
Rebased hopefully :flushed:

lib/service/Service.ts
```diff
@@ -349,6 +350,44 @@ class Service {
...
+  public listOrderHistory = async (args: { orderId: string }): Promise<{
+    trades: TradeInstance[], swapDeals: SwapDealInstance[], status: OrderStatus}> => {
+    const order = await this.orderBook.getOrder(args.orderId);
+
+    if (order === null) {
+      throw errors.PAIRID_NON_EXISTENT();
+    }
+
+    const trades = await this.orderBook.getTradesByOrderId(order.id, true);
+    const swapDeals = await this.swaps.getDeals(true);
```
sangaman: We are pulling back all swap deals here, which is a mistake. We only want the ones for this order.

lib/service/Service.ts
```diff
@@ -349,6 +350,44 @@ class Service {
...
+    trades.forEach((trade) => {
+      quantity = quantity + trade.quantity;
+    });
+
+    if (!isInOrderBook && quantity === order.initialQuantity) {
+      status = OrderStatus.FILLED;
+    } else if (!isInOrderBook && trades.length === 0) {
+      status = OrderStatus.CANCELLED;
+    } else if (isInOrderBook && trades.length === 0) {
+      status = OrderStatus.OPEN;
```
sangaman: `status` is `OPEN` when it's initialized above, I figure we can leave this condition out?

lib/service/Service.ts
```diff
@@ -349,6 +350,44 @@ class Service {
...
+
+    if (order === null) {
+      throw errors.PAIRID_NON_EXISTENT();
+    }
+
+    const trades = await this.orderBook.getTradesByOrderId(order.id, true);
+    const swapDeals = await this.swaps.getDeals(true);
+
+    const isInOrderBook = this.orderBook.getOwnOrder(order.id, order.pairId);
+    let quantity = 0;
```
sangaman: Can we call this `tradedQuantity` or `filledQuantity`? It was confusing to me at first.

lib/service/Service.ts
```diff
@@ -349,6 +350,44 @@ class Service {
...
+    });
+
+    if (!isInOrderBook && quantity === order.initialQuantity) {
+      status = OrderStatus.FILLED;
+    } else if (!isInOrderBook && trades.length === 0) {
+      status = OrderStatus.CANCELLED;
+    } else if (isInOrderBook && trades.length === 0) {
+      status = OrderStatus.OPEN;
+    } else if (isInOrderBook && quantity < order.initialQuantity) {
+      status = OrderStatus.PARTIALLY_FILLED;
```
sangaman: One thought that occurred to me. We may have an order that was partially filled then canceled (no longer in the order book) as well as an order that was partially filled but is still open. Should we have two different statuses for these? `PARTIALLY_FILLED` if it's open but `PARTIALLY_FILLED_CANCELED` if it's closed? Thoughts @kilrau?

test/integration/Service.spec.ts
```diff
@@ -105,6 +105,31 @@ describe('API Service', () => {
     expect(order.id).to.equal(orderId);
   });
 
+  it('should list order history with a trade', async () => {
+    const { status, trades, swapDeals } = await service.listOrderHistory({ orderId: orderId! });
+
+    expect(status).to.equal(2);
+    expect(trades).to.eql([]);
+    expect(swapDeals).to.eql([]);
```
sangaman: Just for legibility purposes, I think it'd be nicer if these all ready `to.equal` instead of `to.eql` so there's no doubt that they're doing the same thing.

lib/cli/commands/listorderhistory.ts
```diff
@@ -0,0 +1,19 @@
+import { callback, loadXudClient } from '../command';
+import { Arguments } from 'yargs';
+import { ListOrderHistoryRequest } from '../../proto/xudrpc_pb';
+
+export const command = 'listorderhistory <order_id>';
+
+export const describe = 'list all events related to the order since entered until fully removed from the orderbook';
```
sangaman: This is a little long for a command line description, how about just "lists all events related to an order"

lib/grpc/GrpcService.ts
```diff
@@ -583,6 +599,43 @@ class GrpcService {
...
+  /**
+   * See [[Service.listOrderHistory]]
+   */
+  public listOrderHistory: grpc.handleUnaryCall<xudrpc.ListOrderHistoryRequest, xudrpc.ListOrderHistoryResponse> = async (call, callback) => {
+    try {
+      const orderHistory = await this.service.listOrderHistory(call.request.toObject());
+      const response = new xudrpc.ListOrderHistoryResponse();
+
+      const trades: xudrpc.Trade[] = [];
+      const swapDeals: xudrpc.SwapSuccess[] = [];
```
sangaman: Maybe rename this `swapSuccesses`?

test/integration/Service.spec.ts
```diff
@@ -105,6 +105,31 @@ describe('API Service', () => {
     expect(order.id).to.equal(orderId);
   });
 
+  it('should list order history with a trade', async () => {
```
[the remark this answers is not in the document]
rsercano: typo! Thanks :)

lib/service/errors.ts
```diff
@@ -30,6 +31,10 @@ const errors = {
     message: `failed to open channel with nodePubKey: ${nodePubKey}, currency: ${currency}, amount: ${amount}, message: ${message}`,
     code: errorCodes.OPEN_CHANNEL_FAILURE,
   }),
+  PAIRID_NON_EXISTENT: () => ({
```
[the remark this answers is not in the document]
rsercano: Does it mean I have to import order book error to the here ? Committing this way but that doesn't feel right to me ?

lib/service/Service.ts
rsercano: waiting @kilrau to answer

kilrau: Good point @sangaman 

> `PARTIALLY_FILLED` if it's open but `PARTIALLY_FILLED_CANCELED` if it's closed?

That actually sound good to me!

test/integration/Service.spec.ts
rsercano: in fact they're doing different things, `eql` is checking deeply as far as I remember, if I use equal it throws with `not equal` error

lib/service/Service.ts
rsercano: will check it

kilrau:
https://github.com/ExchangeUnion/xud/issues/1514 solves the use case "know what happened to an order" for a live-trade use case. We'll see if we need something separate for unattended market maker nodes apart from https://github.com/ExchangeUnion/xud/pull/1243. Closing here for now.

PR #916, opened 2019-04-27
test: simulation framework refactoring, p2p scenarios

### LINKED ISSUE #919: [xud simulation] Implement more tests (p2p)
P2P:

- [ ] Wrong identity: Alice connects to Bob expecting a random public key. Bob should reject the connection due to auth invalid target.

- [ ] Banning: If Alice bans Bob, connection attempts from both directions should fail.

- [ ] Unbanning: If Alice unbanned Bob, connection attempts from both directions should succeed.

- [ ] Duplicate connection: If Alice and Bob are already connected, connection attempts from both directions should get blocked.

moshababo:
* Refactoring to the simulation framework (to be continued):
  * 4 running `xud` nodes instead of 2.
  * Separation of test scenarios from framework/setup code.
  * Avoid static config when handling the circular dependency between `lnd` and `xud` (due to the hash resolver), by breaking `xud` setup into few steps.

* P2P test scenarios:
  * Discovery
  * Incorrect public key 
  * Ban/Unban
  * Already connected

* Non-test code: small bug fix in the p2p discovery. Previously if we heard about a new node, we wouldn't connect to it (I inverted the negative conditions because they became unclear). 

Closes #919


kilrau:
> * Non-test code: small bug fix in the p2p discovery. Previously if we heard about a new node, we wouldn't connect to it

I suspected that was the case. Very nice, that the tests caught it!

.travis.yml
```diff
@@ -15,11 +15,14 @@ script:
...
   - eval "$(GIMME_GO_VERSION=1.12 gimme)"
   - go get -u github.com/golang/dep/cmd/dep
   - go get -u github.com/btcsuite/btcd
   - npm install
-  - npm run compile
   - npm run test:sim
+  - cat test/simulation/xuddatadir-Alice/logs/xud.log
+  - cat test/simulation/xuddatadir-Bob/logs/xud.log
+  - cat test/simulation/xuddatadir-Carol/logs/xud.log
+  - cat test/simulation/xuddatadir-Dave/logs/xud.log
```
sangaman: Do we want to leave all these statements here? Seems like it would be a lot of output.

lib/p2p/Pool.ts
```diff
@@ -245,11 +245,11 @@ class Pool extends EventEmitter {
    * If the node is banned, already connected, or has no listening addresses, then do nothing.
    * Additionally, if we're already trying to connect to a given node also do nothing.
    * @param nodes a collection of nodes with a `forEach` iterator to attempt to connect to
-   * @param ignoreKnown whether to ignore nodes we are already aware of, defaults to false
+   * @param allowKnown whether to allow nodes we are already aware of, defaults to true
```
sangaman: Could you make this a bit clearer by saying "whether to allow *connecting to* nodes we are already aware of"

package.json
```diff
@@ -32,7 +32,7 @@
     "test:p2p": "mocha -r ts-node/register test/p2p/*",
     "test:p2p:watch": "mocha -r ts-node/register test/p2p/*  --watch --watch-extensions ts",
     "test:crypto": "mocha -r ts-node/register test/crypto/*",
-    "test:sim": "(cd test/simulation && GO111MODULE=on go test -v)",
+    "test:sim": "(npm run compile && cd test/simulation && GO111MODULE=on go test -v)",
```
sangaman: Nice, saves me from having to manually recompile every time I run this locally!

package-lock.json
```diff
@@ -3436,8 +3436,7 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true,
-          "optional": true
+          "bundled": true
```
sangaman: I'm guessing these package-lock.json changes were not deliberate, I still don't fully understand why these changes pop up from time to time but I'm thinking it'd be best to leave them out of this PR to minimize the number of times this file changes unnecessarily.

test/simulation/config.go
```diff
@@ -4,15 +4,15 @@ import "time"
 
 const (
 	defaultXudKill    = true
-	defaultXudCleanup = true
+	defaultXudCleanup = false
```
sangaman: Did you mean to flip this?

With https://github.com/ExchangeUnion/xud/pull/923/commits/631d60792913de72e26b8e3feb3f63e6f43cb9d1 I have been using `false` here locally as well and I can run the sim tests without having to manually delete the data dir folders each time, so maybe it would make sense to leave it `false` here, but I wanted to make sure it was intentional.

lib/p2p/Pool.ts
```diff
@@ -245,11 +245,11 @@ class Pool extends EventEmitter {
...
    * @param retryConnecting whether to attempt retry connecting, defaults to false
    * @returns a promise that will resolve when all outbound connections resolve
    */
-  private connectNodes = (nodes: NodeConnectionIterator, ignoreKnown = false, retryConnecting = false) => {
+  private connectNodes = (nodes: NodeConnectionIterator, allowKnown = true, retryConnecting = false) => {
```
sangaman: I'm curious where the bug fix pertaining to discovery actually is. The code here makes sense to me (and is easier to understand than before), but it seems to function the same as before. Could you explain a bit more how it is fixed?

test/simulation/harnessTest.go
```diff
@@ -0,0 +1,91 @@
...
+	"testing"
+)
+
+type testCase struct {
+	name string
+	test func(net *xudtest.NetworkHarness, t *harnessTest)
+}
+
+// harnessTest wraps a regular testing.T providing enhanced error detection
+// and propagation. All error will be augmented with a full stack-trace in
```
sangaman: "error" should be plural.

test/simulation/xudtest/node.go
```diff
@@ -51,13 +51,15 @@ func (cfg nodeConfig) genArgs() []string {
 	var args []string
 
 	args = append(args, "--initdb=false")
+	args = append(args, "--nosanitychecks=true")
```
sangaman: Why reenable this flag? I think it's handy to test the sanity checks automatically, and if they are not passing it's likely a sign that there's a bug in the implementation.

.travis.yml
moshababo: No, I was adding this to debug the certificate generation issue (`travis debug` commits). Will remove it before merging.

package.json
moshababo: This came after I spend an hour trying to figure why the test are failing, after switching branches and not recompiling :\

moshababo: This came after I spend an hour trying to figure why the test are failing with a weird error, after switching branches and not recompiling :\

test/simulation/config.go
moshababo: The reason was only for debugging the certificate generation issue (`travis-debug` commits), so i'm going to remove this.

I'd say not to prefer using memory db, because there's no problem with deleting these folders, and better to have the behavior close as possible to production-like. Also, I sometimes read this database after runs to inspect various issues. But i'll review your solution better and comment there.

test/simulation/xudtest/node.go
moshababo: When I resolved conflicts with master, it looked to me like you added it. 
If that wasn't the case, i'll remove it.

lib/p2p/Pool.ts
moshababo: The previous condition was:
````
const isNotIgnored = this.nodes.has(node.nodePubKey) && !ignoreKnown;
````

It’s checking whether we are not ignoring this node, which logically applies only to nodes we already know (the first condition). But if we simply don't know this node, it will return false, and we won't attempt to connect.

moshababo:
> Do you think this makes our javascript p2p sanity tests obsolete, or is there value in keeping them around?

Although they were useful mostly because we didn't had anything better, I think we should keep them because:

*  it's testing all instances under a single thread, which sometimes emulate stress condition (like with the `Parser` bug).
* they are useful for very easily launch and dispose nodes with different configurations (like in the `networks.spec.ts` tests). It can't be done so fast in the network-simulation.

moshababo:
> Also, not terribly important but I noticed a couple things that I believe would throw golint warnings, namely id capitalized as Id instead of ID and a few exported types/functions without comments. If you could try running golint and see about addressing any warnings it throws that would be nice so they don't come up later.

I saw you already applied it in other PR. So i'l skip that for this branch, to avoid unnecessary potential merge conflicts.

sangaman:
> I saw you already applied it in other PR. So i'l skip that for this branch, to avoid unnecessary potential merge conflicts.

OK that's fine, I'll just rebase that one after this is merged in that case.

test/simulation/xudtest/node.go
sangaman: Ah ok, must have just been some mix up then.

lib/p2p/Pool.ts
sangaman: Ahhhh ok, yes that makes sense thank you. It's clearer now (and working as intended).

PR #918, opened 2019-04-27
refactor: swap client management

### LINKED ISSUE #879: Show channelbalance for all the registered currencies
It currently only works for BTC, LTC and WETH - https://github.com/ExchangeUnion/xud/blob/master/lib/service/Service.ts#L126

### LINKED ISSUE #898: Trading new ERC20 token requires restart
Currently, when adding a new ERC20 token requires a restart because we're associating the swap client (Raiden) with currencies while starting up xud. Currencies that are added at runtime are not immediately associated with the swap client. That will result in the following error message in case of buy/sell orders:
```
Error: 2 UNKNOWN: unable to find swap client for currency WETH
```

ghost:
In this PR we move swap client related logic out of various places in the codebase to `SwapClientsManager`. Still WIP, but good for a concept review.

Closes #898 
Closes #879


lib/swaps/SwapClientsManager.ts
```diff
@@ -0,0 +1,150 @@
...
+import { Models } from '../db/DB';
+import BaseClient from '../BaseClient';
+import LndClient, { LndInfo } from '../lndclient/LndClient';
+import RaidenClient, { RaidenInfo } from '../raidenclient/RaidenClient';
+import { Loggers } from '../Logger';
+import Pool from '../p2p/Pool';
+import { SwapClient } from '../constants/enums';
+import OrderBookRepository from '../orderbook/OrderBookRepository';
+
+class SwapClientsManager {
```
sangaman: Minor nit but I think "SwapClientManager" without the s sounds better.

lib/service/Service.ts
```diff
@@ -121,17 +119,14 @@ class Service extends EventEmitter {
...
     const balances = new Map<string, { balance: number, pendingOpenBalance: number }>();
     const getBalance = async (currency: string) => {
-      const client = this.lndClients[currency.toUpperCase()]
-        // TODO: support all registered tokens for Raiden
-        || (currency.toUpperCase() === 'WETH' && this.raidenClient);
-
-      if (!client) {
-        // TODO: throw an error here indicating that lnd is disabled for this currency
+      try {
+        const client = this.swapClientsManager.get(currency.toUpperCase());
```
sangaman: I think best to call this `swapClient` just to be consistent with how its called elsewhere

lib/service/Service.ts
```diff
@@ -121,17 +119,14 @@ class Service extends EventEmitter {
...
+        const channelBalance = await client.channelBalance();
+        return channelBalance;
+      } catch (e) {
```
sangaman: I think this may be better without the try/catch, because if we get an error on the call it will bubble up to the grpc layer and return a grpc error. With the try/catch, if there's an error we return balances of 0 which is not accurate. I'd leave that TODO in there (and this will remind me to fix it) or if you want you can have it throw an error rather than return 0 balances when there's no swap client.

lib/swaps/SwapClientsManager.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  public async getRaidenInfo() {
+    const raidenInfo = this.raidenClient.isDisabled() ? undefined : await this.raidenClient.getRaidenInfo();
+    return raidenInfo;
+  }
+
+  public get(currency: string): BaseClient {
+    const swapClient = this.swapClients.get(currency);
+    if (!swapClient) {
+      throw new Error(`swapClient for currency ${currency} not found`);
```
sangaman: Maybe this is where we can throw that error I was talking about in my last comment, and ideally make it a defined error type rather than a generic `Error` so we can easily check for it.

lib/swaps/SwapClientsManager.ts
```diff
@@ -0,0 +1,150 @@
...
+import BaseClient from '../BaseClient';
+import LndClient, { LndInfo } from '../lndclient/LndClient';
+import RaidenClient, { RaidenInfo } from '../raidenclient/RaidenClient';
+import { Loggers } from '../Logger';
+import Pool from '../p2p/Pool';
+import { SwapClient } from '../constants/enums';
+import OrderBookRepository from '../orderbook/OrderBookRepository';
+
+class SwapClientsManager {
+  private lndClients: { [currency: string]: LndClient | undefined } = {};
```
sangaman: I think this would be cleaner as a `Map` type.

sangaman (commented):
Made a quick pass with a few comments. One feature that would be really nice to have, and maybe makes sense to implement here, is making the `LndClient` output which currency it's for every time it writes a log statement. That way we can easily tell which `LndClient` is printing to logs, often times right now there's no real way to tell.

Conceptually though I like this and the manager class approach is good i think.

erkarl:
Feedback applied. Ready for round 2. Before merging I definitely want to also add TypeDoc comments to `SwapClientManager`.

I included https://github.com/ExchangeUnion/xud/pull/941 in this PR because I was getting inconsistent results when running the simulation tests.

lib/swaps/SwapClientManager.ts
```diff
@@ -0,0 +1,224 @@
...
+      }
+    }
+  }
+
+  /**
+   * Gets a swap client instance.
+   * @param currency a currency that the swap client is linked to.
+   * @returns swap client instance upon success, throws otherwise.
+   */
+  public get(currency: string): BaseClient {
```
sangaman: Just for consistency's sake, we're using arrow functions (almost) everywhere else so I think it's better to use that here too unless you have a reason not to, `public get = (currency: string): BaseClient {`. Same for other functions in this class.

lib/service/Service.ts
sangaman: Nice solution!

lib/swaps/SwapClientManager.ts
```diff
@@ -0,0 +1,224 @@
...
+import Pool from '../p2p/Pool';
+import { SwapClient } from '../constants/enums';
+import OrderBookRepository from '../orderbook/OrderBookRepository';
+import errors from './errors';
+import { Currency } from '../orderbook/types';
+
+class SwapClientManager {
+  /** A map between currencies and all swap clients */
+  private swapClients = new Map<string, BaseClient>();
+  private raidenClient!: RaidenClient;
```
sangaman: I would either make this nullable with the `?` operator, or create the `RaidenClient` object in the constructor (so that indeed this value will never be undefined) and then initialize it inside the `init` function here and let it become disabled.

lib/swaps/SwapClientManager.ts
```diff
@@ -0,0 +1,224 @@
...
+      }
+    }
+  }
+
+  /**
+   * Removes a new swap client and currency association.
+   * @param currency a currency that should be unlinked from a swap client.
+   * @returns Nothing upon success, throws otherwise.
+   */
+  public remove(currencyId: string): void {
```
sangaman: To match the @param comment above (and how we're usually calling this throughout the codebase) I'd rename the param to `currency`.

lib/swaps/SwapClientManager.ts
```diff
@@ -0,0 +1,224 @@
...
+    }
+  }
+
+  /**
+   * Gets all lnd clients' pubKeys.
+   * @returns An object containing lnd public keys.
+   */
+  public getLndPubKeys(): LndPubKeys {
+    const lndPubKeys: LndPubKeys = {};
+    for (const [currency, swapClient] of this.swapClients.entries()) {
```
sangaman: What do you think about maintaining this list of pubkeys (similar to how we're maintaining the list of raiden currencies) rather than determining it dynamically when needed? I don't think it will be changing very often, so it might be ever so slightly clearer/more performant.

lib/swaps/SwapClientManager.ts
```diff
@@ -0,0 +1,224 @@
...
+      if (swapClient.type === SwapClient.Lnd) {
```
sangaman: In several places we're checking `swapClient.type` and then casting. It'd be nice to create our own type checking functions that will automatically tell the compiler the type, like we for `isPeerOrder` and `isOwnOrder`

```typescript
export function isLndClient(swapClient: BaseClient): swapClient is LndClient {
  return (swapClient.type === SwapClient.Lnd);
}

export function isRaidenClient(swapClient: BaseClient): swapClient is RaidenClient {
  return (swapClient.type === SwapClient.Raiden);
}
```

jest.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
sangaman: Mainly a question, but do you see value in keeping mocha and jest side by side? Or should we consider migrating to jest? Or maybe just the parts that are using stubbing, replacing sinon so to speak?

lib/swaps/Swaps.ts
```diff
@@ -256,7 +257,7 @@ class Swaps extends EventEmitter {
   public executeSanitySwap = async (currency: string, peer: Peer) => {
     const { rPreimage, rHash } = await generatePreimageAndHash();
     const peerPubKey = peer.nodePubKey!;
-    const swapClient = this.swapClients.get(currency);
+    const swapClient = this.swapClientManager.get(currency);
```
sangaman: We probably want a try/catch here (rather than the if statement below) in case we have no swap client for this currency.

lib/swaps/SwapClientManager.ts
```diff
@@ -0,0 +1,224 @@
...
+    }
+    // setup raiden client and connect if configured
+    const raidenEnabled = !this.config.raiden.disable;
+    if (raidenEnabled) {
+      this.raidenClient = new RaidenClient(this.config.raiden, this.loggers.raiden);
+      initPromises.push(this.raidenClient.init());
+    }
+    // bind event listeners
+    this.bind();
+    await Promise.all(initPromises);
```
sangaman: It might make sense to remove swap clients that turned out to be disabled after the `initPromises` resolve, since we currently treat disabled clients as if they don't exist I believe.

lib/swaps/Swaps.ts
```diff
@@ -369,7 +370,7 @@ class Swaps extends EventEmitter {
 
     const { makerCurrency, makerAmount, takerCurrency, takerAmount } = Swaps.calculateMakerTakerAmounts(quantity, price, isBuy, requestBody.pairId);
 
-    const swapClient = this.swapClients.get(takerCurrency);
+    const swapClient = this.swapClientManager.get(takerCurrency);
```
sangaman: Same here for try/catch if the swap client doesn't exist, replacing the if statement below.

lib/swaps/Swaps.ts
```diff
@@ -503,7 +504,7 @@ class Swaps extends EventEmitter {
       }
     }
 
-    const swapClient = this.swapClients.get(deal.makerCurrency);
+    const swapClient = this.swapClientManager.get(deal.makerCurrency);
     if (!swapClient) {
```
sangaman: I think we can just remove this if statement (it will never be falsey), if don't have the swap client after we'd already checked it earlier then it's probably fine to let the exception bubble up.

lib/swaps/Swaps.ts
```diff
@@ -613,7 +614,7 @@ class Swaps extends EventEmitter {
         return rPreimage;
       } else {
         // we need to get the preimage by making a payment
-        const swapClient = this.swapClients.get(currency);
+        const swapClient = this.swapClientManager.get(currency);
```
sangaman: Same here for try/catch if the swap client doesn't exist, replacing the if statement below.

test/integration/Service.spec.ts
```diff
@@ -35,10 +35,10 @@ describe('API Service', () => {
       },
       lnd: {
         LTC: {
-          disable: true,
+          disable: false,
         },
         BTC: {
-          disable: true,
+          disable: false,
```
sangaman: Why are these values changed?

sangaman (commented):
Looking good, this is turning out to be a pretty big change but I think it'll be helpful to organize the logic handling swap clients before it grows even more complex.

jest.config.js
erkarl: I'd do a gradual upgrade as the benefits of doing a full rewrite (less dev-dependencies) don't justify the cost (time consuming).

ghost (commented):
Rebased with some added comments/thoughts:
https://github.com/ExchangeUnion/xud/pull/918/files#diff-207d104426c9b9c110cdd1c78f1fc6ecR137
&&
https://github.com/ExchangeUnion/xud/pull/918/files#diff-e2eea4c0c26a2406fc9624a42aa3b06eR110

Also, I'm aware that simulation tests are failing again. I removed the https://github.com/ExchangeUnion/xud/pull/941 commit from this PR as I want to rebase that and get it merged before this one.

sangaman:
Looks like the tests are failing due to lint errors, should be easy enough to fix.

erkarl:
@sangaman interesting - the same warnings that the linter was complaining about made my `npm run compile` fail on my other machine. Works fine on this machine, though.

sangaman:
Fixing the warnings made the compile fail? Or just that the cause of the warnings were compilation errors on one machine but not on the other?

It might make sense to update the linters, it's been a while.

erkarl:
> Fixing the warnings made the compile fail? Or just that the cause of the warnings were compilation errors on one machine but not on the other?

Those warnings presented as compilation errors. I cannot reproduce the situation after a clean `rm -Rf node_modules && npm i && npm run compile && npm run test` - everything passes. Very weird.

lib/swaps/SwapClientManager.ts
erkarl: I'm not completely sold on the idea because of the added overhead of making sure that list stays up to date with swap clients' values. I added a TODO in the code for further consideration.

lib/swaps/SwapClientManager.ts
erkarl: Great idea!

lib/swaps/SwapClientManager.ts
erkarl: Added a TODO for this since it's not a trivial change. We'd also need to readd the swap clients once their state becomes enabled.

test/integration/Service.spec.ts
erkarl: With the new logic querying for a `swapClient` will throw if it's not found. The service spec relies on the the swap client to be available. With the simulation tests in place I'm beginning to question the need to have these kinds of acceptance tests 🤔 .

lib/swaps/Swaps.ts
erkarl: Why do we need to try catch here if the `SwapClientManager.get` will throw a valid error if the swap client doesn't exist?

erkarl:
@sangaman I believe I've now addressed all your comments. I also cherry-picked https://github.com/ExchangeUnion/xud/pull/941 on top of this PR in order to make the simulation tests pass.

After applying changes to `Swaps.ts` I feel like making the `SwapClientManager.get()` to throw upon non-existent swap client might have been a mistake. I'm good with changing it back to `SwapClient || undefined` and handling the non-existent swap clients with the previous logic.

sangaman (commented):
Apologies, I wrote these comments a few days ago but forgot to actually hit submit so they were still pending.

lib/Xud.ts
```diff
@@ -112,32 +89,28 @@ class Xud extends EventEmitter {
...
         pool: this.pool,
         swaps: this.swaps,
         shutdown: this.beginShutdown,
       });
 
-      if (!this.raidenClient.isDisabled()) {
+      if (!this.config.raiden.disable) {
+        // TODO: Should starting of the httpServer be handled as a callback from
+        // Raiden init/start/ready via the manager?
+        this.swapClientManager.get('WETH');
```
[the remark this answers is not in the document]
erkarl: Good catch - the `get` call was not needed. I forgot it in from previous experiment.

lib/Xud.ts
```diff
@@ -112,32 +89,28 @@ class Xud extends EventEmitter {
...
-        lndClients: this.lndClients,
-        raidenClient: this.raidenClient,
+        swapClientManager: this.swapClientManager,
         pool: this.pool,
         swaps: this.swaps,
         shutdown: this.beginShutdown,
       });
 
-      if (!this.raidenClient.isDisabled()) {
+      if (!this.config.raiden.disable) {
```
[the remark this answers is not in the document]
erkarl: We no longer have `this.raidenClient` within the context of `Xud.ts`. I think it makes sense to enable the `httpServer` based on the configuration value of `raiden.disable` because if we're relying on the dynamic `raidenClient.isDisabled()` it could be in a state of trying to reconnect. However, once the reconnection happens we don't have logic to start the `httpServer`.

lib/swaps/Swaps.ts
```diff
@@ -134,8 +134,12 @@ class Swaps extends EventEmitter {
     this.pool.on('packet.swapAccepted', this.handleSwapAccepted);
     this.pool.on('packet.swapComplete', this.handleSwapComplete);
     this.pool.on('packet.swapFailed', this.handleSwapFailed);
-    this.swapClients.forEach((swapClient, currency) => {
-      swapClient.on('htlcAccepted', async (rHash, amount) => {
+    // TODO: Made swapClientManager.swapClients public to temporarily/quickly
```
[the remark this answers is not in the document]
erkarl: I removed the comment from the code and created https://github.com/ExchangeUnion/xud/issues/972 as I don't want to bloat this PR more than necessary.

lib/Xud.ts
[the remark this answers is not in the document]
sangaman: If it's trying to reconnect, it should have the `Disconnected` status instead of `Disabled`. `Disabled` means it's off and staying off, and although currently the only way raiden is disabled is via config, that could change one day (the lnd client can be disabled if we can't load the tls/macaroon files).

Could we expose the raiden status via the swap client manager? We might need it for other purposes down the road as well.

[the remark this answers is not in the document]
erkarl: As discussed in private I ended up exposing the whole `raidenClient` instance and removing `SwapClientManager.getRaidenInfo/getRaidenAddress` as they were redundant.

sangaman (approved):

PR #920, opened 2019-04-27
feat: discover nodes from a specific peer

moshababo:
Discover nodes from a specific peer, apply new connections, and return the list to the client.


lib/service/Service.ts
```diff
@@ -374,6 +377,14 @@ class Service extends EventEmitter {
     return this.orderBook.removePair(pairId);
   }
 
+  /** Removes a trading pair. */
```
sangaman: Comment doesn't match up with the method.

proto/xudrpc.proto
```diff
@@ -534,6 +542,14 @@ message RemovePairRequest {
 }
 message RemovePairResponse {}
 
+message DiscoverNodesRequest {
+    // The node pub key of the peer to discover nodes from.
+    string peer_pub_key = 1 [json_name = "peer_pub_key"];
+}
+message DiscoverNodesResponse {
+    repeated NodeConnectionInfo nodes = 1;
```
sangaman: Conceptually I'm wondering whether we want to actually return all the connection info we get from the peer. I am not sure whether that would be useful.

I'm thinking this call would be used to tell xud to find new peers, but the caller can't actually act on the nodepubkeys and all the addresses we discover, that's all handled internally. Instead we could just return the number of nodes that we discovered, or maybe just a list of pub keys (which can then be used to call getnodeinfo if we want more details). Thoughts @kilrau?

moshababo: The first option would require some infra work, so I went for the latter, since the use-case if limited.

PR #921, opened 2019-04-28
Pairs thresholds

### LINKED ISSUE #163: Trading pairs thresholds
Define **optional** sanity thresholds for each trading pair:
* minPrice
* maxPrice
* minQuantity
* disable thresholds for specific hosts

The usage of these thresholds:
* Check thresholds on `PlaceOrder`. If not, throw meaningful threshold error.
* Drop peer orders which don't meet threshold.
* Exchanges have e.g. minimum quantities in place and could enforce them with these thresholds.

ImmanuelSegol:
Edit: Leaving #163 open to track future enhancements for thresholds.

@sangaman @kilrau The test are failing because of the thresholds I defined. 
Could you please give me default thresholds we should use?
Or should thresholds be disabled by default?

### DIFF
--- a/lib/Config.ts
+++ b/lib/Config.ts
@@ -8,6 +8,7 @@ import { RaidenClientConfig } from './raidenclient/types';
 import { Level } from './Logger';
 import { XuNetwork } from './constants/enums';
 import { PoolConfig } from './p2p/types';
+import { OrderBookThresholds } from './orderbook/types';
 
 class Config {
   public p2p: PoolConfig;
@@ -20,6 +21,7 @@ class Config {
   public http: { port: number };
   public lnd: { [currency: string]: LndClientConfig | undefined } = {};
   public raiden: RaidenClientConfig;
+  public orderthresholds: OrderBookThresholds;
   public webproxy: { port: number, disable: boolean };
   public instanceid = 0;
   /** Whether to intialize a new database with default values. */
@@ -93,6 +95,10 @@ class Config {
       disable: true,
       port: 8080,
     };
+    // TODO: add dynamic max/min price limits
+    this.orderthresholds = {
+      minQuantity: 0, // 0 = disabled
+    };
     this.lnd.BTC = {
       disable: false,
       certpath: path.join(lndDefaultDatadir, 'tls.cert'),
@@ -157,6 +163,12 @@ class Config {
         this.updateMacaroonPaths();
       }
 
+      if (props.thresholds) {
+        this.orderthresholds = {
+          ...this.orderthresholds,
+          ...props.thresholds,
+        };
+      }
       // merge parsed json properties from config file to the default config
       deepMerge(this, props);
     } catch (err) {}
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -14,7 +14,7 @@ import Swaps from '../swaps/Swaps';
 import { SwapRole, SwapFailureReason, SwapPhase, SwapClientType } from '../constants/enums';
 import { CurrencyInstance, PairInstance, CurrencyFactory } from '../db/types';
 import { Pair, OrderIdentifier, OwnOrder, OrderPortion, OwnLimitOrder, PeerOrder, Order, PlaceOrderEvent,
-  PlaceOrderEventType, PlaceOrderResult, OutgoingOrder, OwnMarketOrder, isOwnOrder, IncomingOrder } from './types';
+  PlaceOrderEventType, PlaceOrderResult, OutgoingOrder, OwnMarketOrder, isOwnOrder, IncomingOrder, OrderBookThresholds } from './types';
 import { SwapRequestPacket, SwapFailedPacket } from '../p2p/packets';
 import { SwapSuccess, SwapDeal, SwapFailure } from '../swaps/types';
 // We add the Bluebird import to ts-ignore because it's actually being used.
@@ -71,6 +71,7 @@ class OrderBook extends EventEmitter {
   /** A map of supported trading pair tickers and pair database instances. */
   private pairInstances = new Map<string, PairInstance>();
   private repository: OrderBookRepository;
+  private thresholds: OrderBookThresholds;
   private logger: Logger;
   private nosanityswaps: boolean;
   private nobalancechecks: boolean;
@@ -91,10 +92,11 @@ class OrderBook extends EventEmitter {
     return this.currencyInstances.keys();
   }
 
-  constructor({ logger, models, pool, swaps, nosanityswaps, nobalancechecks, nomatching = false }:
+  constructor({ logger, models, thresholds, pool, swaps, nosanityswaps, nobalancechecks, nomatching = false }:
   {
     logger: Logger,
     models: Models,
+    thresholds: OrderBookThresholds,
     pool: Pool,
     swaps: Swaps,
     nosanityswaps: boolean,
@@ -109,6 +111,7 @@ class OrderBook extends EventEmitter {
     this.nomatching = nomatching;
     this.nosanityswaps = nosanityswaps;
     this.nobalancechecks = nobalancechecks;
+    this.thresholds = thresholds;
 
     this.repository = new OrderBookRepository(models);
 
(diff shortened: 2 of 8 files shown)


kilrau:
Agree, please check the description in #163 @ImmanuelSegol : all thresholds should be strictly optional.

> Another idea I'm just thinking of now is making the thresholds optional but dynamic, based on the current bid/ask prices. E.g, we automatically reject orders that are an order of magnitude or more removed from the current best bid and ask 

I think this makes a lot of sense for `minPrice` & `maxPrice`. Maintaining these indeed can get quite tedious. Thinking about it, it might be a bit complicated for this rule to work when starting with an empty order book, that needs to be figured out, but basically we want to drop orders which are e.g. 200% of current mid market price and e.g. 50% of current mid market price. Not fixed numbers. Ideas how to go about it when starting with an empty book? @sangaman @ImmanuelSegol 

> A peer that floods us with tons of tiny orders is potentially hurting us, but it's conceivable that they don't realize it's spam. Or maybe it's not conceivable really, and we don't advertise our thresholds but we do punish peers that routinely violate them.

Agree on that too, but would just drop orders for now to keep this issue simple.

ImmanuelSegol:
> Agree, please check the description in #163 @ImmanuelSegol : all thresholds should be strictly optional.
> 
> > Another idea I'm just thinking of now is making the thresholds optional but dynamic, based on the current bid/ask prices. E.g, we automatically reject orders that are an order of magnitude or more removed from the current best bid and ask
> 
> I think this makes a lot of sense for `minPrice` & `maxPrice`. Maintaining these indeed can get quite tedious. Thinking about it, it might be a bit complicated for this rule to work when starting with an empty order book, that needs to be figured out, but basically we want to drop orders which are e.g. 200% of current mid market price and e.g. 50% of current mid market price. Not fixed numbers. Ideas how to go about it when starting with an empty book? @sangaman @ImmanuelSegol
> 
> > A peer that floods us with tons of tiny orders is potentially hurting us, but it's conceivable that they don't realize it's spam. Or maybe it's not conceivable really, and we don't advertise our thresholds but we do punish peers that routinely violate them.
> 
> Agree on that too, but would just drop orders for now to keep this issue simple.

@sangaman @kilrau Made them optional. We should open another issue to take care of your suggestion.

kilrau:
Opened new issue for dynamic price thresholds: https://github.com/ExchangeUnion/xud/issues/945

Ready for another review I think @sangaman @erkarl

lib/Config.ts
```diff
@@ -84,6 +86,13 @@ class Config {
       disable: true,
       port: 8080,
     };
+    this.thresholds = {
+      disable: true,
+      disablefor: [],
+      maxprice: 10,
+      minprice: 0,
+      minquantity: 0.1,
```
sangaman: We're representing quantity with integer satoshis internally. `0.1` whole units seems very high for early testing phases, even for production trading it's common to see orders smaller than that, so maybe this should be like 100 or 1000 if enabled.

lib/Config.ts
```diff
@@ -19,6 +20,7 @@ class Config {
   public rpc: { disable: boolean, host: string, port: number };
   public lnd: { [currency: string]: LndClientConfig | undefined } = {};
   public raiden: RaidenClientConfig;
+  public thresholds: OrderBookThresholds;
```
sangaman: The name "thresholds" is vague, maybe "orderthresholds"?

kilrau:
>  I think it'd be wisest to go straight to dynamic thresholds

I agree. Ideas how to do that when the order book is empty or liquidity is thin? We'd have to ensure a certain order book depth and then apply the dynamic threshold on this. Other ideas? @ImmanuelSegol @sangaman

sangaman:
> I agree. Ideas how to do that when the order book is empty or liquidity is thin? We'd have to ensure a certain order book depth and then apply the dynamic threshold on this. Other ideas? @ImmanuelSegol @sangaman

I think we can ignore the threshold when a side of an order book is empty or under X orders (maybe a configurable number?). If we have enough, we throw away orders that are more than a ratio of Y away from the best bid/ask (also configurable, probably). If Y is 10, for instance, and the best bid is 0.1 and the best ask is 0.2, we'd require buy orders to be at least 0.01 and sell orders to be no more than 2.

kilrau:
Since this is moving a bit slow, let's take price out altogether. `minquantity` only, [dynamic prices will be handled after launch](https://github.com/ExchangeUnion/xud/issues/945). Please remove `minprice` and `maxprice` and then let's get this merged @sangaman

kilrau:
Please remind @sangaman to review once it's ready from your side, it's hard to watch a PR just by new commits and assuming it needs to be reviewed @ImmanuelSegol

kilrau:
tests failing, this needs to be fixed @ImmanuelSegol

lib/orderbook/errors.ts
```diff
@@ -13,7 +13,9 @@ const errorCodes = {
   LOCAL_ID_DOES_NOT_EXIST: codesPrefix.concat('.9'),
   QUANTITY_DOES_NOT_MATCH: codesPrefix.concat('.10'),
   CURRENCY_MISSING_ETHEREUM_CONTRACT_ADDRESS: codesPrefix.concat('.11'),
-  INSUFFICIENT_OUTBOUND_BALANCE: codesPrefix.concat('.12'),
+  SWAP_CLIENT_NOT_FOUND: codesPrefix.concat('.12'),
```
sangaman: I believe the tests are failing here because it's changing the error code for an existing error, `INSUFFICIENT_OUTBOUND_BALANCE` which impacts the tests. This should just create two new errors and assign them the 13 and 14 error codes without modifying existing errors.

rsercano:
Tests are passing, changed configuration parameter to lowercase, tested manually working as expected. Are there any changes which you require on this ?

lib/orderbook/OrderBook.ts
```diff
@@ -555,6 +571,15 @@ class OrderBook extends EventEmitter {
    * @returns `false` if it's a duplicated order or with an invalid pair id, otherwise true
    */
   private addPeerOrder = (order: IncomingOrder): boolean => {
+    if (!this.thresholds.disable) {
+      const shouldApplyThreshold = this.thresholds.disableFor.includes(order.peerPubKey);
```
sangaman: The `disableFor` functionality seems unnecessary. It seems very clunky to have to put pub keys of peers in the config file. I would drop this logic altogether.

lib/orderbook/types.ts
```diff
@@ -1,6 +1,12 @@
 import { SwapClientType } from '../constants/enums';
 import { SwapSuccess, SwapFailure } from '../swaps/types';
 
+export type OrderBookThresholds = {
+  disable: boolean;
```
sangaman: Rather than a separate `disable` property, how about we just treat a `minQuantity` of 0 as the threshold being disabled?

test/jest/Orderbook.spec.ts
```diff
@@ -77,6 +77,13 @@ jest.mock('../../lib/Logger');
 jest.mock('../../lib/nodekey/NodeKey');
 const mockedNodeKey = <jest.Mock<NodeKey>><any>NodeKey;
 
+// TODO: Decide default quantity value
+const thresholds: OrderBookThresholds = {
```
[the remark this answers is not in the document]
rsercano: Somehow jest doesn't initialize config constructor, do you know the reason ?

lib/orderbook/OrderBook.ts
```diff
@@ -555,6 +569,14 @@ class OrderBook extends EventEmitter {
    * @returns `false` if it's a duplicated order or with an invalid pair id, otherwise true
    */
   private addPeerOrder = (order: IncomingOrder): boolean => {
+    if (this.thresholds.minQuantity > 0) {
+      if (!this.checkThresholdCompliance(order)) {
+        this.removePeerOrder(order.id, order.pairId, order.peerPubKey, order.quantity);
+        this.logger.debug(`incoming peer order does not compile to thresholds: ${order.id}`);
```
sangaman: "incoming peer order does not comply with configured thresholds"

lib/orderbook/errors.ts
```diff
@@ -14,6 +14,7 @@ const errorCodes = {
   QUANTITY_DOES_NOT_MATCH: codesPrefix.concat('.10'),
   CURRENCY_MISSING_ETHEREUM_CONTRACT_ADDRESS: codesPrefix.concat('.11'),
   INSUFFICIENT_OUTBOUND_BALANCE: codesPrefix.concat('.12'),
+  MIN_QUANTITY_VIOLATED: codesPrefix.concat('.13'),
```
sangaman: We should add this error to the list of errors that result in a `INVALID_ARGUMENT` grpc error in GrpcService.ts.

rsercano:
I'll hopefully today check the points @sangaman mentioned

rsercano:
Added two new test cases

PR #923, opened 2019-04-29
feat(lnd): use lnd 0.6.1 w/ hold invoices

### LINKED ISSUE #798: Switch to "hodl invoice"
https://github.com/lightningnetwork/lnd/pull/2022/ will be the new standard and our https://github.com/lightningnetwork/lnd/pull/1820 won't get merged. Hence we should switch at some point. We'll definitely wait for 2022 to be in master tho.

sangaman:
This is a major PR that changes the way hashes are resolved and swaps are performed with lnd. Here we switch from using a fork of lnd to the main lnd repository.

Previously, we used the hash resolver service which made a resolve gRPC call from lnd to xud when lnd accepted an htlc with a payment hash it did not recognize. Now, when we prepare for a swap xud will tell lnd about the hash via `AddHoldInvoice` and subscribe to updates to that invoice via `SubscribeSingleInvoice`. The subscription will alert xud when lnd accepts an htlc for that invoice, at which point xud will attempt to resolve the hash by completing the swap and provide the preimage to lnd via `SettleInvoice`. If a swap fails, the invoice in lnd is canceled via `CancelInvoice`.

A new SanitySwapAck packet is necessary for this new approach. Previously, the sanity swap payment and the SanitySwap packet (which informs the peer of the hash for the swap would) be sent at the same time. A node that received a resolve request for an amount of 1 satoshi and a hash it did not recognize would wait a short while to see if it received a SanitySwap packet for that hash. With hold invoices, lnd must be informed of the payment hash via `AddInvoice` before it receives the htlc, otherwise it will fail the htlc immediately. Thus, the SanitySwapAck packet is used to inform the peer that it is ready to accept the sanity swap payment.

Closes #798.


kilrau:
牛逼！

kilrau:
As discussed on the call, simnet should be a great to test this

sangaman:
I rebased the PR onto the new tests. Interestingly, I was hitting a strange error where the BTC lnd node would report that it was out of sync on start up during the simulation tests. I figured out two solutions, either adding `WaitForBlockchainSync` calls after we open channels and generate blocks, or change the number of generated blocks from 100 and 10 to 6 (which is enough for a channel to open). I went with the latter since it was less code.

sangaman:
Tests failing on travis now with something I'm not getting locally and it's too late for me to really dig into this now:

```
    xud_test.go:161: lnd-ltc: finished with error (stderr):
        exit status 1
        loadConfig: The mainnet, testnet, and simnet params can't be used together -- choose one of the three
        
    xud_test.go:161: lnd-ltc: finished with error (stderr):
        exit status 1
        loadConfig: The mainnet, testnet, and simnet params can't be used together -- choose one of the three
        
    xud_test.go:161: lnd-ltc: finished with error (stderr):
        exit status 1
        loadConfig: The mainnet, testnet, and simnet params can't be used together -- choose one of the three
        
    xud_test.go:161: lnd-ltc: finished with error (stderr):
        exit status 1
        loadConfig: The mainnet, testnet, and simnet params can't be used together -- choose one of the three
```

If you have any ideas @moshababo let me know, otherwise I'll try figuring it out another time. Either way this is ready to review.

sangaman:
This can be tested with xud-simnet using this branch which has updated install instructions: https://github.com/sangaman/xud-simnet/tree/lnd-0.6.0

lib/swaps/Swaps.ts
```diff
@@ -669,16 +675,39 @@ class Swaps extends EventEmitter {
...
 
       this.setDealPhase(deal, SwapPhase.AmountReceived);
-      return deal.rPreimage;
+      return deal.rPreimage!;
+    }
+  }
+
+  public handleResolveRequest = async (resolveRequest: ResolveRequest): Promise<string> => {
+    const rHash = resolveRequest.getHash();
+    const amount = resolveRequest.getAmount() / 1000;
```
erkarl: We're dividing by 1000 here because of millisatoshis, right?

sangaman: Yup, although this code is probably outdated now and we should confirm the units that we're getting from raiden. But previously yes the resolve request was in millisatoshis.

erkarl:
In hindsight we should also include a check for test lnd/btcd/ltcd commit hashes in order to prevent weird debugging situations. Doesn't have to be in this PR, but we could at least create an issue for it.

erkarl:
@sangaman I temporarily changed the install script to use a forked version of lnd until https://github.com/lightningnetwork/lnd/pull/3045 gets merged.

kilrau:
> In hindsight we should also include a check for test lnd/btcd/ltcd commit hashes in order to prevent weird debugging situations

For SimNet: The install scripts can take care of that in case of a known bug in current master, just like you did with https://github.com/ExchangeUnion/xud/pull/923/commits/b86c3606555997ac9c40428dc4ed60a8071d0de3 . Don't see a difference a test would make!?

For TestNet/Mainnet: We can control it the same way with docker setup scripts. For existing lnd/full nodes we can check the version via `getInfo` calls. Better ideas?

erkarl:
@kilrau what I meant was to lock down the simulation tests' lnd version so that when we change branches locally it forces a rebuild in case the commit hashes don't match.

I've switched lnd back to master since the PR got merged and added code to lock it down to specific commit.

kilrau:
Got it - makes sense. That merge was lightning-fast, good stuff!

lib/swaps/Swaps.ts
```diff
@@ -274,17 +287,30 @@ class Swaps extends EventEmitter {
...
+    const sanitySwapPacket = new packets.SanitySwapPacket({
       currency,
       rHash,
-    }));
+    });
+
+    try {
+      await Promise.all([
+        swapClient.addInvoice(rHash, 1),
+        peer.sendPacket(sanitySwapPacket),
```
moshababo: `AddInvoice` and `SendPacket` should happen in 2 separated steps, to avoid race, no?

test/simulation/install.sh
```diff
@@ -8,28 +8,38 @@ delete_dir() {
...
+$(echo $LND_VERSION | grep -q "$LND_COMMIT_HASH")
+if [ $? == 0 ]
 then
     echo "lnd already installed"
 else
+    echo "deleting temporary gopath directory"
     delete_dir ${temp_gopath}
 
+    echo "getting dep..."
     GO111MODULE=off GOPATH=${temp_gopath} go get -u github.com/golang/dep/cmd/dep
```
moshababo: Do we need `dep` with the new lnd version?

test/simulation/xudtest/node.go
```diff
@@ -54,6 +55,8 @@ func (cfg nodeConfig) genArgs() []string {
 	args = append(args, "--loglevel=debug")
 
 	args = append(args, fmt.Sprintf("--xudir=%v", cfg.DataDir))
+	args = append(args, fmt.Sprintf("--dbpath=%v", cfg.DBPath))
```
moshababo: `cfg.DBPath` isn't initialized with a value. Do we need it?

lib/lndclient/LndClient.ts
```diff
@@ -383,17 +387,80 @@ class LndClient extends BaseClient {
     return this.unaryCall<lndrpc.SendToRouteRequest, lndrpc.SendResponse>('sendToRouteSync', request);
   }
 
+  public addInvoice = async (rHash: string, amount: number) => {
+    const addHoldInvoiceRequest = new lndinvoices.AddHoldInvoiceRequest();
+    addHoldInvoiceRequest.setHash(hexToUint8Array(rHash));
+    addHoldInvoiceRequest.setValue(amount);
+    addHoldInvoiceRequest.setCltvExpiry(this.cltvDelta); // TODO: use peer's cltv delta
```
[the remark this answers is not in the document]
sangaman: Done in #953.

lib/lndclient/LndClient.ts
```diff
@@ -202,18 +212,12 @@ class LndClient extends BaseClient {
...
-  private getPreimageFromSendResponse = (response: lndrpc.SendResponse) => {
-    return Buffer.from(response.getPaymentPreimage_asB64(), 'base64').toString('hex');
-  }
-
   public sendSmallestAmount = async (rHash: string, destination: string) => {
     const sendRequest = new lndrpc.SendRequest();
     sendRequest.setAmt(1);
     sendRequest.setDestString(destination);
     sendRequest.setPaymentHashString(rHash);
+    sendRequest.setFinalCltvDelta(this.cltvDelta);
```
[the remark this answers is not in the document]
sangaman: I believe we want to match the value that the peer used when creating the invoice, which I believe is determined by the peer. I am not 100% sure so maybe let's go over this, but it's probably more of a question for #953 than this PR.

lib/service/Service.ts
```diff
@@ -425,8 +425,8 @@ class Service extends EventEmitter {
   /**
    * resolveHash resolve hash to preimage.
    */
-  public resolveHash = async (request: lndrpc.ResolveRequest) => {
-    return this.swaps.resolveHash(request);
+  public resolveHash = async (request: ResolveRequest) => {
```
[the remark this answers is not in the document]
sangaman: At the time of working on this PR, I thought we might still be using this for Raiden. Since then it seems that we won't, and I'm removing obsolete code/references like this in #940.

lib/swaps/Swaps.ts
sangaman: I didn't think so, since the sanity swap won't actually begin until we receive the sanity swap ack packet. But thinking about it some more, it's indeed a race that could go wrong if the peer adds its invoice and respond to us faster than our add invoice call. So I'll put in some logic to ensure we don't begin the sanity swap until our invoice has been added, but I think we can still start both these calls in parallel (or if that's too complicated, just await these here separately).

test/simulation/xudtest/node.go
sangaman: Not any longer actually, it's a leftover from using `:memory:`. So I'll just remove this.

test/simulation/install.sh
sangaman: I confirmed that we do not, I'll remove. Thanks.

sangaman:
@erkarl's lnd ltc simnet fix is included in the latest 0.6.1 lnd RC tag, so I'll update the install script to use that tag.

lib/swaps/Swaps.ts
sangaman: Actually looking at the code again, I think this is fine. We are awaiting the `addInvoice` call before we begin the swap with `sendSmallestAmount` a few lines down, so there's no risk that we begin the swap before lnd knows about our invoice.

lib/swaps/Swaps.ts
```diff
@@ -503,17 +542,18 @@ class Swaps extends EventEmitter {
...
-      // We checked that we had a swap client for both currencies involved when the swap was initiated. Still...
+    const makerSwapClient = this.swapClients.get(deal.makerCurrency);
+    const takerSwapClient = this.swapClients.get(deal.takerCurrency);
+    if (!makerSwapClient || !takerSwapClient) {
+      // We checked that we had a swap client for both currencies involved during the peer handshake. Still...
       return;
     }
 
-    // TODO: use timeout on call
+    await takerSwapClient.addInvoice(deal.rHash, deal.takerAmount);
```
erkarl: Think you forgot to wrap this with try/catch.

test/simulation/install.sh
```diff
@@ -8,28 +8,34 @@ delete_dir() {
...
+    echo "deleting temporary gopath directory"
     delete_dir ${temp_gopath}
 
-    GO111MODULE=off GOPATH=${temp_gopath} go get -u github.com/golang/dep/cmd/dep
+    echo "getting btcd..."
     GO111MODULE=off go get -u github.com/btcsuite/btcd
 
     echo "starting lnd clone..."
-    if ! git clone -b resolver-cmd+simnet-ltcd https://github.com/ExchangeUnion/lnd.git ${temp_gopath}/src/github.com/lightningnetwork/lnd > /dev/null 2>&1; then
+    if ! git clone -b v0.6.1-beta-rc2 --depth 1 https://github.com/lightningnetwork/lnd ${temp_lndpath} > /dev/null 2>&1; then
```
erkarl: should be able to use `$LND_COMMIT` here

sangaman: Yeah that makes sense. It won't work if we have to go back to using a commit hash instead of a tag, but in that case this line needs to change anyway. I'll go ahead and make this change and rename the variable to `$LND_TAG`.

sangaman:
@erkarl I handled your feedback, and also updated to use the recently tagged `v0.6.1-beta` of lnd rather than the release candidate.

kilrau:
> use the recently tagged `v0.6.1-beta` of lnd rather than the release candidate

dope that this is released already

kilrau:
As discussed, go ahead and merge this. We need this urgently.
* merge `this`
* deploy on seednodes, check that btc sanity swap is working (apart from LTC) on simnet
* review: https://github.com/ExchangeUnion/xud/pull/940 @erkarl

erkarl:
@sangaman needs a rebase -> merge.

PR #925, opened 2019-04-29
feat: pretty cli listpairs

ImmanuelSegol:
#558 
Formates `listpairs` cli response.


lib/utils/utils.ts
```diff
@@ -115,6 +115,18 @@ export const derivePairId = (pair: Pair) => {
   return `${pair.baseCurrency}/${pair.quoteCurrency}`;
 };
 
+/**
+ * Convert ticker symbole into pair.
+ * @param { string } pair - example: btc/ltc
+ */
+export const getPairId = (pair: string): Pair => {
```
sangaman: This functionality is used a couple other places in the code base, and I think it's most concisely expressed simply with:

```javascript
const [baseCurrency, quoteCurrency] = pairId.split('/');
```

So I'd prefer to use that line instead of this function for consistency and conciseness.

lib/cli/commands/listpairs.ts
```diff
@@ -1,11 +1,38 @@
...
+const HEADERS = [
+  colors.red('PairId'),
+  colors.green('Base'),
+  colors.red('Quote'),
+];
+
+const formatPairs = (pairs: ListPairsResponse.AsObject): string[][] => {
+  const formatted: string[][] = [];
+  pairs.pairsList.forEach((pair) => {
+    const pairString = getPairId(pair);
```
sangaman: This variable isn't actually a string, it's an array, so the naming is misleading here. But again I think you can just use `const [baseCurrency, quoteCurrency] = pairId.split('/');` here and then use those two variables in the line below.

lib/cli/commands/listpairs.ts
```diff
@@ -1,11 +1,38 @@
...
-import { ListPairsRequest } from '../../proto/xudrpc_pb';
+import { ListPairsRequest, ListPairsResponse } from '../../proto/xudrpc_pb';
+import Table, { HorizontalTable } from 'cli-table3';
+import colors from 'colors/safe';
+import { getPairId } from '../../utils/utils';
+
+const HEADERS = [
+  colors.red('PairId'),
+  colors.green('Base'),
+  colors.red('Quote'),
```
sangaman: These colors in a pattern feels a little too "christmasy" to me. I know we're using red/green on the orders table but it makes more sense there where red is "sell" and green is "buy." Personally I'd stick to a single color for all the headers here, not sure which would look nicest, maybe blue which is also sort of the exchange union color.

erkarl:
Looks good. 

@sangaman @ImmanuelSegol
Minor nit. I'm wondering if it makes sense to display the `Base` and `Quote` in a separate column? We're also displaying the same information in the `PairId` column. What do you think about renaming it to `Pair Base/Quote` and losing the `Base` and `Quote` columns?

kilrau:
> What do you think about renaming it to Pair Base/Quote and losing the Base and Quote columns?

`Pair (Base/Quote)` 👍 @ImmanuelSegol @erkarl

lib/cli/commands/listpairs.ts
```diff
@@ -1,11 +1,34 @@
 import { callback, loadXudClient } from '../command';
 import { Arguments } from 'yargs';
-import { ListPairsRequest } from '../../proto/xudrpc_pb';
+import { ListPairsRequest, ListPairsResponse } from '../../proto/xudrpc_pb';
+import Table, { HorizontalTable } from 'cli-table3';
+import colors from 'colors/safe';
+import { getPairId } from '../../utils/utils';
+
+const HEADERS = [colors.red('PairId')];
```
sangaman: This is subjective and minor, but I would propose a different color for our standard header rather than the red that we use for the "sell" columns elsewhere. Again I'd suggest blue or dark blue, but I'm not too picky.

lib/cli/commands/listpairs.ts
```diff
@@ -1,11 +1,34 @@
...
+
+const formatPairs = (pairs: ListPairsResponse.AsObject): string[][] => {
+  const formatted: string[][] = [];
+  pairs.pairsList.forEach((pair) => {
+    const { baseCurrency, quoteCurrency } = getPairId(pair);
+    formatted.push([`${baseCurrency} / ${quoteCurrency}`]);
```
sangaman: It looks like all we are doing here after collapsing this into one column is putting a space before and after the `/` in the pair id, which seems unnecessary. I think we can just put the `pairId` in the output unformatted and keep it simple.

lib/utils/utils.ts
sangaman: The name of this function is confusing, because it accepts what we refer to everywhere else as `pairId` but the name suggests that it returns `pairId`. I'm still of the opinion that we don't need this function since it's basically just an alias for the `split` function, but if we do want it in the future it should be called something like `getPairCurrencies`.

lib/utils/utils.ts
```diff
@@ -115,6 +115,18 @@ export const derivePairId = (pair: Pair) => {
   return `${pair.baseCurrency}/${pair.quoteCurrency}`;
 };
 
+/**
+ * Convert ticker symbole into pair.
+ * @param { string } pair - example: btc/ltc
+ */
+export const getPairCurrencies = (pair: string): Pair => {
```
sangaman: As I've said in previous comments I don't think we need this function. @erkarl what are your thoughts?

If we do keep it I think it can be more concise to match how it's currently being done:

```javascript
const [baseCurrency, quoteCurrency] = pairId.split('/');
return { baseCurrency, quoteCurrency };
```

lib/utils/utils.ts
```diff
@@ -115,6 +115,18 @@ export const derivePairId = (pair: Pair) => {
   return `${pair.baseCurrency}/${pair.quoteCurrency}`;
 };
 
+/**
+ * Convert ticker symbole into pair.
```
sangaman: If we keep this function, this comment has a typo and should say something like "Derives the base and quote currencies from a trading pair ticker."

lib/cli/commands/listpairs.ts
sangaman: You marked this resolved but I don't think anything changed and you didn't leave a comment. What are your thoughts? I still don't see the point of this and in looking around it seems most currency trading pair symbols are expressed without these spaces, eg `EUR/USD`.

In the future please don't mark as resolved making a change or leaving a comment for why you've resolved, it makes it harder to re-review a PR.

lib/cli/commands/listpairs.ts
```diff
@@ -1,11 +1,34 @@
 import { callback, loadXudClient } from '../command';
 import { Arguments } from 'yargs';
-import { ListPairsRequest } from '../../proto/xudrpc_pb';
+import { ListPairsRequest, ListPairsResponse } from '../../proto/xudrpc_pb';
+import Table, { HorizontalTable } from 'cli-table3';
+import colors from 'colors/safe';
+import { getPairCurrencies } from '../../utils/utils';
+
+const HEADERS = [colors.blue('PairId')];
```
sangaman: One more thing, `PairId` is something of an internal term. For end-user purposes I think we can call this `Trading Pair` or something more descriptive like that.

lib/utils/utils.ts
erkarl: I think the code would read better without this function.

ImmanuelSegol: @sangaman I removed it.
 I just don't agree tho we use this same function multiple times throughout the code base. and `getPairCurrencies` is clear to me at least

lib/utils/utils.ts
ImmanuelSegol: We will remove it.

lib/cli/commands/listpairs.ts
ImmanuelSegol: Agree! thanks for the tip.

kilrau:
LGTM

```
kilrau@K-Yoga:~/xud/bin$ ./xucli addpair btc ltc
success
kilrau@K-Yoga:~/xud/bin$ ./xucli listpairs
Trading Pairs:
┌──────────────┐
│ Trading pair │
├──────────────┤
│ LTC / BTC    │
├──────────────┤
│ BTC / LTC    │
```

Issue #643, opened 2018-11-06
Listorders CLI output - peerpubkey


sangaman:
After #600, our `getorders` CLI output looks like this:

![image](https://user-images.githubusercontent.com/3440835/48082856-9e906100-e1c1-11e8-82f9-45a06ad67666.png)

I'm thinking it would be nice to show which peer the owner belongs to. We could probably repurpose the `Own Order` column for this. Maybe change the label to something like "Owner" or "Peer" and print the node's peer pub key for peer orders, and "Own" or "Local" for own orders. It may even be worth taking it a step further and printing the peer's alias if one exists - however we don't have that data currently available from the `getorders` rpc call. If we think it's worth it, we could make cli command also call `listpeers` to get all the aliases and then map them to the pubkeys.

Thoughts @erkarl @kilrau @offerm?

offerm:
@sangaman Initially, I thought that this is a great idea.  Problems that I see:
- do we have xud node alias support? 
- will these aliases be used by the peers? can you trust the alias? will it give valuable information
- do you have room to show it?
- If you show host name if we have it
- not sure if IP address provides value.

After thinking about it more, and having a close look at the table, I think that:
a. we need to sort the buy/sell orders according to prices.
b. We need to aggregate the capacity available at each price level (rounded). For example, in the table above we have 2 BUY orders of at price of 0.87.  These should be combined.
c. once you combine and sort properly you will be able to see the market depth. If we will not aggregate but have a lot of orders this display will not be usable. 
d. I would consider to have this as a new cli `getmarketdept` and leave `getorders` as today.
e. to be considered - aggregate the buy/sell orders into n (say 8) buy buckets and n sell bucket and presenting them over the price line. So sell orders will come first (higher prices) and buy order will come below it (normal view of order book).

With a-e, I don't see a need for `owner` column.

erkarl:
I'm really hesitant in regards to showing aliases. If they are like lnd's aliases they can't be trusted and just cause confusion.

a-b. Yes, definitely. That's something I left out of the original PR intentionally.
c-e. I imagine `getorders` with the current implementation will not be usuable in any production like environment because of too many orders. Maybe the default of `getorders` should be to return the market depth/aggregated results.

sangaman:
Ok, maybe leave the alias aside for now, it's true that it's not implemented (but I'm thinking it would be nice to have locally specified aliases, made a comment on that issue).

[...]

Issue #486, opened 2018-09-16
CLI - addAlias to peer


offerm:
When having multiple peers it is hard to know which peer object represents which real entity (exchange). 

Suggest adding an optional alias to a peer that can help to identify it. 

I see two options:
a. the alias is set by the peer itself and provided as part of the handshake, or
b. the alias is set by the node for a connected peer

Option a is what LND is doing. This, however, is risky since the peer may provide a wrong alias and claims that he is not who he is. setting by the node is more cumbersome but more secured.

Another option to explore is using a secured P2P channel and taking the information from the client certificate.

michael1011:
I would go for option `a` and do it similar to LND. Because when the alias isn't used as identifier of the node there is no risk involved imo.

kilrau:
Good idea. I would not go for self-set alias (the LND way) because of the reason you mentioned. We've all seen the "Google uses Lightning" tweets. Also don't see how the client certificate solves this?

I'd strongly vouch for B and maybe even a generated one as default (google docs "anonymous squirrel" style).

sangaman:
Yeah after thinking of this some I am leaning towards option B, there is little downside there and it seems like it could be a nice feature.

kilrau:
Some more thoughts on option b:
* we can use a list of funny animal names and assign it to it
* drawback: when debugging/communicating with others one and the same peer will have different names so one had to fall back to the pubKey again.

Thoughts from a fresh set of eyes? @erkarl

erkarl:
I'd also go for the option B because I saw how much unnecessary confusion it first created with lightning clients.

kilrau:
Actually I'd like to hold off on this feature. Reasons:
* manually assigning an alias won't be a crucial feature
* an automated assigned alias is complicated
* both approaches have the drawback of being a "local point of view" - the assigned alias for a given peer will be different on other nodes
========================
it's a debug feature and not reason enough to do the effort imo

sangaman:
Should we leave it open tho even if having it low priority? I don't think of it as a debug feature necessarily (for debugging purposes, I have no problem using hex pub keys), but as a way to make it more easily to identify peers at a glance. It's certainly superficial and not urgent in the least, but still something that seems like it's worth tracking.

hatmer:
How about deterministically generating an easy-to-identify name for the peer using the pub key hex? This way it is both easy to read and consistent across all nodes. If this sounds good please assign me since I have a simple algorithm ready.

[...]

PR #926, opened 2019-04-29
feat(cli): enhance listpeers output

ImmanuelSegol:
#558 
Formats cli response for `listpeers`.

### DIFF
--- a/lib/cli/commands/listpeers.ts
+++ b/lib/cli/commands/listpeers.ts
@@ -1,11 +1,85 @@
-import { callback, loadXudClient } from '../command';
+import Table, { HorizontalTable } from 'cli-table3';
+import colors from 'colors/safe';
 import { Arguments } from 'yargs';
-import { ListPeersRequest } from '../../proto/xudrpc_pb';
+import { ListPeersRequest, ListPeersResponse, Peer } from '../../proto/xudrpc_pb';
+import { callback, loadXudClient } from '../command';
+import { generateHeaders } from '../utils';
+
+const HEADERS = [
+  'Peer',
+  'Pairs',
+  'Details',
+];
+
+const createTable = () => {
+  const table = new Table({
+    head: generateHeaders(HEADERS),
+  }) as HorizontalTable;
+  return table;
+};
+
+const trimPubKey = (key: string) => {
+  if (key.length <= 0) {
+    return '';
+  }
+  return `${key.slice(0, 10)}...${key.slice(key.length - 10)}`;
+};
+
+const formatPairList = (pairs: string[]) => {
+  let pairString = '';
+  pairs.forEach((pair) => {
+    /* eslint disable-next-line */
+    pairString = `${pairString}${pairString ? '\n' : ''}${pair}`;
+  });
+  return pairString;
+};
+
+const formatLndPubKeys = (lndKeys: string[][]) => {
+  let str = '';
+  lndKeys.forEach((client) => {
+    /* eslint disable-next-line */
+    str =  `${str}${str ? '\n' : ''}${client[0]} lnd key: ${trimPubKey(client[1])}`;
+  });
+  return str;
+};
+
+const formatPeers = (peers: ListPeersResponse.AsObject) => {
+  const formattedPeers: string[][] = [];
+  peers.peersList.forEach((peer: Peer.AsObject) => {
+
+    const address = `${peer.nodePubKey}
+@${peer.address}` ;
+
+    const details = [
+      `Alias:
+
+Address ([node_key]@[host]:[port]):
+${address}`,
+      formatPairList(peer.pairsList),
+      `inbound: ${peer.inbound}\
+\nversion: ${peer.xudVersion}\
+\ntime connected: ${peer.secondsConnected.toString()} seconds\
+\n${formatLndPubKeys(peer.lndPubKeysMap)}\
+${peer.raidenAddress ? `\nraiden address: ${trimPubKey(peer.raidenAddress)}` : ''}`,
+    ];
+    formattedPeers.push(details);
+  });
+  return formattedPeers;
+};
+
+const displayTables = (peers: ListPeersResponse.AsObject) => {
+  const table = createTable();
+  formatPeers(peers).forEach((peer) => {
+    table.push(peer);
+  });
+  console.log(colors.underline(colors.bold('\nPeers:')));
+  console.log(table.toString());
+};
 
 export const command = 'listpeers';
 
 export const describe = 'list connected peers';
 
 export const handler = (argv: Arguments) => {
-  loadXudClient(argv).listPeers(new ListPeersRequest(), callback(argv));
+  loadXudClient(argv).listPeers(new ListPeersRequest(), callback(argv, displayTables));
 };
--- a/lib/cli/utils.ts
+++ b/lib/cli/utils.ts
@@ -1,3 +1,4 @@
+import colors from 'colors/safe';
 import os from 'os';
 import path from 'path';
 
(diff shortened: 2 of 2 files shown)


kilrau:
> 1. abbreviated pubKey (`023ec5...`)

This one. [Alias' are not a prio right now](https://github.com/ExchangeUnion/xud/issues/486). @ImmanuelSegol please make sure to always use the same way for abbreviation ob pubkeys: the first 6 digits. And `xud` pubkey only. I adjusted comments in https://github.com/ExchangeUnion/xud/issues/643 to "the first 6 digits" @erkarl 

> list of active pairs

👍

lib/cli/commands/listpeers.ts
```diff
@@ -1,11 +1,55 @@
...
+  return table;
+};
+
+const formatPeers = (peers: ListPeersResponse.AsObject) => {
+  const formattedPeers: string[][] = [];
+  peers.peersList.forEach((peer: any) => {
+    const peerInfo: string[] = [];
+    Object.keys(peer).forEach((key) => {
+      if (key === 'pairsList') {
+        colors.green(peer[key]);
```
sangaman: Is there a reason pairslist is being made green? It's not getting pushed to the array in this case, which I believe is the cause of the off by one error karl mentions above.

lib/cli/commands/listpeers.ts
```diff
@@ -1,11 +1,55 @@
...
+  colors.red('node public key'),
+  colors.red('lnd public keys'),
+  colors.red('inbound'),
+  colors.red('pairs list'),
+  colors.red('xud version'),
+  colors.red('seconds connected'),
+  colors.red('raiden address'),
+];
+
+const createTable = () => {
```
sangaman: I'm noticing this function/code in a couple of your cli-related PRs, maybe make this a cli utils function that accepts the headers as an argument?

ImmanuelSegol:
@sangaman @kilrau For some reason the tests are failing, they aren't failing on my local machine and im not able to produce the error Travis is showing on my local machine.

erkarl:
@ImmanuelSegol I get the same (failing) output when running:
`➜  xud git:(pretty-cli-listpeers) ✗ rm -Rf node_modules && npm i && npm run compile`

https://travis-ci.org/ExchangeUnion/xud/builds/532961287#L779

sangaman:
@ImmanuelSegol Can you rebase this on master? It's using an older version of the p2p protocol so I can't connect to updated peers to see the output properly.

That said, the output still looks cramped in a default size terminal for me:

image

Maybe we should keep the `public key`, `address`, and maybe the `pairs` columns and then use an additional column called something like `details` that lists all the other key info line by line, not in json but in a format like:

```
inbound: true
version: 1.0.0-alpha.11
uptime: 3d 5h 17m 53s
BTC lnd pubkey: 03b483...
LTC lnd pubkey: 02a832...
Raiden address: 0x2398c...
```
I figure we're likely to add more fields to the list peers response in the future so this seems like a more flexible approach.

Also minor suggestion, but I think it makes sense to have peers' public keys in the leftmost column, since that their true identifier. Address can be second

lib/cli/commands/listpeers.ts
```diff
@@ -1,11 +1,74 @@
...
 import { Arguments } from 'yargs';
-import { ListPeersRequest } from '../../proto/xudrpc_pb';
+import colors from 'colors/safe';
+import Table, { HorizontalTable } from 'cli-table3';
+import { generateHeaders } from '../utils';
+import { ListPeersRequest, ListPeersResponse, Peer } from '../../proto/xudrpc_pb';
+
+const HEADERS = [
+  'address',
+  'public key',
```
sangaman: Could we call this `node key` actually? I'm thinking it will further differentiate it from the lnd pub keys.

lib/cli/commands/listpeers.ts
```diff
@@ -52,12 +48,9 @@ const formatPeers = (peers: ListPeersResponse.AsObject) => {
...
       peer.address,
       trimPubKey(peer.nodePubKey),
-      `${peer.inbound}`,
-      peer.xudVersion,
-      `${peer.secondsConnected.toString()} seconds`,
       formatPairList(peer.pairsList),
-      formatLndPubKeys(peer.lndPubKeysMap),
-      trimPubKey(peer.raidenAddress),
+      // tslint:disable-next-line:max-line-length
+      `inbound: ${peer.inbound}\nversion: ${peer.xudVersion}\ntime connected: ${peer.secondsConnected.toString()} seconds\nlnd keys: ${formatLndPubKeys(peer.lndPubKeysMap)}\nraiden address: ${trimPubKey(peer.raidenAddress)}`,
```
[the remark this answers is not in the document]
sangaman: Two more thoughts.

1. Can we put each lnd key on a separate line? Like

```
BTC lnd pubkey: 02d693...
LTC lnd pubkey: 02572a...
```

2. The `raiden address` line has some space, and of the 6 characters we show before cutting off, the first two are `0x` leaving only 4 unique characters. I think it would be better to show 8 (or maybe even 10) total characters for the raiden address, including the `0x`.

lib/cli/commands/listpeers.ts
```diff
@@ -1,11 +1,74 @@
...
-import { ListPeersRequest } from '../../proto/xudrpc_pb';
+import colors from 'colors/safe';
+import Table, { HorizontalTable } from 'cli-table3';
+import { generateHeaders } from '../utils';
+import { ListPeersRequest, ListPeersResponse, Peer } from '../../proto/xudrpc_pb';
+
+const HEADERS = [
+  'address',
+  'public key',
+  'pairs list',
```
sangaman: I don't think we actually need "list" here. Just "pairs" is more concise and still descriptive enough I think.

ImmanuelSegol:
@kilrau these tests are not event launching.
Locally they pass.

package-lock.json
```diff
@@ -4379,6 +4379,492 @@
       "requires": {
         "graceful-fs": "^4.1.11",
         "through2": "^2.0.3"
+<<<<<<< HEAD
```
sangaman: Merge issue here is causing the tests to fail, I think you can just discard the changes to this file.

rsercano:
Merged master into it, seems to be working for me.

ImmanuelSegol:
@kilrau This can be merged.

sangaman:
Open to thoughts on how this should display since it's mainly a matter of preference and aesthetics @erkarl @kilrau.

rsercano:
ping @kilrau

kilrau:
> What I had in mind was combining all the columns to the right of `details` into the details column and listing them vertically like so:
> details
> inbound: false
> time connected: 8 seconds
> BTC lnd key: 2823ca2...
> LTC lnd key: 27ab262...

Yes, that seems to be the best way.
 
> I'm also thinking it makes sense not to display "raiden address" at all if it doesn't have a value.

Yes, the rule should be to not display anything without value.

Additionally we should always use the same way to abbreviate the node key. Either 123456... or `1234...n-3n-2n-1n`. Whereas I prefer the latter. Would be nice to straighten that out throughout all your PRs.

rsercano:
@kilrau 
image

rsercano:
image

done, @kilrau

kilrau:
> * Rename the "Address" column to something like "Peer", since we're not only showing the address but also the node pubkey.

Good idea, agree. @rsercano Let's rename the column title to `Peer`.

> * Get rid of the "Alias" column and add the alias (if a non-empty/null alias exists) under the "Peer" column at the top in its own line like `Alias: XU Seed Node #1`. I think total table width would be an issue with Alias in its own column, whereas we already have spare room vertically. I also think that the alias makes sense to go with the other identifies like nodepubkey and IP address/hostname. Lastly I figure it's the first thing we'd want to see.

Good idea, agree. @rsercano Let's get rid of the separate `Alias` column and format the second `Peer` column to contain both:
```
Alias:
happy-rabbit
Address ([node_key]@[host]:[port]):
abc123
@xyz.onion:8885
```

I'd like to establish the term "Address" as the full URI, including node key, that is needed in order to connect to another node. This is in line with https://github.com/ExchangeUnion/xud/pull/1170 - here we also show our own "address" as `[node_key]@[host]:[port]`.

rsercano:
done guys 
image

PR #928, opened 2019-04-29
chore: incremental builds & typescript 3.4

sangaman:
This upgrades typescript to 3.4 to use its incremental build feature. This will speed up build times both locally and on travis by caching the dist directory and .tsbuildinfo file which is necessary for incremental compilation. It also makes the precompile script which deleted the dist folder obsolete.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ typings/
 
 # Typescript compilation output
 dist/
+.tsbuildinfo
 
 # Optional npm cache directory
 .npm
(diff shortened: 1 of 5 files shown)


sangaman:
I'm actually thinking it may not make sense to cache the `dist` folder in travis since it will change with virtually every build, and it looks like repacking and adding a cache takes more time than is saved on the compile step. It'd still be useful to have for local dev though, especially with simulation tests requiring a compile each time.

PR #937, opened 2019-05-02
fix: switch swap capacity check currency

### LINKED ISSUE #932: Order placement verification
I'm trying to place a buy and sell order for WETH/BTC. I see at least two issues here:
```
2019/05/01 05:51:27 [DEBUG] Placing order: {0.0305 350000000 WETH/BTC  BUY} 
2019/05/01 05:51:27 [ERROR] Could not place orders: {0.0305 350000000 0 WETH/BTC} - rpc error: code = Unknown desc = WETH outbound balance of 0 is not sufficient for order amount of 35
00000000000000000
```
- Since this is a BUY order, I would like to get WETH and pay with BTC. For that I don't need to have WETH outbound balance.
- This order should be prevented since I don't have any Raiden channel. 

```
2019/05/01 05:51:27 [DEBUG] Placing order: {0.0315 350000000 WETH/BTC  SELL} 
2019/05/01 05:51:27 [DEBUG] Order added: price:0.0315 quantity:350000000 pair_id:"WETH/BTC" id:"29384370-6bd5-11e9-b365-0db8c15a4f61" local_id:"29384370-6bd5-11e9-b365-0db8c15a4f61" cr
eated_at:1556689887527 side:SELL is_own_order:true 
```
- for this order I do need WETH balance (which I don't have).
- Should not be able to open this order as I don't have raiden channel.

Was done on xud1

sangaman:
This fixes a bug where the outbound capacity sanity check was comparing against the inbound currency and amount rather than outbound. In addition, this commit refactors & renames the methods used to calculate the amounts and currencies involved in a swap to minimize potential confusion. It also adds tests to detect a regression in the calculation logic.

Fixes #932.

### DIFF
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -59,20 +59,6 @@ class Swaps extends EventEmitter {
   }
 
   /**
-   * Derives the maker and taker currency for a swap.
-   * @param pairId The trading pair id for the swap
-   * @param isBuy Whether the maker order in the swap is a buy
-   * @returns An object with the derived `makerCurrency` and `takerCurrency` values
-   */
-  public static deriveCurrencies = (pairId: string, isBuy: boolean) => {
-    const [baseCurrency, quoteCurrency] = pairId.split('/');
-
-    const makerCurrency = isBuy ? baseCurrency : quoteCurrency;
-    const takerCurrency = isBuy ? quoteCurrency : baseCurrency;
-    return { makerCurrency, takerCurrency };
-  }
-
-  /**
    * Checks if a swap request is valid. This is a shallow check that only detects critical
    * inconsistencies and verifies only whether the request can possibly lead to a successful swap.
    * @returns `true` if the request is valid, otherwise `false`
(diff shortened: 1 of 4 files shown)


sangaman:
Note: I had to update the simulation tests as well to get them to pass, but the failure is a sign that this test works. Previously the remove order test was checking the amount of BTC being sent against the LTC capacity which was ok, but testing against the BTC capacity failed, so I had to reduce the quantity (and the price to something more realistic for the LTC/BTC pair).

lib/swaps/Swaps.ts
```diff
@@ -84,19 +70,40 @@ class Swaps extends EventEmitter {
...
+  }
+
+  /**
+   * Calculates the incoming and outgoing currencies and amounts of subunits/satoshis for an order if it is swapped.
+   * @param quantity The quantity of the order
+   * @param price The price of the order
+   * @param isBuy Whether the order is a buy
+   * @returns An object with the calculated incoming and outgoing values.
+   */
+  public static calculateInboundOutboundAmounts = (quantity: number, price: number, isBuy: boolean, pairId: string) => {
```
erkarl: I really like that it's returning `inboundAmount/outboundAmount` now. Much more clearer than `makerAmount/takerAmount`.

Issue #934, opened 2019-05-01
Sanity swap - Raiden


offerm:
The payment request to raiden should include a secrethash having length of 66 chars. 

Should include 0x at the beginning.

Raiden log:
```
{
  "node": "cdb6238c",
  "registry_address": "0x81E09B9Ba95c9D1fcC6Be59D7798eFB8E1EF342E",
  "token_address": "0x4c354C76d5f73A63a90Be776897DC81Fb6238772",
  "target_address": "0x7ed0299Fa1ADA71D10536B866231D447cDFa48b9",
  "amount": 1,
  "payment_identifier": 5321335948005626,
  "secret": null,
  "secret_hash": "6a47a3f0b2fa906a4c3b0bda5261e7b1ff8859546672858fcebf6e85ab6b4245",
  "event": "Initiating payment",
  "logger": "raiden.api.rest",
  "level": "debug",
  "timestamp": "2019-05-01 09:38:23.536048"
}
{
  "errors": "secret_hash length should be 66.",
  "status_code": 409,
  "event": "Error processing request",
  "logger": "raiden.api.rest",
  "level": "error",
  "timestamp": "2019-05-01 09:38:23.536454"
}
```

offerm:
also, The amount here is `1 WETH` ?  this is very high.

offerm:
Another comment here: 
The raiden code is not very mature with regards to error handling.  When things don't work well they just hang until expiration which can be 40-50 blocks away. This can take half an hour. 

To be on the safe side, before we attempt a sanity swap we should check:
- do we have a direct channel with the peer?
- do we have outgoing capacity on the channel?
- does the peer have outgoing capacity on its end? [This may not be needed but the peer will fail the channel anyway]

Only if the above is successful we should go to the sanity swap.

sangaman:
I take it then that we should just be preprending "0x" to all `secret_hash` values sent to Raiden?

>  The amount here is 1 WETH ? this is very high.

Oh if that's the case then, I totally misunderstood the documentation then, based on the description and example here 
https://raiden-network.readthedocs.io/en/stable/rest_api.html#payments. The example used an `amount` of 200, too much to be denominated in WETH I thought, and I don't see any examples of fractional amounts. With sanity swaps I'm trying to send the smallest unit supported, gwei in this case.

> The raiden code is not very mature with regards to error handling. When things don't work well they just hang until expiration which can be 40-50 blocks away. This can take half an hour.
> 
[...]

PR #938, opened 2019-05-02
fix(swaps): prepend 0x to Raiden's sanity swap rHash

ghost:
As mentioned in https://github.com/ExchangeUnion/xud/issues/934 we should prepend "0x" to the `secret_hash` values sent to Raiden. This PR deals only with sanity swap's rHash in order to not block @offerm - will deal with regular `sendPayment` rHash in a separate PR.

### DIFF
--- a/lib/raidenclient/RaidenClient.ts
+++ b/lib/raidenclient/RaidenClient.ts
@@ -7,6 +7,7 @@ import { SwapClient, SwapState, SwapRole } from '../constants/enums';
 import assert from 'assert';
 import OrderBookRepository from '../orderbook/OrderBookRepository';
 import { Models } from '../db/DB';
+import { RaidenClientConfig, RaidenInfo, OpenChannelPayload, Channel, ChannelEvent, TokenPaymentRequest, TokenPaymentResponse } from './types';
 
 /**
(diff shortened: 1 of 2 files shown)


lib/raidenclient/RaidenClient.ts
```diff
@@ -157,7 +157,7 @@ class RaidenClient extends BaseClient {
       throw(errors.TOKEN_ADDRESS_NOT_FOUND);
     }
 
-    const tokenPaymentResponse = await this.tokenPayment(tokenAddress, destination, 1, rHash);
+    const tokenPaymentResponse = await this.tokenPayment(tokenAddress, destination, 1, `0x${rHash}`);
```
sangaman: I assume that we'll always want to prepend 0x when we make a token payment, so maybe we should put this in the `tokenPayment` method itself, line 381 I think. Also I just noticed we're using `Object.assign` there, it might be nice to create a `tokenPaymentRequest` type with a nullable `secret_hash` field so we can just do `payload.secret_hash =  `0x${secret_hash};`.

erkarl: Thanks for the feedback. Great idea - updated.

sangaman (approved):

PR #940, opened 2019-05-03
Raiden resolver http endpoint

### LINKED ISSUE #931: Rest endpoint for Raiden resolve request
In order for Raiden to query for external hash from xud we need a rest endpoint.

@sangaman proposed to not go through the gRPC API since once we move to hold invoices we can deprecate that call.

sangaman:
This is a WIP for an http endpoint that accepts raiden hash resolver requests via JSON over http. I based this off the raiden code at https://github.com/raiden-network/raiden/blob/master/raiden/network/resolver/client.py.

This branch is based off of the hold invoices PR since there is a lot of overlap and potential conflict there.

For now I hardcoded the port to listen on at 8887 and the host as 127.0.0.1. I figure at least the port should be configurable, I just wasn't sure where to put the config option for it. For now the HTTP server is only enabled when raiden is enabled, since that's it's only current purpose, although I tried to structure the http server code in at least a somewhat generic way. Therefore it wouldn't be a big rework if we want to accept resolve requests over http from a different client with a different request/response structure, for instance.

I still need to test this but it's at least ready for a conceptual review. The code changes for the PR as a whole include a lot of hold invoices code that's out of scope here, so you can just look at the last commit or two.

Closes #931.


erkarl:
Conceptually OK.

kilrau:
> enabled when raiden is enabled, since that's it's only current purpose

Sounds good!

> Therefore it wouldn't be a big rework if we want to accept resolve requests over http from a different client with a different request/response structure, for instance.

Well we can't really know how this structure might look like if they this other client doesn't design it according to what we have, so also fine!

offerm:
@sangaman is this ready for testing? Do we need  first the hold invoices?

sangaman:
@offerm @erkarl This branch is ready for testing and deeper review. I haven't tested the whole flow yet, but I've added unit tests and done some basic manual tests.

It still includes the commits from the hold invoices PR which aren't yet in master, so only the last two commits are truly relevant here.

lib/http/HttpServer.ts
```diff
@@ -0,0 +1,108 @@
...
+   * Starts the server and begins listening on the provided port.
+   * @returns true if the server started listening successfully, false otherwise
+   */
+  public listen = async (port: number) => {
+    return new Promise<void>((resolve, reject) => {
+      const listenErrHandler = (err: Error) => {
+        reject(err);
+      };
+
+      this.server.listen(port, '127.0.0.1').once('listening', () => {
```
erkarl: Once this is merged we should also let Offer know that it's not possible to specify the listening interface. I'm assuming this is because there is no authentication.

test/unit/HttpServer.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+      .then((res: ChaiHttp.Response) => {
+        expect(res.status).to.equal(404);
+        done();
+      }).catch((reason) => {
+        done(reason);
+      });
+  });
+
+  after(async () => {
+    await httpServer.close();
```
erkarl: Minor nit: `sandbox.restore()`

test/unit/HttpServer.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+import sinon from 'sinon';
+import Service from '../../lib/service/Service';
+import { ResolveRequest } from '../../lib/swaps/types';
+
+chai.use(chaiAsPromised);
+chai.use(chaiHttp);
+
+const loggers = Logger.createLoggers(Level.Warn);
+
+describe('HttpServer', () => {
```
erkarl: Nice tests :) 👍

lib/http/HttpServer.ts
sangaman: Yes, due to no authentication. Longer term we may want to add some authentication here as well.

erkarl:
Uh, forgot to mention. Update `sample-xud.conf` via `npm run config`.

sangaman:
> Uh, forgot to mention. Update `sample-xud.conf` via `npm run config`.

This will happen automatically the next time we bump versions.

PR #941, opened 2019-05-05
fix(tests): make testNetworkInit wait for synced chains

ghost:
In this PR we change `testNetworkInit` simulation test to wait for `lnd-ltc` and `lnd-btc` to be synced to the chain before proceeding. I also increased the test timeout to 30 seconds as I was occasionally reaching the 5 second limit.

### DIFF
--- a/test/simulation/tests.go
+++ b/test/simulation/tests.go
@@ -21,26 +21,29 @@ import (
 // by their preceding ones.
 func testNetworkInit(net *xudtest.NetworkHarness, ht *harnessTest) {
 	for _, node := range net.ActiveNodes {
-
 		// Verify connectivity.
-		req := &xudrpc.GetInfoRequest{}
-		res, err := node.Client.GetInfo(ht.ctx, req)
-		ht.assert.NoError(err)
-		ht.assert.NotNil(res.Lnd["BTC"])
-		ht.assert.NotNil(res.Lnd["LTC"])
-		ht.assert.Len(res.Lnd["BTC"].Chains, 1)
-		ht.assert.Equal(res.Lnd["BTC"].Chains[0].Chain, "bitcoin")
-		ht.assert.Equal(res.Lnd["BTC"].Chains[0].Network, "simnet")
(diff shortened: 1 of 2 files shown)


test/simulation/tests.go
```diff
@@ -21,23 +21,23 @@ import (
...
-		ht.assert.Equal(res.Lnd["LTC"].Chains[0], "litecoin")
-
-		// Set the node public key.
-		node.SetPubKey(res.NodePubKey)
-
-		// Add pair to the node.
-		ht.act.addPair(node, "LTC", "BTC", xudrpc.AddCurrencyRequest_LND)
+		for {
+			// retry interval
+			time.Sleep(100 * time.Millisecond)
```
moshababo: Better to start with an initial attempt, and do retries on failure. 

Also maybe limit the retries/total time below the default test timeout, specially if you increased it to 30 sec.

erkarl: Good point. I've changed the code to do the initial attempt first and added a timeout of 10 second. Also lowered the context timeout to 15 seconds.

sangaman (approved):
Awesome, thanks for figuring this out.

moshababo (approved):

moshababo:
Ready for merging?

moshababo:
Ready to merge?

sangaman:
Yes I think so if you're ok with your 2nd comment above, making code more go idiomatic.

erkarl:
@moshababo looks like your suggested way of doing the retry/timeout link does not point to the correct loc. Can you provide a new example?

moshababo:
Yes it changed a bit, here's the new link. https://github.com/ExchangeUnion/xud/blob/master/test/simulation/xudtest/node.go#L215.

It's just a suggestion, we can also merge as-is if it's a hassle.

erkarl:
Thanks for the updated link @moshababo - I'm going to merge this as is because I'm now having trouble reproducing the original issue locally which makes it very time consuming to test the changes.

PR #949, opened 2019-05-07
feat(raiden): pass secret hash to raiden

ghost:


sangaman (approved):
Looks good. I'm thinking it might make sense to have the secret hash be a required parameter for our token payment method, since we aren't going to be making that call without specifying a hash as far as I know.

erkarl:
Agreed - updated.

sangaman (approved):

Issue #574, opened 2018-10-09
Don't write files to disk during automated tests


sangaman:
Currently, xud will potentially write files to disk during automated tests, namely the node key file and tls.cert. We should ensure that the tests do not write any files, nor make any other persistent changes.

kilrau:
something for @erkarl or @moshababo ?

kilrau:
Did you look into details here already? An overview what needs to be done and how long it might take? @moshababo

moshababo:
Closed by #950.

PR #950, opened 2019-05-07
test: use os-specific temp dir as xudir

moshababo:
Closing #574.

Since avoid writing to disk isn't practical (it would require different code to be run for production and under test), a temp directory can be used instead.

### DIFF
--- a/test/utils.ts
+++ b/test/utils.ts
@@ -47,7 +47,7 @@ export const waitForSpy = (spy: SinonSpy, property = 'called') => {
 /**
  * Creates and returns an os-specific directory for temp files.
  */
-export const getTempDir = (unique = false) => {
+export const getTempDir = (unique: boolean) => {
   let dir = path.join(os.tmpdir(), 'xud-test');
   if (unique) {
     dir = path.join(dir, uuidv1());
--- a/test/utils.ts
+++ b/test/utils.ts
@@ -1,8 +1,11 @@
 import net from 'net';
 import { SinonSpy } from 'sinon';
+import uuidv1 from 'uuid';
+import os from 'os';
+import path from 'path';
+import fs from 'fs';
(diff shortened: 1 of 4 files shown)


test/utils.ts
```diff
@@ -41,6 +44,22 @@ export const waitForSpy = (spy: SinonSpy, property = 'called') => {
   });
 };
 
+/**
+ * Creates and returns an os-specific directory for temp files.
+ */
+export const getTempDir = (unique: boolean) => {
```
erkarl: Is there a case where we don't want the temporary directory to **NOT** be unique?

test/utils.ts
```diff
@@ -41,6 +44,22 @@ export const waitForSpy = (spy: SinonSpy, property = 'called') => {
...
+  let dir = path.join(os.tmpdir(), 'xud-test');
+  if (unique) {
+    dir = path.join(dir, uuidv1());
+  }
+
+  if (!fs.existsSync(dir)) {
```
sangaman: Minor nit, but you can just `try` the mkdir call and in the `catch` ignore the exception if `err.code === 'EEXIST'`. That will just combine two calls into one and make the tests run ever so slightly faster.

Also you can consider using the async version of the mkdir call that returns a promise, but I figure the tests are being run one at a time anyway so I'm thinking it probably wouldn't make a difference.

moshababo: I noticed that due to the `{recursive: true}` option, that error isn’t thrown. 
Assuming that the implementation is different on that option, I wasn’t sure if removing `existsSync` would cause performance gain, so I left it as it. 

In regards to the async version - I thought it would just create a hassle for the test code.

moshababo:
Thanks for the reviews, and sorry for the delay.

PR #958, opened 2019-05-11
[WIP]: feat: track nodes that have banned us

### LINKED ISSUE #693: Implement the behavior for #152 disconnect package "banned"
This is an offshoot of #152 and implements the behavior for disconnect package "banned"

* if we're getting the #152 packet for being banned instead of `HELLO` on connection attempt, socket gets closed by peer and we stop connecting to the peer (ban/disable outgoing connections to this node). If peer successfully connects to us again, we delete the in-memory entry.
* if peer doesn't send #152 or `HELLO` packet on connection attempt, cancel and don't attempt to reconnect (peer doesn't follow the protocol) (ban/disable outgoing connections to this node). If peer successfully connects to us again, we delete the in-memory entry.

ImmanuelSegol:
closes #693 

@sangaman This PR will ban and disconnect from a peer if we receive a `DisconnectingPacket` before a `SessionInitPacket`.

The next step for this PR will be persisting the fact that the node has banned us in memory or in the DB, 
personally, I think we should persist it in the DB. What are you're opinions?

### DIFF
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -25,7 +25,9 @@ export const reputationEventWeight = {
 
 interface NodeList {
   on(event: 'node.ban', listener: (nodePubKey: string, events: ReputationEvent[]) => void): this;
+  on(event: 'node.unban', listener: (nodePubKey: string) => void): this;
   emit(event: 'node.ban', nodePubKey: string, events: ReputationEvent[]): boolean;
+  emit(event: 'node.unban', nodePubKey: string): boolean;
 }
 
 /** A wrapper class with help methods that represents a list of known nodes. */
@@ -248,6 +250,7 @@ class NodeList extends EventEmitter {
         // If the reputationScore is not below the banThreshold but node.banned
         // is true that means that the node was unbanned
         promises.push(this.setBanned(node, false));
+        this.emit('node.unban', nodePubKey);
       }
 
       promises.push(this.repository.addReputationEvent({ event, nodeId: node.id }));
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -204,7 +204,7 @@ class NodeList extends EventEmitter {
         node.lastAddress = lastAddress;
       }
 
-      await node.save();
+      await this.save(node);
       return true;
     }
 
@@ -269,7 +269,7 @@ class NodeList extends EventEmitter {
       const index = node.addresses.findIndex(existingAddress => addressUtils.areEqual(address, existingAddress));
       if (index > -1) {
         node.addresses = [...node.addresses.slice(0, index), ...node.addresses.slice(index + 1)];
-        await node.save();
+        await this.save(node);
         return true;
       }
 
@@ -282,9 +282,14 @@ class NodeList extends EventEmitter {
     return false;
   }
 
+  private save = async (node: NodeInstance) => {
+    await node.save();
+    this.nodes.set(node.nodePubKey, node);
+  }
+
   private setBanned = async (node: NodeInstance, banned: boolean) => {
     node.banned = banned;
-    await node.save();
+    await this.save(node);
   }
 
   private addNode = (node: NodeInstance) => {
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -5,7 +5,12 @@ import addressUtils from '../utils/addressUtils';
 import { pubKeyToAlias } from '../utils/aliasUtils';
 import errors from './errors';
 import P2PRepository from './P2PRepository';
-import { Address } from './types';
+import { Address, NodeConnectionInfo } from './types';
+
+type NodeReputation = {
+  reputationScore: number;
+  banned?: boolean;
+};
 
 export const reputationEventWeight = {
   [ReputationEvent.ManualBan]: Number.NEGATIVE_INFINITY,
@@ -74,14 +79,13 @@ class NodeList extends EventEmitter {
   /**
    * Persists whether a node has banned us.
    */
-  public setBannedBy = async (nodePubKey: string, bannedBy = true) => {
+  public setBannedBy = async (nodePubKey: string, bannedBy: boolean) => {
     const node = this.nodes.get(nodePubKey);
     if (!node) {
       throw errors.NODE_NOT_FOUND(nodePubKey);
     }
     node.bannedBy = bannedBy;
-
-    await this.repository.updateBannedBy(nodePubKey, bannedBy);
+    await this.save(node);
   }
 
   /**
@@ -93,10 +97,6 @@ class NodeList extends EventEmitter {
 
   public forEach = (callback: (node: NodeInstance) => void) => {
     this.nodes.forEach(callback);
-  }
-
-  public getNode(nodePubKey: string): NodeInstance | undefined {
-    return this.nodes.get(nodePubKey);
   }
 
   /**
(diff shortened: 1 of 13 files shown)


lib/p2p/Peer.ts
```diff
@@ -839,6 +839,12 @@ class Peer extends EventEmitter {
   }
 
   private handleDisconnecting = (packet: packets.DisconnectingPacket): void  => {
+    if (!this.sessionInitPacket && packet.body && packet.body.reason === DisconnectionReason.Banned) {
+      this.recvDisconnectionReason = packet.body.reason;
+      this.emit('reputation', ReputationEvent.ManualBan);
```
sangaman: I don't think we should use the manual ban event here, I think of that for when we are the ones doing the banning. It would prevent the peer from "unbanning" us in the future, since we will have banned them. Rather, we should be making a note that we have been banned by this peer to prevent us from trying to reconnect to them in the future. We could do this with a new database field in the `nodes` table, which we'd toggle in the code below. We'd also need to make sure to toggle it back off if the peer successfully connects with us again in the future.

lib/p2p/Peer.ts
```diff
@@ -839,6 +839,12 @@ class Peer extends EventEmitter {
   }
 
   private handleDisconnecting = (packet: packets.DisconnectingPacket): void  => {
+    if (!this.sessionInitPacket && packet.body && packet.body.reason === DisconnectionReason.Banned) {
```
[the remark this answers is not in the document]
ImmanuelSegol: The reason is if we don't have the init packet, that means we aren't connected to the node and as @kilrau stated [here](https://github.com/ExchangeUnion/xud/issues/693), we should only mark the node as banning us if we don't get the `init packet` on the connection attempt.

That is what I understood from the issue.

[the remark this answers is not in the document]
sangaman: I think that issue is slightly outdated since we aren't using `Hello` packets any more. In any case, in any case I think we will actually get the session init packet even when we are banned, since that is where we communicate node pub keys to peers, and until they know our pub key they won't know if they've banned us. So I'm pretty sure we don't want this check here, unless there's something else I'm missing here.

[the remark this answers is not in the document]
kilrau: @sangaman is right, the issue description is outdated

sangaman:
@ImmanuelSegol The errors here are not something I've seen before. Could you try merging in the latest from master (it includes some changes to the tests) and see if that resolves the travis failures? If not I'll try to recreate locally.

ImmanuelSegol:
@sangaman The tests are only failing on Travis locally they are fine.

sangaman:
@ImmanuelSegol The tests do fail for me when I pull down the branch and run them locally.

kilrau:
Since it's github's standard, let's make passing travis tests also our standard. Only in very rare occasions travis tests fail due to a travis problem.

rsercano:
It took some time but I figured that the failure of test's cause is disconnect packet's checksum somehow doesn't match with header's. Probably the newly added `peerPubKey` field is related with it.

rsercano:
Tests fixed, should be working now.

lib/p2p/packets/types/DisconnectingPacket.ts
```diff
@@ -6,6 +6,7 @@ import { removeUndefinedProps } from '../../../utils/utils';
 
 export type DisconnectingPacketBody = {
   reason: DisconnectionReason;
+  peerPubKey: string;
```
sangaman: We shouldn't have to add this to the disconnecting packet, particularly since this is a a breaking change. The peer should already know our pub key, and should use the previously known pubkey rather than the value from this packet.

Plus I figure this could introduce some weird vulnerability where a peer sends a disconnecting packet with another node's pub key.

lib/p2p/Peer.ts
```diff
@@ -288,7 +290,7 @@ class Peer extends EventEmitter {
   /**
    * Close a peer by ensuring the socket is destroyed and terminating all timers.
    */
-  public close = async (reason?: DisconnectionReason, reasonPayload?: string): Promise<void> => {
+  public close = async (localPubKey?: string, reason?: DisconnectionReason, reasonPayload?: string): Promise<void> => {
```
sangaman: Likewise we shouldn't need the local pub key for this method.

lib/p2p/Pool.ts
```diff
@@ -829,6 +840,13 @@ class Pool extends EventEmitter {
 
     peer.once('close', () => this.handlePeerClose(peer));
 
+    peer.on('bannedBy', async (nodePubKey: string) => {
+      // set peer as banning us
+      await this.nodes.persistBannedBy(nodePubKey);
+      this.nodes.setBannedBy(nodePubKey);
+      this.logger.info(`Peer: ${nodePubKey} has banned you`);
```
sangaman: I think "has banned us" is more in line with how we log such messages elsewhere.

lib/p2p/P2PRepository.ts
```diff
@@ -37,6 +37,14 @@ class P2PRepository {
   public addNodes = async (nodes: NodeFactory[]) => {
     return this.models.Node.bulkCreate(<NodeAttributes[]>nodes);
   }
+
+  public setBannedBy = async (nodePubKey: string, bannedBy: boolean) => {
```
sangaman: Maybe call this `updateBannedBy` to differentiate from the other `setBannedBy` method?

lib/p2p/Pool.ts
```diff
@@ -514,12 +517,20 @@ class Pool extends EventEmitter {
       // the node is known, update its listening addresses
       await this.nodes.updateAddresses(peer.nodePubKey!, addresses, peer.inbound ? undefined : peer.address);
     }
+
+    // Check if peer has previously banned us if so set `bannedBy` to false
+    const node = await this.repository.getNode(peerPubKey);
```
sangaman: We shouldn't need to read from the db again ideally, as we would already have all nodes in the `NodeList` class and can just get the node from there.

lib/p2p/NodeList.ts
```diff
@@ -40,6 +41,31 @@ class NodeList extends EventEmitter {
     super();
   }
 
+  /**
+   * Set status of peer that bans/unbans and persist in DB
+   */
+  public persistBannedBy = async (nodePubKey: string, bannedBy = true) => {
```
sangaman: I don't think we need both a `persistBannedBy` and a `setBannedBy` method, as we're always calling both of them sequentially. I think it'd be cleaner to just have a `setBannedBy` method that updates both the db and the internal list of nodes that have banned us.

lib/p2p/packets/types/DisconnectingPacket.ts
rsercano: I suppose `expectedNodePubKey` is the peer's pub key ?

rsercano:
implemented @sangaman

lib/p2p/Peer.ts
```diff
@@ -932,6 +934,12 @@ class Peer extends EventEmitter {
   }
 
   private handleDisconnecting = (packet: packets.DisconnectingPacket): void  => {
+    if (packet.body && packet.body.reason === DisconnectionReason.Banned) {
```
sangaman: I think this if statement should be moved within the if statement below. The main problem as I see it is that if we set `this.recvDisconnectionReason` here then we won't enter the next `if` statement and instead go to the `else` where we should penalize the peer. i'm thinking it should look like this:

```javascript
if (!this.recvDisconnectionReason && packet.body && packet.body.reason) {
  this.logger.debug(`received disconnecting packet from ${this.label}:${JSON.stringify(packet.body)}`);
  this.recvDisconnectionReason = packet.body.reason;
  if (packet.body.reason === DisconnectionReason.Banned) {
    // if we are banned by the peer set `bannedBy` in DB.
	const nodePubKey = this._nodePubKey || this.expectedNodePubKey;
	if (nodePubKey) {
	  this.emit('bannedBy', this.expectedNodePubKey!);
	}
  }
} else {
  // protocol violation: packet should be sent once only, with body, with `reason` field
  // TODO: penalize peer
}
```

lib/p2p/Peer.ts
```diff
@@ -932,6 +934,12 @@ class Peer extends EventEmitter {
...
+      this.recvDisconnectionReason = packet.body.reason;
+      // if we are banned by the peer change `bannedBy` in DB.
+      this.emit('bannedBy', this.expectedNodePubKey!);
```
sangaman: We may have established the pub key for the peer (such that we're no longer merely "expecting" it) so this should be: `this._nodePubKey || this.expectedNodePubKey` - I'll update the code snippet above with this.

lib/p2p/packets/types/DisconnectingPacket.ts
sangaman: Or the `_nodePubKey` in cases where a peer bans us after we have connected and initialized the session, I figure.

rsercano:
Thanks Daniel, and I'm sorry it's been a while; for which exact case do we require the test ? @sangaman

sangaman:
The simulation tests are failing with an odd panic error, I'll have to look into it more closely.

```
--- FAIL: TestIntegration (254.56s)

    xud_test.go:70: Running 9 integration tests

    --- PASS: TestIntegration/network_initialization (0.60s)

    --- PASS: TestIntegration/raiden_swap (200.13s)

panic: runtime error: invalid memory address or nil pointer dereference [recovered]

	panic: runtime error: invalid memory address or nil pointer dereference

[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0xab9c36]

goroutine 13 [running]:

testing.tRunner.func1(0xc0000f6500)

	/home/travis/.gimme/versions/go1.13.linux.amd64/src/testing/testing.go:874 +0x3a3

panic(0xbad960, 0x1405c90)

	/home/travis/.gimme/versions/go1.13.linux.amd64/src/runtime/panic.go:679 +0x1b2

github.com/ExchangeUnion/xud-simulation.TestIntegration(0xc0000f6500)

	/home/travis/build/ExchangeUnion/xud/test/simulation/xud_test.go:114 +0x886

testing.tRunner(0xc0000f6500, 0xe09128)

	/home/travis/.gimme/versions/go1.13.linux.amd64/src/testing/testing.go:909 +0xc9

created by testing.(*T).Run

	/home/travis/.gimme/versions/go1.13.linux.amd64/src/testing/testing.go:960 +0x350

exit status 2

FAIL	github.com/ExchangeUnion/xud-simulation	523.634s
```

sangaman:
@erkarl I put in a db migration framework with a migration for the new `bannedBy` column, you can take a look at it.

I'm still stumped as to why the simulation tests are failing. One thing I tried changing was making it so we can still attempt manual connections to a peer that has banned us, since the simulation tests do a lot of banning then unbanning to simulate disconnects. It didn't resolve the error, but I left it that way in case it would prevent other sim test failures and since I think it's actually desirable to be attempt to connect to peers manually even if they have banned us in the past.

erkarl:
Restarted tests - looks like mostly issues with the simtests.

kilrau:
I think it's time we get this oldie out of the way, wdyt? @sangaman

sangaman:
I'm going to rebase this and see if it plays nicely with existing tests.

erkarl:
Closing due to inactivity.

PR #959, opened 2019-05-14
test: preserve xud log files from sim tests

sangaman:
This preserves the log files generated from running the simulation tests within the simulation test temp folder. Data directories are also moved to the temp folder and are still cleaned up by default after each run, but separating and keeping the logs will help with diagnosing errors that arise during simulation test runs.

### DIFF
--- a/test/simulation/xudtest/node.go
+++ b/test/simulation/xudtest/node.go
@@ -29,6 +29,7 @@ var (
 type nodeConfig struct {
 	DataDir     string
 	XUDPath     string
+	LogPath     string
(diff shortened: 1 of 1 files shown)


erkarl:
Another welcome addition for better debugging experience. Have you tested this on top of https://github.com/ExchangeUnion/xud/pull/962 where we removed `temp` from the gitignore?

sangaman:
The temp folder is still in the git ignore in that PR, it's just that the go folder has been moved out of there. It shouldn't conflict with this PR.

PR #960, opened 2019-05-14
feat(cli): add orderbook command

### LINKED ISSUE #648: Market depth CLI command
This is an offshoot of #643. @offerm said:

> b. We need to aggregate the capacity available at each price level (rounded). For example, in the table above we have 2 BUY orders of at price of 0.87. These should be combined.
> c. once you combine and sort properly you will be able to see the market depth. If we will not aggregate but have a lot of orders this display will not be usable.
> d. I would consider to have this as a new cli getmarketdept and leave getorders as today.
> e. to be considered - aggregate the buy/sell orders into n (say 8) buy buckets and n sell bucket and presenting them over the price line. So sell orders will come first (higher prices) and buy order will come below it (normal view of order book).

I agree with d that if we do want a market depth command from the cli, it should be separate from `getorders`. I figure the new `getmarketdepth` command could take a parameter to specify precision - we could group prices of 0.871 and 0.869 for example.

I would consider this an enhancement and not high priority, but perhaps a tool for individual traders using the command line.

ghost:
This PR adds a new CLI command `orderbook`. Default `precision` is currently set to 5 (the number of digits following the decimal point) - this value is **configurable** via the CLI. 

Orderbook is divided into 8 "buckets" - this value is currently **not configurable** via the CLI, but changing it in the codebase is trivial.

Some performance tests on my machine:
- ~3000ms @ 1 million orders
- ~160ms @ 100,000 orders
- ~25ms @ 10,000 orders

I consider these metrics acceptable for initial paint since in the future we'll be implementing a streaming option to the orderbook call which will cheap.

Closes #648

### DIFF
--- a/lib/cli/command.ts
+++ b/lib/cli/command.ts
@@ -38,7 +38,7 @@ interface GrpcResponse {
   toObject: Function;
 }
 
-export const callback = (argv: Arguments, formatOutput?: Function) => {
+export const callback = (argv: Arguments, formatOutput?: Function, displayJson?: Function) => {
   return (error: Error | null, response: GrpcResponse) => {
     if (error) {
       console.error(`${error.name}: ${error.message}`);
@@ -47,9 +47,13 @@ export const callback = (argv: Arguments, formatOutput?: Function) => {
       if (Object.keys(responseObj).length === 0) {
         console.log('success');
       } else {
-        !argv.json && formatOutput
-          ? formatOutput(responseObj)
-          : console.log(JSON.stringify(responseObj, undefined, 2));
+        if (!argv.json && formatOutput) {
+          formatOutput(responseObj, argv);
+        } else {
+          displayJson
+            ? displayJson(responseObj, argv)
+            : console.log(JSON.stringify(responseObj, undefined, 2));
+        }
       }
     }
   };
--- a/lib/cli/commands/orderbook.ts
+++ b/lib/cli/commands/orderbook.ts
@@ -1,0 +1,179 @@
+import { Arguments } from 'yargs';
+import { callback, loadXudClient  } from '../command';
+import { ListOrdersRequest, ListOrdersResponse, Order } from '../../proto/xudrpc_pb';
+import Table, { HorizontalTable } from 'cli-table3';
+import colors from 'colors/safe';
+import { satsToCoinsStr } from '../utils';
+
+type FormattedOrderbook = {
+  pairId: string,
+  rows: string[][],
+};
+
+type BucketDepth = {
+  price: number,
+  depth: number;
+};
+
+type OrderbookJson = {
+  pairId: string,
+  sell: BucketDepth[],
+  buy: BucketDepth[],
+};
+
+const COLUMNS = [19, 19, 19, 19];
+const COLUMNS_IN_ORDER_SIDE = COLUMNS.length / 2;
+const HEADER = [
+  { content: colors.green('Buy'), colSpan: 2 },
+  { content: colors.red('Sell'), colSpan: 2 },
+];
+const SECONDARY_HEADER = [
+  colors.green('Price'),
+  colors.green('Depth'),
+  colors.red('Price'),
+  colors.red('Depth'),
+];
+
+const addSide = (buckets: BucketDepth[]): string[] => {
+  const bucket = buckets.pop();
+  if (bucket) {
+    return [
+      bucket.price.toString(),
+      satsToCoinsStr(bucket.depth),
+    ];
+  } else {
+    return Array.from(Array(COLUMNS_IN_ORDER_SIDE)).map(() => '');
+  }
+};
+
+export const createOrderbook = (orders: ListOrdersResponse.AsObject, precision: number) => {
+  const formattedOrderbooks: FormattedOrderbook[] = [];
+  orders.ordersMap.forEach((tradingPair) => {
+    const buy = createOrderbookSide(tradingPair[1].buyOrdersList, precision);
+    const sell = createOrderbookSide(tradingPair[1].sellOrdersList, precision);
+    const totalRows = buy.length < sell.length
+      ? sell.length : buy.length;
+    const orderbookRows = Array.from(Array(totalRows))
+      .map(() => {
+        return addSide(buy).concat(addSide(sell));
+      });
+    formattedOrderbooks.push({
+      pairId: tradingPair[0],
+      rows: orderbookRows,
+    });
+  });
+  return formattedOrderbooks;
+};
+
+const createTable = () => {
+  const table = new Table({
+    colWidths: COLUMNS,
+  }) as HorizontalTable;
+  table.push(HEADER);
+  table.push(SECONDARY_HEADER);
+  return table;
+};
+
+const displayOrderbook = (orderbook: FormattedOrderbook) => {
+  const table = createTable();
+  orderbook.rows.forEach(row => table.push(row));
+  console.log(colors.underline(colors.bold(`\nTrading pair: ${orderbook.pairId}`)));
+  console.log(table.toString());
+};
+
+const displayTables = (orders: ListOrdersResponse.AsObject, argv: Arguments) => {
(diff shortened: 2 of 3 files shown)


kilrau:
Pls check travis

test/unit/Command.orderbook.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+};
+
+describe('Command.orderbook.createOrderbookSide', () => {
+  it('generates orderbook from 1 million orders', () => {
+    const orders: Order.AsObject[] = createOrders(1000000, 0.012160, 100000, true);
+    const startTime = performance.now();
+    createOrderbookSide(orders);
+    const endTime = performance.now();
+    const timeSpent = endTime - startTime;
+    expect(timeSpent < 6000).to.equal(true);
```
erkarl: Think we can lower this test to check for 10k instead of 1M orders in order to not slow down the test suite too much unless someone sees a compelling reason not to do so.

sangaman: I agree, in fact I'm not sure we need to have performance tests here in general since it's so hardware dependent. Tests that pass for us might fail on Travis because of differences in available resources.

lib/cli/command.ts
```diff
@@ -48,8 +48,10 @@ export const callback = (argv: Arguments, formatOutput?: Function) => {
         console.log('success');
       } else {
         !argv.json && formatOutput
-          ? formatOutput(responseObj)
-          : console.log(JSON.stringify(responseObj, undefined, 2));
+          ? formatOutput(responseObj, argv)
```
sangaman: This is slightly hard to read with nested ternary operators. How about 

```javascript
if (!argv.json && formatOutput) {
  formatOutput(responseObj);
} else {
  displayJson ? displayJson(responseObj, argv) : console.log(JSON.stringify(responseObj, undefined, 2));
}
```

lib/cli/commands/orderbook.ts
```diff
@@ -0,0 +1,183 @@
...
+  console.log(colors.underline(colors.bold(`\nTrading pair: ${orderbook.pairId}`)));
+  console.log(table.toString());
+};
+
+const displayTables = (orders: ListOrdersResponse.AsObject, argv: Arguments) => {
+  createOrderbook(orders, argv.precision).forEach(displayOrderbook);
+};
+
+const getAvailablePrices = (orders: Order.AsObject[], sellSide: boolean, count = 8): number[] => {
+  const sortedOrders = [
```
sangaman: I believe xud is returning the orders sorted now, so no need to sort them here.

lib/cli/commands/orderbook.ts
```diff
@@ -0,0 +1,183 @@
...
+  orderbook.rows.forEach(row => table.push(row));
+  console.log(colors.underline(colors.bold(`\nTrading pair: ${orderbook.pairId}`)));
+  console.log(table.toString());
+};
+
+const displayTables = (orders: ListOrdersResponse.AsObject, argv: Arguments) => {
+  createOrderbook(orders, argv.precision).forEach(displayOrderbook);
+};
+
+const getAvailablePrices = (orders: Order.AsObject[], sellSide: boolean, count = 8): number[] => {
```
sangaman: This method is slightly confusing to me, it's just getting the prices of the orders in an array right? I'm not sure what "available" means in this context. I don't know if maybe `getPrices` or `getOrderPrices` might be more straightforward, or you can toss on a brief comment for the method.

lib/cli/commands/orderbook.ts
```diff
@@ -0,0 +1,183 @@
...
+  return sortedOrders.splice(0, count);
+};
+
+const getPriceBuckets = (
+  orders: Order.AsObject[],
+  availablePrices: number[],
+  existingBuckets: PriceBucket[] = [],
+): PriceBucket[] => {
+  // go through all the available price buckets
+  const price = availablePrices.shift();
```
sangaman: I think it's slightly more performant to loop through the array without changing it (as is done with `shift`). I think this would mean abandoning the recursion in place of something like `forEach`, adding orders to a given bucket until you reach the next price point, then starting with a new bucket. It would be interesting to confirm that since you've taken benchmarks already. I think calling `filter` in each loop might be a bit expensive performance-wise too.

lib/cli/commands/orderbook.ts
erkarl: Indeed, it's confusing. It actually returns the "bucket" prices. How about naming it `getBucketPrices`? And the other one below would then be `getBucketDepth`?

sangaman: Yes I like that a lot more.

lib/cli/command.ts
kilrau: Want to change this? @erkarl

lib/cli/commands/orderbook.ts
kilrau: Should be changed if it's indeed already sorted

lib/cli/commands/orderbook.ts
kilrau: also tbd I think

test/unit/Command.orderbook.spec.ts
erkarl: I opted to keep the test for future debugging purposes to have an easy way to benchmark, but reduced the amount of orders and increased the buffer time so that I'll pass even on slow machines. The alternative is to skip the test and enable it when needed. Actually, that sounds better. I'll make the change.

erkarl: Done - skipped that test.

lib/cli/commands/orderbook.ts
erkarl: I went with `getBucketPrices` and `getDepthForBuckets`.

lib/cli/commands/orderbook.ts
erkarl: I tried doing it with `.forEach`, but it didn't read/feel as good as the recursive version. I don't think we'll have performance issues unless we start having hundreds of thousands of orders - I'd be OK addressing the UI/CLI layer performance issues once that happens.

sangaman: Alright, not a huge deal. Like you said though it's maybe something we can revisit in the future.

sangaman (approved):
LGTM!

erkarl:
@sangaman I rebased this on top of master and extended `satsToCoinsStr` with an optional precision parameter.

kilrau:
Great! When this is finally merged, don't forget https://github.com/ExchangeUnion/xud/issues/643 @erkarl , which finally separates
* `listorders` as per-order-view with all sorts of details of the order and the peer 
from
* `orderbook` as anonymous, aggregated view of orders (like on a regular exchange)

lib/cli/utils.ts
```diff
@@ -9,9 +9,9 @@ export const coinsToSats = (coinsQuantity: number) => {
   return Math.round(coinsQuantity * SATOSHIS_PER_COIN);
 };
 
-/** Returns a number of satoshis as a string representation of coins with up to 8 decimal places. */
-export const satsToCoinsStr = (satsQuantity: number) => {
-  return (satsQuantity / SATOSHIS_PER_COIN).toFixed(8).replace(/\.?0+$/, '');
+/** Returns a number of satoshis as a string representation of coins with specified decimal places. */
```
sangaman: Just to clarify, this will return a string with *up to* the specified number of decimal places, if you specify 4 and have 1.2 coins it will return `1.2` and not `1.2000`. I think this is fine, but if this not the behavior you intended then maybe you can change this method. If this is the behavior you intended, maybe we can say "up to the specified number of decimal places."

erkarl: @sangaman 
Thanks for clarifying. I did some tinkering... in hindsight I think the orderbook depth looks nicer if it's up to 8 decimal places regardless of the precision requested from the CLI, only the price will be rounded to the provided precision.

kilrau:
Comments fixed? Let's get this in! @erkarl

PR #973, opened 2019-05-22
feat(seedutil): generate eth keystore from aezeed mnemonic

ghost:
This PR adds support for deriving Ethereum keystore file from aezeed mnemonic seed. 

Related issue: https://github.com/ExchangeUnion/xud/issues/912


seedutil/README.md
```diff
@@ -0,0 +1,8 @@
+# seedutil
+This utility is used to derive Ethereum keystore file from [aezeed](https://github.com/lightningnetwork/lnd/tree/master/aezeed) generated mnemonic seed.
+
+## Build
+`npm run compile:seedutil`
+
+## Usage
+[Tests](/test/jest/SeedUtil.spec.ts)
```
sangaman: Does it make sense to add the command for running this directly from the command line here?

Also for tests, I'm wondering if we want a `test:seedutil` npm script that's separate from the other jest scripts, which we can then put here. I figure eventually as we continue migrating to jest, we'll want to keep our `test:int`, `test:unit`, etc... breakdown and `test:seedutil` would fit in there.

seedutil/README.md
```diff
@@ -0,0 +1,8 @@
+# seedutil
+This utility is used to derive Ethereum keystore file from [aezeed](https://github.com/lightningnetwork/lnd/tree/master/aezeed) generated mnemonic seed.
```
sangaman: used to derive *an* Ethereum keystore file from *a* generated mnemonic seed.

/grammarnit

test/jest/SeedUtil.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  test('it allows custom keystore save path', async () => {
+    const CUSTOM_PATH = `${process.cwd()}/seedutil/custom`;
+    const cmd = `./seedutil/seedutil ${VALID_SEED.seedWords.join(' ')} ${VALID_SEED.seedPassword} ${CUSTOM_PATH}`;
+    await expect(executeCommand(cmd))
+      .resolves.toMatch(SUCCESS_KEYSTORE_CREATED);
+    // cleanup custom path
+    await deleteDir(CUSTOM_PATH);
+  });
+
```
sangaman: Can drop this blank line.

sangaman (commented):
Great work, the built-in tests are great and this will be very handy. Just a couple comments but this looks ready to go.

seedutil/README.md
erkarl: I didn't want to add a command line usage here because of privacy concerns (the history is recorded by default). It does make sense to add it for development purposes. I'll add it with a warning.

As for tests, yes. Good idea - now that we have more than 1 jest test suite it makes sense to start dividing them into categories.

sangaman:
@erkarl Will you get a chance to look at my comments/reorganizing the jest tests soon? It'd be good to merge this so I can use it for Raiden in #1006. I think it's fine tho not to change existing jest tests actually, but at least break off the aezeed tests into their own command.

erkarl:
@sangaman feedback applied as fixup commits.

sangaman (approved):
Looks good, squash -> merge.

PR #975, opened 2019-05-24
fix: load raiden token addresses from db

sangaman:
This fixes the initialization logic for the raiden client and swap manager to load existing raiden currencies from the database.

This should fix the issue observed in xud-simnet where WETH sanity swaps are not being initiated.

### DIFF
--- a/test/jest/SwapClientManager.spec.ts
+++ b/test/jest/SwapClientManager.spec.ts
@@ -5,7 +5,17 @@ import SwapClientManager from '../../lib/swaps/SwapClientManager';
 import Config from '../../lib/Config';
 import { SwapClientType } from '../../lib/constants/enums';
 
-jest.mock('../../lib/db/DB');
+jest.mock('../../lib/db/DB', () => {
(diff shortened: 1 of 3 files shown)


test/jest/SwapClientManager.spec.ts
```diff
@@ -99,7 +109,7 @@ describe('Swaps.SwapClientManager', () => {
 
   test('it initializes lnd-ltc, lnd-btc and raiden', async () => {
```
erkarl: think we can also consider asserting
```
    expect(swapClientManager.raidenClient.tokenAddresses.size).toEqual(1);
    expect(swapClientManager.raidenClient.tokenAddresses.get('WETH')).not.toBeUndefined();
```
to test that the tokenAddress does get set.

erkarl: And remove line 43 that currently adds this value via mock: `tokenAddresses.set('WETH', '0x1234');`

sangaman: Done, thanks!

PR #976, opened 2019-05-24
 fix: swap client initialization

sangaman:
This fixes a bug where the swap clients would fail on their first attempt to initialize because they would try to update `pool.nodeState` which is undefined. `pool.nodeState` is not assigned a value until `pool.init()` is called, which would not happen until after the first attempt to initialize swap clients.

This changes the initialization flow so that the node state is assigned a value in the `Pool` constructor. The properties of `Pool.nodeState` are then updated as the other components (swap clients, order book)
initialize.

### DIFF
--- a/lib/swaps/SwapClientManager.ts
+++ b/lib/swaps/SwapClientManager.ts
@@ -7,8 +7,16 @@ import Logger, { Loggers } from '../Logger';
 import Pool from '../p2p/Pool';
 import { errors } from './errors';
 import { Currency } from '../orderbook/types';
-import { isLndClient, isRaidenClient } from './types';
 import { Models } from '../db/DB';
+import { SwapClientType } from '../constants/enums';
+
+function isRaidenClient(swapClient: SwapClient): swapClient is RaidenClient {
+  return (swapClient.type === SwapClientType.Raiden);
+}
+
+function isLndClient(swapClient: SwapClient): swapClient is LndClient {
+  return (swapClient.type === SwapClientType.Lnd);
+}
 
 class SwapClientManager {
   /** A map between currencies and all swap clients */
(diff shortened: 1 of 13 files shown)


test/integration/OrderBook.spec.ts
```diff
@@ -30,23 +30,23 @@ const initValues = async (db: DB) => {
   ]);
 };
 
+const sandbox = sinon.createSandbox();
```
erkarl: What do you think about moving `nomatching OrderBook` describe block within `Orderbook` instead of moving the scope of sandbox & pool to outer scope?

sangaman: I thought of this same thing, but I kind of liked having the nomatching tests separate. Although in thinking about this a bit more, maybe it messes up how we call `sandbox.restore()` in the `after` block. I can keep nomatching and matching separate, but have each have a separate `sandbox` within their scope. Does that sound good to you?

erkarl: That works.

Minor nit: I also noticed `Orderbook` describe block has 2 after blocks which probably means one of them isn't being executed.

sangaman: Done.

sangaman:
One of the tests is failing, not sure why, I'll investigate further later.

sangaman:
I figured out the issue with the tests, this should be ready to merge pending another review.

lib/swaps/SwapClientManager.ts
```diff
@@ -7,8 +7,16 @@ import Logger, { Loggers } from '../Logger';
 import Pool from '../p2p/Pool';
 import { errors } from './errors';
 import { Currency } from '../orderbook/types';
-import { isLndClient, isRaidenClient } from './types';
 import { Models } from '../db/DB';
+import { SwapClientType } from '../constants/enums';
+
+function isRaidenClient(swapClient: SwapClient): swapClient is RaidenClient {
```
erkarl: Minor nit: do we want to keep `isRaiden`/`isLndClient` here or move them back to `types`?

sangaman: After I changed the method to use a type guard (e.g. `swapClient is RaidenClient`) I was running into circular dependency errors between `SwapClientType` and `SwapClient`, so I reorganized it like this. It seems fine since at least for now since we're not using it anywhere else.

PR #980, opened 2019-05-25
chore: add response logging to http resolver

ghost:
This PR adds additional logging to http resolver to make debugging for Offer easier.


erkarl:
Going to merge this as this is only adding extra logging.
cc @offerm @sangaman

PR #991, opened 2019-05-27
feat: if order placed and raiden disabled throw error

### LINKED ISSUE #983: Can add order for WETH/BTC while raiden is disabled
Disabling raiden in xud.conf still allows WETH/BTC order.

ImmanuelSegol:
closes #983

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -376,12 +376,21 @@ class OrderBook extends EventEmitter {
 
     if (!this.nobalancechecks) {
       // check if sufficient outbound channel capacity exists
-      const { outboundCurrency, outboundAmount } = Swaps.calculateInboundOutboundAmounts(order.quantity, order.price, order.isBuy, order.pairId);
-      const swapClient = this.swaps.swapClientManager.get(outboundCurrency);
-      if (!swapClient) {
+      const { outboundCurrency, inboundCurrency, outboundAmount } =
+          Swaps.calculateInboundOutboundAmounts(order.quantity, order.price, order.isBuy, order.pairId);
+      const outboundSwapClient = this.swaps.swapClientManager.get(outboundCurrency);
+      const inboundSwapClient = this.swaps.swapClientManager.get(inboundCurrency);
+
+        // check if clients exists
+      if (!outboundSwapClient) {
         throw swapsErrors.SWAP_CLIENT_NOT_FOUND(outboundCurrency);
       }
-      const maximumOutboundAmount = swapClient.maximumOutboundCapacity(outboundCurrency);
+
+      if (!inboundSwapClient) {
+        throw swapsErrors.SWAP_CLIENT_NOT_FOUND(inboundCurrency);
+      }
+
+      const maximumOutboundAmount = outboundSwapClient.maximumOutboundCapacity(outboundCurrency);
       if (outboundAmount > maximumOutboundAmount) {
         throw errors.INSUFFICIENT_OUTBOUND_BALANCE(outboundCurrency, outboundAmount, maximumOutboundAmount);
       }
--- a/lib/grpc/GrpcService.ts
+++ b/lib/grpc/GrpcService.ts
@@ -7,6 +7,7 @@ import { Order, isOwnOrder, OrderPortion, PlaceOrderResult, PlaceOrderEvent, Pla
 import { errorCodes as orderErrorCodes } from '../orderbook/errors';
 import { errorCodes as serviceErrorCodes } from '../service/errors';
 import { errorCodes as p2pErrorCodes } from '../p2p/errors';
+import { errorCodes as swapErrors } from '../swaps/errors';
 import { errorCodes as lndErrorCodes } from '../lndclient/errors';
 import { LndInfo } from '../lndclient/types';
 import { SwapSuccess, SwapFailure } from '../swaps/types';
@@ -158,6 +159,8 @@ class GrpcService {
       case orderErrorCodes.CURRENCY_CANNOT_BE_REMOVED:
       case orderErrorCodes.MARKET_ORDERS_NOT_ALLOWED:
       case serviceErrorCodes.NOMATCHING_MODE_IS_REQUIRED:
+      case orderErrorCodes.INSUFFICIENT_OUTBOUND_BALANCE:
+      case swapErrors.SWAP_CLIENT_NOT_FOUND:
         code = status.FAILED_PRECONDITION;
         break;
       case lndErrorCodes.LND_IS_UNAVAILABLE:


lib/constants/enums.ts
```diff
@@ -14,6 +14,10 @@ export enum OrderSide {
   Sell,
 }
 
+export enum RaidenTokens {
```
erkarl: We're storing the added currencies and their swap clients on a database level (Currencies). At runtime `SwapClientManager` associates all the currencies/tokens with their respective clients.

What do you think about [checking if we can get](https://github.com/ExchangeUnion/xud/blob/master/lib/swaps/SwapClientManager.ts#L86) the swap client before placing an order? That way we wouldn't have to keep a hardcoded list in the code and always be up to date even if the currencies change at runtime.

ImmanuelSegol: Thanks didn't know we had that :)

erkarl: @ImmanuelSegol I think we can just use the existing logic that's already in place:
https://github.com/ExchangeUnion/xud/blob/master/lib/orderbook/OrderBook.ts#L347

We just need to make it happen regardless of the value of `nosanitychecks` and also check the swap client for inbound currency.

ImmanuelSegol:
> What I meant from my last comment is that we don't need this additional logic in `Service.ts`. We can use the existing logic from `Orderbook.placeOrder` - https://github.com/ExchangeUnion/xud/blob/master/lib/orderbook/OrderBook.ts#L349.
> 
> We just need to make it happen regardless of the value of `nosanitychecks` (move it out of the conditional) and also check the existence of inbound currency swap client.

Sorry didn't understand you correctly its fixed now

lib/orderbook/OrderBook.ts
```diff
@@ -84,7 +85,7 @@ class OrderBook extends EventEmitter {
     return this.currencyInstances.keys();
   }
 
-  constructor(private logger: Logger, models: Models, public nomatching = false,
+  constructor(private logger: Logger, models: Models, private raidenDisabled: boolean, public nomatching = false,
```
erkarl: I believe we don't need to use this value at all because `SwapClientManager.get` will not return a swap client for the currency if the client is disabled.

lib/orderbook/OrderBook.ts
```diff
@@ -341,14 +345,22 @@ class OrderBook extends EventEmitter {
...
       const { outboundCurrency, outboundAmount } = Swaps.calculateInboundOutboundAmounts(order.quantity, order.price, order.isBuy, order.pairId);
       const swapClient = this.swaps.swapClientManager.get(outboundCurrency);
+
+      // check if client exists
       if (!swapClient) {
         throw swapsErrors.SWAP_CLIENT_NOT_FOUND(outboundCurrency);
       }
-      if (outboundAmount > swapClient.maximumOutboundCapacity) {
+
+      // Check if client is supported
```
erkarl: As mentioned above - we don't need this logic because swap client for currency won't exist when it is disabled.

lib/orderbook/OrderBook.ts
```diff
@@ -341,14 +345,22 @@ class OrderBook extends EventEmitter {
...
     }
 
-    if (!this.nosanitychecks && this.swaps) {
-      // check if sufficient outbound channel capacity exists
+    if (this.swaps) {
       const { outboundCurrency, outboundAmount } = Swaps.calculateInboundOutboundAmounts(order.quantity, order.price, order.isBuy, order.pairId);
       const swapClient = this.swaps.swapClientManager.get(outboundCurrency);
+
+      // check if client exists
       if (!swapClient) {
```
erkarl: Let's name this `outboundSwapClient` and also do a similar check for `inboundSwapClient`.

lib/orderbook/OrderBook.ts
```diff
@@ -194,6 +195,9 @@ class OrderBook extends EventEmitter {
     return tp;
   }
 
+  // Check if client radian client
+  private isRaidenToken = (client: SwapClient): boolean => client.type === SwapClientType.Raiden;
```
sangaman: We have an `isRaidenClient` that exists already you can use.

lib/swaps/errors.ts
```diff
@@ -4,6 +4,7 @@ const codesPrefix = errorCodesPrefix.SWAPS;
 const errorCodes = {
   SWAP_CLIENT_NOT_FOUND: codesPrefix.concat('.1'),
   SWAP_CLIENT_NOT_CONFIGURED: codesPrefix.concat('.2'),
+  SWAP_CLIENT_RAIDEN_DISABLED: codesPrefix.concat('.3'),
```
sangaman: Per @erkarl's comments, it sounds like we'd just want to use the `SWAP_CLIENT_NOT_FOUND` error since clients that are disabled won't be tracked in the swap client manager. More generally speaking though, I think the errors in the swaps module should be generic and not specify particular client types like Raiden.

lib/Xud.ts
```diff
@@ -81,6 +81,7 @@ class Xud extends EventEmitter {
       initPromises.push(this.swaps.init());
 
       this.orderBook = new OrderBook(loggers.orderbook, this.db.models, this.config.nomatching, this.pool, this.swaps, this.config.nosanitychecks);
+
```
sangaman: I think this is more readable without the blank line here.

sangaman:
@ImmanuelSegol Can you look into why the tests are failing? Something to do with the code merge I guess.

ImmanuelSegol:
@sangaman fixed the test

lib/orderbook/OrderBook.ts
```diff
@@ -354,15 +355,28 @@ class OrderBook extends EventEmitter {
       };
     }
 
-    if (!this.nosanitychecks) {
-      // check if sufficient outbound channel capacity exists
-      const { outboundCurrency, outboundAmount } = Swaps.calculateInboundOutboundAmounts(order.quantity, order.price, order.isBuy, order.pairId);
-      const swapClient = this.swaps.swapClientManager.get(outboundCurrency);
-      if (!swapClient) {
+    if (this.swaps && !this.nosanitychecks) {
```
sangaman: We don't need the `this.swaps` check any more since it's no longer nullable. Also don't we want to make the client checks even if `nosanitychecks` is true? We check it again further down anyway, so I think we can just drop this if statement.

lib/orderbook/OrderBook.ts
```diff
@@ -354,15 +355,28 @@ class OrderBook extends EventEmitter {
...
+      const {
+        outboundCurrency,
+        inboundCurrency,
+        outboundAmount,
+        } = Swaps.calculateInboundOutboundAmounts(order.quantity, order.price, order.isBuy, order.pairId);
```
sangaman: minor nit but there's extra indent here relative to how it's written elsewhere in the code, the `}` should line up with `const`.

lib/orderbook/OrderBook.ts
ImmanuelSegol: why remove `this.nosanitychecks` there are sanity checks.

rsercano:
Tested and seems to be working, also merged master into it again.

ImmanuelSegol:
@sangaman @kilrau This is complete as well.
If there are any desired changes please review :)

erkarl:
Let's wait for https://github.com/ExchangeUnion/xud/pull/1069 to be merged as it conflicts slightly with this one.

sangaman:
After conflicts are resolved I'll give this another quick review then we can merge.

rsercano:
Merged @sangaman

PR #992, opened 2019-05-27
fix(orderbook): nosanitycheck enabled order exchange

### LINKED ISSUE #990: `--nosanitychecks` needs restarts to exchange orders
As reported by @offerm : 

* test1 and test2 are running with `--nosanitychecks`
* on test1 I have the order bot, If I restart test2 it will not get orders from test1. But if I restart test1, it will.

ghost:
This commit fixes a bug when `nosanitychecks` enabled needed a restart to exchange orders.

It also starts the process gradually migrating Orderbook integration tests over to jest.

Closes #990

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -700,6 +700,8 @@ class OrderBook extends EventEmitter {
     if (this.nosanitychecks) {
       // we have disabled sanity checks, so assume all pairs should be activated
       pairIds.forEach(pair => peer.activePairs.add(pair));
+      // request peer's orders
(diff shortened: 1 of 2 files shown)


sangaman (approved):
This is a good catch. I think it would make sense to add an `activatePair` method to the `Peer` class (to accompany the existing `deactivatePair`) which will request the peer's orders for that pair. This would have prevented this bug in the first place I think and would similarly prevent it again, and I figure we always will want the orders for a newly activated pair. This can definitely be in a separate PR though, since this is a bug fix we shouldn't hold it up.

erkarl:
@sangaman it's something that I already considered doing and I think I will include it with relaxing the sanity checks PR.

Issue #970, opened 2019-05-21
switch off sanity swaps (as default)


kilrau:
Description below: https://github.com/ExchangeUnion/xud/issues/970#issuecomment-500808616

sangaman:
Thinking about the best approach here, we'll need to track the unbroadcasted orders somehow. I'm thinking we could use another fast priority queue, same as the order book, except ordered by quantity smallest on top. When the capacity increases, we start checking the top item if it's less than the capacity. If so, we broadcast it, remove it from the queue, and repeat for the next item in the list. If not we stop and wait until the next capacity increase. Any unbroadcast orders get added to the queue in real time.

kilrau:
As agreed on the call, we go for temporarily making the checks non-blocking with warnings as log entries and CLI outputs first. Opened https://github.com/ExchangeUnion/xud/issues/971 for better handling and making them blocking again.

kilrau:
And more:
* actual sanity swaps with preimage release should only happen with direct channel peers
* for multi-hop peers: setup htlcs, but don't release preimage. Swap partner has to be aware that this is the sanity swap test to not punish us in future for this behavior. Reason: routing fees

kilrau:
Moved sanity swap fixes to https://github.com/ExchangeUnion/xud/issues/988, this is only about making them non-blocking.

kilrau:
Decision from the call: just switch off sanity swaps on default, making them non-blocking in #988

kilrau:
I just ran into a couple of situations on simnet (where `sanitychecks=false`), where I entered an order without having a channel or sufficient channel balance. The balance checks on `placeorder` would have caught that. And as far as I know, there are no problems with these. So my strong suggestion is to do the following:
* separate config option in the category `#sanity checks` with defaults `sanityswaps=false`, `balancechecks=true`.
* This leaves the balance checks on placeorder enabled and will prevent *many* bug reports

erkarl:
> separate config option in the category

Let's discuss this during the call. In the meanwhile I've opened https://github.com/ExchangeUnion/xud/pull/1026

PR #994, opened 2019-05-28
test(orderbook): add insufficient outbound balance check

ghost:
In this PR we're softening sanity checks for `Orderbook.placeOrder`,
making them non-blocking. Previously, if the user didn't have sufficient
outbound balance then we would not allow the order to enter the
orderbook.

This handles the `placeOrder` part of https://github.com/ExchangeUnion/xud/issues/970


sangaman (commented):
I think based on discussion today around just setting `nosanitychecks` by default and not changing the existing behavior, we'd actually want to leave the code in OrderBook.ts as is. The new test is good though (just need to change it to verify that it does throw). Minor nit on that front is that the commit for the tests should be `test` rather than `chore`. Thanks!

erkarl:
@sangaman as discussed I removed the logger warning and kept the existing logic. The test case is now modify to assert that it throws an error.

sangaman (approved):
Looks great, thanks!

Issue #986, opened 2019-05-26
Swap confirmed to the user but actually failed (and xud crash)


offerm:
The user did this:
```
Trading pair: WETH/BTC
┌────────────────────────────────────────────────────┬────────────────────────────────────────────────────┐
│ Buy                                                │ Sell                                               │
├───────────────────┬───────────────────┬────────────┼───────────────────┬───────────────────┬────────────┤
│ Quantity          │ Price             │ Own Order  │ Quantity          │ Price             │ Own Order  │
├───────────────────┼───────────────────┼────────────┼───────────────────┼───────────────────┼────────────┤
│ 0.00350000        │ 0.0305            │            │ 0.00340000        │ 0.0315            │            │
└───────────────────┴───────────────────┴────────────┴───────────────────┴───────────────────┴────────────┘
xud@xud-test-2:~/scripts/raiden$ xucli buy 0.0001 WETH/BTC 0.0315
swapped 0.0001 WETH with peer order 36a0c280-7fee-11e9-9d3e-99dc63a12b46
```

He got confirmation that the swap was OK. But the log is telling a different story.
```
26/05/2019 19:41:15.327 [ORDERBOOK] debug: matched with peer 02b66438730d1fcdf4a4ae5d3d73e847a272f160fee2938e132b52cab0a0d9cfc6, executing swap on taker 3962c8b0-7fee-11e9-a77b-efa6070
e4bb6 and maker 36a0c280-7fee-11e9-9d3e-99dc63a12b46 for 10000
26/05/2019 19:41:15.358 [LND] debug: BTC: got 1 route(s) to destination 02d693d951ce4fc116b2b5e22e8e4966dd3cf9ad7600b8de24190d6af429f260c0: 292813,,315,320796811054874600,2500000000,31
5,,292813,315000,,02d693d951ce4fc116b2b5e22e8e4966dd3cf9ad7600b8de24190d6af429f260c0,,315000
26/05/2019 19:41:15.361 [SWAPS] debug: New deal: {"takerCltvDelta":1,"rHash":"fd8d5e6c7917e17f6b4523a63f0a35ad7f45b822543fd86d7eebdbbfcf4fbcd4","orderId":"36a0c280-7fee-11e9-9d3e-99dc6
[...]

Issue #984, opened 2019-05-25
LTC/BTC swap failure - Taker's xud crashed


offerm:
Failure to take an order. 
Notes:
1.  no problem with sanity swaps.
2. LTC and BTC have direct channels
2. Error on make side atter receiving the first leg.
3. XUD crashed on the taker side.

Taker issues a buy order:
```
xud@xud-test-2:/opt/xud$ xucli listorders
Trading pair: LTC/BTC
┌────────────────────────────────────────────────────┬────────────────────────────────────────────────────┐
│ Buy                                                │ Sell                                               │
├───────────────────┬───────────────────┬────────────┼───────────────────┬───────────────────┬────────────┤
│ Quantity          │ Price             │ Own Order  │ Quantity          │ Price             │ Own Order  │
├───────────────────┼───────────────────┼────────────┼───────────────────┼───────────────────┼────────────┤
│ 3.50000000        │ 0.0077            │            │ 2.50000000        │ 0.0079            │            │
├───────────────────┼───────────────────┼────────────┼───────────────────┼───────────────────┼────────────┤
│ 15.50000000       │ 0.0076            │            │ 13.00000000       │ 0.008             │            │
├───────────────────┼───────────────────┼────────────┼───────────────────┼───────────────────┼────────────┤
│ 18.00000000       │ 0.0075            │            │ 15.60000000       │ 0.0081            │            │
├───────────────────┼───────────────────┼────────────┼───────────────────┼───────────────────┼────────────┤
│ 21.25000000       │ 0.0074            │            │ 18.10000000       │ 0.0082            │            │
├───────────────────┼───────────────────┼────────────┼───────────────────┼───────────────────┼────────────┤
│ 24.00000000       │ 0.0073            │            │ 22.30000000       │ 0.0083            │            │
└───────────────────┴───────────────────┴────────────┴───────────────────┴───────────────────┴────────────┘
Trading pair: WETH/BTC
┌────────────────────────────────────────────────────┬────────────────────────────────────────────────────┐
│ Buy                                                │ Sell                                               │
├───────────────────┬───────────────────┬────────────┼───────────────────┬───────────────────┬────────────┤
│ Quantity          │ Price             │ Own Order  │ Quantity          │ Price             │ Own Order  │
└───────────────────┴───────────────────┴────────────┴───────────────────┴───────────────────┴────────────┘
xud@xud-test-2:/opt/xud$ xucli buy 0.1 LTC/BTC 0.0079 --stream
```

[...]

PR #997, opened 2019-05-28
fix(swaps): prevent SwapFailurePacket crash

ghost:
In this commit we're fixing a bug where on 2 occasions it was possible
to crash the xud process because of type mismatch when serializing the
packet.

Furthermore, we're changing the call signature of `sendErrorToPeer.sendErrorToPeer` to an object in order to prevent such mistakes in the future.

Related issues: 
https://github.com/ExchangeUnion/xud/issues/984
https://github.com/ExchangeUnion/xud/issues/986

### DIFF
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -174,7 +174,10 @@ class Swaps extends EventEmitter {
   /**
    * Sends an error to peer. Sets reqId if packet is a response to a request.
    */
-  private sendErrorToPeer = async (peer: Peer, rHash: string, failureReason: SwapFailureReason, errorMessage?: string, reqId?: string) => {
+  private sendErrorToPeer = async (
+    { peer, rHash, failureReason = SwapFailureReason.UnknownError, errorMessage, reqId }:
+    { peer: Peer, rHash: string, failureReason?: SwapFailureReason, errorMessage?: string, reqId?: string },
+  ) => {
     const errorBody: packets.SwapFailedPacketBody = {
       rHash,
       failureReason,
@@ -403,7 +406,12 @@ class Swaps extends EventEmitter {
 
     const rHash = requestPacket.body!.rHash;
     if (this.usedHashes.has(rHash)) {
-      await this.sendErrorToPeer(peer, requestPacket.body!.rHash, SwapFailureReason.PaymentHashReuse, undefined, requestPacket.header.id);
+      await this.sendErrorToPeer({
+        peer,
+        rHash,
+        failureReason: SwapFailureReason.PaymentHashReuse,
+        reqId: requestPacket.header.id,
+      });
       return false;
     }
     const requestBody = requestPacket.body!;
@@ -414,7 +422,13 @@ class Swaps extends EventEmitter {
 
     const takerSwapClient = this.swapClientManager.get(takerCurrency);
     if (!takerSwapClient) {
-      await this.sendErrorToPeer(peer, rHash, SwapFailureReason.SwapClientNotSetup, 'Unsupported taker currency', requestPacket.header.id);
+      await this.sendErrorToPeer({
+        peer,
+        rHash,
+        failureReason: SwapFailureReason.SwapClientNotSetup,
+        errorMessage: 'Unsupported taker currency',
+        reqId: requestPacket.header.id,
+      });
       return false;
     }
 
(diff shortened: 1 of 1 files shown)


lib/swaps/Swaps.ts
```diff
@@ -174,7 +174,10 @@ class Swaps extends EventEmitter {
   /**
    * Sends an error to peer. Sets reqId if packet is a response to a request.
    */
-  private sendErrorToPeer = async (peer: Peer, rHash: string, failureReason: SwapFailureReason, errorMessage?: string, reqId?: string) => {
+  private sendErrorToPeer = async ({
+    peer, rHash, failureReason = SwapFailureReason.UnknownError, errorMessage, reqId }:
+    {peer: Peer, rHash: string, failureReason?: SwapFailureReason, errorMessage?: string, reqId?: string,
```
sangaman: These lines are a little bit hard to read, could you put the open and close `{}` on the same line? It looks off to me when one of the two braces is in-line with the properties but the other is not.

lib/swaps/Swaps.ts
```diff
@@ -403,7 +406,12 @@ class Swaps extends EventEmitter {
 
     const rHash = requestPacket.body!.rHash;
     if (this.usedHashes.has(rHash)) {
-      await this.sendErrorToPeer(peer, requestPacket.body!.rHash, SwapFailureReason.PaymentHashReuse, undefined, requestPacket.header.id);
+      await this.sendErrorToPeer({
+        peer,
+        rHash: requestPacket.body!.rHash,
```
sangaman: You can just write this as `rHash,` right? Since we declare it above on line 407.

lib/swaps/Swaps.ts
```diff
@@ -448,21 +462,39 @@ class Swaps extends EventEmitter {
     // Make sure we are connected to lnd for both currencies
     if (!this.isPairSupported(deal.pairId)) {
       this.failDeal(deal, SwapFailureReason.SwapClientNotSetup);
-      await this.sendErrorToPeer(peer, deal.rHash, deal.failureReason!, deal.errorMessage, requestPacket.header.id);
+      await this.sendErrorToPeer({
+        peer,
+        rHash: deal.rHash,
```
sangaman: I think this can also just be `rHash` as well.

lib/swaps/Swaps.ts
```diff
@@ -448,21 +462,39 @@ class Swaps extends EventEmitter {
...
     }
 
     try {
       deal.makerToTakerRoutes = await takerSwapClient.getRoutes(takerAmount, takerPubKey);
     } catch (err) {
       this.failDeal(deal, SwapFailureReason.UnexpectedClientError, err.message);
-      await this.sendErrorToPeer(peer, deal.rHash, deal.failureReason!, deal.errorMessage, requestPacket.header.id);
+      await this.sendErrorToPeer({
+        peer,
+        rHash: deal.rHash,
```
sangaman: Same here for `rHash`, and lines 493 and 508.

lib/swaps/Swaps.ts
```diff
@@ -565,7 +603,11 @@ class Swaps extends EventEmitter {
       await takerSwapClient.addInvoice(deal.rHash, deal.takerAmount);
     } catch (err) {
       this.failDeal(deal, SwapFailureReason.UnexpectedClientError, err.message);
-      await this.sendErrorToPeer(peer, rHash, err.message);
+      await this.sendErrorToPeer({
+        peer,
+        rHash,
+        errorMessage: err.message,
```
sangaman: I think we can set the failure reason to `SwapFailureReason.UnexpectedClientError` here.

lib/swaps/Swaps.ts
```diff
@@ -582,7 +624,11 @@ class Swaps extends EventEmitter {
       await peer.sendPacket(new packets.SwapCompletePacket(responseBody));
     } catch (err) {
       this.failDeal(deal, SwapFailureReason.SendPaymentFailure, err.message);
-      await this.sendErrorToPeer(peer, rHash, err.message);
+      await this.sendErrorToPeer({
+        peer,
+        rHash,
+        errorMessage: err.message,
```
sangaman: Similarly we can set `SwapFailureReason.SendPaymentFailure` for the reason here.

sangaman (commented):
Thanks for diagnosing this. I just have some comments/suggestions that are mainly stylistic.

erkarl:
@sangaman thanks for the feedback. Changes made.

sangaman (approved):
Awesome, thank you!

Issue #998, opened 2019-05-28
Raiden secret back to LND


offerm:
@sangaman @karl need your help: Situation is like this - taker is doing
```
xucli buy 0.0001 WETH/BTC 0.0315
```
sending BTC, getting WETH.

maker received the BTC via hold invoice and now doing second leg (raiden payment).
Assume that this payment is good. XUD needs the result tokenPaymentResponse.secret to be able to provide it back to LND to settle the invoice.
My question - Does the code ready for the raiden response in this format:
```
{
  "secret_hash": "0x2eb110722be384d1af597f5e58de7a7b0e7bbf4c475c0f6511246d162d2aa86a",
  "identifier": 12760061994021652000,
  "initiator_address": "0x4F6799E2341379eA52213919622B9bEc80a1b655",
  "secret": "0x2ff886d47b156de00d4cad5d8c332706692b5b572adfe35e6d2f65e92906806e",
  "amount": 10,
  "token_address": "0x387d35a432Ee58a03034db5976c8412f573C387b",
  "hashalgo": "SHA256",
  "target_address": "0x133499f529A8D7cF00D8C5f7E73841DDB62b9Ca3"
}
```
**A hex string with 0x prefix?**

On the maker log I see this:
```
28/05/2019 21:46:39.925 [P2P] debug: received swapRequest from 028599d05b18c0c3f8028915a17d603416f7276c822b6b2d20e71a3502bd0f9e0a: {"proposedQuantity":10000,"pairId":"WETH/BTC","or
derId":"0e8d9980-8192-11e9-8b52-5f7ac5ad7904","rHash":"b8a0243672b503714822b454405de879e2b8300c7579d60295c35607ffd5613e","takerCltvDelta":1}
28/05/2019 21:46:39.927 [ORDERBOOK] debug: added hold of 10000 on order 0e8d9980-8192-11e9-8b52-5f7ac5ad7904
28/05/2019 21:46:39.930 [SWAPS] debug: New deal: {"proposedQuantity":10000,"pairId":"WETH/BTC","orderId":"0e8d9980-8192-11e9-8b52-5f7ac5ad7904","rHash":"b8a0243672b503714822b454405
de879e2b8300c7579d60295c35607ffd5613e","takerCltvDelta":1,"takerPubKey":"0x2B88992DEd5C96aa7Eaa9CFE1AE52350df7dc5DF","price":0.0315,"isBuy":false,"quantity":10000,"makerAmount":315
[...]

PR #999, opened 2019-05-28
fix(raiden): sendPayment secret response

ghost:
As per @offerm's feedback in https://github.com/ExchangeUnion/xud/issues/998 this commit changes the `RaidenClient.sendPayment` logic to take into account to remove `0x` from the received `tokenPaymentResponse.secret`.

Related issue: https://github.com/ExchangeUnion/xud/issues/998

### DIFF
--- a/lib/raidenclient/types.ts
+++ b/lib/raidenclient/types.ts
@@ -45,8 +45,8 @@ export type ChannelEvent = {
 /**
  * The response for tokenPayment call.
  */
-export type TokenPaymentResponse = {
-  secret: string;
+export type TokenPaymentResponse = TokenPaymentRequest & {
+  secret?: string;
 };
 
 /**
--- a/lib/raidenclient/RaidenClient.ts
+++ b/lib/raidenclient/RaidenClient.ts
@@ -95,7 +95,7 @@ class RaidenClient extends SwapClient {
       amount: 1,
       secret_hash: rHash,
     });
-    return tokenPaymentResponse.secret;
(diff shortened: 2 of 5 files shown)


sangaman (approved):

offerm:
@erkarl this is working for me (havn't looked at the change itself so review by @daniel is still pending/needed).

Can you please also add to this the handling of 0x in the when Raiden is making the http call to xud to resolve the secret_hash?

Thanks!

erkarl:
@sangaman I've updated this PR with proper error handling for `RaidenPaymentResponse`. We're assuming that the `response.secret` can be empty. In the negative case we'll throw an error. The sanitization now also includes `RaidenClient.sendSmallestAmount`.

I've also created additional test coverage for the new logic.

lib/raidenclient/types.ts
```diff
@@ -45,8 +45,10 @@ export type ChannelEvent = {
 /**
  * The response for tokenPayment call.
  */
-export type TokenPaymentResponse = {
-  secret: string;
+export type TokenPaymentResponse = TokenPaymentRequest & {
+  secret?: string;
+  initiator_address?: string;
+  hashalgo?: string;
```
sangaman: Are we going to use these 2 new fields? If not it might be better to leave them out of the codebase for simplicity's sake.

erkarl: You're right, it doesn't make sense to keep them. Updated.

sangaman (approved):
Looks good to merge to me.

PR #1000, opened 2019-05-29
feat(raiden): remove 0x from RaidenResolveRequest

### LINKED ISSUE #981: bug: http-resolver doesn't account for 0x at the beginning of secret_hash
The `secret_hash` value we receive from Raiden to http resolver always starts with `0x`. As a result the http resolver never matches the internal hash.

ghost:
In this commit we're removing "0x" from the request that Raiden sends to
our resolver in order to ensure that it matches with our internal hash
map.

Closes: https://github.com/ExchangeUnion/xud/issues/981

### DIFF
--- a/lib/swaps/errors.ts
+++ b/lib/swaps/errors.ts
@@ -4,6 +4,7 @@ const codesPrefix = errorCodesPrefix.SWAPS;
 const errorCodes = {
   SWAP_CLIENT_NOT_FOUND: codesPrefix.concat('.1'),
   SWAP_CLIENT_NOT_CONFIGURED: codesPrefix.concat('.2'),
+  PAYMENT_HASH_NOT_FOUND: codesPrefix.concat('.3'),
 };
 
 const errors = {
@@ -14,6 +15,10 @@ const errors = {
   SWAP_CLIENT_NOT_CONFIGURED: (currency: string) => ({
     message: `swapClient for currency ${currency} is not configured`,
     code: errorCodes.SWAP_CLIENT_NOT_CONFIGURED,
+  }),
+  PAYMENT_HASH_NOT_FOUND: (rHash: string) => ({
+    message: `deal for rHash ${rHash} not found`,
+    code: errorCodes.PAYMENT_HASH_NOT_FOUND,
   }),
 };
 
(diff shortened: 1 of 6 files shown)


lib/swaps/errors.ts
```diff
@@ -15,6 +16,10 @@ const errors = {
     message: `swapClient for currency ${currency} is not configured`,
     code: errorCodes.SWAP_CLIENT_NOT_CONFIGURED,
   }),
+  SECRET_HASH_NOT_FOUND: (hash: string) => ({
```
sangaman: As far as naming goes, "not found" sounds like we're unable to find that secret hash internally, instead of it being unspecified on a request that we receive. Also the "secret hash" terminology seems exclusive to raiden, everywhere else we are using `rHash` and in fact there is a check in the `Service.resolveHash()` method to ensure that `rHash` has a value (so it would catch the times that it's not present). Having this in `swaps/errors.ts` seems slightly misplaced.

In #999 you make a raiden error called `INVALID_TOKEN_PAYMENT_RESPONSE` for when the secret hash is missing. Maybe you can use a raiden error called `INVALID_RESOLVE_REQUEST` here?

sangaman (commented):
Looks good, just 1 comment regarding the new error.

lib/swaps/errors.ts
erkarl: @sangaman you're right - the naming for this was bad. Just to clarify - this is not checking the inputs from Raiden, just the output.

I changed it to `SWAP_DEAL_NOT_FOUND` and moved the error throwing logic into `Swaps`. I still feel like `swaps/errors` is the correct place for this because it's now being thrown from `Swaps.resolveHash`. I think the "not found" terminology is correct because Raiden requests a resource (secret) from xud. If we cannot find the resource it makes sense to respond with 404 not found.

sangaman: Ok it makes much more sense now, maybe I misunderstood earlier. Ironically tho, I now think that something more generic like `PAYMENT_HASH_NOT_FOUND` might be more appropriate, since we use this error for sanity swaps as well which don't have "deal" involved per se. Just a suggestion, definitely not a deal breaker, and otherwise this all looks good.

erkarl: Updated the error to `PAYMENT_HASH_NOT_FOUND`.

sangaman (approved):

PR #1001, opened 2019-05-30
refactor: make peer activePairs private

sangaman:
This commit adds an `activatePair` public method for the `Peer` class and makes the `activePairs` property private. `activatePair` will both add a pair to the set of active pairs as well as request existing orders for that pair from the peer.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -15,7 +15,7 @@ import { SwapRole, SwapFailureReason, SwapPhase, SwapClientType } from '../const
 import { CurrencyInstance, PairInstance, CurrencyFactory } from '../db/types';
 import { Pair, OrderIdentifier, OwnOrder, OrderPortion, OwnLimitOrder, PeerOrder, Order, PlaceOrderEvent,
   PlaceOrderEventType, PlaceOrderResult, OutgoingOrder, OwnMarketOrder, isOwnOrder, IncomingOrder } from './types';
-import { SwapRequestPacket, SwapFailedPacket, GetOrdersPacket } from '../p2p/packets';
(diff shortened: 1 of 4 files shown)


lib/orderbook/OrderBook.ts
```diff
@@ -745,15 +742,9 @@ class OrderBook extends EventEmitter {
     pairIds.forEach(async (pairId) => {
       const [baseCurrency, quoteCurrency] = pairId.split('/');
       if (peer.verifiedCurrencies.has(baseCurrency) && peer.verifiedCurrencies.has(quoteCurrency)) {
-        peer.activePairs.add(pairId);
-        verifiedPairs.push(pairId);
+        await peer.activatePair(pairId);
```
erkarl: Minor nit: we could do a `pairsToActivate.push(pairId)` -> `await Promise.all(pairsToActivate)` here for slightly better performance in exchange for slightly worse readability. Up to you - I don't mind the way it currently is.

test/simulation/tests.go
```diff
@@ -189,7 +189,7 @@ func testOrderMatchingAndSwap(net *xudtest.NetworkHarness, ht *harnessTest) {
 		Price:    0.02,
 		Quantity: 1000000,
 		PairId:   "LTC/BTC",
-		Side:     xudrpc.OrderSide_BUY,
+		Side:     xudrpc.OrderSide_SELL,
```
erkarl: Why is this change required?

sangaman: That's an unintentional leftover, thanks for catching.

lib/orderbook/OrderBook.ts
sangaman: I changed it to use `Promise.all` - not a big deal either way I agree but I like sticking to a pattern of awaiting promises in parallel when they don't depend on each other.

Issue #948, opened 2019-05-07
Limit trade size


kilrau:
Hardcoded limits based on lnd/raiden client hard-coded limits, which we maintain in xud, enforced on:
* on placeorder
* on swap

Simnet installer needs to adjust these values to allow higher limits @offerm @reliveyy

PR #1002, opened 2019-05-30
Limit trade & order sizes for mainnet

ImmanuelSegol:
#948

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -8,15 +8,16 @@ import { errors as swapsErrors } from '../swaps/errors';
 import Pool from '../p2p/Pool';
 import Peer from '../p2p/Peer';
 import Logger from '../Logger';
-import { ms, derivePairId, setTimeoutPromise } from '../utils/utils';
+import { derivePairId, ms, setTimeoutPromise } from '../utils/utils';
 import { Models } from '../db/DB';
 import Swaps from '../swaps/Swaps';
-import { SwapRole, SwapFailureReason, SwapPhase, SwapClientType } from '../constants/enums';
-import { CurrencyInstance, PairInstance, CurrencyFactory } from '../db/types';
-import { Pair, OrderIdentifier, OwnOrder, OrderPortion, OwnLimitOrder, PeerOrder, Order, PlaceOrderEvent,
-  PlaceOrderEventType, PlaceOrderResult, OutgoingOrder, OwnMarketOrder, isOwnOrder, IncomingOrder, OrderBookThresholds } from './types';
-import { SwapRequestPacket, SwapFailedPacket } from '../p2p/packets';
-import { SwapSuccess, SwapDeal, SwapFailure } from '../swaps/types';
+import limits from '../constants/limits';
+import { SwapClientType, SwapFailureReason, SwapPhase, SwapRole, XuNetwork } from '../constants/enums';
+import { CurrencyFactory, CurrencyInstance, PairInstance } from '../db/types';
+import { IncomingOrder, isOwnOrder, Order, OrderBookThresholds, OrderIdentifier, OrderPortion, OutgoingOrder, OwnLimitOrder, OwnMarketOrder,
+  OwnOrder, Pair, PeerOrder, PlaceOrderEvent, PlaceOrderEventType, PlaceOrderResult } from './types';
+import { SwapFailedPacket, SwapRequestPacket } from '../p2p/packets';
+import { SwapDeal, SwapFailure, SwapSuccess } from '../swaps/types';
 // We add the Bluebird import to ts-ignore because it's actually being used.
 // @ts-ignore
 import Bluebird from 'bluebird';
@@ -374,25 +375,37 @@ class OrderBook extends EventEmitter {
       };
     }
 
-    if (!this.nobalancechecks) {
-      // check if sufficient outbound channel capacity exists
-      const { outboundCurrency, inboundCurrency, outboundAmount } =
-          Swaps.calculateInboundOutboundAmounts(order.quantity, order.price, order.isBuy, order.pairId);
-      const outboundSwapClient = this.swaps.swapClientManager.get(outboundCurrency);
-      const inboundSwapClient = this.swaps.swapClientManager.get(inboundCurrency);
+    const { outboundCurrency, inboundCurrency, outboundAmount, inboundAmount } =
+        Swaps.calculateInboundOutboundAmounts(order.quantity, order.price, order.isBuy, order.pairId);
+    const outboundSwapClient = this.swaps.swapClientManager.get(outboundCurrency);
+    const inboundSwapClient = this.swaps.swapClientManager.get(inboundCurrency);
 
-        // check if clients exists
+    if (!this.nobalancechecks) {
+      // check if clients exists
       if (!outboundSwapClient) {
         throw swapsErrors.SWAP_CLIENT_NOT_FOUND(outboundCurrency);
       }
-
       if (!inboundSwapClient) {
         throw swapsErrors.SWAP_CLIENT_NOT_FOUND(inboundCurrency);
       }
 
+      // check if sufficient outbound channel capacity exists
       const maximumOutboundAmount = outboundSwapClient.maximumOutboundCapacity(outboundCurrency);
       if (outboundAmount > maximumOutboundAmount) {
         throw errors.INSUFFICIENT_OUTBOUND_BALANCE(outboundCurrency, outboundAmount, maximumOutboundAmount);
+      }
+    }
+
+    if (this.pool.getNetwork() === XuNetwork.MainNet) {
+      // check if order abides by limits
+      const outboundCurrencyLimit = limits[outboundCurrency];
+      if (outboundCurrencyLimit && outboundAmount > outboundCurrencyLimit) {
+        throw errors.EXCEEDING_LIMIT(outboundCurrency, outboundAmount, outboundCurrencyLimit);
+      }
+
+      const inboundCurrencyLimit = limits[inboundCurrency];
+      if (inboundCurrencyLimit && inboundAmount > inboundCurrencyLimit) {
+        throw errors.EXCEEDING_LIMIT(inboundCurrency, inboundAmount, inboundCurrencyLimit);
       }
     }
 
--- a/lib/orderbook/errors.ts
+++ b/lib/orderbook/errors.ts
@@ -15,6 +15,7 @@ const errorCodes = {
   CURRENCY_MISSING_ETHEREUM_CONTRACT_ADDRESS: codesPrefix.concat('.11'),
   INSUFFICIENT_OUTBOUND_BALANCE: codesPrefix.concat('.12'),
   MIN_QUANTITY_VIOLATED: codesPrefix.concat('.13'),
+  EXCEEDING_LIMIT: codesPrefix.concat('.14'),
 };
 
 const errors = {
@@ -69,6 +70,10 @@ const errors = {
   MIN_QUANTITY_VIOLATED: (id: string) => ({
     message: `order ${id} has surpassed the minimum quantity`,
     code: errorCodes.MIN_QUANTITY_VIOLATED,
+  }),
+  EXCEEDING_LIMIT: (currency: string, amount: number, limit: number) => ({
+    message: `outbound amount ${amount} is exceeding limit of ${limit} set for ${currency}`,
+    code: errorCodes.EXCEEDING_LIMIT,
   }),
 };
 
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -367,6 +367,15 @@ class Pool extends EventEmitter {
   }
 
   /**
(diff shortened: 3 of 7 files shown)


ImmanuelSegol:
@sangaman again the integration test are failing on Travis locally they pass.

lib/Config.ts
```diff
@@ -112,6 +114,10 @@ class Config {
       host: 'localhost',
       port: 5001,
     };
+    this.limits = {
+      lnd: 0.04 ,
+      raiden: 1,
```
sangaman: These should be in the smallest base units (since they get compared to the amounts of the payments which are expressed in these terms), not whole units, and also on a per currency basis. For btc the maximum is 4294967 satoshis, I'm actually not sure about ltc but you should be able to research it. For WETH @moshababo should be able to confirm.

lib/Config.ts
```diff
@@ -152,6 +158,11 @@ class Config {
         this.updateMacaroonPaths();
       }
 
+      if (props.limits) {
```
sangaman: Is this something we want to make configurable? That's not the impression I was under and it's not mentioned in the issue. @kilrau

lib/orderbook/types.ts
```diff
@@ -1,6 +1,11 @@
 import { SwapClientType } from '../constants/enums';
 import { SwapSuccess, SwapFailure } from '../swaps/types';
 
+export type Limits = {
```
sangaman: I don't think we need a new type for this. We can just create an object with hard-coded properties and the compiler will know that they're there.

test/integration/OrderBook.spec.ts
```diff
@@ -62,7 +67,7 @@ describe('OrderBook', () => {
       }
       return client;
     };
-    orderBook = new OrderBook(loggers.orderbook, db.models, false, pool, swaps);
+    orderBook = new OrderBook(loggers.orderbook, db.models, limits, false, pool, swaps);
```
sangaman: Rather than pass the limits around in constructors (which already are taking lots of parameters) I think we can put the limits in the `constants` folder and `import` them where they are needed.

kilrau:
Update? @ImmanuelSegol Raiden limit is 1 wETH.

lib/Config.ts
ImmanuelSegol: @sangaman correct I don't think we want this to be configurable for now.

kilrau:
only on mainnet, not testnet & simnet

rsercano:
I've managed to merge master into the branch. Afterwards tried to execute `npm run test` and as @ImmanuelSegol mentioned it successfully works locally. But fails with one single test in travis with the following issue;

```
 1) Parser
       test TCP segmentation/concatenation support
         should parse encrypted Ping SessionInit Ping concatenated and split on byte 17 from each packet beginning:
      AssertionError: expected '"{\\"body\\":{\\"ephemeralPubKey\\":\\"f475ef72-a54a-47df-baab-52f2911f8a0b\\",\\"nodeState\\":{\\"addresses\\":[{\\"host\\":\\"1.1.1.1\\",\\"port\\":8885},{\\"host\\":\\"2.2.2.2\\",\\"port\\":8885}],\\"lndPubKeys\\":{\\"BTC\\":\\"263f01e8-2040-4891-a2d2-a131f001b00e\\",\\"LTC\\":\\"e0e29dcd-83eb-4521-93d1-f3d6e34befb6\\"},\\"nodePubKey\\":\\"6b4ac35e-d5f3-47fc-8c04-99f18fd4cdf7\\",\\"pairs\\":[\\"ee6416d1-457e-40ec-8e85-85c69c7e991e\\"],\\"raidenAddress\\":\\"05a212bd-919b-4b2e-bd49-0aba861c2b88\\",\\"version\\":\\"1.0.0\\"},\\"peerPubKey\\":\\"077a101c-fd11-43ad-877c-27925044c79b\\",\\"sign\\":\\"d5fd71ab-4a4e-486f-9755-4ee1c134708f\\"},\\"header\\":{\\"id\\":\\"608f4ab9-9742-11e9-95f8-d39f4e011977\\"}}"' to equal '"{\\"header\\":{\\"id\\":\\"608f4ab8-9742-11e9-95f8-d39f4e011977\\"}}"'
      + expected - actual
      -"{\"body\":{\"ephemeralPubKey\":\"f475ef72-a54a-47df-baab-52f2911f8a0b\",\"nodeState\":{\"addresses\":[{\"host\":\"1.1.1.1\",\"port\":8885},{\"host\":\"2.2.2.2\",\"port\":8885}],\"lndPubKeys\":{\"BTC\":\"263f01e8-2040-4891-a2d2-a131f001b00e\",\"LTC\":\"e0e29dcd-83eb-4521-93d1-f3d6e34befb6\"},\"nodePubKey\":\"6b4ac35e-d5f3-47fc-8c04-99f18fd4cdf7\",\"pairs\":[\"ee6416d1-457e-40ec-8e85-85c69c7e991e\"],\"raidenAddress\":\"05a212bd-919b-4b2e-bd49-0aba861c2b88\",\"version\":\"1.0.0\"},\"peerPubKey\":\"077a101c-fd11-43ad-877c-27925044c79b\",\"sign\":\"d5fd71ab-4a4e-486f-9755-4ee1c134708f\"},\"header\":{\"id\":\"608f4ab9-9742-11e9-95f8-d39f4e011977\"}}"

```

Seems like a headache..

rsercano:
Ok. After restarting the build in travis it successfully worked, but it's failing on sim tests now, I guess at least the first mentioned issue is related with master branch. I'm now trying to execute simnet tests locally.

lib/Config.ts
sangaman: Wouldn't we want to remove this code then?

sangaman:
@rsercano See my original comments above, @ImmanuelSegol  marked them as resolved but none of the comments were actually addressed. I've since resolved them, and they should all be addressed. The first comment of mine above should explain at least in part why the tests are failing.

The failing tests definitely should be handled in this repo, in its current state we couldn't make any swaps at all with this branch because the hardcoded limit of `0.04` for lnd is currently less than 1 satoshi.

kilrau:
Limits should be (please double check)
* BTC: 4294967 satoshis = 0.04294967 BTC (enforced by LND/LN)
* LTC: 257698020 litoshis = 2.57 LTC (x60 according to https://github.com/lightningnetwork/lnd/pull/1246/files, enforced by LND/LN)
* ETH: 1 WETH, 500 DAI (set freely by us)

rsercano:
I assume this's done after finding the maximum transfer amounts for currencies; **LTC, WETH, DAI**. 
**BTC** is set to **4294967** satoshies as suggested by @sangaman

rsercano:
I guess this's done, passing tests, all revisions suggested by @sangaman have been implemented.

lib/orderbook/errors.ts
```diff
@@ -65,6 +66,10 @@ const errors = {
     message: `${currency} outbound balance of ${availableAmount} is not sufficient for order amount of ${amount}`,
     code: errorCodes.INSUFFICIENT_OUTBOUND_BALANCE,
   }),
+  EXCEEDING_LIMIT: (currency: string, amount: number, limit: number) => ({
+    message: `outbound amount: ${amount} is exceeding limit of: ${limit} set for ${currency}`,
```
sangaman: The extra `:` in here make this message kind of hard to read, I think we can just drop both of them.

lib/orderbook/errors.ts
```diff
@@ -65,6 +66,10 @@ const errors = {
     message: `${currency} outbound balance of ${availableAmount} is not sufficient for order amount of ${amount}`,
     code: errorCodes.INSUFFICIENT_OUTBOUND_BALANCE,
   }),
+  EXCEEDING_LIMIT: (currency: string, amount: number, limit: number) => ({
```
sangaman: We should add this to the `INVALID_ARGUMENT` case in `getGrpcError` in GrpcService.ts.

lib/constants/enums.ts
```diff
@@ -138,3 +138,10 @@ export enum DisconnectionReason {
   AuthFailureInvalidSignature = 12,
   WireProtocolErr = 13,
 }
+
+export enum Limits {
```
sangaman: An `enum` doesn't really make sense here, and will break down once we have 2 currencies with the same limit. This should just be a plain object mapping currency (upper case like `BTC`) to these values. You can move it to a `limits.ts` file that just exports that object.

lib/orderbook/OrderBook.ts
```diff
@@ -857,6 +863,25 @@ class OrderBook extends EventEmitter {
       }, requestPacket.header.id));
     }
   }
+
+  /**
+   * Returns maximum transfer amount for a swap operation for the given outbound currency.
+   * @param outboundCurrency the outbound currency during swap
+   */
+  private getLimit = (outboundCurrency: string): any => {
```
sangaman: This method should be unnecessary with the change above to replace the `enum` with a plain object. Just give that object a `{ [currency: string]: number }` type and we can refer to the `limits` object that we import directly like `limits[currency]`.

test/simulation/go.mod
```diff
@@ -14,10 +16,11 @@ require (
 	github.com/roasbeef/btcd v0.0.0-20180418012700-a03db407e40d
 	github.com/roasbeef/btcutil v0.0.0-20180406014609-dfb640c57141
 	github.com/stretchr/testify v1.3.0
-	golang.org/x/crypto v0.0.0-20190513172903-22d7a77e9e5f // indirect
-	golang.org/x/net v0.0.0-20190514140710-3ec191127204
-	golang.org/x/sys v0.0.0-20190514135907-3a4b5fb9f71f // indirect
+	golang.org/x/crypto v0.0.0-20190621222207-cc06ce4a13d4 // indirect
+	golang.org/x/net v0.0.0-20190620200207-3b0461eec859
+	golang.org/x/sys v0.0.0-20190624142023-c5567b49c5d0 // indirect
```
sangaman: I think your go version might be slightly different or maybe there have been some updates, but I don't think we should update this as part of the PR. Please revert or remove any changes to the go.mod/go.sum.

rsercano:
Added requested changes @sangaman

kilrau:
As per call today we would need
1) this feature for the bug bounty release to set limits for mainnet only. Currently limits apply for all networks.
2) and lower the limits to
42949 satoshis (0.00042949 BTC, ~5 USD)
50000 litoshi (0.05 LTC, ~ 5 USD)
0.02 eth (~ 5 USD)
5 DAI (set freely by us)

rsercano:
Added requested changes @kilrau

lib/p2p/Pool.ts
```diff
@@ -366,6 +366,15 @@ class Pool extends EventEmitter {
     return false;
   }
 
+  /**
+   * Current network information that comes from xud configuration
+   *
+   * @return returns currently configured network
```
sangaman: This shouldn't need "return" twice so just `@returns the active xu network` or similar.

lib/orderbook/OrderBook.ts
```diff
@@ -385,6 +386,12 @@ class OrderBook extends EventEmitter {
       if (outboundAmount > maximumOutboundAmount) {
         throw errors.INSUFFICIENT_OUTBOUND_BALANCE(outboundCurrency, outboundAmount, maximumOutboundAmount);
       }
+
+      // Check if order abides to limits
```
sangaman: Minor nit: this should be "abides by"

test/simulation/go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/ExchangeUnion/xud-simulation
 
 require (
 	github.com/Roasbeef/btcutil v0.0.0-20180406014609-dfb640c57141 // indirect
-	github.com/btcsuite/btcd v0.0.0-20190629003639-c26ffa870fd8
```
sangaman: Revert this file please.

test/simulation/go.sum
```diff
@@ -20,6 +20,8 @@ github.com/btcsuite/btcd v0.0.0-20190614013741-962a206e94e9 h1:18Pe+JPyglNO9lzLr
 github.com/btcsuite/btcd v0.0.0-20190614013741-962a206e94e9/go.mod h1:3J08xEfcugPacsc34/LKRU2yO7YmuT8yt28J8k2+rrI=
 github.com/btcsuite/btcd v0.0.0-20190629003639-c26ffa870fd8 h1:mOg8/RgDSHTQ1R0IR+LMDuW4TDShPv+JzYHuR4GLoNA=
 github.com/btcsuite/btcd v0.0.0-20190629003639-c26ffa870fd8/go.mod h1:3J08xEfcugPacsc34/LKRU2yO7YmuT8yt28J8k2+rrI=
+github.com/btcsuite/btcd v0.0.0-20190807005414-4063feeff79a h1:We35J+0yvVFrEXbtViYUW8H/wNOhqjIF3PsrW4yYmGw=
```
sangaman: And this as well.

kilrau:
> but it looks like the simulation tests go module files changed

Looks like this is fixed. Good to go!

PR #1005, opened 2019-06-03
refactor(orderbook): pass object to constructor

sangaman:
This refactors the constructor for `OrderBook` to accept a single object rather than 6 parameters. This brings the constructor in line with other recently changed constructors/methods that accept 4+ parameters.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -61,6 +61,8 @@ interface OrderBook {
 class OrderBook extends EventEmitter {
   /** A map between active trading pair ids and trading pair instances. */
   public tradingPairs = new Map<string, TradingPair>();
+  public nomatching: boolean;
+
   /** A map between own orders local id and their global id. */
   private localIdMap = new Map<string, OrderIdentifier>();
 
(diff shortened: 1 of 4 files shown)


lib/orderbook/OrderBook.ts
```diff
@@ -84,10 +90,27 @@ class OrderBook extends EventEmitter {
...
+    pool: Pool,
+    swaps: Swaps,
+    nosanitychecks?: boolean,
+  }) {
     super();
 
+    this.logger = logger;
+    this.pool = pool;
+    this.swaps = swaps;
+    if (nomatching !== undefined) {
```
erkarl: What do you think about removing this and `nosanitychecks !== undefined` conditional and assigning them a default value when destructuring?

```
...
  constructor(
    { logger, models, pool, swaps, nomatching = false, nosanitychecks = false }:
    {
      logger: Logger,
      models: Models,
      pool: Pool,
      swaps: Swaps,
      nomatching: boolean,
      nosanitychecks: boolean,
    },
  ) {
...
```

erkarl: Up to you!

sangaman: OK I went ahead and changed it back and decided it's a little bit cleaner, going with your approach above.

PR #1006, opened 2019-06-04
 feat: xud master password

sangaman:
This is a major commit that introduces new functionality to encrypt the xud node key with a password which can also be used to encrypt lnd and (with a separate PR) Raiden.

A new config option `noencrypt` is added and for now defaults to `true`, meaning that this functionality is off by default. This is to minimize disruption with existing applications of xud. When `noencrypt` is true, starting up xud occurs as it does before this PR.

When `noencrypt` is false, we first check if there is an existing node key file saved to disk. If one exists, we wait to receive a new `UnlockNode` call to provide us with a password. We use this password to decrypt the node key file and to call `UnlockWallet` on each configured lnd instance that is waiting to be unlocked.

If none is found, we wait to receive a new `CreateNode` call to provide us with a password. We then attempt to call `GenSeed` on a configured lnd instance to generate an aezeed 24 word mnemonic. We use this mnemonic to generate our node key, and then encrypt that with our password and save to disk. We then call `InitWallet` on each configured lnd instance that is waiting to be unlocked, using the same mnemonic and password from earlier steps.

As part of this PR, we introduce a new grpc `XudInitService` for calls when xud is in a state of waiting for a password before it can complete initialization. New `create` and `unlock` commands are added for `xucli` as well.

This is a major step towards #912.

Note: I'm still diagnosing an issue with the simulation tests whereby the `GetInfo` call is not returning the lnd information, I'll comment when I've fixed it.


sangaman:
TODO items related to this PR, either to be addressed in this or future PRs:

- Regression tests for new functionailty
- Simulations tests to use encryption
- Better error handling for incorrect passwords and other edge cases
- Allow user to provide own seed when creating xud node
- `CreateNodeResponse` and/or `UnlockNodeResponse` to indicate which swap clients were initialized/unlocked?
- Hide password input for `xucli unlock` and `xucli create`
- Valid arg checks for InitService calls
- Alternative mnemonic generation if we don't have an lnd client available for GenSeed? Right now we won't get a mnemonic without a working lnd.
- Recover xud key from mneominc (extend seedutil functionality?)
- Generate raiden address (after seedutil merged)
- Support aezeed passphrase encryption?

sangaman:
I rebased the PR on master, added some unit tests for `NodeKey`, and fixed the issue that was causing sim tests to fail.

kilrau:
Well thought-out todos :1st_place_medal: ! Here my thoughts on what to do within the PR, what separate and what later @sangaman @erkarl 

> TODO items related to this PR, either to be addressed in this or future PRs:
> 
>     * Regression tests for new functionailty
I think this should be done in this PR.

>     * Simulations tests to use encryption
Separate issue

>     * Better error handling for incorrect passwords and other edge cases
This PR

>     * Allow user to provide own seed when creating xud node
Separate issue, `post-1.0.0`

>     * `CreateNodeResponse` and/or `UnlockNodeResponse` to indicate which swap clients were initialized/unlocked?
Separate issue

>     * Hide password input for `xucli unlock` and `xucli create`
This PR

>     * Valid arg checks for InitService calls
This PR 
>     * Alternative mnemonic generation if we don't have an lnd client available for GenSeed? Right now we won't get a mnemonic without a working lnd.
Separate issue, `post-1.0.0`

>     * Recover xud key from mneominc (extend seedutil functionality?)
Separate issue

>     * Generate raiden address (after seedutil merged)
This PR

>     * Support aezeed passphrase encryption?
This PR

>     * CLI enhancement, right now `CreateNode` just spits out the 24 word seed in JSON, but we should probably have a message instructing people to write it down, keep it safe, what they can recover with it and how, etc etc...
Separate issue

>     * Update README/wiki/docs with basic instructions on how to set up and use the master password
Separate issue

erkarl:
Some general feedback without digging deep into the code:

When trying to create when already created:
```
➜  xud git:(master-key) ✗ ./bin/xucli create hi2
Error: 12 UNIMPLEMENTED: call is not supported by the current state of xud
```

When entering a wrong password:
```
➜  xud git:(master-key) ✗ ./bin/xucli unlock wrongpw
Error: 2 UNKNOWN: error:06065064:digital envelope routines:EVP_DecryptFinal_ex:bad decrypt
```

When lnd is already unlocked:
```
➜  xud git:(master-key) ✗ ./bin/xucli create mypasswordisusername
Error: 2 UNKNOWN: 12 UNIMPLEMENTED: unknown service lnrpc.WalletUnlocker
➜  xud git:(master-key) ✗ ./bin/xucli unlock hi
Error: 12 UNIMPLEMENTED: call is not supported by the current state of xud
```
Should we change these error messages to be more user-friendly?

When I start lnd in a completely fresh state it switches its state to disabled, falling back to generating without seed.
```
07/06/2019 20:08:27.902 [LND] error: BTC: could not load lnd macaroon, is lnd installed?
07/06/2019 20:08:27.903 [LND] info: BTC: LndClient status: Disabled
➜  xud git:(master-key) ✗ ./bin/xucli create test
{
  "seedMnemonicList": []
}
```

lib/cli/commands/create.ts
```diff
@@ -0,0 +1,20 @@
+import { Arguments } from 'yargs';
+import { callback, loadXudInitClient } from '../command';
+import { CreateNodeRequest } from '../../proto/xudrpc_pb';
+
+export const command = 'create <password>';
```
erkarl: As you mentioned in your TODO list we could hide this input and perhaps also confirm the password?

lib/Xud.ts
```diff
@@ -65,8 +67,34 @@ class Xud extends EventEmitter {
...
+        const initRpcServer = new GrpcServer(loggers.rpc);
+        initRpcServer.addXudInitService(initService);
+        await initRpcServer.listen(
+          this.config.rpc.port,
+          this.config.rpc.host,
+          path.join(this.config.xudir, 'tls.cert'),
+          path.join(this.config.xudir, 'tls.key'),
+        );
+
+        this.logger.info('awaiting password...');
```
erkarl: For better user experience I'd mention the `xucli unlock/create` commands here like lnd does.

lib/cli/commands/create.ts
sangaman: Yup, although I am thinking it should be for a separate PR since the CLI is a separate tool plus I don't think it would be a trivial change.

sangaman:
> Should we change these error messages to be more user-friendly?

Yeah I think so, I will do that soon.

> When I start lnd in a completely fresh state it switches its state to disabled, falling back to generating without seed.

sangaman:
> Well thought-out todos 1st_place_medal ! Here my thoughts on what to do within the PR, what separate and what later @sangaman @erkarl . If you don't agree on some, let me know, otherwise I'll go ahead and open these new issues soonish

I agree with your breakdown except I think I'd leave the CLI enhancements (including hidden password input) for a separate issue as well, it doesn't seem trivial to me plus the CLI is a standalone tool so logically it's simple to keep separate.

Also for aezeed passphrase encryption, I wasn't so clear on how we should go about that. Should we ask for a separate password? Or encrypt the mnemonic with the same password we use to encrypt the lnd wallets and xud node key?

kilrau:
### Here the open todo's for this PR:
>     * Regression tests for new functionailty
>     * Better error handling for incorrect passwords and other edge cases
>     * Valid arg checks for InitService calls
>     * Generate raiden address (after seedutil merged)
>     * Support aezeed passphrase encryption (same master password)

#### Here the separate issues for `1.0.0`:
>     * Simulations tests to use encryption
#1016
>     * `CreateNodeResponse` and/or `UnlockNodeResponse` to indicate which swap clients were initialized/unlocked?
#1018
>     * Hide password input for `xucli unlock` and `xucli create`
#1021 
>     * CLI enhancement, right now `CreateNode` just spits out the 24 word seed in JSON, but we should probably have a message instructing people to write it down, keep it safe, what they can recover with it and how, etc etc...
#1021 
>     * Recover xud key from mneominc (extend seedutil functionality?)
#1020 
>     * Update README/wiki/docs with basic instructions on how to set up and use the master password
#1022

#### And `post-1.0.0`:
>     * Allow user to provide own seed when creating xud node
#1017
>     * Alternative mnemonic generation if we don't have an lnd client available for GenSeed? Right now we won't get a mnemonic without a working lnd.
#1019

erkarl:
> I agree with your breakdown except I think I'd leave the CLI enhancements (including hidden password input) for a separate issue as well, it doesn't seem trivial to me plus the CLI is a standalone tool so logically it's simple to keep separate.

Agreed.

> Also for aezeed passphrase encryption, I wasn't so clear on how we should go about that. Should we ask for a separate password? Or encrypt the mnemonic with the same password we use to encrypt the lnd wallets and xud node key?

I think a good practice would be to keep `xud unlock` password separate from seed restore password (can also be empty in which case the pw will be `aezeed`).

lib/lndclient/LndClient.ts
```diff
@@ -387,6 +414,34 @@ class LndClient extends SwapClient {
...
 
+  public genSeed = async (): Promise<lndrpc.GenSeedResponse.AsObject> => {
+    const genSeedResponse = await this.unaryWalletUnlockerCall<lndrpc.GenSeedRequest, lndrpc.GenSeedResponse>(
+      'genSeed', new lndrpc.GenSeedRequest(),
+    );
+    return genSeedResponse.toObject();
+  }
+
+  public initWallet = async (walletPassword: string, seedMnemonic: string[]): Promise<lndrpc.InitWalletResponse.AsObject> => {
+    const request = new lndrpc.InitWalletRequest();
```
erkarl: What do you think about extending `seedutil` to generate the seed for us? We already have `aezeed` package imported and it would be really trivial to extend `seedutil` to support `seedutil create [optinal-pw]`. That way we wouldn't have to rely on lnd to be enabled, connected AND in a specific state.

erkarl: Nevermind, we still need to `init` lnd wallet somehow.

erkarl: This comment is still valid for `swapClientManager.genSeed`, though.

kilrau:
> I think a good practice would be to keep xud unlock password separate from seed restore password (can also be empty in which case the pw will be aezeed).

What exact attack scenario does it prevent? Best practise would be to have separate passwords and seeds for all clients.

lib/swaps/SwapClientManager.ts
```diff
@@ -103,6 +103,63 @@ class SwapClientManager extends EventEmitter {
     }
   }
 
+  /**
+   * Generates a cryptographically random 24 word seed mnemonic from an lnd client.
+   */
+  public genSeed = async () => {
+    // loop through swap clients until we find a connected lnd client
+    for (const swapClient of this.swapClients.values()) {
+      // if (isLndClient(swapClient) && swapClient.isConnected()) {
```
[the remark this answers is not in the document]
sangaman: Actually I believe I meant to use this line and delete the line below, I'll do that, thanks.

package.json
```diff
@@ -118,7 +118,6 @@
     "cli-table3": "^0.5.1",
     "colors": "^1.3.3",
     "cross-os": "^1.3.0",
-    "crypto-js": "^3.1.9-1",
```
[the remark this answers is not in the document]
sangaman: Yup I figured it wasn't really necessary and I could reuse the same approach to encryption we're using for the p2p layer.

lib/service/InitService.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (this.pendingCall) {
+      throw errors.PENDING_CALL_CONFLICT;
+    }
+
+    this.pendingCall = true;
+    const seed = await this.swapClientManager.genSeed();
+    let nodeKey: NodeKey;
+
+    if (seed) {
+      const seedBytes = typeof seed.encipheredSeed === 'string' ?
```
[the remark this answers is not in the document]
sangaman: I actually mainly wrote it this way for the compiler. According to the type definitions we're using currently, `encipheredSeed` (or any field that's a `bytes` protobuf type) will have a `string | Uint8Array` type. I'm not sure why this is - it always seems to be a base64 string in practice. However, I figured it made sense to check the type just in case.

lib/service/InitService.ts
```diff
@@ -0,0 +1,74 @@
...
+        Buffer.from(seed.encipheredSeed, 'base64') :
+        Buffer.from(seed.encipheredSeed);
+      assert.equal(seedBytes.length, 33);
+
+      // the seed is 33 bytes, the first byte of which is the version
+      // so we use the remaining 32 bytes to generate our private key
+      const privKey = Buffer.from(seedBytes.slice(1));
```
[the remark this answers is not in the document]
sangaman: Yeah that might make sense. This might be a little bit hacky, and we're not getting a full 32 bytes of entropy this way, but it so happens that the aezeed key is 33 bytes of which the 1st is a near-static version byte, and the remaining 32 bytes are enough for our priv key. I'll add a TODO here.

sangaman:
> What exact attack scenario does it prevent? Best practise would be to have separate passwords and seeds for all clients, this is about UX.

Maybe we can discuss this for tomorrow, but if we are just encrypting the mnemonic data with the same password as we use to unlock the wallet, I'm not sure how much that adds in the way of security either, and in that case maybe we should just not use encryption. Karl's suggestion of asking for a separate, optional password will leave the default UX easy with a single password while creating a separate one for those who think it's worth the trouble. You wouldn't need the mnemonic password to start xud, only in case of recovering from seed.

lib/lndclient/LndClient.ts
sangaman: You mean using the seedutil to generate a seed rather than returning `undefined`? I guess in that case I'm still thinking there might not be much value since we couldn't use that seed for any lnd instances. Maybe I'll just do that if we're only connected to raiden.

erkarl: Exactly - fallback to `seedutil create`. I'm good with leaving a TODO right now as our current setup assumes lnd is always there.

kilrau:
As discussed in the call, moved Aezeed encryption to a separate `post-1.0.0` issue: #1030

Updated [summary comment](https://github.com/ExchangeUnion/xud/pull/1006#issuecomment-500349786) above.

sangaman:
@erkarl I addressed the your review comments, if this looks good to you I can squash and merge and open separate PRs for the priority TODO items we've identified.

PR #1007, opened 2019-06-04
Master key refactor

sangaman:
This splits out commits related to refactoring code and fixing/modifying tests in preparation for the master key changes in #1006. These commits can stand on their own and be merged separately to simplify the review process.

### DIFF
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -33,15 +33,10 @@ class Xud extends EventEmitter {
   private orderBook!: OrderBook;
   private rpcServer?: GrpcServer;
   private httpServer?: HttpServer;
-  private nodeKey!: NodeKey;
   private grpcAPIProxy?: GrpcWebProxyServer;
   private swaps!: Swaps;
   private shuttingDown = false;
   private swapClientManager!: SwapClientManager;
-
-  public get nodePubKey() {
-    return this.nodeKey.nodePubKey;
-  }
 
   /**
    * Create an Exchange Union daemon.
@@ -67,13 +62,22 @@ class Xud extends EventEmitter {
     this.logger.info('config loaded');
 
     try {
+      const nodeKey = await NodeKey.load(this.config.xudir, this.config.instanceid);
+      this.logger.info(`Local nodePubKey is ${nodeKey.nodePubKey}`);
+
       const initPromises: Promise<any>[] = [];
       // TODO: wait for decryption of existing key or encryption of new key, config option to disable encryption
-      initPromises.push(NodeKey.load(this.config.xudir, this.config.instanceid));
 
       this.db = new DB(loggers.db, this.config.dbpath);
       await this.db.init(this.config.network, this.config.initdb);
-      this.pool = new Pool(this.config.p2p, this.config.network, loggers.p2p, this.db.models, version);
+      this.pool = new Pool({
+        nodeKey,
+        version,
+        config: this.config.p2p,
+        xuNetwork: this.config.network,
+        logger: loggers.p2p,
+        models: this.db.models,
+      });
       this.swapClientManager = new SwapClientManager(this.config, loggers, this.pool);
       await this.swapClientManager.init(this.db.models);
 
@@ -91,12 +95,10 @@ class Xud extends EventEmitter {
       initPromises.push(this.orderBook.init());
 
       // wait for components to initialize in parallel
-      const initPromisesResults = await Promise.all(initPromises);
-      this.nodeKey = initPromisesResults[0]; // the first init promise in the array was NodeKey.load
-      this.logger.info(`Local nodePubKey is ${this.nodeKey.nodePubKey}`);
+      await Promise.all(initPromises);
 
       // initialize pool and start listening/connecting only once other components are initialized
-      await this.pool.init(this.nodeKey);
+      await this.pool.init();
 
       this.service = new Service({
         version,
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -62,14 +62,12 @@ class Xud extends EventEmitter {
     this.logger.info('config loaded');
 
     try {
+      this.db = new DB(loggers.db, this.config.dbpath);
+      await this.db.init(this.config.network, this.config.initdb);
+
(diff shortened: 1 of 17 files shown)


lib/Xud.ts
```diff
@@ -67,29 +62,36 @@ class Xud extends EventEmitter {
...
+        version,
+        config: this.config.p2p,
+        xuNetwork: this.config.network,
+        logger: loggers.p2p,
+        models: this.db.models,
+      });
+      this.swapClientManager = new SwapClientManager(this.config, loggers);
       this.swaps = new Swaps(loggers.swaps, this.db.models, this.pool, this.swapClientManager);
+
+      await this.swapClientManager.init(this.db.models);
```
erkarl: nit: this line could be right below the creation of `SwapClientManager` (line 79)

lib/db/DB.ts
```diff
@@ -40,14 +40,16 @@ class DB {
    */
   public init = async (network = XuNetwork.SimNet, initDb = false): Promise<void> => {
     this.models = await this.loadModels();
-    let newDb = !this.storage;
-    if (this.storage) {
+    let newDb = true;
+    if (this.storage && this.storage !== ':memory:') {
```
erkarl: Since we're adding new functionality here - what do you think about extracting it from `init` to something like `shouldCreateNewDB` and unit testing it?

lib/swaps/Swaps.ts
```diff
@@ -147,16 +148,17 @@ class Swaps extends EventEmitter {
...
-      swapClient.on('htlcAccepted', async (rHash, amount) => {
-        try {
-          const rPreimage = await this.resolveHash(rHash, amount, currency);
-          await swapClient.settleInvoice(rHash, rPreimage);
-        } catch (err) {
-          this.logger.error('could not settle invoice', err);
-        }
-      });
+
+    this.swapClientManager.on('htlcAccepted', async (swapClient, rHash, amount, currency) => {
```
erkarl: Really like this approach :)

test/jest/Orderbook.spec.ts
```diff
@@ -90,8 +92,15 @@ describe('OrderBook', () => {
...
     db = new DB(loggers.db, config.dbpath);
-    pool = new Pool(config.p2p, config.network, loggers.p2p, db.models, '1.0.0');
-    swapClientManager = new SwapClientManager(config, loggers, pool);
+    pool = new Pool({
+      config: config.p2p,
+      xuNetwork: config.network,
+      logger: loggers.p2p,
+      models: db.models,
+      version: '1.0.0',
+      nodeKey: new NodeKey(new Buffer(1)),
```
erkarl: We should be able to avoid passing arguments to `NodeKey` since it's mocked.

Consider casting NodeKey to any and then to `jest.Mock`. Example:
https://github.com/ExchangeUnion/xud/blob/master/test/jest/HttpService.spec.ts#L5

lib/Xud.ts
sangaman: So actually this was a necessary change (one that I spent a while figuring out) because `pool` needs to know about the lnd pub keys and raiden address that we get only after `swapClientManager.init` completes. `swapClientManager` notifies `swaps` whenever we get a new pubkeyor address, and `swaps` updates `pool`, and with this change that happens on initialization. If we initialize `swapClientManager` before creating `swaps`, it doesn't happen.

But in looking at this again, this seems less than ideal and fragile. What I can do is put logic in the `Swaps` constructor to update `pool` with the pub keys that we already have. That way it doesn't matter in what order we put these lines, and I can go ahead and reorder this as you've suggested. How does that sound?

lib/db/DB.ts
sangaman: That's fine with me, I'll do that.

lib/swaps/Swaps.ts
sangaman: Yup definitely fewer ways that things can go wrong like this.

test/jest/Orderbook.spec.ts
sangaman: Sounds good, I'll try this out.

lib/Xud.ts
erkarl: I'm not sure about adding the listeners in `Swaps` constructor. In the long run I'm beginning to think a pub-sub pattern between our services might be a better approach compared to dependency injection, but it's definitely out of scope of this PR.

I'm good with just leaving a comment right now that explains why `swapClientManager` is initalized before creating `swaps`.

lib/db/DB.ts
sangaman: Done and I actually refactored it a bit more (we don't always need to make this check for one thing, only when `initDb` is true). It's in a separate commit so you can easily see the changes.

lib/Xud.ts
sangaman: There wouldn't be any new listeners being added, just a check in the constructor to set existing lnd pubkeys and raiden addresses from `swapClientManager` to `pool`. Although I'd also be OK leaving it as is with a comment to explain.

erkarl: Up to you :).

sangaman: Putting that much code in the constructor didn't seem right - and in general I don't think calling constructors should be making changes to other objects - so I moved that logic in `init`. It works now (and they can be initialized in parallel) and it's less fragile I believe so I'll go with this if it looks good to you as well.

test/jest/Orderbook.spec.ts
```diff
@@ -90,8 +93,15 @@ describe('OrderBook', () => {
...
     db = new DB(loggers.db, config.dbpath);
-    pool = new Pool(config.p2p, config.network, loggers.p2p, db.models, '1.0.0');
-    swapClientManager = new SwapClientManager(config, loggers, pool);
+    pool = new Pool({
+      config: config.p2p,
+      xuNetwork: config.network,
+      logger: loggers.p2p,
+      models: db.models,
+      version: '1.0.0',
+      nodeKey:  new mockedNodeKey(),
```
erkarl: nit: extra whitespace

lib/db/DB.ts
erkarl: Perfect, looks much better now.

PR #1024, opened 2019-06-10
fix: blockheight prop of `undefined` in GetInfo

### LINKED ISSUE #1011: Cannot read property 'blockheight' of undefined
Branch: `origin/listTrades-grpc`

With LND and Raiden enabled in config but unavailable, `xucli getinfo` throws:
```
~/xud/bin$ ./xucli getinfo
Error: 2 UNKNOWN: Cannot read property 'blockheight' of undefined
```

Other calls work. Log clearly shows that this is because of raiden with a failed `ECONNREFUSED` for every getinfo RPC call:
```
07/06/2019 18:33:33.736 [RPC] debug: received call /xudrpc.Xud/GetInfo
07/06/2019 18:33:33.736 [RAIDEN] error: Error: connect ECONNREFUSED 127.0.0.1:5001
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1106:14)
07/06/2019 18:33:33.736 [RPC] error: Cannot read property 'blockheight' of undefined
```

sangaman:
This fixes a bug where xud tries to read a property of `undefined` when an instance of `LndClient` is disabled. This refactors the code to use a `Map` when retrieving a mapping of `LndInfo` objects, and also calls `GetInfo` in parallel on all `LndClient`s.

Fixes #1011.

### DIFF
--- a/lib/swaps/SwapClientManager.ts
+++ b/lib/swaps/SwapClientManager.ts
@@ -1,7 +1,7 @@
 import Config from '../Config';
 import SwapClient from './SwapClient';
 import LndClient from '../lndclient/LndClient';
-import { LndLogger, LndInfos } from '../lndclient/types';
+import { LndLogger, LndInfo } from '../lndclient/types';
 import RaidenClient from '../raidenclient/RaidenClient';
 import Logger, { Loggers } from '../Logger';
 import { errors } from './errors';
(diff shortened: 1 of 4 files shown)


lib/swaps/SwapClientManager.ts
```diff
@@ -169,19 +169,23 @@ class SwapClientManager extends EventEmitter {
...
     // (similar to how we're maintaining the list of raiden currencies)
     // rather than determining it dynamically when needed. The benefits
     // would be slightly improved performance.
+    const getInfoPromises: Promise<void>[] = [];
     for (const [currency, swapClient] of this.swapClients.entries()) {
       if (isLndClient(swapClient)) {
-        lndInfos[currency] = swapClient.isDisabled()
-          ? undefined
-          : await swapClient.getLndInfo();
+        if (!swapClient.isDisabled()) {
```
erkarl: Think it's possible to use 1 conditional instead of 2 here:
`if (isLndClient(swapClient) && !swapClient.isDisabled())`

kilrau:
Great! Sure this can't also happen with Raiden[ as the log in my case](https://github.com/ExchangeUnion/xud/issues/1011) somewhat suggested?

lib/swaps/SwapClientManager.ts
sangaman: Indeed it is, I changed it.

sangaman:
> Great! Sure this can't also happen with Raiden[ as the log in my case](https://github.com/ExchangeUnion/xud/issues/1011) somewhat suggested?

Yes pretty sure, I don't think we get back or attempt to read a `blockheight` from any Raiden calls. I suspect that case also had a disabled lnd client.

PR #1026, opened 2019-06-11
refactor: disable sanity checks by default

ghost:
Disables sanity checks in the default configuration:
https://github.com/ExchangeUnion/xud/issues/970

### DIFF
--- a/bin/xud
+++ b/bin/xud
@@ -57,8 +57,13 @@ const { argv } = require('yargs')
       type: 'boolean',
       default: undefined,
     },
-    nosanitychecks: {
-      describe: 'Whether to disable order sanity checks',
+    nosanityswaps: {
+      describe: 'Whether to disable sanity swaps',
+      type: 'boolean',
+      default: undefined,
+    },
+    nobalancechecks: {
+      describe: 'Whether to disable balance checks',
       type: 'boolean',
       default: undefined,
     },
--- a/lib/Config.ts
+++ b/lib/Config.ts
@@ -32,10 +32,15 @@ class Config {
   public noencrypt = true; // TODO: enable encryption by default
 
   /**
-   * Whether to disable sanity checks that verify that the orders can possibly be swapped
-   * before adding them to the order book, can be enabled for testing & debugging purposes.
+   * Whether to disable sanity swaps that verify that the orders can possibly be swapped
+   * before adding trading pairs as active.
    */
-  public nosanitychecks = false;
+  public nosanityswaps = true;
+  /**
+   * Whether to disable balance checks that verify that the orders can possibly be swapped
+   * before adding them to the order book.
+   */
+  public nobalancechecks = false;
 
   constructor() {
     const platform = os.platform();
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -120,7 +120,8 @@ class Xud extends EventEmitter {
         nomatching: this.config.nomatching,
         pool: this.pool,
         swaps: this.swaps,
-        nosanitychecks: this.config.nosanitychecks,
+        nosanityswaps: this.config.nosanityswaps,
+        nobalancechecks: this.config.nobalancechecks,
       });
       initPromises.push(this.orderBook.init());
 
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -72,7 +72,8 @@ class OrderBook extends EventEmitter {
   private pairInstances = new Map<string, PairInstance>();
   private repository: OrderBookRepository;
   private logger: Logger;
-  private nosanitychecks: boolean;
+  private nosanityswaps: boolean;
+  private nobalancechecks: boolean;
   private pool: Pool;
   private swaps: Swaps;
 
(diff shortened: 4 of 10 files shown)


sangaman:
My only question is if we are going to go with separate config options for the capacity checks and the sanity swaps, we might want to do that first.

erkarl:
@sangaman as a result of your comment and https://github.com/ExchangeUnion/xud/issues/970#issuecomment-500808616 I've changed the PR to split `nosanitychecks` into 2 configuration options:

`sanityswaps` - false by default
`balancechecks` - true by default

erkarl:
Tests are failing, will look into it shortly.

kilrau:
> The only question that remains - do we want the `balancechecks` to throw or just display a warning in the console?

Definitely throw. "Just displaying a warning" will result in a swap failure in 99.99% of the cases (unless I'm missing something).

bin/xud
```diff
@@ -57,8 +57,13 @@ const { argv } = require('yargs')
...
       default: undefined,
     },
-    nosanitychecks: {
-      describe: 'Whether to disable order sanity checks',
+    sanityswaps: {
+      describe: 'Whether to enable sanity swaps',
+      type: 'boolean',
+      default: undefined,
+    },
+    balancechecks: {
```
sangaman: I think it might be good to keep the "no" in the config option naming if the goal long term is to enable these by default, which is my understanding. Otherwise we may want to rename them in the future and that would be a hassle if people need to rewrite their confg files. So I'd go with `nosanityswaps` and `nobalancechecks`.

lib/Config.ts
```diff
@@ -32,10 +32,15 @@ class Config {
   public noencrypt = true; // TODO: enable encryption by default
 
   /**
-   * Whether to disable sanity checks that verify that the orders can possibly be swapped
-   * before adding them to the order book, can be enabled for testing & debugging purposes.
+   * Whether to enable sanity swaps that verify that the orders can possibly be swapped
+   * before adding trading pairs as active.
    */
-  public nosanitychecks = false;
+  public sanityswaps: boolean;
```
sangaman: Some of the other boolean config options have their default value set here where they are declared (like `nomatching` above). It's less code that way, although I suppose the benefit of setting all default values in the constructor is that they're all in one place, since some values can't be initialized outside the constructor. I think I prefer having default values where they're initialized (like `public sanityswaps = true;`) where it's possible to do so, but either way I think it's best to be consistent, just wondering what you think is best going foward.

bin/xud
erkarl: The reason I opted to go for `balancechecks` and `sanityswaps` was that for me it felt very weird to read double negative code. Consider the previous code snippet:
```
if (!this.nosanitychecks) {
```
This reads in my head as if no-no-sanitychecks which causes confusion whereas `if (this.sanitychecks)` or `if (!this.sanitychecks)` is much clear to me.

lib/Config.ts
erkarl: I agree - let's be consistent. Think your way of doing it is better because it's less verbose.

bin/xud
sangaman: That's a fair point, but I'm thinking mainly about the implications for passing args to xud. I think it's more convenient to be able to do `xud --nosanitychecks` as opposed to `xud --sanitychecks false`. Internally at some point we could `sanitychecks = !nosanitychecks` and use that variable in our logic.

sangaman:
One more nit, I think by the commit template and changelog generator, `chore` is for build related changes but this is a code change, probably better suited as a `refactor`.

bin/xud
erkarl: I went ahead and changed it to `nosanityswaps` and `nobalancechecks`. I feel like adding an abstraction layer between the CLI flags and the actual code would add unnecessary complexity.

sangaman:
@erkarl Looks like the tests are failing, could you take a look?

erkarl:
@sangaman should be fixed now. I managed to run into flaky tests for ~8 runs straight.

bin/xud
```diff
@@ -57,10 +57,15 @@ const { argv } = require('yargs')
       type: 'boolean',
       default: undefined,
     },
-    nosanitychecks: {
-      describe: 'Whether to disable order sanity checks',
+    nosanityswaps: {
+      describe: 'Whether to disable sanity swaps',
       type: 'boolean',
-      default: undefined,
+      default: true,
```
sangaman: One last thing, we should keep the default here as `undefined`, otherwise this will override the value in the config file. If we have `nosanityswaps = false` in the config file, and then run `xud` without passing any argument for `nosanityswaps`, `nosanityswaps` will wind up as true which is obviously undesirable. It's a quirk of yargs.

bin/xud
```diff
@@ -57,10 +57,15 @@ const { argv } = require('yargs')
...
+    },
+    nobalancechecks: {
+      describe: 'Whether to disable balance checks',
+      type: 'boolean',
+      default: false,
```
sangaman: Same here for a default of undefined.

sangaman (changes requested):
One last thing then good to merge.

sangaman:
One more thing I'm thinking is that it probably makes sense to leave sanity swaps on for the simulation tests - or maybe better yet have Alice use sanity swaps and Bob not use them so that both ways are tested.

sangaman (approved):

sangaman:
Builds passing on Travis but still show pending on Github for some reason, going to merge.

PR #1032, opened 2019-06-14
feat: add backup daemon

### LINKED ISSUE #917: Backup
* Channel State backups: [SCB](https://github.com/lightningnetwork/lnd/blob/master/docs/recovery.md). LND fires an event whenever there is an update to the SCB file. Call: https://api.lightning.community/#subscribechannelbackups 
xud should then copy it to [location set by user](https://github.com/ExchangeUnion/xud/issues/246). Raiden: the channel state db should bin `~/.raiden` and be backed up the same way. Not sure if events are available.
* `xud.db`, `xud.conf`, `lnd.conf`, `raiden conf`, `/root/.xud/tor` (xud container)

Anything I forgot?

michael1011:
Closes #917 

This PR adds `xu-backup` that starts listening for new LND SCBs and writing them to the disk when executed.

I am not finished yet and have some things to do before this backup daemon can be used but please have a look at this PR nevertheless to acknowledge the concept and give me some feedback on what I have been doing in this branch. 

Todo:
- [x] Raiden database backup (just copying the file is the recommended way of the Raiden devs)
- [x] XUD database backup
- [x] write SCB of all LNDs and Raiden database backup to disk on startup
- [x] writing unit/integration tests
- [ ] intensive manual testing

@kilrau we agreed on not implementing a restoring feature for now, right?

### DIFF
--- a/bin/xud-backup
+++ b/bin/xud-backup
@@ -1,0 +1,79 @@
+#!/usr/bin/env node
+
+const Backup = require('../dist/backup/Backup').default;
+
+const { argv } = require('yargs')
+  .options({
+    backupdir: {
+      describe: 'Data directory for backups',
+      type: 'string',
+      alias: 'b',
+    },
+    xudir: {
+      describe: 'Data directory for xud',
+      type: 'string',
+      alias: 'x',
+    },
+    dbpath: {
+      describe: 'Path to the XUD database',
+      type: 'string',
+      alias: 'd',
+    },
+    loglevel: {
+      describe: 'Verbosity of the logger',
+      type: 'string',
+      alias: 'l',
+    },
+    logpath: {
+      describe: 'Path to the log file',
+      type: 'string',
+    },
+    logdateformat: {
+      describe: 'Format of the logger date',
+      type: 'string',
+    },
+    'lnd.[currency].certpath': {
+      describe: 'Path to the SSL certificate for lnd',
+      type: 'string',
+    },
+    'lnd.[currency].cltvdelta': {
+      describe: 'CLTV delta for the final timelock',
+      type: 'number',
+    },
+    'lnd.[currency].disable': {
+      describe: 'Disable lnd integration',
+      type: 'boolean',
+      default: undefined,
+    },
+    'lnd.[currency].host': {
+      describe: 'Host of the lnd gRPC interface',
+      type: 'string',
+    },
+    'lnd.[currency].macaroonpath': {
+      describe: 'Path of the admin macaroon for lnd',
+      type: 'string',
+    },
+    'lnd.[currency].nomacaroons': {
+      describe: 'Whether to disable macaroons for lnd',
+      type: 'boolean',
+      default: undefined,
+    },
+    'lnd.[currency].port': {
+      describe: 'Port for the lnd gRPC interface',
+      type: 'number',
+    },
+    'raiden.dbpath': {
+      describe: 'Path to the database of Raiden',
+      type: 'string',
+    },
+  });
+
+// delete non-config keys from argv
+delete argv._;
+delete argv.version;
+delete argv.help;
+delete argv.$0;
+
+const backup = new Backup();
+
+backup.start(argv);
--- a/lib/backup/Backup.ts
+++ b/lib/backup/Backup.ts
@@ -1,0 +1,181 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { createHash } from 'crypto';
+import Config from '../Config';
+import Logger, { Context } from '../Logger';
+import LndClient from '../lndclient/LndClient';
+
+class Backup {
+  private logger!: Logger;
+  private config = new Config();
+
+  private backupDir!: string;
+
+  private fileWatchers: fs.FSWatcher[] = [];
+  private lndClients: LndClient[] = [];
+
+  public start = async (args: { [argName: string]: any }) => {
+    await this.config.load(args);
+
+    this.backupDir = args.backupdir || this.getDefaultBackupDir();
+
+    this.logger = new Logger({
+      context: Context.Backup,
+      level: this.config.loglevel,
+      filename: this.config.logpath,
+      instanceId: this.config.instanceid,
+      dateFormat: this.config.logdateformat,
+    });
+
+    if (!fs.existsSync(this.backupDir)) {
+      try {
+        fs.mkdirSync(this.backupDir);
+      } catch (error) {
+        this.logger.error(`Could not create backup directory: ${error}`);
+        return;
+      }
+    }
+
+    try {
+      await this.startLndSubscriptions();
+    } catch (error) {
+      this.logger.error(`Could not connect to LNDs: ${error}`);
+    }
+
+    // Start the Raiden database filewatcher
+    if (args.raiden) {
+      this.startFilewatcher('raiden', args.raiden.dbpath);
+    } else {
+      this.logger.warn('Raiden database file not specified');
+    }
+
+    // Start the XUD database filewatcher
+    this.startFilewatcher('xud', this.config.dbpath);
+
+    this.logger.info('Started backup daemon');
+  }
+
+  public stop = async () => {
+    this.fileWatchers.forEach((watcher) => {
+      watcher.close();
+    });
+
+    for (const lndClient of this.lndClients) {
+      await lndClient.close();
+    }
+  }
+
+  private startLndSubscriptions = async () => {
+    // Start the LND SCB subscriptions
+    for (const currency in this.config.lnd) {
+      const config = this.config.lnd[currency]!;
+
+      // Ignore the LND client if it is disabled or not configured
+      if (!config.disable && Object.entries(config).length !== 0) {
+        const lndClient = new LndClient({
+          config,
+          currency,
+          logger: Logger.DISABLED_LOGGER,
+        });
+
+        this.lndClients.push(lndClient);
+
+        await lndClient.init();
+
+        const backupPath = this.getBackupPath('lnd', lndClient.currency);
+
+        this.logger.verbose(`Writing initial ${lndClient.currency} LND channel backup to: ${backupPath}`);
+
+        const channelBackup = await lndClient.exportAllChannelBackup();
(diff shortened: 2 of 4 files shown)


bin/xu-backup
```diff
@@ -0,0 +1,70 @@
...
+      type: 'boolean',
+      default: undefined,
+    },
+    'lnd.[currency].port': {
+      describe: 'Port for the lnd gRPC interface',
+      type: 'number',
+    },
+  });
+
+// delete non-config keys from argv
```
erkarl: is `config.load` expecting sanitized data?

michael1011: I guess that the config would work fine without these lines but [`bin/xud` is doing the same thing](https://github.com/ExchangeUnion/xud/blob/master/bin/xud#L196) and therefore I did too.

kilrau:
> @kilrau we agreed on not implementing a restoring feature for now, right?

Correct, since it can be restored using lnd/raiden directly.

As far as I see it we'd have to implement `lnd`s https://api.lightning.community/#restorechannelbackups in `xud` for that, track differences between btc & ltc ... but why not open an issue to keep track of it `post-1.0.0` @michael1011 :)

lib/backup/Backup.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    this.logger.info('Started backup daemon');
+  }
+
+  private listenToChannelBackups = (lndClient: LndClient) => {
+    lndClient.on('channelBackup', (channelBackup) => {
+      this.logger.verbose(`${lndClient.currency} channel backup: ${channelBackup}`);
+
+      fs.writeFileSync(
```
sangaman: I doubt performance will really be an issue with this tool but still you can use the promisified version of the `fs` API (there are examples elsewhere in the code) and `await` it. Also it would probably be good to check to handle errors in case we can't write to the file here, maybe try writing to a separate file with the date & time in the filename? That might be useful in case we ever get near simultaneous backups and file write attempts.

michael1011: The arrow function of the callback isn't `async` so the events should be handled sequentially and parallel writes shouldn't be a problem (apart from that it is really unlikely with the current behavior of LND that something like this would happen).

I agree on the error handling and will implement that.

lib/backup/Backup.ts
```diff
@@ -0,0 +1,151 @@
...
+
+        this.listenToChannelBackups(lndClient);
+
+        lndClient.subscribeChannelBackups();
+        this.logger.verbose(`Listening to ${currency} LND channel backups`);
+      }
+    }
+
+    // Start the Raiden database filewatcher
+    if (args.raiden && args.raiden.database !== undefined) {
```
erkarl: think the `!== undefined` check here is redundant

lib/backup/Backup.ts
```diff
@@ -0,0 +1,151 @@
...
+            // Compare the MD5 hash of the current content of the file with hash of the content when
+            // it was backed up the last time to ensure that the content of the file has changed
+            if (hash !== previousRaidenMd5) {
+              this.logger.debug('Raiden database changed');
+
+              previousRaidenMd5 = hash;
+              this.writeBackup('raiden', content);
+            }
+          }
+        })  ;
```
erkarl: `});`

lib/backup/Backup.ts
sangaman: Sounds good.

lib/backup/Backup.ts
```diff
@@ -0,0 +1,151 @@
...
+  private fileWatcher!: fs.FSWatcher;
+  private lndClients: LndClient[] = [];
+
+  public start = async (args: { [argName: string]: any }) => {
+    await this.config.load(args);
+
+    this.backupDir = args.backupdir || this.getDefaultBackupDir();
+
+    if (!fs.existsSync(this.backupDir)) {
+      fs.mkdirSync(this.backupDir);
```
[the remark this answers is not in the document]
michael1011: Oh I overlooked that. It's now wrapped in a try-catch block which should handle all errors gracefully.

michael1011:
@erkarl regarding the style: you are right about the inconsistencies with the rest of the codebase but I didn't think that it mattered that much when actually writing the code since some other files also have some blank lines. And `npm run lint` didn't throw any warnings or messages so I just left it as it is now.

What do you think? Should I remove some blank lines or it is fine to leave it as is?

erkarl:
> What do you think? Should I remove some blank lines or it is fine to leave it as is?

Just leave it as is for now. I'll take a look at our linter configuration at some point and unify the formatting style.

kilrau:
Added `xud.db`, since we start writing important things into that (swap states). Loosing these means potentially loosing funds.

kilrau:
Todo's need to be updated as per https://github.com/ExchangeUnion/xud/issues/917

michael1011:
> Can we rename xu-backup -> xud-backup?

Done!

lib/lndclient/LndClient.ts
```diff
@@ -817,6 +828,25 @@ class LndClient extends SwapClient {
...
+  /**
+   * Subscribes to channel backups
+   */
+  public subscribeChannelBackups = () => {
+    if (!this.lightning) {
+      throw errors.UNAVAILABLE(this.currency, this.status);
+    }
+
+    if (this.channelBackupSubscription) {
+      this.channelBackupSubscription.cancel();
```
sangaman: Does it make sense just to return if a channel subscription already exists, rather than cancel it and restart it? Just not sure if there's any value to doing the latter.

test/jest/Backup.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  test('should resolve backup directory', () => {
+    expect(backup['getDefaultBackupDir']()).toContainEqual('/');
+  });
+
+  test('should write LND backups on startup', () => {
+    expect(
+      fs.readFileSync(
+        path.join(backupdir, 'lnd-BTC'),
+        'utf-8',
```
sangaman: I'm guessing `utf-8` might just be an alias for `utf8` internally in nodejs such that it doesn't matter, but I noticed again that it'd different from what I'm used to. I won't comment on the other instances.

lib/backup/Backup.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  private startFilewatcher = (client: string, dbPath: string) => {
+    if (fs.existsSync(dbPath)) {
+      let previousDatabaseHash = '';
+
+      this.logger.verbose(`Writing initial ${client} database backup`);
+      const { content, hash } = this.readDatabase(dbPath);
+
+      previousDatabaseHash = hash;
```
sangaman: Minor nit: I think `previousDatabaseHash` can be declared here rather than being initialized above.

kilrau:
let's definitely do some manual testing @sangaman @erkarl

lib/backup/Backup.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    // Start the Raiden database filewatcher
+    if (args.raiden) {
+      this.startFilewatcher('raiden', args.raiden.database);
+    } else {
+      this.logFileNotSpecified('raiden');
+    }
+
+    // Start the XUD database filewatcher
+    if (args.dbpath !== undefined && args.dbpath !== '') {
```
sangaman: @erkarl and I are thinking it would make sense not to reference `args` here (and above for raiden) but rather `this.config`. That way we can use values from the config file or default values and not have to pass in all of these values via command line argument. That might make the `logFileNotSpecified` function obsolete since there will always be a default value.

bin/xud-backup
```diff
@@ -0,0 +1,74 @@
...
+    'lnd.[currency].nomacaroons': {
+      describe: 'Whether to disable macaroons for lnd',
+      type: 'boolean',
+      default: undefined,
+    },
+    'lnd.[currency].port': {
+      describe: 'Port for the lnd gRPC interface',
+      type: 'number',
+    },
+    'raiden.database': {
```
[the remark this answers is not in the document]
michael1011: I didn't do that initially because the path to the database of raiden depends not only on the network it is running on but also on a node specific id and looks like this: `exchange/node_1a8e5ab4/netid_4321/network_de8a2bdd/v23_log.db`.

Any ideas on how to determine a reasonable default value?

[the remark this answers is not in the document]
erkarl: I think `raiden.dbpath` could be the only value where we don't specify a default value. Thoughts @sangaman?

michael1011:
> I'm not seeing the debug statements for when lnd channel backups are written - it should happen after swaps occur right? I tested with a BTC-LTC swap and I did see this.

I am not backing up the database of LND but the SCBs (static channel backups). These backups are done once for a channel and are therefore static. This also means that a new backup will be done once you close an existing or open a new channel. When you do that it should print you a debug statement @sangaman

michael1011:
The tests fail only on Travis and because of this error:

```
 FAIL  test/jest/Backup.spec.ts

  ● Backup › should write Raiden backups on new event

    RangeError: Maximum call stack size exceeded

        at Function.[Symbol.hasInstance] (<anonymous>)

      at WriteStream (../../node_modules/graceful-fs/graceful-fs.js:200:14)

      at WriteStream (../../node_modules/graceful-fs/graceful-fs.js:201:29)

      at WriteStream (../../node_modules/graceful-fs/graceful-fs.js:201:29)

      at WriteStream (../../node_modules/graceful-fs/graceful-fs.js:201:29)

      at WriteStream (../../node_modules/graceful-fs/graceful-fs.js:201:29)

      at WriteStream (../../node_modules/graceful-fs/graceful-fs.js:201:29)

      at WriteStream (../../node_modules/graceful-fs/graceful-fs.js:201:29)

      at WriteStream (../../node_modules/graceful-fs/graceful-fs.js:201:29)

      at WriteStream (../../node_modules/graceful-fs/graceful-fs.js:201:29)
```

It is an bug of one of the dependencies of Jest (https://github.com/facebook/jest/issues/9062) and was fixed recently.

What should be done about this? Wait for the next Jest release and continue working on the restore in this branch or should we update the dependency in `npm-shrinkwrap.json` ourselves?

@sangaman @erkarl

PR #902, opened 2019-04-24
chore: disable current v12 node travis build

sangaman:
This disables the travis build for the current Node.js version (12.0.0) until all dependencies support it. See: https://github.com/grpc/grpc-node/issues/834

PR #1042, opened 2019-06-18
chore: reenable v12 current node build

sangaman:
This reenables the current node build in our travis configuration, reversing #902, and updates packages that did not previously support Node.js v12.

This also includes the results of running `npm audit fix` (reverting `grpc-tools` to 1.6.0 because of a regression bug in newer versions) and `npm dedupe` after updating the necessary packages.


sangaman:
Good question, I am not entirely sure. Updating them before a major release like we're trying to have soon seems like a reasonable idea, although generally speaking I am reluctant to do so without a particular reason in case it breaks something.

I have also been thinking of creating an [npm-shrinkwrap.json](https://docs.npmjs.com/files/shrinkwrap.json) file to pin down versions.

Thoughts @erkarl @moshababo ?

kilrau:
I suggest to update before the testnet release since we are these versions are locked in for quite some time after. Goes without saying, after `1.0.0-testnet` release no updates allowed anymore.

Simulation test passing + (hopefully) all the manual testing going on until mainnet should catch things.

package-lock.json
```diff
@@ -4620,7 +4630,7 @@
       "dependencies": {
         "pify": {
           "version": "2.3.0",
-          "resolved": "https://registry.npmjs.org/pify/-/pify-2.3.0.tgz",
+          "resolved": "http://registry.npmjs.org/pify/-/pify-2.3.0.tgz",
```
erkarl: did we ever figure out what was causing the `https <-> http` diffs?

erkarl:
> Updating them before a major release like we're trying to have soon seems like a reasonable idea, although generally speaking I am reluctant to do so without a particular reason in case it breaks something.

Same here, I'd upgrade when we have a good reason to do it (definitely not in this PR).

> I have also been thinking of creating an npm-shrinkwrap.json file to pin down versions.

Seems like a good idea.

kilrau:
> Same here, I'd upgrade when we have a good reason to do it (definitely not in this PR).

K, majority wins - no dependency upgrades @michael1101

package-lock.json
sangaman: I really do not know, googling it looks like it's still a known issue. I can go back and replace http with https manually though, the links I've read suggested that.

Issue #909, opened 2019-04-26
Channel management


kilrau:
### Motivation:
Our [new docker setup](https://github.com/ExchangeUnion/xud/issues/246) will ask for a "channel strategy" - direct or multi-hop. Due to a serious raiden routing issue (raiden uses a new secret for every new route it tries) we probably have to make direct channels the default for now and only allow multi-hop for btc and ltc.

### Specification:
This issue is about creating a new config option for the two modes. Docker setup will manipulate this config based on user input. Also there should be a place where "Open <coin> channel with <x> amount to xud <nodepubkey>. This will also be the place where the docker setup persist this user input.

sangaman:
I don't see why we would need a config option for direct channel mode. Even given the raiden bug, if we can't swap (whether due to lack of a direct channel or for any other reason) we will fail the sanity swaps and therefore not send or accept any orders with peers for the trading pairs that can't be swapped.

kilrau:
Ok, where else would you persist this?

erkarl:
As discussed during the call this issue is now about:
* enabling/disabling autopilot for swap clients that have that functionality (currently only lnd) via our gRPC
* adding a gRPC call to open channels by providing `xudPubKey`, `currency` and `amount`

PR #1044, opened 2019-06-19
feat(service): add open channel support

ghost:
Still WIP and not ready for in depth review. Putting this out here for early feedback.

TODO:
- [x] finish skipped tests for `LndClient`
- [x] add `OpenChannel` gRPC call
- [x] consume gRPC call via CLI `xucli openchannel [nodePubKey] [amount] [currency]`
- [x] manual lnd channel testing
- [x] share lnd pubKeys with peers
- [x] manual raiden channel testing


kilrau:
If this is supposed to close https://github.com/ExchangeUnion/xud/issues/909, it also should include an RPC call to set lnd's autopilot to true/false. Can also be done in a separate PR.

erkarl:
This is ready for review, but I suggest we deal with #1055 first because this is built on top of this. Afterwards I'll rebase this against master.

lib/cli/commands/openchannel.ts
```diff
@@ -0,0 +1,31 @@
...
+import { OpenChannelRequest } from '../../proto/xudrpc_pb';
+import { coinsToSats } from '../utils';
+
+export const command = 'openchannel <nodePubKey> <currency> <amount>';
+
+export const describe = 'open a payment channel';
+
+export const builder = {
+  nodePubKey: {
+    description: 'The peer public key which to open channel with',
```
sangaman: How about "The node pub key of the peer to open the channel with"?

lib/swaps/SwapClient.ts
```diff
@@ -17,9 +17,14 @@ type ChannelBalance = {
   pendingOpenBalance: number,
 };
 
+export interface ConnectionVerified {
```
sangaman: We've been using `type` instead of `interface`, I think they are effectively the same but better to be consistent. Also `ConnectionVerified` is a bit confusing of a name for a type to me, how about `ConnectionInfo` or `SwapClientInfo`?

lib/lndclient/LndClient.ts
```diff
@@ -209,10 +215,14 @@ class LndClient extends SwapClient {
 
           /** The new lnd pub key value if different from the one we had previously. */
           let newPubKey: string | undefined;
+          let newUris: string[] | undefined;
           if (this.identityPubKey !== getInfoResponse.getIdentityPubkey()) {
             newPubKey = getInfoResponse.getIdentityPubkey();
             this.logger.debug(`pubkey is ${newPubKey}`);
             this.identityPubKey = newPubKey;
+            newUris = getInfoResponse.getUrisList();
```
sangaman: Right now we'll only update `newUris` if the identity pub key changes. Maybe we should have a separate if statement that checks if the new uris list is different from the one we currently have and set `newUris` if so.

lib/swaps/SwapClient.ts
```diff
@@ -17,9 +17,14 @@ type ChannelBalance = {
...
+  newIdentifier?: string;
+  uris?: string[];
```
sangaman: Could call this `newUris` as well for consistency.

lib/lndclient/LndClient.ts
```diff
@@ -382,19 +395,76 @@ class LndClient extends SwapClient {
   /**
    * Connects to another lnd node.
    */
-  public connectPeer = (pubkey: string, host: string, port: number): Promise<lndrpc.ConnectPeerResponse> => {
+  public connectPeer = (pubkey: string, host: string): Promise<lndrpc.ConnectPeerResponse> => {
```
sangaman: I know lnd is using `host` to refer to both the host but it's a little confusing to me. Maybe we could rename parameter here to `uri` or `address` instead?

lib/lndclient/LndClient.ts
```diff
@@ -382,19 +395,76 @@ class LndClient extends SwapClient {
...
     return this.unaryCall<lndrpc.ConnectPeerRequest, lndrpc.ConnectPeerResponse>('connectPeer', request);
   }
 
+  /**
+   * Opens a channel given peerPubKey and amount.
+   */
+  public openChannel = async (
+    peerPubKey: string,
+    amount: number,
+    currency: string,
```
sangaman: It's a little weird to me to pass `currency` in as a parameter here. It seems like the logic to use the right lnd client for the currency should be handled outside of this class. Technically we could have this currency check in other functions as well, but I don't think that's the best approach.

Edit: I see that this is mainly used for Raiden and we want to keep a common interface between that and lnd, so I see the reasoning here. Still, I might prefer making `currency` an optional parameter for the `openChannel` method and just ignore it for the lnd client.

lib/lndclient/LndClient.ts
```diff
@@ -382,19 +395,76 @@ class LndClient extends SwapClient {
...
+      const CONNECT_TIMEOUT = 10000;
+      setTimeout(() => {
+        return reject({ message: 'tryConnectPeerAddresses has timed out' });
+      }, CONNECT_TIMEOUT);
+      const peerAddresses = peerListeningUris
+        .map((listeningUri) => {
+          return listeningUri.split('@')[1];
+        });
+      let tryAnotherAddress = true;
+      while (tryAnotherAddress) {
```
sangaman: This while loop is a little hard to read for me. I think we could just do a `for of` loop here that `return resolve()`s on a successful connection, followed by a `return reject()` after the loop in case we don't manage to connect successfully.

Actually, it may be best for this method not to handle an unsuccessful connection attempt as a rejection. Instead it could return `true` when successful (call it `connectPeerAddreses` if so), `false` when unsuccessful, and reserve errors/rejections for unhandled cases like grpc erors or messages we aren't expecting. That way I think we could just write this method as a regular `async` function, with `return true` and `return false` statements instead of the resolve and reject lines.

proto/xudp2p.proto
```diff
@@ -26,6 +26,7 @@ message NodeState {
     string raiden_address = 5;
     map<string, string> lnd_pub_keys = 6;
     map<string, string> token_identifiers = 7;
+    map<string, string> lnd_uris = 8;
```
sangaman: As I mentioned you should take a look at using `map<string, ListValue>` instead of a comma delimited string here. It might be a cleaner approach.

lib/raidenclient/RaidenClient.ts
```diff
@@ -285,9 +292,26 @@ class RaidenClient extends SwapClient {
...
+      settle_timeout: 500,
+    });
+  }
+
+  /**
+   * Creates a payment channel request.
    * @returns The channel_address for the newly created channel.
    */
-  public openChannel = async (payload: OpenChannelPayload): Promise<string> => {
+  public openChannelRequest = async (payload: OpenChannelPayload): Promise<string> => {
```
sangaman: I think it would make sense to just combine this with the `openChannel` method above, although if you'd prefer not to I figure this method could at least be private.

lib/service/Service.ts
```diff
@@ -161,6 +161,29 @@ class Service {
...
+    argChecks.VALID_CURRENCY({ currency });
+    try {
+      const peer = this.pool.getPeer(nodePubKey);
+      await this.swapClientManager.openChannel({
+        peer,
+        amount,
+        currency,
+      });
+    } catch (e) {
+      const errorMessage = e.message || 'unknown';
```
sangaman: Maybe log the error here?

proto/xudrpc.proto
```diff
@@ -313,6 +320,17 @@ message ChannelBalanceResponse {
   map<string, ChannelBalance> balances = 1 [json_name = "orders"];
 }
 
+message OpenChannelRequest {
+  // The node pub key of the peer which to open channel with.
```
sangaman: "peer with which"

proto/xudrpc.proto
```diff
@@ -313,6 +320,17 @@ message ChannelBalanceResponse {
...
+  string node_pub_key = 1 [json_name = "peer_pub_key"];
+  // The ticker symbol of the currency to open channel for.
```
sangaman: "open the channel for"

proto/xudrpc.proto
```diff
@@ -313,6 +320,17 @@ message ChannelBalanceResponse {
...
+  string currency = 2 [json_name = "currency"];
+  // The amount of the channel in the smallest base unit of the currency.
```
sangaman: I think we should denominate this in satoshis rather than the smallest base unit, otherwise we will quickly run into integer overflow issues with the raiden currencies using 18 digits after the decimal place as we've encountered with the other rpc calls. This would also require changes to the logic elsewhere.

proto/xudrpc.proto
```diff
@@ -83,6 +83,13 @@ service Xud {
     };
   }
 
+  /* Opens a payment channel given pub key, amount and currency. */
```
sangaman: How about "Opens a payment channel to a peer with the given node pub key for the specified amount and currency."

sangaman (changes requested):
I've reviewed everything except the tests, made comments below. This will be a very handy feature, thanks!

lib/cli/commands/openchannel.ts
erkarl: Yep, that sounds better :).

lib/lndclient/LndClient.ts
erkarl: My thinking was that `newUris` don't change at runtime so there's no need to add logic to compare array contents of `newUris`.

lib/lndclient/LndClient.ts
erkarl: I changed the param to `address` and the actual request to `lightningAddress`.

lib/lndclient/LndClient.ts
erkarl: Agreed. I changed the function to use destructuring and also changed `peerPubKey` to `peerIdentifier` because in the context of Raiden we don't have `peerPubKey`.

lib/raidenclient/RaidenClient.ts
erkarl: Thanks for catching this. I changed the method to be private.

proto/xudrpc.proto
erkarl: Thanks for noticing. Yep, in the context of RPC we're always denominating amounts in satoshis per coin. Will implement the required logic.

erkarl:
@sangaman 
Thanks for the detailed review. I've applied most of the changes requested to the fixup commit.

Things left to do:
- [ ] simplify `tryConnectPeerAddresses` loop/function
- [ ] use map<string, LndUris> instead of map<string, string> on p2p level
- [ ] inject logger to Service and log error when `openchannel` call fails
- [ ] convert `Service.openChannel` amounts to correct units for the currency

lib/lndclient/LndClient.ts
sangaman: But they would change if lnd stops and starts while xud is running, right? In that case I'm thinking we'd want to update our lnd uris within xud.

lib/service/Service.ts
erkarl: Turns out that we're already RPC logging on the `GrpcService` layer:
https://github.com/ExchangeUnion/xud/blob/master/lib/grpc/GrpcService.ts#L178

lib/lndclient/LndClient.ts
erkarl: Would it be OK if I create an issue for it and deal with in a separate PR?

sangaman: OK with me.

lib/lndclient/LndClient.ts
```diff
@@ -382,19 +395,72 @@ class LndClient extends SwapClient {
...
+          peerPubKey: splitUri[0],
+          address: splitUri[1],
+        };
+      });
+    const CONNECT_TIMEOUT = 4000;
+    for (const uri of splitListeningUris) {
+      const { peerPubKey, address } = uri;
+      let timeout;
+      try {
+        timeout = setTimeout(() => {
```
sangaman: I was thinking this might be a good spot to use `Promise.race` - but after looking at this a bit more the grpc calls actually take an options parameter with a `deadline` field to specify the max time for the call to complete. Doesn't have to be in this PR, but it'd be nice to use that (and maybe reuse it elsewhere down the road). I can create a quick followup PR perhaps.

lib/p2p/packets/utils.ts
```diff
@@ -18,10 +19,31 @@ export const convertNodeState = (nodeState: pb.NodeState.AsObject) => {
     addresses: nodeState.addressesList,
     raidenAddress: nodeState.raidenAddress,
     lndPubKeys: convertKvpArrayToKvps(nodeState.lndPubKeysMap),
+    lndUris: convertLndUris(nodeState.lndUrisMap),
     tokenIdentifiers: convertKvpArrayToKvps(nodeState.tokenIdentifiersMap),
   });
 };
 
+const convertLndUris = <T>(kvpArray: [string, T][]): { [key: string]: T } => {
```
sangaman: Why the generic `<T>` here? Is this in case the type of `LndUris` changes in the future from a repeated string? I figure we can just replace `T` with `string[]` right? Would that also eliminate the need for `@ts-ignore` below?

sangaman (approved):
Looks good, I don't think we should hold this up much longer. `UnitConverter` is a good first step towards handling raiden currency amounts without hardcoding too.

lib/p2p/packets/utils.ts
erkarl: Thanks for noticing. I meant to do that, but somehow missed this. I've removed the `@ts-ignore` and replaced `T` with `string[]`.

lib/lndclient/LndClient.ts
erkarl: Hmm... perhaps. The benefit of having the timeout here would be a more clear error message.

I initially tried it with `Promise.race`, but I ran into some issues (can't remember the specifics).

Issue #1039, opened 2019-06-17
xud doesn't try to connect to lnd if it's syncing


kilrau:
Urgent off shoot from https://github.com/ExchangeUnion/xud/issues/684 since this makes things quite complicated for https://github.com/ExchangeUnion/xud-docker.

## Issue
If LND is not synced:
```Restarting xud while lnd is syncing ("synced_to_chain": false)does not make it connect.
Attaching to testnet_lndltc_1
lndltc_1     | 2019-06-17 13:36:33.097 [WRN] LTND: open /root/.lnd/lnd.conf: no such file or directory
lndltc_1     | 2019-06-17 13:36:33.097 [INF] LTND: Version: 0.6.1-beta commit=v0.6.1-beta, build=production, logging=default
lndltc_1     | 2019-06-17 13:36:33.097 [INF] LTND: Active chain: Litecoin (network=testnet)
lndltc_1     | 2019-06-17 13:36:33.098 [INF] CHDB: Checking for schema update: latest_version=8, db_version=8
lndltc_1     | 2019-06-17 13:36:33.108 [INF] RPCS: password gRPC proxy started at 0.0.0.0:8080
lndltc_1     | 2019-06-17 13:36:33.109 [INF] RPCS: password RPC server listening on 0.0.0.0:10009
lndltc_1     | 2019-06-17 13:36:33.109 [INF] LTND: Waiting for wallet encryption password. Use `lncli create` to create a wallet, `lncli unlock` to unlock an existing wallet, or `lncli changepassword` to change the password of an existing wallet and unlock it.
lndltc_1     | spawn lncli -n testnet -c litecoin create
lndltc_1     | Input wallet password: 
lndltc_1     | Confirm wallet password: 
lndltc_1     | 
lndltc_1     | Do you have an existing cipher seed mnemonic you want to use? (Enter y/n): n
lndltc_1     | 
lndltc_1     | Your cipher seed can optionally be encrypted.
lndltc_1     | Input your passphrase if you wish to encrypt it (or press enter to proceed without a cipher seed passphrase): 
lndltc_1     | 
lndltc_1     | Generating fresh cipher seed...
lndltc_1     | 
lndltc_1     | [lncli] unable to generate seed: rpc error: code = Unknown desc = wallet already exists
lndltc_1     | spawn lncli -n testnet -c litecoin unlock
lndltc_1     | Input wallet password: 12345678
lndltc_1     | 
[...]

PR #1046, opened 2019-06-19
feat(logger): optional subcontext

sangaman:
This adds an optional subcontext for loggers to print on every log statement. This is used in place of wrapping the log methods for the lnd loggers to display currency, but can also be reused in a generic fashion going forward.

Before:

```
19/06/2019 15:44:43.634 [LND] info: BTC: trying to verify connection...
```

After:

```
19/06/2019 15:44:43.634 [LND-BTC] info: trying to verify connection...
```

This is an offshoot of #1039.

### DIFF
--- a/lib/Logger.ts
+++ b/lib/Logger.ts
@@ -50,15 +50,34 @@ class Logger {
 
   private level: string;
   private context: Context;
+  private subcontext?: string;
   private logger?: winston.Logger;
+  private filename?: string;
   private instanceId: number;
+  private dateFormat?: string;
 
-  constructor({ level, filename, context, instanceId, disabled, dateFormat }:
-    {instanceId?: number, level?: string, filename?: string, context?: Context, disabled?: boolean, dateFormat?: string}) {
-
-    this.level = level || Level.Trace;
-    this.context = context || Context.Global;
-    this.instanceId = instanceId || 0;
+  constructor({
+    level = Level.Trace,
+    filename,
+    context = Context.Global,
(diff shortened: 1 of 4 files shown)


lib/Logger.ts
```diff
@@ -50,15 +50,26 @@ class Logger {
...
-
-  constructor({ level, filename, context, instanceId, disabled, dateFormat }:
-    {instanceId?: number, level?: string, filename?: string, context?: Context, disabled?: boolean, dateFormat?: string}) {
-
-    this.level = level || Level.Trace;
-    this.context = context || Context.Global;
-    this.instanceId = instanceId || 0;
+  private dateFormat?: string;
+
+  constructor({ level = Level.Trace, filename, context = Context.Global, subcontext, instanceId = 0, disabled, dateFormat }: {
```
erkarl: minor style nit: I feel like this line is too long.

lib/swaps/SwapClientManager.ts
```diff
@@ -41,21 +41,6 @@ class SwapClientManager extends EventEmitter {
     this.raidenClient = new RaidenClient(config.raiden, loggers.raiden);
   }
 
-  /**
```
erkarl: Beautiful! Loving that we get to delete this code.

lib/Logger.ts
sangaman: I was on the fence here too, so I broke it down to one variable per line.

PR #1047, opened 2019-06-20
test: breakswap

moshababo:
Added test cases which try to break the protocol via an adversarial (custom) xud client.

Tests are divided to two types: these which are relying on channels to be open for them beforehand, and these which are doing it themselves, because they simulate a scenario where one of the channels remains in unsettled state, due to a pending HTLC.

Channels and wallet balance is being checked after each test, as this is one of the main purposes of these tests.

New npm script was added for running them:

```
$ npm run test:sim:security
```

Due to the long execution time, it was not added to Travis.


test/simulation/tests-security.go
```diff
@@ -0,0 +1,470 @@
...
+
+	err = closeLtcChannel(ht.ctx, net.LndLtcNetwork, net.Bob.LndLtcNode, bobLtcChanPoint, false)
+	ht.assert.EqualError(err, "rpc error: code = Unknown desc = cannot co-op close channel with active htlcs")
+
+	// Close the channel uncooperatively.
+
+	err = closeLtcChannel(ht.ctx, net.LndLtcNetwork, net.Bob.LndLtcNode, bobLtcChanPoint, true)
+	ht.assert.NoError(err)
+
+	//_, err = net.LndLtcNetwork.LtcMiner.Node.Generate(1100)
```
sangaman: Should these lines be removed? I figure this is where you're trying to see if the balance will show up in lnd, but ultimately it shouldn't be needed because it should show up in the unconfirmed and total balance numbers immediately.

sangaman:
We do need to make sure the regular simulation tests aren't affected tho, I'm retrying on travis but i see an error I don't normally see of an insufficient balance, although it passes for me locally.

Also, when I run the security tests I sometimes get a bunch of ports that remain in use after the test is over used by the `go` process that prevent me from running any simulation tests again unless I kill them manually.

```
listen tcp 127.0.0.1:19558: bind: address already in use

2019/07/02 10:42:38 lnd-ltc: finished with error (stderr):
exit status 1
listen tcp 127.0.0.1:19555: bind: address already in use

2019/07/02 10:42:38 lnd-ltc: finished with error (stderr):
exit status 1
listen tcp 127.0.0.1:19564: bind: address already in use

2019/07/02 10:42:38 lnd-ltc: finished with error (stderr):
exit status 1
listen tcp 127.0.0.1:19561: bind: address already in use
```

kilrau:
Then better don't merge until these are resolved..

sangaman:
This is the error I'm seeing in Travis for the simulation tests that I haven't seen before:

```
    --- FAIL: TestIntegration/order_broadcast_and_invalidation (0.22s)
        require.go:794: 
            	Error Trace:	actions.go:121
            	            				tests-integration.go:180
            	            				harnessTest.go:89
            	            				xud_test.go:72
            	            	rpc error: code = Unknown desc = BTC outbound balance of 0 is not sufficient for order amount of 20000
            	Test:       	TestIntegration/order_broadcast_and_invalidation
    require.go:794: 
        	Error Trace:	xud_test.go:111
        	            	rpc error: code = Unknown desc = cannot co-op close channel with active htlcs
        	Test:       	TestIntegration
```

test/simulation/tests-security.go
moshababo: Yes, it was now removed.

test/simulation/tests-security.go
```diff
@@ -268,37 +271,39 @@ func testMakerShutdownAfter1stHTLC(net *xudtest.NetworkHarness, ht *harnessTest)
...
 
-	//_, err = net.LndLtcNetwork.LtcMiner.Node.Generate(1100)
-	//ht.assert.NoError(err)
-	//err = net.Bob.LndLtcNode.WaitForBlockchainSync(ht.ctx)
-	//ht.assert.NoError(err)
-	//
-	//time.Sleep(5 * time.Second)
+	time.Sleep(5 * time.Second)
+
+	// Wait for HTLC output to be spentfully confirmed.
```
sangaman: "Spentfully" is a new word for me :stuck_out_tongue:. I'm guessing this means to just say "to be fully confirmed"?

moshababo: I'm not familiar with it as well ;)
Will fix.

PR #1055, opened 2019-06-24
feat: token identifiers for currencies

### LINKED ISSUE #910: Currency primary key
How to generically set a unique primary key for coins & ERC20 tokens to prevent intentionally or unintentionally tricking a trading partner into thinking my contract is WETH on xud p2p message layer, even though it's not. Swaps will fail in the end, but that would be annoying.

Short-term:
Use token contract address as unique identifier for ERC20.

Long-term:
A way might be to call getinfo from lnd/raiden and use this as chain id. Token chains like ethereum **additionally** need the token address (https://github.com/ExchangeUnion/xud/issues/897) to be part of the primary key.

sangaman:
At a high level, this adds a concept of a token identifier for each currency supported by xud that is consistent across the entire network.

NOTE: Although this is a breaking change in terms of the p2p messaging, I wound up implementing this without any changes to the database. We already store token addresses in the database, and the lnd chain is determined dynamically after we call `GetInfo`, so no db changes are made and there's no need to drop existing databases after this PR is merged.

"BTC" can mean mainnet tokens or testnet tokens (or simnet, etc...) and can even be misconfigured to point to the LTC lightning network. Meanwhile, Raiden ERC20 tokens refer to a particular token contract, and there's no guarantee that peers won't be using different (and therefore incompatible) contracts for the same currency such as WETH.

Here, we use an identifier to ensure peers are referring to the same token as we are. For currencies that use lnd, the token identifier is determined dynamically using the `chain` and `network` returned by lnd on a `GetInfo` call. For example, BTC mainnet would be "bitcoin-mainnet" and LTC testnet would be "litecoin-testnet". For currencies that use Raiden, we use the token contract address as the identifier.

These token identifiers are shared with peers during the handshake and on any subsequent node state updates. The p2p messaging format is changed to accommodate this. This also drops static node values like the `nodePubKey` and `version` from the node state p2p message, as these values can not change mid-session for a peer.

A future improvement may be to detect and support discrepancies in the currency symbol used by a peer for the same token. For example, a peer that advertises "XBT" with a token identifier of "bitcoin-mainnet" should be compatible with our "BTC" that also is identified as "bitcoin-mainnet". This is not included in this commit as it would require further refactoring and complexity.

Another follow-up may be creating a test suite for the `Peer` class or refactoring it to move logic elsewhere (along with tests for that logic). There is some logic around handling node state updates where information regarding the peer's supported currencies may change that is currently not tested.

Closes #910.

BREAKING CHANGE: Changed p2p messaging structure for `SessionInit` and `NodeStateUpdate` packets.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -121,6 +121,26 @@ class OrderBook extends EventEmitter {
     return outgoingOrder ;
   }
 
+  /**
+   * Checks that a currency advertised by a peer are known to us, have a swap client identifier,
+   * and that their token identifier matches ours.
+   */
+  private isPeerCurrencySupported = (peer: Peer, currency: string) => {
+    const currencyAttributes = this.getCurrencyAttributes(currency);
+    if (!currencyAttributes) {
+      return false; // we don't know about this currency
+    }
+
+    if (!peer.getIdentifier(currencyAttributes.swapClient, currency)) {
+      return false; // peer did not provide a swap client identifier for this currency
+    }
+
+    // ensure that our token identifiers match
+    const ourTokenIdentifier = this.pool.getTokenIdentifier(currency);
+    const peerTokenIdentifier = peer.getTokenIdentifier(currency);
+    return ourTokenIdentifier === peerTokenIdentifier;
+  }
+
   private bindPool = () => {
     this.pool.on('packet.order', this.addPeerOrder);
     this.pool.on('packet.orderInvalidation', this.handleOrderInvalidation);
@@ -128,19 +148,15 @@ class OrderBook extends EventEmitter {
     this.pool.on('packet.swapRequest', this.handleSwapRequest);
     this.pool.on('peer.close', this.removePeerOrders);
     this.pool.on('peer.pairDropped', this.removePeerPair);
-    this.pool.on('peer.pairsAdvertised', this.verifyPeerPairs);
+    this.pool.on('peer.verifyPairs', this.verifyPeerPairs);
     this.pool.on('peer.nodeStateUpdate', (peer) => {
-      // remove any trading pairs for which we no longer have both swap client identifiers
-      peer.forEachActivePair((activePairId) => {
-        const [baseCurrency, quoteCurrency] = activePairId.split('/');
-        const isCurrencySupported = (currency: string) => {
-          const currencyAttributes = this.getCurrencyAttributes(currency);
-          return currencyAttributes && peer.getIdentifier(currencyAttributes.swapClient, currency);
-        };
+      const advertisedCurrencies = peer.getAdvertisedCurrencies();
 
-        if (!isCurrencySupported(baseCurrency) || !isCurrencySupported(quoteCurrency)) {
-          // this peer's node state no longer supports at least one of the currencies for this trading pair
-          peer.deactivatePair(activePairId);
+      advertisedCurrencies.forEach((advertisedCurrency) => {
+        if (!this.isPeerCurrencySupported(peer, advertisedCurrency)) {
+          peer.disableCurrency(advertisedCurrency);
+        } else {
+          peer.enableCurrency(advertisedCurrency);
         }
       });
     });
@@ -170,10 +186,7 @@ class OrderBook extends EventEmitter {
 
   /** Loads the supported pairs and currencies from the database. */
   public init = async () => {
-    const promises: PromiseLike<any>[] = [this.repository.getPairs(), this.repository.getCurrencies()];
-    const results = await Promise.all(promises);
-    const pairs = results[0] as PairInstance[];
-    const currencies = results[1] as CurrencyInstance[];
+    const [pairs, currencies] = await Promise.all([this.repository.getPairs(), this.repository.getCurrencies()]);
 
     currencies.forEach(currency => this.currencyInstances.set(currency.id, currency));
     pairs.forEach((pair) => {
@@ -184,8 +197,8 @@ class OrderBook extends EventEmitter {
     this.pool.updatePairs(this.pairIds);
   }
 
-  public getCurrencyAttributes(symbol: string) {
-    const currencyInstance = this.currencyInstances.get(symbol);
+  public getCurrencyAttributes(currency: string) {
+    const currencyInstance = this.currencyInstances.get(currency);
     return currencyInstance ? currencyInstance.toJSON() : undefined;
   }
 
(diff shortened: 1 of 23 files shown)


kilrau:
:arrow_up: quickly created these follow-up issues

lib/orderbook/OrderBook.ts
```diff
@@ -167,10 +183,7 @@ class OrderBook extends EventEmitter {
 
   /** Loads the supported pairs and currencies from the database. */
   public init = async () => {
-    const promises: PromiseLike<any>[] = [this.repository.getPairs(), this.repository.getCurrencies()];
-    const results = await Promise.all(promises);
-    const pairs = results[0] as PairInstance[];
-    const currencies = results[1] as CurrencyInstance[];
+    const [pairs, currencies] = await Promise.all([this.repository.getPairs(), this.repository.getCurrencies()]);
```
erkarl: Nice refactor!

lib/p2p/Peer.ts
```diff
@@ -197,12 +226,19 @@ class Peer extends EventEmitter {
...
    * Prepares a peer for use by establishing a socket connection and beginning the handshake.
    * @param ownNodeState our node state data to send to the peer
-   * @param nodeKey our identity node key
+   * @param ownNodeKey our identity node key
    * @param expectedNodePubKey the expected nodePubKey of the node we are opening a connection with
    * @param retryConnecting whether to retry to connect upon failure
    * @returns the session init packet from beginning the handshake
    */
-  public beginOpen = async (ownNodeState: NodeState, nodeKey: NodeKey, expectedNodePubKey?: string, retryConnecting = false):
+  public beginOpen = async ({ ownNodeState, ownNodeKey, ownVersion, expectedNodePubKey, retryConnecting = false }:
```
erkarl: nit: ownVersion is missing from the typedoc

lib/p2p/Peer.ts
```diff
@@ -216,25 +252,25 @@ class Peer extends EventEmitter {
...
   /**
    * Finishes opening a peer for use by marking the peer as opened, completing the handshake,
    * and setting up the ping packet timer.
    * @param ownNodeState our node state data to send to the peer
-   * @param nodeKey our identity node key
+   * @param ownNodeKey our identity node key
    * @param sessionInit the session init packet we received when beginning the handshake
    */
-  public completeOpen = async (ownNodeState: NodeState, nodeKey: NodeKey, sessionInit: packets.SessionInitPacket) => {
+  public completeOpen = async (ownNodeState: NodeState, ownNodeKey: NodeKey, ownVersion: string, sessionInit: packets.SessionInitPacket) => {
```
erkarl: nit: same here, `ownVersion` missing from the typedoc

lib/p2p/Peer.ts
```diff
@@ -794,22 +847,19 @@ class Peer extends EventEmitter {
...
+      await this.initSession(ownNodeState, ownNodeKey, ownVersion, sessionInit.body!.nodePubKey);
     }
   }
 
   private checkPairs = () => {
-    const inactivePairs = this.advertisedPairs.filter(pair => !this.activePairs.has(pair));
-    if (inactivePairs.length) {
-      this.emit('pairsAdvertised', inactivePairs);
-    }
+    this.emit('verifyPairs');
```
erkarl: would it make sens to inline this on 279 since the function is now a 1-liner?

lib/p2p/Pool.ts
```diff
@@ -31,8 +31,8 @@ interface Pool {
   on(event: 'packet.getOrders', listener: (peer: Peer, reqId: string, pairIds: string[]) => void): this;
   on(event: 'packet.orderInvalidation', listener: (orderInvalidation: OrderPortion, peer: string) => void): this;
   on(event: 'peer.close', listener: (peerPubKey?: string) => void): this;
-  /** Adds a listener to be called when a peer has newly advertised pairs. */
-  on(event: 'peer.pairsAdvertised', listener: (peer: Peer, pairIds: string[]) => void): this;
+  /** Adds a listener to be called when a peer's advertised but inactive pairs should be verified. */
```
erkarl: nit: `advertised, but`

lib/p2p/Pool.ts
```diff
@@ -68,8 +68,12 @@ interface NodeConnectionIterator {
  * interface for other modules to interact with the p2p layer.
  */
 class Pool extends EventEmitter {
+  /** The version of xud  we are using. */
```
erkarl: nit: extra space before `we`

lib/p2p/Pool.ts
```diff
@@ -45,8 +45,8 @@ interface Pool {
   emit(event: 'packet.getOrders', peer: Peer, reqId: string, pairIds: string[]): boolean;
   emit(event: 'packet.orderInvalidation', orderInvalidation: OrderPortion, peer: string): boolean;
   emit(event: 'peer.close', peerPubKey?: string): boolean;
-  /** Notifies listeners that a peer has newly activated pairs. */
-  emit(event: 'peer.pairsAdvertised', peer: Peer, pairIds: string[]): boolean;
+  /** Notifies listeners that a peer's advertised but inactive pairs should be verified. */
```
erkarl: nit: same here - `advertised, but`

lib/orderbook/OrderBook.ts
```diff
@@ -118,26 +118,42 @@ class OrderBook extends EventEmitter {
...
   private bindPool = () => {
     this.pool.on('packet.order', this.addPeerOrder);
     this.pool.on('packet.orderInvalidation', this.handleOrderInvalidation);
     this.pool.on('packet.getOrders', this.sendOrders);
     this.pool.on('packet.swapRequest', this.handleSwapRequest);
     this.pool.on('peer.close', this.removePeerOrders);
     this.pool.on('peer.pairDropped', this.removePeerPair);
-    this.pool.on('peer.pairsAdvertised', this.verifyPeerPairs);
+    this.pool.on('peer.verifyPairs', this.verifyPeerPairs);
     this.pool.on('peer.nodeStateUpdate', (peer) => {
```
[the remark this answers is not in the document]
sangaman: Rename `verifyPeerPairs` you mean? Yeah I'm not in love with the current naming, but maybe I'd call it `checkAdvertisedPairs`. It's technically checking both pairs and currencies, but I think the end purpose is to determine which trading pairs to activate.

lib/p2p/Peer.ts
sangaman: I noticed that with destructuring, typedoc is no longer showing the comments for the parameters via `@param`. I'm going to switch to the suggestion in this comment since that seems to work: https://github.com/TypeStrong/typedoc/issues/653#issuecomment-504444480

lib/p2p/Peer.ts
sangaman: Thought about doing that but left it like that in case `checkPairs` takes on more logic, but on some further thought I don't know when that would happen, so I went and made it an inline function like you suggested.

lib/p2p/Pool.ts
sangaman: I think grammatically it's correct either way. To me it reads better without the punctuation, but if this was confusing to you I can change it.

sangaman:
Responded to the feedback and also added debug logging statements to `Peer.enableCurrency` and `Peer.disableCurrency`/

lib/orderbook/OrderBook.ts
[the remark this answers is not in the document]
erkarl: I meant moving the callback of `peer.nodeStateUpdate` to a named function (instead of having it anon. inline) like all the other listeners for pool currently are.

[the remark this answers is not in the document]
sangaman: Ohhh gotcha, I'll do that in a separate PR/commit.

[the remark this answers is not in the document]
erkarl: Sounds good.

Issue #1054, opened 2019-06-24
UNITS_PER_CURRENCY to be fetched from db


kilrau:
Reported by @offerm :

Trying to swap LTC/DAI. Failed.
I see this in the log
```
23/06/2019 13:59:15.460 [SWAPS] debug: New deal: {"takerCltvDelta":576,"rHash":"4f6572529ebf0b858c2458280a459e7e7dff023af30269037083fb2c4797f4f5","orderId":"0b6b3be0-95bf-11e9-9818-8f149c949153","pairId":"LTC/DAI","proposedQuantity":10000000,"rPreimage":"431bb884c5dde90d20d5a664a01ccea41b74dfff7b39ffdd57a1c272bd9d0cfe","takerCurrency":"LTC","makerCurrency":"DAI","takerAmount":10000000,"makerAmount":null,"destination":"0x7ed0299Fa1ADA71D10536B866231D447cDFa48b9","peerPubKey":"02b66438730d1fcdf4a4ae5d3d73e847a272f160fee2938e132b52cab0a0d9cfc6","localId":"16213d00-95bf-11e9-976d-eb69bb9b1bb9","price":86.1,"isBuy":false,"phase":0,"state":0,"role":0,"createTime":1561298355457}
```
 
note the `makerAmount":null`
diving into the code I found this:
```
Swaps.UNITS_PER_CURRENCY = {
    BTC: 1,
    LTC: 1,
    WETH: 10 ** 10,
};
```
Clearly DAI is missing. Also addcurrency can't be really used like that.
Not sure I understand why WETH has 10**10 units per currency.

erkarl:
I added https://github.com/ExchangeUnion/xud/pull/1058 as a quick fix. A more permanent/dynamic solution would be to fetch these values from the database, but I think we can lower the priority of this to after testnet/before mainnet.

kilrau:
TODO in `lib/raidenclient/RaidenClient.ts`:
`Populate the mapping from the database (Currency.decimalPlaces)`

PR #1058, opened 2019-06-24
fix(swaps): add hardcoded support for DAI

ghost:
In this PR we're addressing the symptoms of https://github.com/ExchangeUnion/xud/issues/1054 and add a hardcoded support for DAI. A more dynamic fix where we fetch these values from the database will be added in a separate PR.


sangaman (approved):
Good for a temporary fix, thanks.

PR #1064, opened 2019-06-25
refactor(raiden): update channel methods

sangaman:
This updates the methods around querying channels to match the latest stable raiden API.

PR #1065, opened 2019-06-25
feat(raiden): check for direct channels

### LINKED ISSUE #1027: Direct Channel Checks (Updated)
- similar to sanity swaps

sangaman:
Closes #1027.

This adds logic to check that we have a direct channel with a peer before enabling trading for a given Raiden token. Currently multi-hop Raiden payments are not well supported, and this will prevent any attempts to swap Raiden tokens without a direct channel.

Note, this branch is based off of #1055 (which has useful refactoring for determining whether to disable specific currencies trading) and #1064 (which updates raiden calls used in this PR) so it must be rebased and merged only after those two PRs are merged. The key commit to review for the purposes of this PR is the last one.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -129,7 +129,7 @@ class OrderBook extends EventEmitter {
     return order.quantity >= minQuantity;
   }
   /**
-   * Checks that a currency advertised by a peer are known to us, have a swap client identifier,
+   * Checks that a currency advertised by a peer is known to us, has a swap client identifier,
    * and that their token identifier matches ours.
    */
   private isPeerCurrencySupported = (peer: Peer, currency: string) => {
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -64,10 +64,10 @@ class Peer extends EventEmitter {
   public active = false;
   /** Timer to periodically call getNodes #402 */
   public discoverTimer?: NodeJS.Timer;
-  /** Currencies that we have verified that we can swap for this peer. */
+  /** Currencies that we have verified we can swap with this peer. */
   public verifiedCurrencies = new Set<string>();
   /**
-   * Currencies that we cannot swap because we are missing a swap client identifier or because our
+   * Currencies that we cannot swap because we are missing a swap client identifier or because the
    * peer's token identifier for this currency does not match ours - for example this may happen
    * because a peer is using a different raiden token contract address for a currency than we are.
    */
@@ -128,6 +128,10 @@ class Peer extends EventEmitter {
 
   public get addresses(): Address[] | undefined {
     return this.nodeState ? this.nodeState.addresses : undefined;
+  }
+
+  public get raidenAddress(): string | undefined {
+    return this.nodeState ? this.nodeState.raidenAddress : undefined;
   }
 
   /** Returns a list of trading pairs advertised by this peer. */
--- a/lib/raidenclient/RaidenClient.ts
+++ b/lib/raidenclient/RaidenClient.ts
@@ -51,13 +51,14 @@ class RaidenClient extends SwapClient {
   private disable: boolean;
   private unitConverter: UnitConverter;
   private maximumOutboundAmounts = new Map<string, number>();
+  private directChannelChecks: boolean;
 
   /**
    * Creates a raiden client.
    */
   constructor(
-    { config, logger, unitConverter }:
-    { config: RaidenClientConfig, logger: Logger, unitConverter: UnitConverter },
+    { config, logger, unitConverter, directChannelChecks = false }:
+    { config: RaidenClientConfig, logger: Logger, unitConverter: UnitConverter, directChannelChecks: boolean },
   ) {
     super(logger);
     const { disable, host, port } = config;
@@ -66,6 +67,7 @@ class RaidenClient extends SwapClient {
     this.host = host;
     this.disable = disable;
     this.unitConverter = unitConverter;
+    this.directChannelChecks = directChannelChecks;
   }
 
   /**
@@ -201,12 +203,35 @@ class RaidenClient extends SwapClient {
     // not implemented, raiden does not use invoices
   }
 
-  public getRoutes =  async (_amount: number, _destination: string) => {
-    // stub placeholder, query routes not currently implemented in raiden
-    // assume a fixed lock time of 100 Raiden's blocks
-    return [{
+  public getRoutes = async (amount: number, destination: string, currency: string) => {
+    // a query routes call is not currently provided by raiden
+
+    /** A placeholder route value that assumes a fixed lock time of 100 Raiden's blocks. */
+    const placeholderRoute = {
       getTotalTimeLock: () => 101,
-    }];
+    };
+
+    if (this.directChannelChecks) {
+      // temporary check for a direct channel in raiden
+      const tokenAddress = this.tokenAddresses.get(currency);
+      const channels = await this.getChannels(tokenAddress);
+      for (const channel of channels) {
+        if (channel.partner_address && channel.partner_address === destination) {
+          const balance = channel.balance;
+          if (balance >= amount) {
+            this.logger.debug(`found a direct channel for ${currency} to ${destination} with ${balance} balance`);
+            return [placeholderRoute];
+          } else {
+            this.logger.warn(`direct channel found for ${currency} to ${destination} with balance of ${balance} is insufficient for ${amount})`);
+            return []; // we have a direct channel but it doesn't have enough balance, return no routes
+          }
+        }
+      }
+      this.logger.warn(`no direct channel found for ${currency} to ${destination}`);
+      return []; // no direct channels, return no routes
+    } else {
+      return [placeholderRoute];
+    }
   }
 
   public getHeight = async () => {
--- a/bin/xud
+++ b/bin/xud
@@ -72,6 +72,11 @@ const { argv } = require('yargs')
       type: 'string',
       alias: 'x',
     },
+    'debug.raidenDirectChannelChecks': {
+      describe: 'Whether to require direct channels for raiden payments',
+      type: 'boolean',
+      default: undefined,
+    },
     'http.port': {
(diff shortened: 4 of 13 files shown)


sangaman:
To be clear, this disables a trading pair entirely with a given peer if the pair involves a raiden currency for which we don't have a direct channel. I can certainly make this configurable so it can easily be turned off, and also make the logging more clear. 

Making the check happen only when an order is placed doesn't really work since we may have channels with some peers but not with others, and may connect to other peers later on for whom we don't have direct channels.

Another option is to check right before we attempt or accept a swap, but this involves more code and potentially a bad user experience for users to be shown orders that are guaranteed not to work.

erkarl:
> this disables a trading pair entirely with a given peer if the pair involves a raiden currency for which we don't have a direct channel

I feel this would provide much better user experience than displaying a warning upon placing an order or executing a swap. Also, you have already done the hard work and implemented it. 

The concern that Offer had was that we want to be able to turn it off it starts to cause problems. I'd make it configurable under debug section. `debugRaidenDirectChannelChecks=true`?

kilrau:
> The concern that Offer had was that we want to be able to turn it off it starts to cause problems. I'd make it configurable under debug section. `debugRaidenDirectChannelChecks=true`?

Thanks @erkarl . After sleeping on this, I also think we should go ahead with it. Make it configurable, default to `true`. I'd rather call it `raidenDirectChannelChecks=true` in a [debug] section in the config, but that's not important.

Reason: everything else I could come up with is not better or quite some coding work. Also, simnet can support direct channels in an automated manner. And I'll put a clear warning on this limitation in the wiki install guides.

lib/orderbook/OrderBook.ts
```diff
@@ -118,26 +119,58 @@ class OrderBook extends EventEmitter {
...
+      const channels = await swapClient.getChannels(swapClient.tokenAddresses.get(currency));
+      let doesDirectChannelExist = false;
+      for (const channel of channels) {
+        if (channel.partner_address && channel.partner_address === peer.raidenAddress) {
+          doesDirectChannelExist = true;
+          break;
+        }
+      }
+      if (!doesDirectChannelExist) {
+        return false;
```
michael1011: I don't know if it is just me but this looks more complicated than it should be. Why not move:
```
    // ensure that our token identifiers match
    const ourTokenIdentifier = this.pool.getTokenIdentifier(currency);
    const peerTokenIdentifier = peer.getTokenIdentifier(currency);
    return ourTokenIdentifier === peerTokenIdentifier;
```

Into the the for loop so that it gets executed when a channel is found and if no channel matches this function should just return false. That way there is no need for `doesDirectChannelExist` and the little extra logic to handle it.

lib/p2p/Peer.ts
```diff
@@ -64,8 +64,14 @@ class Peer extends EventEmitter {
   public active = false;
   /** Timer to periodically call getNodes #402 */
   public discoverTimer?: NodeJS.Timer;
-  /** Currencies that we have verified that we can swap with for this peer. */
+  /** Currencies that we have verified that we can swap for this peer. */
```
michael1011: I would say that I swap *with* someone or do you mean that we can do swaps *for* this peer instance?

lib/p2p/Peer.ts
```diff
@@ -64,8 +64,14 @@ class Peer extends EventEmitter {
...
   public verifiedCurrencies = new Set<string>();
+  /**
+   * Currencies that we cannot swap because we are missing a swap client identifier or because our
```
michael1011:
```suggestion
   * Currencies that we cannot swap because we are missing a swap client identifier or because their
```

lib/p2p/Peer.ts
```diff
@@ -82,6 +88,10 @@ class Peer extends EventEmitter {
   private readonly responseMap: Map<string, PendingResponseEntry> = new Map();
   private connectTime!: number;
   private connectionRetriesRevoked = false;
+  /** The version of xud this peer is using. */
+  private _version?: string;
+  /** The node pub key of this peer. */
+  private _nodePubKey?: string;
```
michael1011: Why do these new variables have an underscore as prefix?

lib/p2p/Pool.ts
```diff
@@ -68,8 +68,12 @@ interface NodeConnectionIterator {
  * interface for other modules to interact with the p2p layer.
  */
 class Pool extends EventEmitter {
+  /** The version of xud  we are using. */
```
michael1011:
```suggestion
  /** The version of xud we are using. */
```

erkarl:
Will review and test this manually once it is rebased.

lib/p2p/Pool.ts
sangaman: Good catch, this was fixed in #1055.

lib/p2p/Peer.ts
sangaman: I made getter functions for these two variables and needed to use different naming for the variables themselves, this was the cleanest way I could think to do it. It was done as part of #1055. I can change it in a separate PR if you have a better approach in mind.

lib/p2p/Peer.ts
sangaman: I was thinking in the sense that we have verified that we can swap the currency, and we have verified it for this peer. But I actually like your way better. "Currencies that we have verified we can swap with this peer."

lib/orderbook/OrderBook.ts
sangaman: We'll want to make this check for LND based currencies as well, which won't enter the for loop because it is raiden only. That's why I have the logic above outside the loop. Does that make sense?

sangaman:
I updated the commit to check for a direct channel *with sufficient balance* before each swap attempt. I put the logic in the `getRoutes` function for the raiden client, if we don't have a channel with enough balance the method will now return an empty array and the swap will fail due to `NoRouteFound` before we ever attempt to make payments for the swap.

lib/orderbook/OrderBook.ts
sangaman: This is also out of date given the recent change to check for direct channels before each swap.

lib/raidenclient/RaidenClient.ts
```diff
@@ -201,12 +203,31 @@ class RaidenClient extends SwapClient {
...
+      for (const channel of channels) {
+        if (channel.partner_address && channel.partner_address === destination) {
+          if (channel.balance >= amount) {
+            return [placeholderRoute];
+          } else {
+            return []; // we have a direct channel but it doesn't have enough balance, return no routes
+          }
+        }
+      }
+      return []; // no direct channels, return no routes
```
kilrau: I definitely would print a `NoDirectChannels` error here, not generic `NoRouteFound`. Maybe sth along the lines. `Swap failed due to no direct raiden channel, please make sure to have a direct raiden channel with this peers and try swapping again.`

sangaman: I can log messages like that here explaining that either there is no direct channel with this peer or that we have one but it doesn't have enough balance, but note that either way this will go in the log output and not in a grpc response where it's front and center for a user who has just issued an `xucli` command can see.

sangaman: I added log statements - I'm hesitant though to introduce a new swap failure reason tho since it would require some non-trivial refactoring and since this is a temporary safeguard that we may remove later. `NoRouteFound` I think is descriptive enough, it tells the user that there's no suitable route to make a payment and the standard corrective action for that in any case is to open a channel to the peer we want to swap with.

lib/raidenclient/RaidenClient.ts
```diff
@@ -201,12 +203,35 @@ class RaidenClient extends SwapClient {
     // not implemented, raiden does not use invoices
   }
 
-  public getRoutes =  async (_amount: number, _destination: string) => {
-    // stub placeholder, query routes not currently implemented in raiden
-    // assume a fixed lock time of 100 Raiden's blocks
-    return [{
+  public getRoutes =  async (amount: number, destination: string, currency: string) => {
```
erkarl: Normally I'd advocate an integration test here, but since this is temporary logic... :satisfied:

lib/raidenclient/RaidenClient.ts
```diff
@@ -201,12 +203,35 @@ class RaidenClient extends SwapClient {
...
+    };
+
+    if (this.directChannelChecks) {
+      // temporary check for a direct channel in raiden
+      const tokenAddress = this.tokenAddresses.get(currency);
+      const channels = await this.getChannels(tokenAddress);
+      for (const channel of channels) {
+        if (channel.partner_address && channel.partner_address === destination) {
+          const balance = channel.balance;
+          if (balance >= amount) {
```
erkarl: Not a major issue, but my first reaction here was that we're comparing  amount to units, but upon further inspection it turned out not to be the case.

We might want to consider calling these `unitAmounts` or `units`.

sangaman: Hmmm these are in satoshis though, whereas `units` elsewhere refer to the smallest base unit for the currency. If it's still confusing to you maybe I can clarify it another way, maybe just a line or two of comments explaining exactly what these lines do?

erkarl: Interesting... the `getRoutes` calls are made with units though:
https://github.com/ExchangeUnion/xud/blob/master/lib/swaps/Swaps.ts#L268
https://github.com/ExchangeUnion/xud/blob/master/lib/swaps/Swaps.ts#L529

erkarl: Think we need to convert the incoming units to satoshis then.

sangaman: Ahhh yes, good point, I will straighten this out.

sangaman: Per our discussion, I'll handle renaming these variables in a separate PR. These are both comparing base units as you originally stated.

lib/raidenclient/RaidenClient.ts
kilrau: Alright. We'll make it clear in the documentation.

PR #1066, opened 2019-06-25
feat(raiden): channel balance by currency

### LINKED ISSUE #1051: channel balance (raiden)
I have two tokens defined (WETH, DAI) and one channel with another node. The channel balance is 1e18.

When I use xucli channelbalance I see the balance reported for both tokens.
```
xud@xud-test-1:~/scripts/raiden/exchange$ xucli channelbalance
{
  "balancesMap": [
    [
      "BTC",
      {
        "balance": 3750157448,
        "pendingOpenBalance": 0
      }
    ],
    [
      "DAI",
      {
        "balance": 1000000000000000000,
        "pendingOpenBalance": 0
      }
    ],
    [
      "LTC",
      {
        "balance": 226339963038,
        "pendingOpenBalance": 0
      }
    ],
    [
      "WETH",
      {
        "balance": 1000000000000000000,
        "pendingOpenBalance": 0
      }
    ]
  ]
}
```

sangaman:
Closes #1051.

Note: This is based off of #1064.

This modifies the logic around querying raiden for channel balances. Rather than summing up all balances across all tokens, it allows for specifying a currency and retrieving only the balance for that currency. Calling `ChannelBalance` without specifying a currency will return each currency separately.

It also scales the balance returned by raiden to satoshis (10^-8). This is because the default units of 10^-18 for currencies such as WETH can exceed the maximum value of a `uint64` used by the gRPC layer.

### DIFF
--- a/lib/raidenclient/RaidenClient.ts
+++ b/lib/raidenclient/RaidenClient.ts
@@ -41,6 +41,12 @@ class RaidenClient extends SwapClient {
   private host: string;
   private disable: boolean;
 
+  // TODO: Populate the mapping from the database (Currency.decimalPlaces).
+  private static readonly UNITS_PER_CURRENCY: { [key: string]: number } = {
+    WETH: 10 ** 10,
+    DAI: 10 ** 10,
+  };
+
   /**
    * Creates a raiden client.
    */
@@ -272,14 +278,18 @@ class RaidenClient extends SwapClient {
   }
 
   /**
-   * Returns the total balance available across all channels.
+   * Returns the total balance available across all channels for a specified currency.
    */
-  public channelBalance = async (): Promise<ChannelBalance> => {
-    // TODO: refine logic to determine balance per token rather than all combined
-    const channels = await this.getChannels();
+  public channelBalance = async (currency?: string): Promise<ChannelBalance> => {
+    if (!currency) {
+      return { balance: 0, pendingOpenBalance: 0 };
+    }
+
+    const channels = await this.getChannels(this.tokenAddresses.get(currency));
     const balance = channels.filter(channel => channel.state === 'opened')
       .map(channel => channel.balance)
-      .reduce((acc, sum) => sum + acc, 0);
+      .reduce((sum, acc) => sum + acc, 0)
+      / (RaidenClient.UNITS_PER_CURRENCY[currency] || 1);
     return { balance, pendingOpenBalance: 0 };
   }
 
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -119,23 +119,27 @@ class Service {
   public channelBalance = async (args: { currency: string }) => {
     const { currency } = args;
     const balances = new Map<string, { balance: number, pendingOpenBalance: number }>();
-    const getBalance = async (currency: string) => {
-      const swapClient = this.swapClientManager.get(currency.toUpperCase());
-      if (swapClient) {
-        const channelBalance = await swapClient.channelBalance();
-        return channelBalance;
-      } else {
-        throw swapsErrors.SWAP_CLIENT_NOT_FOUND(currency);
-      }
-    };
 
     if (currency) {
       argChecks.VALID_CURRENCY(args);
-      balances.set(currency, await getBalance(currency));
-    } else {
-      for (const currency of this.orderBook.currencies) {
-        balances.set(currency, await getBalance(currency));
+
+      const swapClient = this.swapClientManager.get(currency.toUpperCase());
+      if (swapClient) {
+        const channelBalance = await swapClient.channelBalance(currency);
+        balances.set(currency, channelBalance);
+      } else {
+        throw swapsErrors.SWAP_CLIENT_NOT_FOUND(currency);
       }
+    } else {
+      const balancePromises: Promise<any>[] = [];
+      this.swapClientManager.swapClients.forEach((swapClient, currency) => {
+        if (swapClient.isConnected()) {
+          balancePromises.push(swapClient.channelBalance(currency).then((channelBalance) => {
+            balances.set(currency, channelBalance);
+          }));
+        }
+      });
+      await Promise.all(balancePromises);
     }
 
     return balances;
--- a/lib/swaps/SwapClient.ts
+++ b/lib/swaps/SwapClient.ts
@@ -44,8 +44,10 @@ abstract class SwapClient extends EventEmitter {
 
   /**
    * Returns the total balance available across all channels.
+   * @param currency the currency whose balance to query for, otherwise all/any
+   * currencies supported by this client are included in the balance.
    */
-  public abstract channelBalance(): Promise<ChannelBalance>;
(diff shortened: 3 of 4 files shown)


kilrau:
Needs rebase

sangaman:
Rebased and tested locally, the new logic works for me.

kilrau:
Restarted the failed tests - passed

lib/raidenclient/RaidenClient.ts
```diff
@@ -41,6 +41,12 @@ class RaidenClient extends SwapClient {
   private host: string;
   private disable: boolean;
 
+  // TODO: Populate the mapping from the database (Currency.decimalPlaces).
```
michael1011: Maybe open an issue for this `TODO` so that you don't forget about it

sangaman: I think #1054 is tracking this.

lib/raidenclient/RaidenClient.ts
```diff
@@ -272,14 +278,17 @@ class RaidenClient extends SwapClient {
...
+  public channelBalance = async (currency?: string): Promise<ChannelBalance> => {
+    if (!currency) {
+      return { balance: 0, pendingOpenBalance: 0 };
+    }
+
+    const channels = await this.getChannels(this.tokenAddresses.get(currency));
     const balance = channels.filter(channel => channel.state === 'opened')
       .map(channel => channel.balance)
-      .reduce((acc, sum) => sum + acc, 0);
+      .reduce((acc, sum) => sum + (acc / RaidenClient.UNITS_PER_CURRENCY[currency]), 0);
```
erkarl: Think we want to switch places with `sum` and `acc` here (and rename them to `sum` and `channelBalance`. 
`arr.reduce(callback(accumulator, currentValue[, index[, array]]), [, initialValue])`

Maybe do the conversion after reduce so that it's only going to be calculated once?

I think this will cause confusion on the CLI level if we show satoshis per coin as it's meant to be our internal unit only. What do you think about converting Raiden channel balance values to `satsToCoinsStr` and then doing additional conversion to smallest unit of the currency on CLI level?

lib/raidenclient/RaidenClient.ts
```diff
@@ -272,14 +278,17 @@ class RaidenClient extends SwapClient {
   }
 
   /**
-   * Returns the total balance available across all channels.
+   * Returns the total balance available across all channels for a specified currency.
    */
-  public channelBalance = async (): Promise<ChannelBalance> => {
-    // TODO: refine logic to determine balance per token rather than all combined
-    const channels = await this.getChannels();
+  public channelBalance = async (currency?: string): Promise<ChannelBalance> => {
```
erkarl: Let's add a test for this one to:
https://github.com/ExchangeUnion/xud/blob/master/test/jest/RaidenClient.spec.ts

lib/service/Service.ts
```diff
@@ -119,23 +119,27 @@ class Service {
   public channelBalance = async (args: { currency: string }) => {
```
erkarl: Would be nice to get test coverage for this public function as well. I've got a boilerplate service spec available in my open channel branch:
https://github.com/ExchangeUnion/xud/pull/1044/files#diff-560e7ab3ce0b2935a5eb94ee3c9be6d7R1

lib/raidenclient/RaidenClient.ts
sangaman: I'm a bit confused by what you mean in the first paragraph, could you clarify? I went ahead and made the conversion happen after reduce though.

We can have the cli print whole units for the "formatted" output, that'd be consistent with how we're displaying amounts most other places. That should be a separate PR though I think.

lib/service/Service.ts
sangaman: Is it ok if I add the test case in a separate PR after that PR is merged?

lib/raidenclient/RaidenClient.ts
sangaman: Done.

lib/service/Service.ts
kilrau: yes, merging this is prio

lib/raidenclient/RaidenClient.ts
erkarl: What I was trying to say was that the first argument for the callback of reduce is `sum` and the second `acc`.

With the previous logic:
```
.reduce((acc, sum) => {
  console.log('sum is', sum, 'and converted', (acc / RaidenClient.UNITS_PER_CURRENCY[currency]));
  return sum + (acc / RaidenClient.UNITS_PER_CURRENCY[currency]);
}, 0);
```

Printed:
```
[1] sum is 75000000000000000 and converted 0
```

As a result converting back to satoshis per unit didn't have any effect:
```
➜  xud git:(channel-mgmt-plus-raiden-balance) ✗ ./bin/xucli channelbalance
{
  "balancesMap": [
    ...
    [
      "WETH",
      {
        "balance": 75000000000000000,
        "pendingOpenBalance": 0
      }
    ]
  ]
}
```

With the new logic this isn't an issue, but I'd still rename the arguments.

As for converting back to smallest units on the CLI level - let's open an issue for that to discuss. I think it will cause confusion for the end user:
```
"WETH",
{
  "balance": 7500000,
  "pendingOpenBalance": 0
}
```

lib/service/Service.ts
erkarl: That's fine. Let's do that.

lib/raidenclient/RaidenClient.ts
sangaman: Gotcha.

lib/raidenclient/RaidenClient.ts
```diff
@@ -272,14 +278,18 @@ class RaidenClient extends SwapClient {
...
+    if (!currency) {
+      return { balance: 0, pendingOpenBalance: 0 };
+    }
+
+    const channels = await this.getChannels(this.tokenAddresses.get(currency));
     const balance = channels.filter(channel => channel.state === 'opened')
       .map(channel => channel.balance)
-      .reduce((acc, sum) => sum + acc, 0);
+      .reduce((sum, acc) => sum + acc, 0)
+      / (RaidenClient.UNITS_PER_CURRENCY[currency] || 1);
```
erkarl: One last nit - when commenting out this line the tests are still passing. We might want to consider covering the conversion in `channelBalance calculates the total balance of open channels for a currency` test as well. Up to you :).

sangaman: Makes sense, I'll add this test case and merge.

sangaman: Actually, I remember now that I tried to do this originally but in its current state I'd have to mock a static value and I'm not really sure how to do that. Maybe revisit this when we get the units per currency from the database?

erkarl: Fair enough.

PR #1072, opened 2019-06-29
Prevent Raiden from being first swap leg

offerm:
This PR prevents Raiden from being first leg of swap.
It better calculates the maker CLTV by using cltv-delta of 5760 in the case of raiden.
This is only a basis for a better change

### DIFF
--- a/lib/http/HttpService.ts
+++ b/lib/http/HttpService.ts
@@ -5,6 +5,7 @@ class HttpService {
   constructor(private service: Service) {}
 
   public resolveHashRaiden = async (resolveRequest: RaidenResolveRequest): Promise<RaidenResolveResponse> => {
+    // TODO: add reveal_timeout, settle time out,  token, etc
     const secret = await this.service.resolveHash({
       rHash: resolveRequest.secrethash.slice(2),
       amount: resolveRequest.amount,
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -12,7 +12,7 @@ import { SwapDealInstance } from '../db/types';
 import { SwapDeal, SwapSuccess, SanitySwap, ResolveRequest } from './types';
 import { generatePreimageAndHash, setTimeoutPromise } from '../utils/utils';
 import { PacketType } from '../p2p/packets';
-import SwapClientManager from './SwapClientManager';
+import SwapClientManager, { isRaidenClient } from './SwapClientManager';
 import { errors } from './errors';
 
 export type OrderToAccept = Pick<SwapDeal, 'quantity' | 'price' | 'localId' | 'isBuy'> & {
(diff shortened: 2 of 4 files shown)


kilrau:
Nice! Tests need to be fixed.

lib/swaps/Swaps.ts
```diff
@@ -681,12 +693,16 @@ class Swaps extends EventEmitter {
...
     let destination: string;
-
+    // TODO: check cltv value
     switch (deal.role) {
       case SwapRole.Maker:
         expectedAmount = deal.makerAmount;
         source = 'Taker';
         destination = 'Maker';
+        if (deal.makerCltvDelta! > 50 * 2) {
+          this.failDeal(deal, SwapFailureReason.UnknownError, 'Wrong CLTV received on first leg');
```
sangaman: Maybe use `SwapFailureReason.InvalidResolveRequest` instead of the unknown error?

lib/http/HttpService.ts
```diff
@@ -5,6 +5,7 @@ class HttpService {
   constructor(private service: Service) {}
 
   public resolveHashRaiden = async (resolveRequest: RaidenResolveRequest): Promise<RaidenResolveResponse> => {
+    // TODO: add reveal_timeout, settle time out,  token, etc
```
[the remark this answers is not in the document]
offerm: we need to verify that we got the expected reveal_timeout (as we agreed during the swap preparation (if we are not preventing raiden to be the first leg))

We also need to check that we got the payment with the correct token.

lib/swaps/Swaps.ts
```diff
@@ -433,6 +433,19 @@ class Swaps extends EventEmitter {
 
     const { makerCurrency, makerAmount, takerCurrency, takerAmount } = Swaps.calculateMakerTakerAmounts(quantity, price, isBuy, requestBody.pairId);
 
+    const makerSwapClient = this.swapClientManager.get(makerCurrency)!;
```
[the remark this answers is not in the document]
offerm: Agree

lib/swaps/Swaps.ts
offerm: Agree

offerm:
@sangaman I created this PR as a starting point for the change. Please take ownership and make changes as needed.

sangaman:
Gotcha, I updated the PR with some minor changes per my comments to make the tests pass.

lib/http/HttpService.ts
erkarl: nit: extra whitespace before `token`

lib/swaps/Swaps.ts
```diff
@@ -433,6 +433,29 @@ class Swaps extends EventEmitter {
...
+    if (!makerSwapClient) {
```
erkarl: It's interesting that we need this extra logic here since we already have a passing test in master that covers this:
https://github.com/ExchangeUnion/xud/blob/master/test/jest/integration/Swaps.spec.ts#L123

lib/swaps/Swaps.ts
```diff
@@ -433,6 +433,29 @@ class Swaps extends EventEmitter {
...
+        peer,
+        rHash,
+        failureReason: SwapFailureReason.SwapClientNotSetup,
+        errorMessage: 'Unsupported maker currency',
+        reqId: requestPacket.header.id,
+      });
+      return false;
+    }
+
+    if (isRaidenClient(makerSwapClient)) {
```
erkarl: I assume this is temporary logic that's going to be removed in #1082?

sangaman: Yes.

lib/swaps/Swaps.ts
sangaman: Yes it's only needed due to what I described here: https://github.com/ExchangeUnion/xud/pull/1072#discussion_r299194032

It's basically just to check for a null value.

PR #1076, opened 2019-07-02
fix(orderbook): show aggregated quantity

ghost:
As discussed with @kilrau in this PR we remove the `depth` column from the orderbook and replace it with `volume` instead.

Previously we displayed `price * quantity` per price point, but what we actually want is all available quantity at one price point aggregated.

### DIFF
--- a/lib/cli/commands/orderbook.ts
+++ b/lib/cli/commands/orderbook.ts
@@ -10,15 +10,15 @@ type FormattedOrderbook = {
   rows: string[][],
 };
 
-type BucketDepth = {
+type Bucket = {
   price: number,
-  depth: number;
+  quantity: number;
 };
 
 type OrderbookJson = {
   pairId: string,
-  sell: BucketDepth[],
-  buy: BucketDepth[],
+  sell: Bucket[],
+  buy: Bucket[],
 };
 
 const COLUMNS = [19, 19, 19, 19];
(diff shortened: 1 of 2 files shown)


kilrau (changes requested):
Tested, works as it should! Minor nit: let's swap the price and volume column for 'Buy', that puts the prices next to each other like on other orderbooks and one can see easier.

```
Trading pair: LTC/BTC
│   Volume            │ Price           │ Price             │ Volume            │
│  2           │ 0.0078                 │ 0.00785           │ 1                 │
│ 3.5           │ 0.0077               │ 0.0079            │ 2                 │
```

kilrau:
Reference orderbook: https://www.kraken.com/prices

sangaman (commented):
"Volume" in a trading context sounds to me like the amount that has been traded, not the amount that's available to trade. I know that's the term Kraken is using, but that still seems like non-standard usage of the word (reference: https://www.investopedia.com/terms/v/volume.asp).

Coinbase uses "Market Size" for this term and bitfinex uses "Amount" for this purpose. I'd be fine with either of those, or maybe just "Quantity". So I'd prefer to use one of those terms and just use the new way of calculating it (without multiplying by price).

I also agree with Kilian's suggestion to have the two price columns adjacent to each other.

kilrau:
> I'd be fine with either of those, or maybe just "Quantity".

Agree, let's stick with "Quantity" then, because that's what we use in `listorders` already.

Also, I don't think our `xucli help` description of `listorders` and `orderbook` is distinctive/descriptive enough. I'd go for sth along the lines
`xucli listorders [pair_id] [include_own_orders] [limit]    list orders from the order book` (no change)
`xucli orderbook [pair_id] [precision]    display the order book, with orders aggregated per price point`

Summary of quick fixes:
- [ ] Volume -> Quantity
- [ ] Switch `Price` and `Quantity` column for `Buy`
- [ ] `xucli help` `orderbook` command description

erkarl:
Rebased and updated.

kilrau (approved):
LGTM!

moshababo (approved):

sangaman (approved):
Thanks!

PR #1077, opened 2019-07-03
fix: clean shutdown

sangaman:
This removes unnecessary code that created duplicate swap client managers and interfered with the graceful shutdown procedure.


erkarl:
Thanks for catching/dealing with this :).

Issue #1079, opened 2019-07-03
Persist swap phase in db


kilrau:
Basics: all measures to handle failures/attacks in a swap flow gracefully (#942 ) require the node to know which phase the swap is in. This issue is about persisting the `SwapPhase` & `SwapState` (reason to be in that state as per [types.ts](https://github.com/ExchangeUnion/xud/blob/master/lib/swaps/types.ts#L12)) in db after each step in the protocol, read it after restarts/crashes and display swaps in critical phases in `getinfo`.

### Status Quo
Currently these [`SwapPhase`](https://github.com/ExchangeUnion/xud/blob/master/lib/constants/enums.ts#L57)s are already present, but only success/failure is persisted in the `swapdeals` table in `xud.db`:
```
export enum SwapPhase {
  /** The swap has been created locally. */
  SwapCreated = 0,
  /** The swap has been sent to a peer to request approval. */
  SwapRequested = 1,
  /** The terms of the swap have been agreed to, and we will attempt to execute it. */
  SwapAgreed = 2,
  /**
   * We have commanded swap client to send payment according to the agreed terms. The payment (and swap)
   * could still fail due to no route with sufficient capacity, lack of cooperation from the
   * receiver or any intermediary node along the route, or an unexpected error from swap client.
   */
  SendingAmount = 3,
  /** We have received the agreed amount of the swap, and the preimage is now known to both sides. */
  AmountReceived = 4,
  /** The swap has been formally completed and both sides have confirmed they've received payment. */
  SwapCompleted = 5,
}
```

### TODO
- [x] Persist each phase change in db, **only continue with next step if phase was successfully stored in db**
- [x] New swap phases as per [Swap Flow](https://github.com/ExchangeUnion/xud/pull/1025#issuecomment-502047537) without breaking changes:
```
export enum SwapPhase {
  /** Role Taker: The swap has been created locally. */
  SwapCreated = 0,

  /** Role Taker: The swap, stating currencies and amounts has been sent to the maker to request approval. */
  SwapRequested = 1,

  /**
  * Role Maker: We set the agreed currency and amount (can be be less than requested) of the swap and we successfully sent a `SwapAccepted` packet with these details to the Taker. Or successfully declined swap with `SwapFailed` packet.*/
  * Role Taker: The currency and an amount of the swap have been agreed to or declined (we received the `SwapAccepted` or `SwapFailed` packet from the Maker)
*/
  SwapAgreed = 2,

  /**
[...]

PR #1080, opened 2019-07-05
feat(swap): persist active swap deals to db

sangaman:
This commit persists the state of swap deals to the database upon each phase change once a deal has been accepted to. This is done to be able to prevent loss of funds and recover gracefully from any active swaps should `xud` crash unexpectedly.

This is the first step towards closing #1079.

I wound up leaving the existing phases as is, and just updated the comments instead and renamed `SwapAgreed` to `SwapAccepted` to align with our terminology elsewhere. I started adding separate phases for maker/taker but it would've required changing quite a few lines of code, and ultimately for recovery purposes we can determine the precise state by looking at both `phase` and `role`.

### DIFF
--- a/lib/constants/enums.ts
+++ b/lib/constants/enums.ts
@@ -55,19 +55,23 @@ export enum SwapRole {
 }
 
 export enum SwapPhase {
-  /** The swap has been created locally. */
+  /** The swap deal has been created locally. */
   SwapCreated = 0,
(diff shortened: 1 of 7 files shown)


kilrau:
Test: what if writing to db fails - what's the behavior? Definitely swap should be failed.

kilrau:
needs your review @erkarl

Issue #1050, opened 2019-06-20
security: maker can lose funds due to a crash


moshababo:
Scenario:
* taker is paying the maker the 1st leg of the swap.
* maker receives the taker HTLC, and is paying the taker the 2nd leg of the swap.
* taker receives the maker HTLC, and settles his invoice as expected.
* maker 2nd leg payment completed and he receives the preimage, but his xud crashes before he manages to settle his own invoice.

Outcome: maker already paid but cannot redeem his pending payment because he lost the preimage. Taker funds are still locked but he can get them back after the HTLC will expire, leaving the maker with none.

sangaman:
As we discussed, the solution here is to write the preimage and hash for every pending swap to database as soon as they become known to us. If xud crashes, when it restarts we check the database for pending swaps and try to settle any invoices for which we known the preimage. If we don't have the preimage, we call `ListPayments` on lnd to see if we've made a payment for any of the pending hashes, and thus learned of the preimage for that hash, and then try to settle an invoice using that preimage. If we don't have the preimage after all that, we cancel any pending invoices to avoid funds from being lucked up unnecessarily.

sangaman:
Resolved by #1081, which implements my approach described above.

PR #1081, opened 2019-07-05
feat(swaps): recover crashed swap deals

### LINKED ISSUE #1079: Persist swap phase in db

sangaman:
Closes #1079 & #1183. This builds off of PR #1080.

This commit attempts to recover swap deals that were interrupted due to a system or `xud` crash. In the case where we are the maker and have attempted to send payment for the second leg of the swap, we attempt to query the swap client for the preimage of that payment in case it went through. We can then use that preimage to try to claim the payment from the first leg of the swap. In all other cases, we simply attempt to close any open invoices and mark the swap deal as having errored.

Raiden currently does not expose an API call to query for the preimage of a completed payment. Any pending swaps are listed in the `GetInfo` response.

The recovery attempts happen once on `xud` startup by looking for any swap deals in the database that have an `Active` state.

I haven't tested this code yet but it is ready to review, if I have time I will also start adding some test cases. This may be a good candidate for a simulation test, since recovering funds in the case of xud crashing while waiting for the `sendPayment` response depends substantially on lnd.

This commit includes a suite of test cases for the newly added functionality. It also adds a new suite of simulation tests to test how `xud` responds to instability. It simulates crashes immediately after the maker sends payment as part of a swap. When the maker comes back online, it will fetch the preimage for the successful payment and use it to settle the incoming payment from the interrupted swap.

This tests two key cases, one where the maker's payment goes through while the maker is offline, and another where the payment goes through only after the maker has come back online. In the latter case, the maker must detect that the payment is in a pending state and monitor it in case it goes through.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -1,6 +1,6 @@
 import grpc, { ChannelCredentials, ClientReadableStream } from 'grpc';
 import Logger from '../Logger';
-import SwapClient, { ClientStatus, SwapClientInfo } from '../swaps/SwapClient';
+import SwapClient, { ClientStatus, SwapClientInfo, PaymentState } from '../swaps/SwapClient';
 import errors from './errors';
 import { errors as swapErrors } from '../swaps/errors';
 import { LightningClient, WalletUnlockerClient } from '../proto/lndrpc_grpc_pb';
@@ -703,11 +703,12 @@ class LndClient extends SwapClient {
   }
 
   public settleInvoice = async (rHash: string, rPreimage: string) => {
+    const settleInvoiceRequest = new lndinvoices.SettleInvoiceMsg();
+    settleInvoiceRequest.setPreimage(hexToUint8Array(rPreimage));
+    await this.settleInvoiceLnd(settleInvoiceRequest);
+
     const invoiceSubscription = this.invoiceSubscriptions.get(rHash);
     if (invoiceSubscription) {
-      const settleInvoiceRequest = new lndinvoices.SettleInvoiceMsg();
-      settleInvoiceRequest.setPreimage(hexToUint8Array(rPreimage));
-      await this.settleInvoiceLnd(settleInvoiceRequest);
       this.logger.debug(`settled invoice for ${rHash}`);
       invoiceSubscription.cancel();
     }
@@ -718,10 +719,52 @@ class LndClient extends SwapClient {
     if (invoiceSubscription) {
       const cancelInvoiceRequest = new lndinvoices.CancelInvoiceMsg();
       cancelInvoiceRequest.setPaymentHash(hexToUint8Array(rHash));
-      await this.cancelInvoice(cancelInvoiceRequest);
-      this.logger.debug(`canceled invoice for ${rHash}`);
+      try {
+        await this.cancelInvoice(cancelInvoiceRequest);
+        this.logger.debug(`canceled invoice for ${rHash}`);
+      } catch (err) {
+        // handle errors due to attempting to remove an invoice that doesn't exist
+        if (err.message === 'unable to locate invoice') {
+          this.logger.debug(`attempted to cancel non-existent invoice for ${rHash}`);
+        } else if (err.message === 'invoice already canceled') {
+          this.logger.debug(`attempted to cancel already canceled invoice for ${rHash}`);
+        } else {
+          throw err;
+        }
+      }
       invoiceSubscription.cancel();
     }
+  }
+
+  public lookupPayment = async (rHash: string) => {
+    const payments = await this.listPayments(true);
+    for (const payment of payments.getPaymentsList()) {
+      if (payment.getPaymentHash() === rHash) {
+        switch (payment.getStatus()) {
+          case lndrpc.Payment.PaymentStatus.SUCCEEDED:
+            const preimage = payment.getPaymentPreimage();
+            return { preimage, state: PaymentState.Succeeded };
+          case lndrpc.Payment.PaymentStatus.IN_FLIGHT:
+            return { state: PaymentState.Pending };
+          default:
+            this.logger.warn(`unexpected payment state for payment with hash ${rHash}`);
+            /* falls through */
+          case lndrpc.Payment.PaymentStatus.FAILED:
+            return { state: PaymentState.Failed };
+        }
+      }
+    }
+
+    // if no payment is found, we assume that the payment was never attempted by lnd
+    return { state: PaymentState.Failed };
+  }
+
+  private listPayments = (includeIncomplete?: boolean): Promise<lndrpc.ListPaymentsResponse> => {
+    const request = new lndrpc.ListPaymentsRequest();
+    if (includeIncomplete) {
+      request.setIncludeIncomplete(includeIncomplete);
+    }
+    return this.unaryCall<lndrpc.ListPaymentsRequest, lndrpc.ListPaymentsResponse>('listPayments', request);
   }
 
   private addHoldInvoice = (request: lndinvoices.AddHoldInvoiceRequest): Promise<lndinvoices.AddHoldInvoiceResp> => {
--- a/lib/raidenclient/RaidenClient.ts
+++ b/lib/raidenclient/RaidenClient.ts
@@ -1,22 +1,25 @@
-import http from 'http';
-import Logger from '../Logger';
-import SwapClient, { ClientStatus, ChannelBalance } from '../swaps/SwapClient';
-import errors from './errors';
-import { SwapDeal } from '../swaps/types';
-import { SwapClientType, SwapState, SwapRole } from '../constants/enums';
 import assert from 'assert';
-import {
-  RaidenClientConfig,
-  RaidenInfo,
-  OpenChannelPayload,
-  Channel,
-  TokenPaymentRequest,
-  TokenPaymentResponse,
-} from './types';
-import { UnitConverter } from '../utils/UnitConverter';
+import http from 'http';
+import { SwapClientType, SwapRole, SwapState } from '../constants/enums';
 import { CurrencyInstance } from '../db/types';
+import Logger from '../Logger';
+import SwapClient, { ChannelBalance, ClientStatus, PaymentState } from '../swaps/SwapClient';
+import { SwapDeal } from '../swaps/types';
+import { UnitConverter } from '../utils/UnitConverter';
+import errors from './errors';
+import { Channel, OpenChannelPayload, PaymentEvent, RaidenClientConfig, RaidenInfo, TokenPaymentRequest, TokenPaymentResponse } from './types';
 
 type RaidenErrorResponse = { errors: string };
+
+type PendingTransfer = {
+  initiator: string;
+  locked_amount: string;
+  payment_identifier: string;
+  role: string;
+  target: string;
+  token_address: string;
+  transferred_amount: string;
+};
 
 /**
  * A utility function to parse the payload from an http response.
@@ -74,6 +77,14 @@ class RaidenClient extends SwapClient {
 
   public get minutesPerBlock() {
     return 0.25; // 15 seconds per block target
+  }
+
+  /**
+   * Derives an integer identifier using the first 4 bytes of a provided payment hash in hex.
+   * @param rHash a payment hash in hex
+   */
+  private static getIdentifier(rHash: string) {
+    return parseInt(rHash.substr(0, 8), 16);
   }
 
   /**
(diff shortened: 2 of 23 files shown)


kilrau:
TBD: raiden

kilrau:
Also: in the recovery cases which are not handled here, e.g. lnd crashing this oldie comes in relevant I think: https://github.com/ExchangeUnion/xud/issues/684 @sangaman

sangaman:
Here's the summary of what I'm trying to cover in this PR, I'm going to push out updated code later today that covers the pending payment cases more thoroughly. The notable gap that I have is how we can claim or settle a raiden payment upon resuming xud. I think in terms of how to handle edge cases where lnd or raiden crash while xud is running (if a payment goes through without us getting a response, for instance), that should be covered in a separate PR since the scope of this PR is already quite large.

## Phases

### `SwapCreated`,  `SwapRequested`, & `SwapAgreed`

No payments have begun and the swap cannot currently be recovered or resumed, so we mark the swap deal as errored in the database.

### `SendingPayment`

#### Taker

The taker is not at risk of losing funds, as only they know the preimage for any pending payments. In this case, the appropriate action is to cancel any invoices (lnd only) for the swap and mark it as errored in the database.

Canceling the invoice for lnd should cancel any pending HTLCs that the maker would have made to us.

#### Maker

The maker is at risk of losing funds. If there is a crash on either xud or lnd while we are attempting to send payment, we need to determine whether our payment went through or not. If it goes through, we need to claim our payment on the other chain before it expires. Once we're certain it has not and will not go through, we can cancel our invoice on the receiving network (lnd only).

To determine the status of our payment attempt we call `ListPayments` in lnd and `/api/(version)/payments/(token_address)/(target_address)` as well as as `/api/(version)/pending_transfers`in raiden.

- For failed payments or payments that are not found (indicating that they were never actually attempted), we cancel any invoices (lnd only) and mark the swap as errored in the database.
- For successful payments, we retrieve the preimage from the `Payment` message (for lnd) or `EventPaymentReceivedSuccess` event (for raiden). 
	- For lnd, we can claim the payment on the other network immediately using `SettleInvoice`.
	- For raiden, there is not currently a way to "push" a preimage to raiden so that it can settle a payment - more research/development is needed on this issue and for now we will log a warning.
- For pending payments, we will schedule a timer to continue querying for the status of pending payments until they have all resolved.

### `PaymentReceived`

We have received payment for our side of the swap, and the only further action would be to mark the swap as completed in the database.

sangaman:
The updated approach tracks any outgoing payments that are still pending upon xud restarting. If an outgoing payment is still in flight and we do not have the preimage for it, we add it to a set of "pending" swaps and check on it on a scheduled interval until we can determine whether it has failed or succeeded. I set the timer to trigger every 5 minutes to check on any swaps that are still pending.

A new `SwapRecovery` class is introduced to contain the logic for recovering interrupted swap deals and for tracking swaps that are still pending.

Swaps that time out during the `SendingPayment` phase also get handled by by the swap recovery logic, since we will need to continue checking on them to determine whether our payment ever goes through.

The key missing piece is that Raiden currently does not expose an API call to push a preimage to claim an incoming payment, instead we print a warning to the log for now. One solution to this shortcoming would be to have Raiden continuously query the resolver endpoint if it does not receive a response, in which case once xud learns of the preimage it can pass it to raiden the next time there is a resolve request for that hash. 

Another consideration would be to list any pending swaps via the rpc layer as a way to inform users that xud is monitoring pending payments. This could be added to the `GetInfo` call and/or a separate call to list pending swaps (this could also return swaps that are ongoing in xud during normal operation).

lib/swaps/SwapRecovery.ts
```diff
@@ -0,0 +1,118 @@
...
+class SwapRecovery {
+  /** A set of swaps where we have a pending outgoing payment for swaps where we don't know the preimage. */
+  public pendingSwaps: Set<SwapDealInstance> = new Set();
+  private pendingSwapsTimer?: NodeJS.Timeout;
+
+  constructor(private swapClientManager: SwapClientManager, private logger: Logger) { }
+
+  public beginTimer = () => {
+    if (!this.pendingSwapsTimer) {
+      this.pendingSwapsTimer = setInterval(this.checkPendingSwaps, 300000);
```
kilrau: Maybe add a comment that this is in ms

kilrau:
Moshe's test case(https://github.com/ExchangeUnion/xud/issues/1050) should pass after this is merged @sangaman @erkarl

lib/lndclient/LndClient.ts
```diff
@@ -597,6 +596,38 @@ class LndClient extends SwapClient {
     }
   }
 
+  public lookupPayment = async (rHash: string) => {
```
erkarl: What do you think about adding an integration test to this once since this is a new public method that is full of logic?

lib/lndclient/LndClient.ts
```diff
@@ -597,6 +596,38 @@ class LndClient extends SwapClient {
...
+    const payments = await this.listPayments(true);
+    for (const payment of payments.getPaymentsList()) {
+      if (payment.getPaymentHash() === rHash) {
+        const paymentStatus = payment.getStatus();
+        if (payment.getStatus() === lndrpc.Payment.PaymentStatus.SUCCEEDED) {
```
erkarl: nit: would a switch statement here look better?

lib/raidenclient/RaidenClient.ts
```diff
@@ -203,6 +235,57 @@ class RaidenClient extends SwapClient {
     // not implemented, raiden does not use invoices
   }
 
+  public lookupPayment = async (rHash: string, currency?: string, destination?: string) => {
```
erkarl: Same comment as for lnd's `lookupPayment` -> I think we should consider integration testing this in addition to upcoming simulation tests.

lib/swaps/SwapRecovery.ts
```diff
@@ -0,0 +1,118 @@
...
+import { SwapDealInstance } from '../db/types';
+import Logger from '../Logger';
+import { SwapPhase, SwapState, SwapFailureReason, SwapRole, SwapClientType } from '../constants/enums';
+import SwapClient, { PaymentState } from './SwapClient';
+
+/**
+ * A class that's responsible for recovering swap deals that were interrupted due to a system or xud crash,
+ * ensuring that we do not lose funds on a partially completed swap.
+ */
+class SwapRecovery {
```
erkarl: Would it make sense to add a separate test suite for this new class?

lib/swaps/SwapRecovery.ts
```diff
@@ -0,0 +1,118 @@
...
+      clearInterval(this.pendingSwapsTimer);
+      this.pendingSwapsTimer = undefined;
+    }
+  }
+
+  private failDeal = async (deal: SwapDealInstance, receivingSwapClient?: SwapClient) => {
+    if (receivingSwapClient) {
+      try {
+        await receivingSwapClient.removeInvoice(deal.rHash);
+      } catch (err) {}
```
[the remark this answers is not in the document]
sangaman: Yeah that would be good.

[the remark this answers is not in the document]
sangaman: Actually looking into this a bit more, and remembering why I did it like this in the first place, lnd throws an error if you remove a non-existing invoice which will frequently be the case. But what I'll do is check for that specific error message and only ignore that one, logging any other type of error.

lib/lndclient/LndClient.ts
sangaman: Indeed it would.

lib/lndclient/LndClient.ts
sangaman: I felt there wasn't much for me to test in the LndClient case (we always just make a single grpc call and return a value based on that call), but I did add cases for the Raiden `lookupPayment` logic since it is a bit more involved.

lib/raidenclient/RaidenClient.ts
```diff
@@ -18,6 +18,30 @@ import { CurrencyInstance } from '../db/types';
 
 type RaidenErrorResponse = { errors: string };
 
+type PaymentEvent = {
```
erkarl: Can we move these new types to `raidenclient/types.ts`?

sangaman:
@erkarl I moved the type and fixed the issue you ran into, thanks for catching that.

erkarl:
@sangaman after testing this with your updated code I'm getting the following output for maker's xud:
```
[1] 09/09/2019 12:11:20.721 [SWAPS] info: recovered LTC swap payment of 1000 using preimage f8a7cd8560654c7a64855ff5a769ae309505d347431cce07c446873a8d524a14
```

However, when I check the channel balances it looks like the maker did not receive any LTC. I also checked the balance on lnd level.

sangaman:
> @sangaman after testing this with your updated code I'm getting the following output for maker's xud:
> 
> ```
> [1] 09/09/2019 12:11:20.721 [SWAPS] info: recovered LTC swap payment of 1000 using preimage f8a7cd8560654c7a64855ff5a769ae309505d347431cce07c446873a8d524a14
> ```
> 
> However, when I check the channel balances it looks like the maker did not receive any LTC. I also checked the balance on lnd level.
> 
> Looking at the `listchannels` for LTC the HTLC is still showing up as pending.

That's strange, as this means setle invoice should have been called. I'll see if I can reproduce.

sangaman:
@erkarl @kilrau I fixed the issue with the recovered payment not settling correctly, and added a new suite of "instability" simulation tests that use a new [instability](https://github.com/ExchangeUnion/xud/tree/instability) branch. It wasn't easy but it simulates exactly the cases we are trying to address with this PR and the tests pass consistently. See the second commit message and the updated top post of this PR for more details.

kilrau:
Very nice @sangaman , let's wait for @erkarl 's review

sangaman:
I think keeping the `instability` branch separate is fine, the code changes are minimal and it should be simple enough to just merge the latest without conflicts most of the time. One key change is it shortens the delay between checking on pending swaps from 30 seconds to 1 second to make the tests go faster, but I think every second would be too often for actual use.

I'm neutral on whether it should be part of the travis tests, I followed the `security` tests which aren't currently run in travis.

PR #1084, opened 2019-07-08
chore: update dependencies and shrinkwrap

sangaman:
This updates project dependencies and removes unused dependencies. It also creates an npm shrinkwrap file to peg all dependencies and nested dependencies to specific version for fresh installs.


package.json
```diff
@@ -37,7 +37,7 @@
     "test:jest": "jest",
     "test:seedutil": "jest seedutil",
     "test:jest:watch": "jest --watch",
-    "typedoc": "typedoc --out typedoc --module commonjs --target es6 lib --readme none",
+    "typedoc": "typedoc --out typedoc --module commonjs --target es6 lib --readme none --ignoreCompilerErrors",
```
michael1011: Why `--ignoreCompilerErrors`?

sangaman: Good question I actually meant to mention this in the PR. When I was testing things, I noticed typedoc is pinned to using an older version of typescript (3.2.x) which raises a false positive compilation error in the codebase which doesn't exist with newer versions of typescript. I added this flag so that the `typedoc` script wouldn't fail. I figure that we already run the `compile` script as part of the `version` and `publish` scripts, so we don't need the potentially stale version of typescript that comes with typedoc to check for compilation errors.

Issue #915, opened 2019-04-27
Limit routing fees, avoid routing fees for sanity swaps


kilrau:
EDIT:
* Sanity swaps: shouldn't consume any routing fees, e.g. by cancelling the swap before releasing the preimage. Should be limited to e.g. 1 sat and one sanity swap per pubkey to avoid an exploit using this to lock up funds of other peers.
* Swap deals: dynamic 3% max of send amount. Make it configurable, [currently 3% are already hardcoded](https://github.com/ExchangeUnion/xud/pull/1086/).

kilrau:
I think a limit on routing fees is good enough for `1.0.0`. It looks like https://github.com/ExchangeUnion/xud/blob/master/proto/lndrpc.proto#L733 is setting 8 satoshis and
https://github.com/ExchangeUnion/xud/blob/master/proto/lndrpc.proto#L1355 is setting 5 satoshis as limits.

Is that correct? How about ltc? @sangaman @erkarl

kilrau:
New insight: in a multi-hop scenario routing fees for our sanity swaps become problematic.

We have to
a) set a max threshold for sanity swaps
a) avoid an attacker to trigger sanity swaps at will and drain funds from victim via routing fees

kilrau:
As discussed on the call this issue is about setting LND:
* Sanity swaps: 10 sat
* Swap deals: 3%

kilrau:
An idea to avoid routing fees at all for sanity swaps:
* unknown payment hash

PR #1086, opened 2019-07-09
add initial fee control. Allow a up to 3% fee

offerm:


offerm:
This is needed to handle https://github.com/ExchangeUnion/xud-simnet/issues/34

I suggest to accept it as is and create a new issue for having the max fee from  configuration file.

kilrau:
I think the change to 0.03 somehow didn't go through.

erkarl:
No need to update `package-lock.json` within this PR.

kilrau:
> I can make a followup to this to replace the `MAXFEE` variable with a configurable option, but this is fine as a first step. Thanks.

Covered in https://github.com/ExchangeUnion/xud/issues/915 :thumbsup:

PR #1087, opened 2019-07-09
fix(raiden): outbound capacity check

ghost:
In this PR we fix issues related to raiden channel balance caching and checks:
- raiden maximum outbound capacity was always 0 because we didn't specify the currency
- channel balance check before placing an order was comparing units to amounts (it is now comparing amounts because raiden's channelbalance method returns us the amount, satoshis per coin)

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -51,6 +51,7 @@ class LndClient extends SwapClient {
   private chainIdentifier?: string;
   private channelSubscription?: ClientReadableStream<lndrpc.ChannelEventUpdate>;
   private invoiceSubscriptions = new Map<string, ClientReadableStream<lndrpc.Invoice>>();
+  private maximumOutboundAmount = 0;
 
   /**
    * Creates an lnd client.
@@ -108,6 +109,19 @@ class LndClient extends SwapClient {
 
   public get chain() {
     return this.chainIdentifier;
+  }
+
+  public maximumOutboundCapacity = () => {
+    return this.maximumOutboundAmount;
+  }
+
+  protected updateCapacity = async () => {
+    try {
+      this.maximumOutboundAmount = (await this.channelBalance()).balance;
+    } catch (e) {
+      // TODO: Mark client as disconnected
+      this.logger.error(`failed to fetch channelbalance: ${e}`);
+    }
   }
 
   private unaryCall = <T, U>(methodName: string, params: T): Promise<U> => {
--- a/lib/raidenclient/RaidenClient.ts
+++ b/lib/raidenclient/RaidenClient.ts
@@ -14,6 +14,7 @@ import {
   TokenPaymentResponse,
 } from './types';
 import { UnitConverter } from '../utils/UnitConverter';
+import { CurrencyInstance } from '../db/types';
 
 type RaidenErrorResponse = { errors: string };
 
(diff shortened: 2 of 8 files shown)


lib/lndclient/LndClient.ts
```diff
@@ -103,6 +104,19 @@ class LndClient extends SwapClient {
     return this.chainIdentifier;
   }
 
+  public maximumOutboundCapacity = () => {
+    return this.maximumOutboundAmount;
```
sangaman: Thoughts on an `assert` here to check that the currency passed to the method matches the currency for this client?

lib/swaps/SwapClientManager.ts
```diff
@@ -85,6 +85,7 @@ class SwapClientManager extends EventEmitter {
           this.swapClients.set(currency.id, this.raidenClient);
         }
       });
+      await this.raidenClient.updateCapacity();
```
sangaman: I'm a little confused as to why this line is necessary, shouldn't https://github.com/ExchangeUnion/xud/blob/beb238da6dfc102cf35138f2f75d9dcf9e292c9a/lib/swaps/SwapClient.ts#L61 take care of this by calling `updateCapacity` immediately the first time that we set the timer? If you're not sure and it's not apparent after a few minutes we can maybe investigate in a future issue/pr.

If we drop this line, we can go back to making `updateCapacity` `protected` instead of `public`.

lib/raidenclient/RaidenClient.ts
```diff
@@ -70,6 +70,22 @@ class RaidenClient extends SwapClient {
...
   }
 
+  public maximumOutboundCapacity = (currency: string): number => {
+    return this.maximumOutboundAmounts.get(currency) || 0;
+  }
+
+  public updateCapacity = async () => {
+    try {
+      for (const [currency] of this.tokenAddresses) {
+        const maximumOutboundAmount = (await this.channelBalance(currency)).balance;
```
sangaman: It would be nice to `await` these in parallel with `Promise.all`, particularly since this is something that will be called often.

sangaman (approved):
Made a few comments although nothing critical. Thanks for catching this.

lib/raidenclient/RaidenClient.ts
erkarl: Think it won't matter in the context of Raiden because all the API calls are blocking. Nevertheless, I've added the recommended changes so that we're covered on xud side.

lib/lndclient/LndClient.ts
erkarl: I'm not sure about this. If we do it here then it would also make sense to do it for the `channelBalance` method.

lib/swaps/SwapClientManager.ts
erkarl: This was necessary because the first `updateCapacity` calls happens when the swap client has been initialized and `ConnectionVerified` status is set. In the case of Raiden by that time we had not yet set the associated `tokenAddreses`.

I didn't like the fact that we made `updateCapacity` public so I moved the initial setting of `tokenAddresses` into the `init` function of Raiden.

moshababo (approved):

lib/raidenclient/RaidenClient.ts
sangaman: Oh good point, but ok yeah since you've made the change it certainly doesn't hurt and one would hope these become non-blocking at some point in the future.

lib/lndclient/LndClient.ts
sangaman: OK good point, fine as is.

sangaman (approved):

PR #1093, opened 2019-07-11
refactor: remove detectexternalip

### LINKED ISSUE #1092: Remove `detectexternalip`
this is handled by docker now and code and config option should be removed from xud

michael1011:
Closes #1092 

I removed the feature itself and all mentions of it from the code. This involved changing some error codes which is fine I think because I discovered that two distinct errors had the same code.

### DIFF
--- a/bin/xud
+++ b/bin/xud
@@ -124,11 +124,6 @@ const { argv } = require('yargs')
       describe: 'Interval minutes for discover option',
       type: 'number',
     },
-    'p2p.detectexternalip': {
-      describe: 'Detect & advertise external ip address',
-      type: 'boolean',
-      default: undefined,
-    },
     'p2p.port': {
       describe: 'Port to listen for incoming peers',
       type: 'number',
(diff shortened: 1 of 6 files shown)


sangaman:
As discussed in the issue I think we may want to keep this code, so I'll hold off on reviewing for now.

kilrau:
Let's quickly discuss this in the call.

I tested this some days ago: enabled the feature via `xud.conf`, but it didnt' detect my ip/set it in the URI. No log entries. If it works, we can keep it.

kilrau:
If somebody can confirm this is still working (it does not for me) and does NOT conflict with our TOR setup and manual setting URI's, we can close this PR and keep the feature.

sangaman:
@kilrau I just confirmed that external ip detection works for me while not using tor. @erkarl would you have time to try running `xud` with `--p2p.detectexternalip` while using tor to make sure there are no issues?

sangaman:
I am thinking however that even if this does cause issues with tor, we may still want to keep this logic and just add some checks to skip or gracefully fail the external ip detection procedure when we are using tor. So my inclination would be to close this PR.

erkarl:
> So my inclination would be to close this PR.

Agreed.

PR #1101, opened 2019-07-17
refactor(lnd): wait for grpc when verifying conn

sangaman:
This refactors the `verifyConnection` logic to use the grpc `waitForReady` function to wait for the client to be ready before attempting a `GetInfo` call to lnd. This cuts down on the number of calls we need to make to an offline lnd instance to determine when it has come back online.


kilrau:
Restarted builds since both failed. If they are still failing, probably some fixes needed.

PR #1102, opened 2019-07-17
feat(raiden-resolver): add configuration option for listening interface

ghost:
In this PR we add a new configuration option for the raiden's http resolver endpoint. We now allow the user to specify the listening interface.

Previously it was hardcoded to `127.0.0.1` which assumed that raiden and xud are always running on the same machine.

### DIFF
--- a/lib/http/HttpServer.ts
+++ b/lib/http/HttpServer.ts
@@ -83,14 +83,14 @@ class HttpServer {
   /**
    * Starts the server and begins listening on the provided port.
    */
-  public listen = async (port: number) => {
+  public listen = async (port: number, host: string) => {
     return new Promise<void>((resolve, reject) => {
       const listenErrHandler = (err: Error) => {
         reject(err);
       };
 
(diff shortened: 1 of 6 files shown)


kilrau (approved):

lib/http/HttpServer.ts
```diff
@@ -83,14 +83,14 @@ class HttpServer {
   /**
    * Starts the server and begins listening on the provided port.
    */
-  public listen = async (port: number) => {
+  public listen = async (host: string, port: number) => {
```
sangaman: Minor nit but elsewhere in the code for `listen` methods the first parameter is port and second is host, similar to the `server.listen` method that's called below, so I think it'd be better to match that ordering here.

sangaman (changes requested):
We should also add this option to the list of arguments in `bin/xud`.

erkarl:
Thanks for the feedback @sangaman - updated.

sangaman (approved):

PR #1106, opened 2019-07-18
Revert "fix(raiden): getRoutes amount comparison"

sangaman:
Reverts ExchangeUnion/xud#1103

`channel.balance` is in units, so we should not be converting to amount.


erkarl:
Thanks for catching it. My bad.

PR #1109, opened 2019-07-18
fix: round unitsToAmount conversion result

### LINKED ISSUE #1097: Higher Raiden balance causes "Assertion failed" error (`channelbalance`)
```
xud@xud-test-1:~$ xucli channelbalance
Error: 13 INTERNAL: Assertion failed
```

Reliably reproducable with 25btc in dai. That is 25 times 11500 times 1e18. Looks like something still violates the int64 limit.

sangaman:
This prevents unwanted fractional values from the `unitsToAmount` method return value.

Fixes #1097.


erkarl:
Nice catch!

PR #1117, opened 2019-07-24
fix(lnd): wait for invoice subscription

ghost:
In this commit we adjust the `LndClient.addInvoice` logic so that
it waits for the hold invoice open before proceeding with the swap flow.

This prevents scenarios where:
- maker sends a `SwapAcceptedPacket` before `MakerInvoice` is ready to be paid
- taker sends a payment before `TakerInvoice` is ready to be paid

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -10,7 +10,11 @@ import assert from 'assert';
 import { promises as fs } from 'fs';
 import { SwapState, SwapRole, SwapClientType } from '../constants/enums';
 import { SwapDeal } from '../swaps/types';
-import { base64ToHex, hexToUint8Array } from '../utils/utils';
+import {
+  base64ToHex,
+  hexToUint8Array,
+  setTimeoutPromiseReject,
+} from '../utils/utils';
 import { LndClientConfig, LndInfo, ChannelCount, Chain } from './types';
 
 interface LightningMethodIndex extends LightningClient {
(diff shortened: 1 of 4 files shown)


kilrau:
Nice catch!

kilrau:
needs your review @sangaman

lib/lndclient/LndClient.ts
```diff
@@ -610,25 +623,31 @@ class LndClient extends SwapClient {
...
-      this.invoiceSubscriptions.delete(rHash);
-      this.logger.debug(`deleted invoice subscription for ${rHash}`);
-    };
-    invoiceSubscription.on('data', (invoice: lndrpc.Invoice) => {
-      if (invoice.getState() === lndrpc.Invoice.InvoiceState.ACCEPTED) {
-        // we have accepted an htlc for this invoice
-        this.emit('htlcAccepted', rHash, invoice.getValue());
+    return new Promise((resolve) => {
+      if (!this.invoices) {
+        throw errors.LND_IS_UNAVAILABLE(this.status);
```
sangaman: Wouldn't we want this to `reject` rather than `throw` since we are inside a promise?

lib/lndclient/LndClient.ts
```diff
@@ -572,7 +576,16 @@ class LndClient extends SwapClient {
     addHoldInvoiceRequest.setCltvExpiry(cltvExpiry);
     await this.addHoldInvoice(addHoldInvoiceRequest);
     this.logger.debug(`added invoice of ${amount} for ${rHash} with cltvExpiry ${cltvExpiry}`);
-    this.subscribeSingleInvoice(rHash);
+    const SUBSCRIBE_INVOICE_TIMEOUT = 5000;
```
sangaman: I'm wondering whether we need a separate timeout here or whether we can just rely on the existing time limit for the entire swap. Practically speaking I can't think of a reason this should ever take nearly 5 seconds, probably something would be very wrong if so.

sangaman (changes requested):
The changes look fine, but I'm wondering whether this is really necessary? The invoice state isn't well documented within lnd unfortunately, and I do see a test suite within lnd that explicitly waits for newly created invoices to reach the `OPEN` state, but looking deeper into the code it looks like all newly created invoices immediately are `OPEN` by default. We wait to get a response from the add hold invoice call, and I believe that as soon as that call is complete that the invoice is ready to be paid.

It wouldn't make much sense to have to make a second call to know when the first call is truly complete, and I'm pretty sure that such a thing isn't necessary for the regular `SubscibeInvoices` call. In there, the `OPEN` state is meaningful because it indicates that a new invoice has been added, but in our case we already know the invoice exists.

kilrau:
Since it seems somewhat controversial if we need this, I'm taking this out of the bounty release.

erkarl:
> but looking deeper into the code it looks like all newly created invoices immediately are OPEN by default

Thanks for digging deep into this. If you say that's the case then this PR is redundant.

PR #1119, opened 2019-07-24
List currencies enhancements

### LINKED ISSUE #1067: `listcurrencies` enhancements
- return local ticker symbol, [global identifier](https://github.com/ExchangeUnion/xud/pull/1055), digits
- add to cli

rsercano:
This should resolve #1067, I'm not really sure about global identifiers tho. So a review from your side would be great.

### DIFF
--- a/lib/cli/commands/listcurrencies.ts
+++ b/lib/cli/commands/listcurrencies.ts
@@ -1,0 +1,48 @@
+import { callback, loadXudClient } from '../command';
+import { Arguments } from 'yargs';
+import colors from 'colors/safe';
+import Table, { HorizontalTable } from 'cli-table3';
+import { ListCurrenciesRequest, ListCurrenciesResponse } from '../../proto/xudrpc_pb';
+import { SwapClientType } from '../../constants/enums';
+
+const HEADERS = [
+  colors.blue('Ticker'),
+  colors.blue('Digits'),
+  colors.blue('Token Address'),
+  colors.blue('Swap Client'),
+];
+
+const formatCurrencies = (currencies: ListCurrenciesResponse.AsObject) => {
+  const formatted: any[] = [];
+  currencies.currenciesList.forEach((currency) => {
+    const element = [];
+    element.push(currency.currency, currency.decimalPlaces, currency.tokenAddress, SwapClientType[currency.swapClient]);
+    formatted.push(element);
+  });
+  return formatted;
+};
+
+const createTable = () => {
+  const table = new Table({
+    head: HEADERS,
+  }) as HorizontalTable;
+  return table;
+};
+
+const displayTable = (response: ListCurrenciesResponse.AsObject) => {
+  const table = createTable();
+
+  formatCurrencies(response).forEach((currency) => {
+    table.push(currency);
+  });
+  console.log(colors.underline(colors.bold('\nCurrencies:')));
+  console.log(table.toString());
+};
+
+export const command = 'listcurrencies';
+
+export const describe = 'list available currencies';
+
+export const handler = (argv: Arguments) => {
+  loadXudClient(argv).listCurrencies(new ListCurrenciesRequest(), callback(argv, displayTable));
+};
--- a/lib/grpc/GrpcService.ts
+++ b/lib/grpc/GrpcService.ts
@@ -12,7 +12,7 @@ import { errorCodes as lndErrorCodes } from '../lndclient/errors';
 import { LndInfo } from '../lndclient/types';
 import { SwapSuccess, SwapFailure } from '../swaps/types';
 import { SwapFailureReason } from '../constants/enums';
-import { TradeInstance, OrderInstance } from '../db/types';
+import { TradeInstance, OrderInstance, CurrencyInstance } from '../db/types';
 
 /**
  * Creates an xudrpc Order message from an [[Order]].
@@ -227,7 +227,7 @@ class GrpcService {
   /**
    * See [[Service.addCurrency]]
    */
-  public addCurrency: grpc.handleUnaryCall<xudrpc.AddCurrencyRequest, xudrpc.AddCurrencyResponse> = async (call, callback) => {
+  public addCurrency: grpc.handleUnaryCall<xudrpc.Currency, xudrpc.AddCurrencyResponse> = async (call, callback) => {
     try {
       await this.service.addCurrency(call.request.toObject());
       const response = new xudrpc.AddCurrencyResponse();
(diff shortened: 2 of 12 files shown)


proto/xudrpc.proto
```diff
@@ -407,8 +407,8 @@ message ListOrdersResponse {
 
 message ListCurrenciesRequest {}
 message ListCurrenciesResponse {
-  // A list of ticker symbols of the supported currencies.
-  repeated string currencies = 1 [json_name = "currencies"];
+  // The list of available currencies in orderbook.
```
sangaman: "in the order book"

proto/xudrpc.proto
```diff
@@ -467,6 +467,15 @@ message Order {
   uint64 hold = 10 [json_name = "hold"];
 }
 
+message Currency {
+  // Ticker symbol of this currency (e.g. BTC)
+  string ticker_symbol = 1 [json_name = "ticker_symbol"];
+  // The places to the right of the decimal point of the smallest subunit (e.g. 8, 18)
+  uint64 digits = 2 [json_name = "digits"];
```
sangaman: `uint64` is definitely overkill here, `uint32` would be better.

proto/xudrpc.proto
```diff
@@ -467,6 +467,15 @@ message Order {
...
+  // Global identifier of this currency
+  string global_identifier = 3 [json_name = "global_identifier"];
```
sangaman: For LND currencies, we don't actually use a "global" identifier, just descriptors like "bitcoin-mainnet". So maybe let's just call this `identifier` and drop the global references.

proto/xudrpc.proto
sangaman: We call this `decimal_places` elsewhere, I'm thinking we should stick to that same name here even tho it's a bit longer.

lib/proto/xudrpc.swagger.json
```diff
@@ -21,7 +21,7 @@
         "operationId": "AddCurrency",
         "responses": {
           "200": {
-            "description": "",
+            "description": "A successful response.",
```
sangaman: Looks like you have a different `protoc-gen-swagger` from what was used last, I need to update our instructions to use a specific version but for now can you try this and then regenerating protos?

```
git clone https://github.com/grpc-ecosystem/grpc-gateway $GOPATH/src/github.com/grpc-ecosystem/grpc-gateway
cd $GOPATH/src/github.com/grpc-ecosystem/grpc-gateway
git reset --hard f2862b476edcef83412c7af8687c9cd8e4097c0f
go install ./protoc-gen-swagger
```

rsercano:
Thanks Daniel @sangaman

proto/xudrpc.proto
```diff
@@ -467,6 +467,15 @@ message Order {
   uint64 hold = 10 [json_name = "hold"];
 }
 
+message Currency {
+  // Ticker symbol of this currency (e.g. BTC)
+  string ticker_symbol = 1 [json_name = "ticker_symbol"];
+  // The places to the right of the decimal point of the smallest subunit (e.g. 8, 18)
+  uint32 decimal_places = 2 [json_name = "decimal_places"];
+  // Global identifier of this currency
```
sangaman: Can we drop the "global" term here in the comment as well?

lib/cli/commands/listcurrencies.ts
```diff
@@ -0,0 +1,46 @@
+import { callback, loadXudClient } from '../command';
+import { Arguments } from 'yargs';
+import colors from 'colors/safe';
+import Table, { HorizontalTable } from 'cli-table3';
+import { ListCurrenciesRequest, ListCurrenciesResponse } from '../../proto/xudrpc_pb';
+
+const HEADERS = [
+  colors.blue('Ticker'),
+  colors.blue('Digits'),
+  colors.blue('Global Identifier'),
```
sangaman: Drop global here too.

proto/xudrpc.proto
```diff
@@ -467,6 +467,15 @@ message Order {
   uint64 hold = 10 [json_name = "hold"];
 }
 
+message Currency {
```
sangaman: I'm also thinking it would be nice to have the swap client used by the currency here, either lnd or raiden for now. We include this when we add the currency, so it makes sense to have it when we list them.

proto/xudrpc.proto
rsercano: my bad

rsercano:
Implemented @sangaman

proto/xudrpc.proto
```diff
@@ -467,6 +467,17 @@ message Order {
...
```
sangaman: I was going to say that I think it would be better to use the `SwapClient` enum we already have defined above in the proto definition for the `swap_client` field, but then I realized that the structure of this message is actually identical to the `AddCurrencyRequest` message. So I think it would be best to actually just rename `AddCurrencyRequest` to `Currency` and then use it for both the `AddCurrency` request and the `ListCurrencies` response. That way we don't have to maintain two message definitions for the same thing.

For the enum swap client field, in the cli you can print the string representation.

lib/cli/commands/listcurrencies.ts
```diff
@@ -0,0 +1,47 @@
+import { callback, loadXudClient } from '../command';
+import { Arguments } from 'yargs';
+import colors from 'colors/safe';
+import Table, { HorizontalTable } from 'cli-table3';
+import { ListCurrenciesRequest, ListCurrenciesResponse } from '../../proto/xudrpc_pb';
+
+const HEADERS = [
+  colors.blue('Ticker'),
+  colors.blue('Digits'),
+  colors.blue('Identifier'),
```
sangaman: Since this will only ever contain raiden token addresses (which is fine I think), it'd probably be better to just call this `Token address`.

README.md
```diff
@@ -57,4 +57,4 @@ All files in this repository are licensed under the [GNU Affero General Public L
 
 ## Discussion
 
-Comments, questions, and any development-related discussions are welcome in our [Discord chat](https://discord.gg/YgDhMSn)!
\ No newline at end of file
+Comments, questions, and any development-related discussions are welcome in our [Discord chat](https://discord.gg/YgDhMSn)!
```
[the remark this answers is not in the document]
rsercano: Yes, you're right :)

rsercano:
Implemented @sangaman

lib/grpc/GrpcService.ts
```diff
@@ -499,9 +499,17 @@ class GrpcService {
...
-      const listCurrenciesResponse = this.service.listCurrencies();
+      const currencies = this.service.listCurrencies();
       const response = new xudrpc.ListCurrenciesResponse();
-      response.setCurrenciesList(listCurrenciesResponse);
+
+      currencies.currencies.forEach((currency: CurrencyInstance) => {
+        const resultCurrency = new xudrpc.Currency();
+        resultCurrency.setDecimalPlaces(currency.decimalPlaces);
+        resultCurrency.setCurrency(currency.id);
+        resultCurrency.setTokenAddress(currencies.tokenIdentifiers[currency.id] || '');
```
sangaman: I believe you can just use `currency.tokenAddress` here.

lib/grpc/GrpcService.ts
```diff
@@ -499,9 +499,17 @@ class GrpcService {
...
+        resultCurrency.setSwapClient(currency.swapClient === SwapClientType.Lnd ? xudrpc.Currency.SwapClient.LND : xudrpc.Currency.SwapClient.RAIDEN);
```
sangaman: You can just do currency.swapClient as number here - our internal enum matches the enum defined in the rpc proto.

lib/p2p/Pool.ts
```diff
@@ -141,6 +141,10 @@ class Pool extends EventEmitter {
     return this.nodeState.addresses;
   }
 
+  public getTokenIdentifiers() {
```
sangaman: Don't believe this is necessary if my comments above work.

rsercano:
image

PR #1121, opened 2019-07-25
refactor(lnd): use SendPayment instead of SendToRoute

ghost:
In this PR we refactor the logic of `LndClient.sendPayment` to use `SendPayment` instead of `SendToRoute`.

The maximum length/duration of maker's payment is enforced by specifying the `cltvLimit`.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -295,15 +295,95 @@ class LndClient extends SwapClient {
     return this.unaryCall<lndrpc.GetInfoRequest, lndrpc.GetInfoResponse>('getInfo', new lndrpc.GetInfoRequest());
   }
 
-  public sendSmallestAmount = async (rHash: string, destination: string) => {
-    const sendRequest = new lndrpc.SendRequest();
-    sendRequest.setAmt(1);
-    sendRequest.setDestString(destination);
-    sendRequest.setPaymentHashString(rHash);
-    sendRequest.setFinalCltvDelta(this.cltvDelta);
+  public sendSmallestAmount = async (rHash: string, destination: string): Promise<string> => {
+    const request = this.buildSendRequest({
+      rHash,
+      destination,
+      amount: 1,
+      // In case of sanity swaps we don't know the
+      // takerCltvDelta or the makerCltvDelta. Using our
+      // client's default.
+      finalCltvDelta: this.cltvDelta,
+    });
+    const preimage = await this.executeSendRequest(request);
+    return preimage;
+  }
+
+  public sendPayment = async (deal: SwapDeal): Promise<string> => {
+    assert(deal.state === SwapState.Active);
+    let request: lndrpc.SendRequest;
+    assert(deal.makerCltvDelta, 'swap deal must have a makerCltvDelta');
+    if (deal.role === SwapRole.Taker) {
+      // we are the taker paying the maker
+      assert(deal.destination, 'swap deal as taker must have a destination');
+      request = this.buildSendRequest({
+        rHash: deal.rHash,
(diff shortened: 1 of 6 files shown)


kilrau:
needs your review @sangaman

lib/lndclient/LndClient.ts
```diff
@@ -295,84 +295,64 @@ class LndClient extends SwapClient {
...
+    const request = this.buildSendRequest({
+      rHash,
+      destination,
+      amount: 1,
+      // In case of sanity swaps we don't know the
+      // takerCltvDelta or the makerCltvDelta. Using our
+      // client's default.
+      finalCltvDelta: this.cltvDelta,
+    });
+    const preImage = await this.executeSendRequest(request);
```
sangaman: Everywhere else we're just calling this `preimage' without the capitalized I.

lib/lndclient/LndClient.ts
```diff
@@ -295,84 +295,64 @@ class LndClient extends SwapClient {
...
-      return '';
+      // we are the maker paying the taker
+      if (!deal.takerPubKey) {
+        assert.fail('swap deal as maker must have a takerPubKey');
+      }
+      if (!deal.takerCltvDelta) {
+        assert.fail('swap deal as maker must have a takerCltvDelta');
+      }
+      if (!deal.makerCltvDelta) {
+        assert.fail('swap deal as maker must have a makerCltvDelta');
```
sangaman: All these checks can be on one line like `assert(deal.makerCltvDelta, 'swap deal as maker must have a makerCltvDelta')` - no need to have the if statements. I think it'd be a bit more readable.

lib/lndclient/LndClient.ts
```diff
@@ -295,84 +295,64 @@ class LndClient extends SwapClient {
...
+        amount: deal.takerAmount,
+        // Using the agreed upon takerCltvDelta. Taker won't accept
+        // our payment if we provide a smaller value.
+        finalCltvDelta: deal.takerCltvDelta,
+        // Enforcing the maximum duration/length of the payment by
+        // specifying the cltvLimit.
+        cltvLimit: deal.makerCltvDelta,
+      });
     }
+    const preImage = await this.executeSendRequest(request);
```
sangaman: Same here for lower case `preimage`.

lib/lndclient/LndClient.ts
```diff
@@ -295,84 +295,64 @@ class LndClient extends SwapClient {
...
+      finalCltvDelta: this.cltvDelta,
+    });
+    const preImage = await this.executeSendRequest(request);
+    return preImage;
   }
 
   public sendPayment = async (deal: SwapDeal): Promise<string> => {
     assert(deal.state === SwapState.Active);
-
-    if (deal.makerToTakerRoutes && deal.role === SwapRole.Maker) {
```
sangaman: Should we drop the `makerToTakerRoutes` field from the swap deal type?

sangaman (commented):
Looks good, just a few comments and a question about the `makerToTakerRoutes` property.

lib/lndclient/LndClient.ts
erkarl: I'm not sure about this since it's still being used in `Swaps` to derive `makerCltvDelta`. If we decide to do this I'd do it in a separate PR.

erkarl:
@sangaman thanks for the review - feedback applied.

lib/lndclient/LndClient.ts
sangaman: OK that's fine, it can be done separately.

sangaman (approved):

Issue #1091, opened 2019-07-11
xud tries to connect to 0.0.0.0 addresses


kilrau:
### Background

xud tries to connect to 0.0.0.0 addresses 

### Your environment

* version of `xud`: 1.0.0-alpha.11
* which operating system (`uname -a` on *Nix): ubuntu 18.04
* version of `btcd`, `bitcoind`, `ltcd`, `litecoind`, `geth`, `parity`, `raiden`, `lnd` or other chain backend: xud simnet setup

### Steps to reproduce

Run xud simnet, check xud log

### Expected behaviour

xud tries to connect to real IP addresses only

### Actual behaviour

xud tries to connect to 0.0.0.0:8885, which is not a valid address and includes itself:
```
11/07/2019 10:04:54.113 [P2P] debug: Connected pre-handshake to peer 0.0.0.0:8885
11/07/2019 10:04:54.117 [P2P] debug: received disconnecting packet from 0.0.0.0:8885:{"reason":11}
11/07/2019 10:04:54.118 [P2P] info: Socket closed prior to handshake (0.0.0.0:8885)
```
Oldie issue back again, see https://github.com/ExchangeUnion/xud/issues/177

sangaman:
Do you have advertised addresses? If so, we attempt a connection to ourselves for each advertised address once only when starting up. Just want to make sure that's not the case here, did those messages come shortly after a `Verifying reachability of advertised address:` line in the logs?

kilrau:
Correct, that's exactly the case, see below @sangaman . Not a big issue then, but maybe we should just exclude trying to connect to "special addresses" like 127.0.01 and 0.0.0.0.

```
xud_1        | 23/07/2019 16:21:42.422 [P2P] info: Connecting to testnet XU network
xud_1        | 23/07/2019 16:21:42.426 [P2P] info: p2p server listening on 0.0.0.0:8885
xud_1        | 23/07/2019 16:21:42.428 [P2P] debug: Verifying reachability of advertised address: 0.0.0.0:8885
xud_1        | 23/07/2019 16:21:42.434 [HTTP] info: http server listening on 0.0.0.0:8887
xud_1        | 23/07/2019 16:21:42.441 [P2P] debug: Connected pre-handshake to peer 127.0.0.1:44446
[...]

PR #1134, opened 2019-08-08
feat: not connecting to its own addresses

michael1011:
Closes https://github.com/ExchangeUnion/xud/issues/1091

After thinking a little bit about it I am of the opinion that only checking whether the address belongs to itself when trying to connect is sufficient. All other checks could cause weird edge cases and inconsistent behaviors and that single check is enough to fix the issue.

### DIFF
--- a/test/jest/Pool.spec.ts
+++ b/test/jest/Pool.spec.ts
@@ -4,6 +4,7 @@ import Config from '../../lib/Config';
 import DB from '../../lib/db/DB';
 import { XuNetwork } from '../../lib/constants/enums';
 import NodeKey from '../../lib/nodekey/NodeKey';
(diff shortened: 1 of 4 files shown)


test/jest/Pool.spec.ts
```diff
@@ -66,4 +71,20 @@ describe('P2P Pool', () => {
     expect(pool.getTokenIdentifier('DAI')).toEqual(daiTokenAddress);
     expect(pool['nodeState'].raidenAddress).toEqual(raidenAddress);
   });
+
+  test('should reject connecting to its own addresses', async () => {
```
erkarl: :)

erkarl:
Working as advertised...

Issue #1139, opened 2019-08-09
Create a list of most common build failures


erkarl:
Our Travis builds are failing quite often with "false positives" - meaning that if you were to restart the build it would eventually pass. 

The first step to go through old build logs and compile a list of failing log outputs to this issue (and how many times they occur):
https://travis-ci.org/ExchangeUnion/xud/builds

erkarl:
Alternative approach is to create a new test PR and restart it many times.

kilrau:
I believe we can close this. Please re-open if there's anything more you think should be done @erkarl

Issue #955, opened 2019-05-08
simulation tests: cannot start xud network


erkarl:
I'm experiencing a possible race condition error in our simulation tests that causes the build to occasionally fail with the following output:
```
xud_test.go:92: xud process (4-Dave) did not shutdown gracefully. process (13835) killed
harnessTest.go:64: Error outside of test: *errors.errorString cannot start xud network: credentials: failed to append certificates
    /home/travis/build/ExchangeUnion/xud/test/simulation/xud_test.go:251 (0xa2df53)
      TestExchangeUnionDaemon: ht.Fatalf("cannot start xud network: %v", err)
    /home/travis/.gimme/versions/go1.12.linux.amd64/src/testing/testing.go:865 (0x4f96a0)
      tRunner: fn(t)
    /home/travis/.gimme/versions/go1.12.linux.amd64/src/runtime/asm_amd64.s:1337 (0x45d371)
      goexit: BYTE	$0x90	// NOP
```

kilrau:
can you reliably reproduce that? @erkarl

erkarl:
@kilrau no, that's what makes this difficult to debug and occasionally makes the simulation tests fail. I consider this a low priority, but it would be nice to have simulation tests produce consistent results.

kilrau:
> I consider this a low priority, but it would be nice to have simulation tests produce consistent results.

Agree, that's why moved to `post-1.0.0`

kilrau:
Did you see this too? @rsercano

rsercano:
yes, in fact this's the reason why we get automated build fails; #1142 but I really don't know where it comes from. 

IMHO, this's a **file read** issue, if there's already a working test in VM, it opens the certificate file(s), if there's another build in the meantime, it tries to open it too, but since it's already opened by another build, it fails. But that's not a certain proved truth, that's just what I can think of about it.

rsercano:
I implemented a simple logic to try to read cert files of node 3 times before exiting with an error.

kilrau:
I assume this is gone in the recent simulation test fixes, if not please reopen @erkarl

PR #1142, opened 2019-08-09
test(simulation): retry reading tls cert

rsercano:
trials for #1139


rsercano:
At the first glance, there seems to be a default timeout for `mocha` tests which are set to 2000, I've increased it to `10000` and first error seem to be gone. 

But still there's an issue with sim tests, I'm checking it.

rsercano:
Looks like simulation tests are somehow using network to connect to running node instances, but when we create a PR, travis creates two builds instead of one. 

The first one tries to build pushed commit, and the second one tries to build PR which is unnecessary. And during these two builds, somehow network got stuck since travis provides only one VM per project I suppose. 

So I suggest to disable building PRs and it should do the trick I suppose. I don't have required permissions to disable PR build on travis @kilrau

rsercano:
After disabling `build pushed branches` (after suggestion of @michael1011 disabling PR wasn't the case since forks wouldn't be built) there's still one last issue. There're to jobs on travis, one is for `latest stable node` and one is for `latest LTS node` and this exactly creates the same situation. 

I disabled lts for now to see how it goes.

rsercano:
Seems way better !

erkarl:
Thanks for working on this. I'm not sure about single VM/disabling LTS builds issue :thinking:.

Could you please move the mocha timeout increase to a separate PR?

rsercano:
Done @erkarl #1156  :)

rsercano:
As per our discussion, we'll try this by dynamically selecting ports in go sim tests not by removing one of the builds.

rsercano:
I've implemented dynamically picking free ports logic (definitely needs a review) but after a few restart & commits it still fails with `2019/08/20 12:16:30 cannot start xud network: credentials: failed to append certificates` 

I'll investigate it too, but seems like the error rate dramatically reduced after mocha timeout fix.

rsercano:
I realized that there's a function named `NewNetworkHarness` which creates the harness and it returns the above error. But I really need some redirection to fix it in `go` code.

kilrau:
What exactly do you need? @rsercano

test/simulation/lntest/node.go
```diff
@@ -76,6 +82,15 @@ var (
 func generateListeningPorts() (int, int, int) {
 	var p2p, rpc, rest int
 	if numActiveNodes == 0 {
+	    freePorts, err := getFreePorts(3)
+		if err != nil {
+		    fmt.Println(fmt.Errorf("could not find free ports %v", err))
```
moshababo: You should return the error here, and not merely print it (method signature need to be changed to `int, int, int, error` which is a bit weird, but probably fine).

test/simulation/lntest/node.go
```diff
@@ -76,6 +82,15 @@ var (
...
+		} else{
+            defaultNodePort = freePorts[0]
```
moshababo: No need to assign it to the default ports IMO. Just assign it to `p2p, rpc, rest` respectively.

test/simulation/lntest/node.go
```diff
@@ -88,6 +103,25 @@ func generateListeningPorts() (int, int, int) {
 	return p2p, rpc, rest
```
moshababo: Regarding 
```
		p2p = defaultNodePort + (3 * numActiveNodes)
		rpc = defaultClientPort + (3 * numActiveNodes)
		rest = defaultRestPort + (3 * numActiveNodes)
```
`numActiveNodes` is now redundant, since it's only applied when `numActiveNodes` is 0.

test/simulation/lntest/node.go
```diff
@@ -76,6 +82,15 @@ var (
 func generateListeningPorts() (int, int, int) {
 	var p2p, rpc, rest int
 	if numActiveNodes == 0 {
+	    freePorts, err := getFreePorts(3)
```
moshababo: This need to be applied to `xudtest/node.go newNode()` as well.

test/simulation/lntest/node.go
```diff
@@ -88,6 +103,25 @@ func generateListeningPorts() (int, int, int) {
 	return p2p, rpc, rest
 }
 
+// gets unused TCP ports from the system for given count
```
moshababo: a comment in `go` should start with the name being declared ("getFreePorts [...]").

test/simulation/lntest/node.go
rsercano: Then next node port's will get `19555 + (3 * numActiveNodes)` , I assigned it to the default ports just to make next node port finding dynamic too, am I wrong ?

test/simulation/lntest/node.go
rsercano: what do you mean, actually it's only applied when numActiveNodes is different than 0?

test/simulation/lntest/node.go
rsercano: But the next node's port then won't be dynamic ? Am I wrong ?

rsercano: or should I use getFreePorts function whenever generate function is called ?

rsercano:
I actually don't know what I need to change for certificate issue @kilrau

kilrau:
> I actually don't know what I need to change for certificate issue @kilrau

Also not sure what causes this, maybe @sangaman has an idea.

test/simulation/lntest/node.go
kilrau: https://github.com/ExchangeUnion/xud/pull/1182

rsercano: then I assume we can close this pR

rsercano: it's totally okay with me :)

kilrau:
Am I the only one annoyed by the flaky tests?? What's holding this back?

rsercano:
Believe me, it's annoying as hell to everybody, but nobody knows what causes to `failed to append certificate` issue on sim tests

sangaman:
@erkarl Should have a PR soon that should fix the context timeout errors.

rsercano:
I don't think there're timeout errors anymore after #1156

erkarl:
> I don't think there're timeout errors anymore after #1156

There are context deadline exceeded errors/timeouts in our simulation tests. That PR was extending mocha timeouts. I'll have a fix for the context timeouts soon.

I think what's left is to figure out the root cause of `failed to append certificate`. I recommend setting up a virtual machine with low resources to reproduce.

rsercano:
Ah I see, sorry for misunderstanding, my bad.

rsercano:
The root cause of `failed to append certificate` comes from the x509 go package, it throws the mentioned error whenever it tries to read the key file and fails to do so. 

Failing can happen because of various reasons such as `file is already opened`, `there's no read access to the file` etc.

rsercano:
Testing this in detail, seems like issue comes from `grpc go module`, I'm thinking of creating certificate files on the fly instead of using file path.

rsercano:
Related issue #955

kilrau:
I believe it simply takes some time for the certificate to be created in a low-spec environment. Since @rsercano latest commit retries reading the certificate for 3 times, this should help. Any reason why we should **wait** with merging this one? @erkarl @sangaman 

I am very much looking forward to travis improvments.

README.md
```diff
@@ -66,4 +66,6 @@ All files in this repository are licensed under the [GNU Affero General Public L
 
 ## Discussion
 
-Comments, questions, and any development-related discussions are welcome in our [Discord chat](https://discord.gg/YgDhMSn)!
\ No newline at end of file
+Comments, questions, and any development-related discussions are welcome in our [Discord chat](https://discord.gg/YgDhMSn)!
+
+t
```
michael1011: This `t` shouldn't be there, should it?

test/simulation/xudtest/node.go
```diff
@@ -253,6 +253,18 @@ func (hn *HarnessNode) ConnectRPC(useMacs bool) (*grpc.ClientConn, error) {
 	}
 
 	tlsCreds, err := credentials.NewClientTLSFromFile(hn.Cfg.TLSCertPath, "")
+	var tries = 3
+	for tries > 0 {
+		if err != nil {
+			fmt.Printf("failed to create TLS from file with error: %v, remaining tries: %v\n", err, tries-1)
```
michael1011:
```suggestion
			fmt.Printf("Failed to read TLS certificate: %v, remaining tries: %v\n", err, tries-1)
```
This log message is a little unclear

README.md
rsercano: oh god how did it go to there, thanks

test/simulation/xudtest/node.go
rsercano: do you have anything else in your mind ?

michael1011: Just:

```go
fmt.Printf("Failed to read TLS certificate: %v, remaining tries: %v\n", err, tries-1)
```

should be fine imo.

test/simulation/xudtest/node.go
```diff
@@ -253,6 +253,18 @@ func (hn *HarnessNode) ConnectRPC(useMacs bool) (*grpc.ClientConn, error) {
 	}
 
 	tlsCreds, err := credentials.NewClientTLSFromFile(hn.Cfg.TLSCertPath, "")
+	var tries = 3
+	for tries > 0 {
```
sangaman: This is an odd for loop, why not:

```go
for tries := 3; tries > 0; tries-- {
```

test/simulation/xudtest/node.go
```diff
@@ -253,6 +253,18 @@ func (hn *HarnessNode) ConnectRPC(useMacs bool) (*grpc.ClientConn, error) {
...
+		if err != nil {
```
sangaman: Shouldn't this be `if err == nil` ? We only want to retry on errors.

Or we could check and break up front without the need for if/else.

```go
if err == nil {
	break;
}
// retry logic below
```

test/simulation/xudtest/node.go
```diff
@@ -253,6 +253,18 @@ func (hn *HarnessNode) ConnectRPC(useMacs bool) (*grpc.ClientConn, error) {
...
+			fmt.Printf("failed to create TLS from file with error: %v, remaining tries: %v\n", err, tries-1)
+			time.Sleep(3 * time.Second)
```
sangaman: 3 seconds seems like kind of a long time between tries, would it make sense to have more tries with just a 1 second wait between tries?

kilrau:
Please review and let's get this in asap @erkarl @michael1011 @sangaman 

 I believe after last week I don't have to justify why we need a travis fix asap :)

kilrau:
Needs @sangaman

PR #1143, opened 2019-08-10
feat(swapclient): add timeout to initialization

sangaman:
This adds a timeout to the initialization procedure for swap clients. Previously, initialization would wait for as long as it took to receive a response from the client. In cases where raiden or lnd hangs on the response, xud would hang as well.

Now, xud will wait a fixed time limit before proceeding with initialization, leaving the client in a not connected state. The swap client will be marked as connected if and when we get a successful response.

### DIFF
--- a/lib/swaps/SwapClient.ts
+++ b/lib/swaps/SwapClient.ts
@@ -52,6 +52,8 @@ abstract class SwapClient extends EventEmitter {
   protected static readonly RECONNECT_TIMER = 5000;
 
   private updateCapacityTimer?: NodeJS.Timer;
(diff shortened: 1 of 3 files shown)


lib/swaps/SwapClient.ts
```diff
@@ -70,6 +72,21 @@ abstract class SwapClient extends EventEmitter {
   public abstract maximumOutboundCapacity(currency?: string): number;
   protected abstract updateCapacity(): Promise<void>;
 
+  public verifyConnectionWithTimeout = () => {
+    // don't wait longer than the alotted time for the connection to
```
erkarl:
```suggestion
    // don't wait longer than the allotted time for the connection to
```

Issue #1135, opened 2019-08-08
`xud`-level P2P reconnections are attempted from both sides


kilrau:
### Background

`xud`-level P2P reconnections are attempted from both sides of a former p2p connection.

### Your environment

* version of `xud`: latest master
* which operating system (`uname -a` on *Nix): ubuntu 18
* version of `btcd`, `bitcoind`, `ltcd`, `litecoind`, `geth`, `parity`, `raiden`, `lnd` or other chain backend:
* any other relevant environment details:

### Steps to reproduce
Connect two xud instances via manual connect on xud level. Restart both instances with e.g. network off and watch both instances trying to reconnect.

### Expected behaviour

Only the connection initiator tries to reconnect

### Actual behaviour

Both peers try to reconnect

sangaman:
@kilrau Are both nodes being restarted at the same time? Consider the following situation:

1. Node A connects to Node B and advertises a listening connection.
2. Node B is shut down.
3. Node B comes back online but no longer accepts incoming connections.
4A. (Current behavior) Node B reconnects to Node A using the advertised address from step 1.
4B. (Proposed behavior) The nodes will not reconnect.

I'm wondering what is the downside of having potentially two connection attempts in a case like this. The second one should fail gracefully due to the nodes already being connected, if not that seems like the real issue.

See my comments on #1145 for more details about my concerns.

erkarl:
Thank you @sangaman for your input. What you said makes a lot of sense and I think we shouldn't restrict connection attempts to outbound connections only.

kilrau:
Even though originally we did not intend to **re**-connect to advertised URIs we never connected to before, now with TOR this behavior is actually desirable (since probability of advertised URIs being reachable is high) and this improves reconnections as per @sangaman description above. No need to change anything.

PR #1145, opened 2019-08-11
fix(p2p): prevent inbound connection attempting to reconnect

krrprr:
This PR solves the issue described in https://github.com/ExchangeUnion/xud/issues/1135.

### DIFF
--- a/lib/db/models/Node.ts
+++ b/lib/db/models/Node.ts
@@ -34,6 +34,7 @@ export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes) 
       },
     },
     banned: { type: DataTypes.BOOLEAN, allowNull: true },
+    inbound: { type: DataTypes.BOOLEAN, allowNull: true },
   };
 
   const indexes: Sequelize.DefineIndexesOptions[] = [{
--- a/lib/db/types.ts
+++ b/lib/db/types.ts
@@ -84,6 +84,7 @@ export type NodeAttributes = NodeFactory & {
   addressesText: string;
   lastAddressText: string;
   lastAddress: Address;
+  inbound: boolean;
 };
 
 export type NodeInstance = NodeAttributes & Sequelize.Instance<NodeAttributes> & {
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -328,6 +328,9 @@ class Pool extends EventEmitter {
    * Attempt to create an outbound connection to a node using its known listening addresses.
    */
   private tryConnectNode = async (node: NodeConnectionInfo, retryConnecting = false) => {
+    if (node.inbound) {
+      return;
+    }
     if (!await this.tryConnectWithLastAddress(node)) {
       if (!await this.tryConnectWithAdvertisedAddresses(node) && retryConnecting) {
         await this.tryConnectWithLastAddress(node, true);
@@ -521,6 +524,7 @@ class Pool extends EventEmitter {
         addresses,
         nodePubKey: peerPubKey,
         lastAddress: peer.inbound ? undefined : peer.address,
+        inbound: peer.inbound,
       });
     } else {
       // the node is known, update its listening addresses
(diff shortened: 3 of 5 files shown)


proto/xudp2p.proto
```diff
@@ -18,6 +18,7 @@ message Order {
 message Node {
     string node_pub_key = 1;
     repeated Address addresses = 2;
+    bool inbound = 3;
```
erkarl: I think we should be keeping `node.inbound` as internal and not send it within the `NodesPacket`.

lib/p2p/Pool.ts
```diff
@@ -796,6 +801,7 @@ class Pool extends EventEmitter {
         connectedNodesInfo.push({
           nodePubKey: connectedPeer.nodePubKey!,
           addresses: connectedPeer.addresses,
+          inbound: connectedPeer.inbound,
```
erkarl: I believe this is redundant - no need to send a `NodesPacket` containing the `inbound` field.

lib/p2p/Pool.ts
```diff
@@ -875,7 +881,7 @@ class Pool extends EventEmitter {
 
     if (!peer.inbound && peer.nodePubKey && shouldReconnect && (addresses.length || peer.address)) {
```
erkarl: Think we can remove `!peer.inbound` check here since we're doing the same check in `tryConnectNode`.

lib/db/models/Node.ts
```diff
@@ -34,6 +34,7 @@ export default (sequelize: Sequelize.Sequelize, DataTypes: Sequelize.DataTypes)
       },
     },
     banned: { type: DataTypes.BOOLEAN, allowNull: true },
+    inbound: { type: DataTypes.BOOLEAN, allowNull: true },
```
sangaman: I don't think it makes sense to store this in the database. A node isn't permanently inbound nor permanently outbound, at least as I see it. We shouldn't want to permanently make the connection one-way only. What if a peer connects to us, but then we lose the ability to receive incoming connections? We'd still want to be able to make an outgoing connection to that node.

Not to mention, I believe this would be a breaking change that would require a migration script or deleting existing databases, so we should avoid if possible.

lib/db/types.ts
```diff
@@ -84,6 +84,7 @@ export type NodeAttributes = NodeFactory & {
   addressesText: string;
   lastAddressText: string;
   lastAddress: Address;
+  inbound: boolean;
```
sangaman: Likewise we wouldn't need this if we're not storing it in the database.

lib/p2p/Pool.ts
```diff
@@ -328,6 +328,9 @@ class Pool extends EventEmitter {
    * Attempt to create an outbound connection to a node using its known listening addresses.
    */
   private tryConnectNode = async (node: NodeConnectionInfo, retryConnecting = false) => {
+    if (node.inbound) {
```
sangaman: Same here, also I think this would have the undesired affect of making it impossible to create an outbound connection to a given node, even if we try to do it manually.

lib/p2p/Pool.ts
```diff
@@ -521,6 +524,7 @@ class Pool extends EventEmitter {
         addresses,
         nodePubKey: peerPubKey,
         lastAddress: peer.inbound ? undefined : peer.address,
+        inbound: peer.inbound,
```
sangaman: Same here for the database persistence.

lib/db/models/Node.ts
erkarl: Thanks for the review @sangaman - I think what you said makes a lot sense. I'll look into the reconnection/peer discovery logic today to see if we can develop a solution without a breaking change.

erkarl:
Thank you for the PR. After discussing this I feel like the suggested behaviour would be too restrictive. Please see @sangaman's comment in: <https://github.com/ExchangeUnion/xud/issues/1135>

Thanks again for working on this :). I'm looking forward to your future contributions.

Issue #1114, opened 2019-07-22
Add default pairs


kilrau:
I think we should add

Trading pair: `BTC/DAI`
Trading pair: `LTC/DAI`
Trading pair: `WETH/BTC`

to the hardcoded standard pairs with the simnet/testnet & mainnet contract addresses. This is a first step towards https://github.com/ExchangeUnion/xud/issues/1111, which will eliminate the need for restarts entirely when adding pairs after being connected to peers.

PR #1147, opened 2019-08-12
feat(db): add simnet default currencies and pairs

erkarl:
This PR enables `BTC/DAI`, `LTC/DAI` and `WETH/BTC` trading pairs by default when starting xud in simnet mode. I did not add `WETH` and `DAI` for testnet and mainnet since I don't know the contract addresses. Adding them in the future should be trivial.

Closes: https://github.com/ExchangeUnion/xud/issues/1114


erkarl:
This PR should be merged in tandem with: https://github.com/ExchangeUnion/xud-docker/pull/124

lib/p2p/seeds/index.ts
```diff
@@ -1,15 +1,41 @@
...
-  [XuNetwork.MainNet]: mainnet,
-  [XuNetwork.TestNet]: testnet,
-  [XuNetwork.SimNet]: simnet,
+const nodeSeeds: { [network: string]: db.NodeAttributes[] | undefined } = {
+  [XuNetwork.MainNet]: mainnetNodes,
+  [XuNetwork.TestNet]: testnetNodes,
+  [XuNetwork.SimNet]: simnetNodes,
+};
+
+const currencySeeds: { [network: string]: db.CurrencyAttributes[] | undefined } = {
```
michael1011: I am not sure whether `seeds` is the correct terminology for this. I always associate `seed` with a node

erkarl: Interesting - I associate it with initial seed data for the DB. I'm open to changing the naming though if you feel like this would be confusing. Can you provide some alternative names?

michael1011: I was thinking about something like `defaultNodes`, `defaultCurrencies` and `defaultPairs` but I am not opposed to the current naming scheme.

erkarl: Both look good from my perspective. @sangaman? @kilrau?

kilrau: Agree with @michael1011 `defaultNodes`, `defaultCurrencies` and `defaultPairs`

erkarl:
@kilrau @michael1011 feedback applied.

PR #1156, opened 2019-08-16
Mocha timeout extended #1139

rsercano:
Follow up from #1142


erkarl:
Thanks. 1 more nit - can you please change the commit message following this template:
https://github.com/ExchangeUnion/xud/blob/master/.git-commit-template.txt

rsercano:
done @erkarl

sangaman:
How much have we tested this change? I could always go and restart this build in travis several times to see if we get any failures during the mocha tests.

rsercano:
I did approx. 10-15 restarts and a few commits to test this. But as I mentioned in #1142 this doesn't fix the build completely. But let's say without this change we get 60-70% fail, with this change it reduces to 20-30%.

erkarl:
> without this change we get 60-70% fail, with this change it reduces to 20-30%

I'd consider that a quick.

@rsercano for the commit message I'd use `test: extend mocha timeouts`

rsercano:
done @erkarl

Issue #1158, opened 2019-08-16
cltvLimit is set incorrectly


erkarl:
@offerm's debugging result:
> The maker is doing queryroute to get the route cltv (LTC) it adds a buffer 144*4. In out case the result is 1312. Maker convers it to BTC blocks (328) and sends to taker. When maker is making the second leg payment the cltvLimit should be 1312 (LTC blocks) and not 328 (BTC blocks). With 328, LND fails to find a route for the send payment

PR #1160, opened 2019-08-17
test: multi-hop sim test

sangaman:
This adds a simulation test case for multi-hop swaps. Pairs of LTC & BTC channels are established between Alice & Bob, Bob & Carol, and Carol & Dave. Then a swap is performed between Alice & Dave.

This also modifies the channel setup to use a single channel with balance on both sides rather than two unbalanced channels to better resemble how channels are constructed during live usage.

### DIFF
--- a/test/simulation/actions.go
+++ b/test/simulation/actions.go
@@ -400,7 +400,7 @@ func getInfo(ctx context.Context, n *xudtest.HarnessNode) (*xudrpc.GetInfoRespon
 }
 
 func openBtcChannel(ctx context.Context, ln *lntest.NetworkHarness, srcNode, destNode *lntest.HarnessNode) (*lnrpc.ChannelPoint, error) {
-	openChanStream, err := ln.OpenChannel(ctx, srcNode, destNode, 15000000, 0, false)
+	openChanStream, err := ln.OpenChannel(ctx, srcNode, destNode, 15000000, 7500000, false)
 	if err != nil {
 		return nil, err
 	}
@@ -430,7 +430,7 @@ func closeBtcChannel(ctx context.Context, ln *lntest.NetworkHarness, node *lntes
 }
 
 func openLtcChannel(ctx context.Context, ln *lntest.NetworkHarness, srcNode, destNode *lntest.HarnessNode) (*lnrpc.ChannelPoint, error) {
-	openChanStream, err := ln.OpenChannel(ctx, srcNode, destNode, 15000000, 0, false)
+	openChanStream, err := ln.OpenChannel(ctx, srcNode, destNode, 15000000, 7500000, false)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


sangaman:
Added a test case where the maker sells instead of buys LTC, which fails currently given the bug in #1158. I'll fix that bug in a separate PR so we can merge this afterwards.

erkarl:
LGTM once we get production code that makes this test pass :).

test/simulation/actions.go
```diff
@@ -400,7 +400,7 @@ func getInfo(ctx context.Context, n *xudtest.HarnessNode) (*xudrpc.GetInfoRespon
 }
 
 func openBtcChannel(ctx context.Context, ln *lntest.NetworkHarness, srcNode, destNode *lntest.HarnessNode) (*lnrpc.ChannelPoint, error) {
-	openChanStream, err := ln.OpenChannel(ctx, srcNode, destNode, 15000000, 0, false)
+	openChanStream, err := ln.OpenChannel(ctx, srcNode, destNode, 15000000, 7500000, false)
```
erkarl: I think in the future we might want to open channels using xud's `openchannel` API. I'm good with not doing it in this PR. Xud's openchannel doesn't currently support push amount.

sangaman: Yup that would be a good idea indeed.

moshababo:
Looks good, waiting for the bug fix to approve.

> This also modifies the channel setup to use a single channel with balance on both sides rather than two unbalanced channels to better resemble how channels are constructed during live usage.

Wouldn't 2 unbalanced channel would resemble real usage more? where each side opens a channel when needed, without pushing any amount.

sangaman:
> Wouldn't 2 unbalanced channel would resemble real usage more? where each side opens a channel when needed, without pushing any amount.

Hm I guess it depends, currently simnet is opening a single balanced channel between two nodes per currency. Presently LN doesn't have many pairs of channels between nodes AFAIK, although many of them remain unbalanced for a while after opening. It's hard to arrange for a balanced channel off the bat nowadays, but I think longer term it will be easier to create jointly funded channels or to balance channels after opening.

For the sim test perspectives, I think probably either way is fine but this way cuts down a bit on the code and operations needed to perform the tests.

sangaman:
Updated the simulation README to include the new test case.

sangaman:
Rebased onto master to include the fix that should make this test case pass.

kilrau:
> Wouldn't 2 unbalanced channel would resemble real usage more? where each side opens a channel when needed, without pushing any amount.

Actually we have something in ideation phase which would make a single, balanced channel the standard on mainnet too. So let's say the tests are fine like this. @sangaman @moshababo

kilrau:
> Wouldn't 2 unbalanced channel would resemble real usage more? where each side opens a channel when needed, without pushing any amount.

Actually we have something in ideation phase which would make a single, balanced channel the standard on mainnet too. The tests are fine like this. @sangaman @moshababo

Issue #1068, opened 2019-06-27
Enable multiple routing attempts = multi-hop raiden routes


kilrau:
### How it is

Currently raiden changes preimages for payments when one route fails and the next is tried. That's breaking swaps.

To be done:
1. https://github.com/ExchangeUnion/raiden/issues/15
2. enable and test multi-hop raiden trades in xud

kilrau:
> understand why raiden's implementation behaves that way

That's clear by now: because of the "refund payments" raiden uses instead of cancelling HTLCs. For the refund payment the preimage ahs to be revealed and thus can't be reused in the second routing attempt. Waiting for the raiden team to come back to us on that.

PR #1163, opened 2019-08-19
fix(swaps): set cltvLimit correctly

### LINKED ISSUE #1158: cltvLimit is set incorrectly

sangaman:
Fixes #1158.

This fixes a bug whereby the `makerCltvDelta` was erroneously used for the `cltvLimit` for the payment from maker to taker. `makerCltvDelta` is irrelevent to the payment to the taker. Instead, this value is set to the time lock of the route found from maker to taker for lnd, raiden currently has no equivalent to `cltvLimit`.

In tests, payments would still fail occasionally due to no route found. Adding 3 blocks to the `cltvLimit` value consistently resolved these failures - any fewer than 3 and the payments would still fail. More investigation is warranted into why this is necessary, it's possible that there is a bug in the `cltvLimit` implementation within lnd.

This fix makes the new multihop simulation test case in #1160 pass.

### DIFF
--- a/lib/raidenclient/RaidenClient.ts
+++ b/lib/raidenclient/RaidenClient.ts
@@ -165,26 +165,28 @@ class RaidenClient extends SwapClient {
   public sendPayment = async (deal: SwapDeal): Promise<string> => {
     assert(deal.state === SwapState.Active);
     assert(deal.destination);
-    let amount = 0;
-    let tokenAddress;
+    let amount: number;
+    let tokenAddress: string;
+    let lock_timeout: number | undefined;
     if (deal.role === SwapRole.Maker) {
       // we are the maker paying the taker
       amount = deal.takerUnits;
-      tokenAddress = this.tokenAddresses.get(deal.takerCurrency);
+      tokenAddress = this.tokenAddresses.get(deal.takerCurrency)!;
     } else {
       // we are the taker paying the maker
       amount = deal.makerUnits;
-      tokenAddress = this.tokenAddresses.get(deal.makerCurrency);
+      tokenAddress = this.tokenAddresses.get(deal.makerCurrency)!;
+      lock_timeout = deal.makerCltvDelta!;
     }
     if (!tokenAddress) {
       throw(errors.TOKEN_ADDRESS_NOT_FOUND);
     }
     const tokenPaymentResponse = await this.tokenPayment({
       amount,
+      lock_timeout,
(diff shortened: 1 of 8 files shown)


kilrau:
Definitely needs your review @offerm

lib/raidenclient/RaidenClient.ts
```diff
@@ -165,26 +165,28 @@ class RaidenClient extends SwapClient {
...
-    let amount = 0;
-    let tokenAddress;
+    let amount: number;
+    let tokenAddress: string;
+    let lock_timeout: number | undefined;
     if (deal.role === SwapRole.Maker) {
       // we are the maker paying the taker
       amount = deal.takerUnits;
-      tokenAddress = this.tokenAddresses.get(deal.takerCurrency);
+      tokenAddress = this.tokenAddresses.get(deal.takerCurrency)!;
```
erkarl: I think we also need to use `lock_timout=deal.takerMaxTimeLock` here to ensure that the maker's payment duration.

sangaman: I don't think so, because we're using `lock_timeout` as the delay extended to the recipient of the payment, whereas the `takerMaxTimeLock` is an upper bound on the total delay for the route. Also when the maker to taker payment is raiden, we use a hardcoded value for the route total time lock of 101 (which is then set to `takerMaxTimelock`) so I don't think it would be of much use here anyway.

Still, this is a good point to consider going forward, if we're going to be using multihop raiden payments with variable total time locks then we need to know what the maximum time lock will be for properly executing the swap.

erkarl: Ah, right - makes sense. I forgot that we're only supporting direct channels for now.

kilrau: > Still, this is a good point to consider going forward, if we're going to be using multihop raiden payments with variable total time locks then we need to know what the maximum time lock will be for properly executing the swap.

Please open an issue for this with https://github.com/ExchangeUnion/xud/issues/1068 as dependency @sangaman

Issue #1129, opened 2019-08-07
xud crashes when connecting to peer timeouts


erkarl:
```
xud_1     | 07/08/2019 08:30:17.442 [P2P] info: Connecting to known / previously connected peers
xud_1     | 07/08/2019 08:30:32.466 [P2P] error: Peer (127.0.0.1:34058): error: response timeout (SessionInit)
xud_1     | 07/08/2019 08:30:32.469 [P2P] debug: Peer (127.0.0.1:34058): closing socket. reason: ResponseStalling
xud_1     | 07/08/2019 08:30:32.478 [P2P] warn: could not open connection to peer (127.0.0.1:34058): response timeout (SessionInit)
xud_1     | Error: response timeout (95f82ba0-b8ed-11e9-a682-c9e653865293)
xud_1     |     at Peer.emitError (/app/dist/p2p/Peer.js:534:36)
xud_1     |     at Peer.<anonymous> (/app/dist/p2p/Peer.js:382:26)
xud_1     |     at Generator.next (<anonymous>)
xud_1     |     at /app/dist/p2p/Peer.js:7:71
xud_1     |     at new Promise (<anonymous>)
xud_1     |     at __awaiter (/app/dist/p2p/Peer.js:3:12)
xud_1     |     at Timeout.Peer.checkTimeout [as _onTimeout] (/app/dist/p2p/Peer.js:376:35)
xud_1     |     at ontimeout (timers.js:436:11)
xud_1     |     at tryOnTimeout (timers.js:300:5)
xud_1     |     at listOnTimeout (timers.js:263:5)
xud_1     |     at Timer.processTimers (timers.js:223:10)
```

erkarl:
Looks like this can only reproduced when attempting the initial connection to peer (not reconnecting).

erkarl:
I experienced a crash with similar stack trace on latest master `365b52bd3001b1ab229b7bdcb33791c53645ecca`:

```
xud_1     | 19/09/2019 12:22:46.443 [P2P] warn: could not open connection to peer (127.0.0.1:50958): response timeout (SessionInit)
xud_1     | Error: response timeout (SessionAck)
xud_1     |     at Peer.emitError (/app/dist/p2p/Peer.js:534:36)
xud_1     |     at Peer.<anonymous> (/app/dist/p2p/Peer.js:382:26)
xud_1     |     at Generator.next (<anonymous>)
xud_1     |     at /app/dist/p2p/Peer.js:7:71
[...]

PR #1166, opened 2019-08-19
fix(p2p): prevent xud from crashing when connecting to peer times out

krrprr:
Fix for https://github.com/ExchangeUnion/xud/issues/1129

xud was crashing because GrpcService::getGrpcError requires an error object that should have code and message properties, but got a string instead.


krrprr:
> I think the only other thing left to do is to add the proper handling in `getGrpcError` for this `RESPONSE_TIMEOUT` error.

Thanks for the feedback! I was actually wondering if there should be a code for this case as well. I'll add the suggested one.

PR #1167, opened 2019-08-20
feat(swaps): improved remote error handling

sangaman:
This improves the way swap errors are communicated between participants of a swap. Previously, any error during an attempt to send payment was communicated to the peer via a `SwapFailedPacket`. However, in cases where the maker to taker (2nd leg) payment fails, the maker will reject the 1st leg payment because they were unable to acquire the preimage. This results in two near-simultaneous payment errors on both peers, who send `SwapFailedPacket`s to each other for already failed deals which results in an undesirable `received swap failed packet for unknown deal` log messages.

This change detects an lnd payment that failed due to being rejected (`UnknownPaymentHash` error message) and skips sending the `SwapFailedPacket`. Instead it fails the swap due to a new `RemoteError` swap failure reason, and updates the error details of the failed swap if it receives a `SwapFailedPacket` for swaps with this failure reason.

### DIFF
--- a/lib/constants/enums.ts
+++ b/lib/constants/enums.ts
@@ -125,6 +125,8 @@ export enum SwapFailureReason {
   DealTimedOut = 11,
   /** The swap failed due to an unrecognized error. */
   UnknownError = 12,
+  /** The swap failed due to an error or unexpected behavior on behalf of the remote peer. */
+  RemoteError = 12,
 }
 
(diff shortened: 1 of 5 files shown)


sangaman:
Added a check for the `IncorrectOrUnknownPaymentDetails` payment error message per the latest changes to lnd.

PR #1170, opened 2019-08-22
feat(cli): getinfo enhancements

### LINKED ISSUE #1059: `getinfo` fixes
- [ ] Rename "error" -> "status"
- [ ] New status "ready" when L2 client is up and ready
- [ ] Show Raiden version

```
      "BTC",
      {
        "status": "ready",
        "channels": {
          "active": 1,
          "inactive": 0,
          "pending": 0
        },
        "chainsList": [
          {
            "chain": "bitcoin",
            "network": "simnet"
          }
        ],
        "blockheight": 30750,
        "urisList": [],
        "version": "0.6.1-beta commit=v0.6.1-beta-dirty",
        "alias": "lndbtc"
```
```
  "raiden": {
    "status": "ready",
    "channels": 0,
    "address": "0xa65dda2A2889263F6E254d1E91E1F7c44211CFD2",
    "version": "v0.100.3 - Rosemary"
    "alias": "raiden"
  }
}
```

rsercano:
resolves #1059


rsercano:
`actions.go` line 36 there're chain & network information checks in these lines, but I didn't change anything related with chain as you can see in the changed files. 

Still, somehow it doesn't receive chain & network information.. Could you please help me with that @moshababo

rsercano:
This PR also includes changes from #1174. I have a few quick questions;

- Where should I get alias of xud, is there an algorithm to generate it?
- Somehow address of the local xud is empty in pool class, is there something wrong with my setup or with these changes ?
- Raiden doesn't provide chains as LND does, so I set `XUD <network>` hardcoded, is that okay ?

kilrau:
>     * Where should I get alias of xud, is there an algorithm to generate it?

There will be soon: https://github.com/ExchangeUnion/xud/issues/486#issuecomment-528392857

> Somehow address of the local xud is empty in pool class, is there something wrong with my setup or with these changes ?

Check `xud.conf`, if nothing configured there and you are running a lonely `xud`, it's empty I believe. The SimNet setup has to give you a TOR address.

> Raiden doesn't provide chains as LND does, so I set XUD <network> hardcoded, is that okay ?

No... Worst case you can judge by the registered `DAI` and `WETH` addresses which network it is (see ` lib/db/seeds`).

> Here's how it seems with these changes; @kilrau

Address should have a line break before the @ to make it fit in a regular CLI terminal size. Rest looks good.

rsercano:
Added line break just before the `@` as you said Kilian but I can't see how I can find out raiden chain ? The address printed by get info is different from registered addresses in seed folder ?

rsercano:
Nvm, I suppose I found it, needs a review tho.

lib/cli/commands/getinfo.ts
```diff
@@ -3,109 +3,87 @@ import { Arguments } from 'yargs';
...
-    displayChannels(info.channels, asset);
-  }
+  const address = info.urisList[0] ? `${info.urisList[0].substring(0, info.urisList[0].indexOf('@'))}
+${info.urisList[0].substring(info.urisList[0].indexOf('@'))}` : '';
 
-  if (!info.error) {
-    displayChainsList(info.chainsList, asset);
-  }
+  basicInfotable.push(
+    { [colors.blue('Version')]: info.version ? info.version : ''   },
```
michael1011:
```suggestion
    { [colors.blue('Version')]: info.version || ''   },
```
This can be simplified a little

lib/cli/commands/getinfo.ts
```diff
@@ -3,109 +3,87 @@ import { Arguments } from 'yargs';
...
+    { [colors.blue('Address')]: address },
+    { [colors.blue('Alias')] : info.alias ? info.alias : '' },
```
michael1011:
```suggestion
    { [colors.blue('Alias')] : info.alias || '' },
```
Same here

lib/cli/commands/getinfo.ts
```diff
@@ -3,109 +3,87 @@ import { Arguments } from 'yargs';
...
+    { [colors.blue('Channels')] :  `Active: ${info.channels ? info.channels['active'] : 0} | Pending: ${info.channels ? info.channels['pending'] : 0}\
+ | Closed: ${info.channels ? info.channels['inactive'] : 0}` },
```
michael1011: Inactive channels are not `closed`. That just means that the peer is not online right now

lib/cli/commands/getinfo.ts
```diff
@@ -3,109 +3,87 @@ import { Arguments } from 'yargs';
...
+  table.push({ [colors.blue('Status')]: SwapClientStatus[info.status] });
+  if (info.error) {
+    table.push({ [colors.blue('Error')]: info.error });
+  }
   table.push(
     { [colors.blue('Version')]: info.version },
     { [colors.blue('Address')]: info.address },
-    { [colors.blue('Channels')]: info.channels },
-    { [colors.blue('Error')]: info.error },
+    { [colors.blue('Channels')]:  `Active: ${info.channels ? info.channels : 0} | Pending: 0 | Closed: 0` },
```
michael1011: Raiden has a `closed` channel state. This could be shown here.

lib/lndclient/LndClient.ts
```diff
@@ -243,17 +244,23 @@ class LndClient extends SwapClient {
...
-        blockheight = lnd.getBlockHeight(),
-        uris = lnd.getUrisList(),
+        blockheight = lnd.getBlockHeight();
+        uris = lnd.getUrisList();
         version = lnd.getVersion();
         alias = lnd.getAlias();
+        if (channels.active > 0) {
+          status = SwapClientStatus.Ready;
+        } else {
+          error = errors.LND_IS_UNAVAILABLE(ClientStatus.NotInitialized).message;
```
michael1011: I am not sure whether we should call a client with no clients `NotInitialized`. @kilrau what do you think?

lib/raidenclient/RaidenClient.ts
```diff
@@ -251,20 +251,29 @@ class RaidenClient extends SwapClient {
     let channels: number | undefined;
     let address: string | undefined;
     let error: string | undefined;
-    const version = ''; // Intentionally left blank until Raiden API exposes it
+    let status: SwapClientStatus = SwapClientStatus.Error;
+    const version = 'v0.100.3 - Rosemary'; // Intentionally left hardcoded until Raiden API exposes it
```
michael1011: We have to remember to update this version regularly if we want to hardcode it.

rsercano:
Thanks @michael1011 and @kilrau I'll check the rest on Monday

lib/raidenclient/RaidenClient.ts
kilrau: Let's never hardcode anything unless there is no other way. I checked the Raiden Client endpoints and indeed I couldn't find one returning the version. BUT each raiden log starts by printing the version in line 0: `Welcome to Raiden, version 0.100.5a1.dev157+geb2af878d!`

lib/lndclient/LndClient.ts
kilrau: Agree, this is not the correct error message. How about  `ClientStatus.NoChannels` ?

lib/raidenclient/RaidenClient.ts
```diff
@@ -251,20 +251,29 @@ class RaidenClient extends SwapClient {
...
+    const chain = Object.keys(this.tokenAddresses).find(key => this.tokenAddresses.get(key) === this.address);
 
     if (this.isDisabled()) {
       error = errors.RAIDEN_IS_DISABLED.message;
     } else {
       try {
         channels = (await this.getChannels()).length;
         address = this.address;
+        if (channels > 0) {
```
kilrau: Same here.

lib/cli/commands/getinfo.ts
rsercano: Then how can I receive the closed ones ?

lib/raidenclient/RaidenClient.ts
rsercano: I couldn't find the mentioned version description in any response of raiden API, did you get it from somewhere else @kilrau ?

rsercano:
Added requested changes @kilrau @michael1011. Also I've realized that I could get closed channels from LND api via `ClosedChannels` call, so I've implemented it too.

image

lib/raidenclient/RaidenClient.ts
michael1011: I think @kilrau means that the first row of the **log** of Raiden shows the version number. But reading that isn't a viable long term solution. I would just leave the Raiden version blank or set it to `v1` (because we are using the version 1 of their REST API)

kilrau: Agree, that it is not a long term solution but would want to use it, as it's the only way we have right now. I'll open an issue on raiden side to have something similar to `getinfo` with version and so on. Once that's there, we can switch. Deal?

michael1011: @rsercano we just agreed on the call to implement a Raiden endpoint that shows the version. Because the swaps will only be working on our Raiden fork we can easily and quickly merge that new endpoint in there. I will implement it and ping you again once it is ready.

rsercano: Great, thanks @michael1011 :)

michael1011: @rsercano it was easier than expected: https://github.com/ExchangeUnion/raiden/commit/2ac24318407084866bd81c0f212ae63a2cbfab33
image

rsercano:
Added fetching version function, it should work as soon as https://github.com/ExchangeUnion/raiden/pull/10 is merged & deployed.

lib/swaps/SwapClient.ts
```diff
@@ -10,6 +10,7 @@ enum ClientStatus {
   ConnectionVerified,
   OutOfSync,
   WaitingUnlock,
+  NoChannels,
```
sangaman: I don't think this is the place for this `NoChannels` status, we're never actually setting the status of a client to `NoChannels` and if we did we'd have to alter logic in other places that depends on `ConnectionVerified` to ensure we can make calls to lnd (including calls to open channels).

proto/xudrpc.proto
```diff
@@ -380,20 +385,24 @@ message SwapFailure {
 
 message GetInfoRequest {}
 message GetInfoResponse {
-  // The version of this instance of xud.
-  string version = 1 [json_name = "version"];
```
sangaman: Not a huge deal since we are still in pre-release stages, but generally speaking going forward we should be careful not to change numbers of existing fields because it would break client code that's using the old proto definition. In this case we already have `version`. It's also just reducing unnecessary code changes. Same for some of the other fields (`num_peers`, `num_pairs`, etc).

proto/xudrpc.proto
```diff
@@ -455,16 +464,19 @@ message LndChannels {
...
 
 message LndInfo {
-  string error = 1 [json_name = "error"];
-  LndChannels channels = 2 [json_name = "channels"];
-  repeated Chain chains = 3 [json_name = "chains"];
-  uint32 blockheight = 4 [json_name = "blockheight"];
-  repeated string uris = 5 [json_name = "uris"];
-  string version = 6 [json_name = "version"];
-  string alias = 7 [json_name = "alias"];
+  SwapClientStatus status = 1 [json_name = "status"];
```
sangaman: Same here for renumbering all these fields.

proto/xudrpc.proto
```diff
@@ -455,16 +464,19 @@ message LndChannels {
   uint32 inactive = 2 [json_name = "inactive"];
   // The number of channels that are pending on-chain confirmation before they can be used.
   uint32 pending = 3 [json_name = "pending"];
+  // The number of channels that are pending on-chain confirmation before they can be used.
```
sangaman: This comment should say "The number of channels that have been closed" or something like that.

proto/xudrpc.proto
```diff
@@ -596,11 +608,19 @@ message PlaceOrderEvent {
   }
 }
 
+message RaidenChannels {
```
sangaman: Can we rename `LndChannels` to `Channels` and just use that single message?

proto/xudrpc.proto
```diff
@@ -596,11 +608,19 @@ message PlaceOrderEvent {
...
+  uint32 active = 1 [json_name = "active"];
+  uint32 settled = 2 [json_name = "settled"];
+  uint32 closed = 3 [json_name = "closed"];
+}
+
 message RaidenInfo {
-  string error = 1 [json_name = "error"];
```
sangaman: Same here for renumbering fields.

proto/xudrpc.proto
```diff
@@ -596,11 +608,19 @@ message PlaceOrderEvent {
...
-  string address = 2 [json_name = "address"];
-  uint32 channels = 3 [json_name = "channels"];
-  string version = 4 [json_name = "version"];
+  SwapClientStatus status = 1 [json_name = "status"];
+  string error = 2 [json_name = "error"];
```
sangaman: I thought we wanted to remove `error` and replace with `status`?

rsercano:
Thanks for your reviews @sangaman, in fact I'm not available before Monday, due to a travel abroad, sorry about that.

lib/swaps/SwapClient.ts
rsercano: Moved it to errors which makes more sense, thanks Daniel

proto/xudrpc.proto
rsercano: Good point

proto/xudrpc.proto
rsercano: At first I thought we want the requested order, but you're right.

proto/xudrpc.proto
rsercano: I'm commenting out setting `settled` for raiden then ?

rsercano:
Changes have been implemented @sangaman, this's the current view;
image
image

kilrau:
Looks great! Questions:
* Why is `xud`'s address field empty?
* Can you run our current raiden version,which has a version endpoint? Would also like to see the fields in there.
* I suggest renaming `Chain` -> `Network` for lndbtc, lndltc, raiden too. To be consistent with xud.

rsercano:
> Why is xud's address field empty?

My configuration was wrong, seems okay now.
image

> Can you run our current raiden version,which has a version endpoint? Would also like to see the fields in there.

image

> I suggest renaming Chain -> Network for lndbtc, lndltc, raiden too. To be consistent with xud.

Consider it done.

kilrau:
> > Why is xud's address field empty?
> 
> My configuration was wrong, seems okay now.
> image

Can we do the line break before the `@` here too (just ? If it's a long .onion address it'll break things again.

> > Can you run our current raiden version,which has a version endpoint? Would also like to see the fields in there.
> 
> image

Don't really see it here, but I assume `raiden.version` calls the new version endpoint. Also, can we call it `raiden simnet` to be in line with `bitcoin simnet` and `litecoin simnet` below? 

> > I suggest renaming Chain -> Network for lndbtc, lndltc, raiden too. To be consistent with xud.
> 
> Consider it done.

:+1:

rsercano:
image

> Don't really see it here, but I assume raiden.version calls the new version endpoint. Also, can we call it raiden simnet to be in line with bitcoin simnet and litecoin simnet below?

Yes it does. Added `raiden` hardcoded just before the network if the `tokenAddress` doesn't have a name.
image

kilrau:
Damn, I just found one more thing: In https://github.com/ExchangeUnion/xud/pull/1203 we finally settled on "Node Key" to describe `xud`'s public key. Can we "Public Key" -> "Node Key"? Thanks.

rsercano:
done @kilrau

kilrau:
Final ACK & merge by @sangaman

PR #1172, opened 2019-08-22
fix(swaps): validate Raiden's resolve request

### LINKED ISSUE #1149: Raiden lock time follow-ups
This tracks open issues persisting after #1082 for us to properly have swaps with raiden as the first leg.

The main issues for now are:

- `expiration` is currently an absolute block number, which can't be used by xud because we don't currently have a way to get to the ethereum block height. Passing in the height on the resolve request would allow us to calculate a delta.

- Ensure that raiden is properly setting the lock time. In testing with @erkarl, the values we got for the settle and reveal timeouts, as well as expiration, were lower than expected and lower than the `lock_timeout` we set on the raiden token payment call.

ghost:
In this commit we validate Raiden's resolve request to make sure the
payment we're receiving has the previously agreed upon `deal.makerCltvDelta` as the
minimum duration of the lock.

In addition we change the error handling to return a status code 400
(Bad Request) to prevent Raiden from crashing. Previously we returned a
status code of 200 with the response of `{ secret: 'error message' }`.

Closes: https://github.com/ExchangeUnion/xud/issues/1149

### DIFF
--- a/lib/http/HttpServer.ts
+++ b/lib/http/HttpServer.ts
@@ -63,19 +63,25 @@ class HttpServer {
         default:
           res.writeHead(404);
           res.end();
+          break;
       }
       res.writeHead(200, { 'Content-Type': 'application/json' });
       res.end(JSON.stringify(resJson));
     } catch (err) {
-      if (err.code === errorCodes.INVALID_ARGUMENT) {
-        res.writeHead(400);
-        res.end(err.message);
-      } else if (err.code === swapErrorCodes.PAYMENT_HASH_NOT_FOUND) {
-        res.writeHead(404);
-        res.end(err.message);
-      } else {
-        res.writeHead(500);
-        res.end(JSON.stringify(err));
+      switch (err.code) {
+        case swapErrorCodes.INVALID_RESOLVE_REQUEST:
+        case errorCodes.INVALID_ARGUMENT:
+          res.writeHead(400);
+          res.end(err.message);
+          break;
+        case swapErrorCodes.PAYMENT_HASH_NOT_FOUND:
+          res.writeHead(404);
+          res.end(err.message);
+          break;
+        default:
+          res.writeHead(500);
+          res.end(JSON.stringify(err));
+          break;
       }
     }
   }
--- a/lib/raidenclient/types.ts
+++ b/lib/raidenclient/types.ts
@@ -64,15 +64,10 @@ export type RaidenResolveRequest = {
   secrethash: string;
   /** The amount of the incoming payment pending resolution, in the smallest units supported by the token. */
   amount: number;
-  /** The maximum number of blocks allowed between the setting of a hashlock and the revealing of the related secret. */
-  reveal_timeout: number;
-  /** The lock expiration for the incoming payment. */
+  /** The lock expiration for the incoming payment (absolute block number). */
   expiration: number;
-  // 'settle_timeout': raiden.config['settle_timeout'],
-  // unused fields on the raiden request listed below, taken from raiden codebase
-  // 'payment_identifier': secret_request_event.payment_identifier,
-  // 'payment_sender': to_hex(secret_request_event.recipient),
-  // 'payment_recipient': to_hex(raiden.address),
+  /** The current height of the chain */
+  chain_height: number;
 };
 
 export type RaidenResolveResponse = {
--- a/tslint.json
+++ b/tslint.json
@@ -67,6 +67,7 @@
       }
     ],
     "one-line": [true, "check-catch", "check-finally", "check-open-brace", "check-whitespace"],
-    "no-switch-case-fall-through": true
+    "no-switch-case-fall-through": true,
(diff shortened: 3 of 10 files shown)


kilrau:
Please open an issue on raiden side for the crash @erkarl

erkarl:
> Please open an issue on raiden side for the crash @erkarl

There already is: https://github.com/ExchangeUnion/raiden/issues/1

kilrau:
I successfully tested this branch on native simnet, which since ExchangeUnion/xud#1080 changed the behaviour of the XUD resolver when it gets an unfamiliar rHash and caused raiden to crash. With this branch, this is handled gracefully and raiden doesn't crash.

lib/raidenclient/types.ts
```diff
@@ -68,6 +68,7 @@ export type RaidenResolveRequest = {
   reveal_timeout: number;
```
sangaman: Should we be removing reveal and settle timeout as part of this PR? Or are you planning to do this separately? I'm thinking if it's done together this can be marked as closing #1149.

If so, you can also remove the commented out "unused" fields below.

lib/http/HttpServer.ts
```diff
@@ -67,7 +67,10 @@ class HttpServer {
       res.writeHead(200, { 'Content-Type': 'application/json' });
       res.end(JSON.stringify(resJson));
     } catch (err) {
-      if (err.code === errorCodes.INVALID_ARGUMENT) {
+      if (
+        err.code === errorCodes.INVALID_ARGUMENT ||
+        err.code === swapErrorCodes.INVALID_RESOLVE_REQUEST
```
sangaman: I'm thinking I should've made this a `switch` statement when I wrote this originally, but now I think it's even more clearly the nicer choice. Would you mind rewriting these if/elses as a switch?

sangaman (commented):
Looks nice, the 400/404/500 HTTP error codes here are much more sensible than returning plain error message strings in place of the preimage. Just a few questions/suggestions below.

lib/raidenclient/types.ts
erkarl: Yes, absolutely - good idea.

lib/http/HttpServer.ts
erkarl: Rewrote this as switch - looks much better now, thanks.

lib/http/HttpServer.ts
```diff
@@ -67,15 +67,22 @@ class HttpServer {
...
-        res.end(err.message);
-      } else {
-        res.writeHead(500);
-        res.end(JSON.stringify(err));
+      switch (err.code) {
+        case errorCodes.INVALID_ARGUMENT:
+          res.writeHead(400);
+          res.end(err.message);
+          break;
+        case swapErrorCodes.INVALID_RESOLVE_REQUEST:
```
sangaman: This can just flow with the previous case like so we don't need to write the code twice:

```javascript
        case errorCodes.INVALID_ARGUMENT:
        case swapErrorCodes.INVALID_RESOLVE_REQUEST:
          res.writeHead(400);
          res.end(err.message);
          break;
```

lib/http/HttpServer.ts
```diff
@@ -67,15 +67,22 @@ class HttpServer {
...
+          res.writeHead(400);
+          res.end(err.message);
+          break;
+        case swapErrorCodes.PAYMENT_HASH_NOT_FOUND:
+          res.writeHead(404);
+          res.end(err.message);
+          break;
+        default:
+          res.writeHead(500);
+          res.end(JSON.stringify(err));
```
sangaman: So a `break` here for the last case is not technically necessary, but I kind of like it anyway in case the cases ever get reordered or if we add another case at the end and forget to add the `break`.

There's actually a [tslint rule](https://palantir.github.io/tslint/rules/switch-final-break/) for this I just found that I think might be worth adding, I believe we have the final `break` in use elsewhere in the code so it would only enforce this rule going forward.

sangaman (changes requested):
Thanks, one last (or maybe two) change re: the new switch statement then this should be good to merge.

tslint.json
```diff
@@ -67,6 +67,8 @@
       }
     ],
     "one-line": [true, "check-catch", "check-finally", "check-open-brace", "check-whitespace"],
-    "no-switch-case-fall-through": true
+    "no-switch-case-fall-through": true,
+    "switch-final-break": [true, "always"]
+
```
sangaman: Extra blank line here.

sangaman (approved):
Looks good! Do you think you could put the tslint rule in a separate commit tho?

moshababo (approved):

michael1011:
@erkarl @sangaman could you please revisit this PR and merge it asap? The raiden in my local simnet instance is crashing a lot which causes some issues with the balancing of the initial deposit.

michael1011 (approved):

sangaman (approved):

PR #1177, opened 2019-08-23
refactor(swaps): change getRoutes to getRoute

sangaman:
This refactors the `getRoutes` call to `getRoute` to match the behavior of lnd which only returns a single route at most for the `QueryRoutes` call.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -542,7 +542,7 @@ class LndClient extends SwapClient {
     return this.unaryCall<lndrpc.ListChannelsRequest, lndrpc.ListChannelsResponse>('listChannels', new lndrpc.ListChannelsRequest());
   }
 
-  public getRoutes = async (units: number, destination: string, _currency: string, finalCltvDelta = this.lockBuffer) => {
+  public getRoute = async (units: number, destination: string, _currency: string, finalCltvDelta = this.lockBuffer) => {
     const request = new lndrpc.QueryRoutesRequest();
     request.setAmt(units);
     request.setFinalCltvDelta(finalCltvDelta);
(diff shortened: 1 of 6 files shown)


lib/lndclient/LndClient.ts
```diff
@@ -551,21 +551,27 @@ class LndClient extends SwapClient {
...
     request.setFeeLimit(fee);
 
+    let route: lndrpc.Route | undefined;
+
     try {
-      const routes = (await this.queryRoutes(request)).getRoutesList();
-      this.logger.debug(`got ${routes.length} route(s) to destination ${destination}: ${routes}, finalCltvDelta: ${finalCltvDelta}`);
-      return routes;
+      // QueryRoutes no longer returns more than one route
+      route = (await this.queryRoutes(request)).getRoutesList()[0];
```
kilrau: Why continue using a list if it has only one entry?

erkarl: This is because lnd still returns us a list: https://api.lightning.community/#grpc-response-queryroutesresponse

lib/lndclient/LndClient.ts
```diff
@@ -551,21 +551,27 @@ class LndClient extends SwapClient {
...
       )) {
-        return [];
-      } else {
         this.logger.error(`error calling queryRoutes to ${destination}, amount ${units}, finalCltvDelta ${finalCltvDelta}`, err);
         throw err;
       }
     }
+
+    if (route) {
+      this.logger.debug(`found a route to ${destination} for ${units} with finalCltvDelta ${finalCltvDelta}: ${route}`);
```
erkarl:
```suggestion
      this.logger.debug(`found a route to ${destination} for ${units} units with finalCltvDelta ${finalCltvDelta}: ${route}`);
```

lib/lndclient/LndClient.ts
```diff
@@ -551,21 +551,27 @@ class LndClient extends SwapClient {
...
+    } else {
+      this.logger.debug(`could not find a route to ${destination} for ${units} with finalCltvDelta ${finalCltvDelta}: ${route}`);
```
erkarl:
```suggestion
      this.logger.debug(`could not find a route to ${destination} for ${units} units with finalCltvDelta ${finalCltvDelta}: ${route}`);
```

PR #1179, opened 2019-08-24
feat(swaps): dynamic lock buffer

### LINKED ISSUE #1164: Make lockBuffer dynamic + additional lockDelta for taker+maker
Off-shoot of https://github.com/ExchangeUnion/xud/issues/1157

> IMHO, The lockBuffer should be dynamically computed based on the lockDuration of the second leg. Assuming that the average block time on the second leg increases 100% while at the same time the first leg decrease by 50% is very drastic. I agree that there should be a lockBufferFactor (>1) that we should use.

TBD:
* what are reasonable values for this dynamic `lockBuffer`? Lower bound, %. Upper bound needed?
* do we still need an additional, configurable `cltvDelta` for taker and maker?

sangaman:
Closes #1164.

This modifies the logic around calculating the `makerCltvDelta` value for swaps which specifies the minimum expected lock duration for the final hop of the first leg. We use the poisson quantile function to determine a very high probability that the first leg final lock won't expire before the second leg payment.

This also removes the recently added `lockBuffer` config option in favor of a `cltvDelta` for lnd that specifies the lock delta that is used for the final hop of the second leg and gets added to the lock buffer to determine the final lock delta for the first leg.

~~Note that this currently adds type definitions for the poisson quantile module to xud. I'm seeing if I can get these merged directly into that package (https://github.com/distributions-io/poisson-quantile/pull/1), or if not that then in the [DefinitelyTyped repo](https://github.com/DefinitelyTyped/DefinitelyTyped), so that it need not be part of the xud codebase.~~

I wrote types for the `distributions-poisson-quantile` package and got it [merged into DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37955), sp I've updated the PR to use the newly published types.


kilrau:
@sangaman  on :fire: :tada:

kilrau:
If we want this for `1.0.0-bounty` (which would be nice), this needs to be reviewed and tested soon @erkarl . But we have an indirect lockduration buffer already, so no must. I'll leave it in `1.0.0-beta` for now.

PR #1182, opened 2019-08-26
test: use dynamic ports for simulation tests

sangaman:
This dynamically selects unused ports for the lnd and xud nodes used during the simulation tests. This should reduce the likelihood of tests failing due to other running processes on the machine.


rsercano:
Although this doesn't fix the automated builds entirely, probably it'll reduce the fail rate. But the main reason now for failing tests are the "certificate thing"

kilrau:
Please test and merge if good @rsercano

sangaman:
I don't think we need any testing per se - the tests pass. I was just leaving this open for a few days in case anyone else wanted to comment on my approach, but we can go ahead and merge it now.

rsercano:
@kilrau the only difference that Daniel has implemented from my approach is using an external go module, so since tests are passing as Daniel said, that can be merged to me

PR #1185, opened 2019-08-27
test: install geth in simtest pipeline

ghost:
In this PR we're adding geth to our simulation tests' install script pipeline. This is part 1 of a series of PRs to enable simulation tests for raiden.

The geth version we're installing is based on `1.9.2` and has support for fast mining.

### DIFF
--- a/test/simulation/install-geth.sh
+++ b/test/simulation/install-geth.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+set -e
+GETH_SOURCE=$1
+GETH_PATH=$2
+EXPECTED_HASH=$3
+install_geth () {
+  echo "installing geth"
+  git clone --depth 1 "$GETH_SOURCE" "$GETH_PATH"
+  cd "$GETH_PATH"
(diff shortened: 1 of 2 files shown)


test/simulation/install-geth.sh
```diff
@@ -0,0 +1,26 @@
...
+  echo "installing geth"
+  git clone --depth 1 "$GETH_SOURCE" "$GETH_PATH"
+  cd "$GETH_PATH" || exit 1
+  GOPATH=$GO_PATH CI=false GO111MODULE=off go run build/ci.go install ./cmd/geth
+}
+if [ ! -d "$GETH_PATH" ]; then
+  install_geth
+else
+  ROOT=$PWD
+  cd "$GETH_PATH" || exit 1
```
michael1011: What does the `|| exit 1` do? Exit the script with the code 1 if `cd "$GETH_PATH"` fails?

erkarl: Exactly. https://github.com/koalaman/shellcheck/wiki/SC2164

Since we're using `set -e` at the beginning of the script and will immediately exit the script with a non-zero status upon command failure it doesn't make sense to have these. I removed the redundant code.

PR #1192, opened 2019-08-29
fix(lnd): wait for macaroon before wallet created

sangaman:
This fixes the logic around initializing `LndClient` for a new lnd instance (before wallet creation) and new xud instance (before node key creation). The lnd macaroon file is not created until after its wallet is created, so we relax the check that disables the client if a macaroon is not found during initialization. Instead, we mark the client as `WaitingUnlock` and wait for a successful xud `CreateNode` call to initialize the lnd wallet. After we've successfully loaded the admin macaroon, we then continue with attempting to verify the connection.

I've tested this manually and the create & unlock calls are now interacting with lnd as expected.

### DIFF
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -72,15 +72,22 @@ class Xud extends EventEmitter {
       this.unitConverter = new UnitConverter();
       this.unitConverter.init();
 
-      this.swapClientManager = new SwapClientManager(this.config, loggers, this.unitConverter);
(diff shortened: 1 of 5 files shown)


sangaman:
> Out of scope of this PR, but I'm wondering if we should also be handling the case of the swap client coming online after xud has already been unlocked? For example the swap client could crash and restart itself (our docker setup does this).
> 
> The current solution is to restart xud and unlock again.

I'll respond in #1196.

PR #1200, opened 2019-09-04
feat: show git commit hash in version

### LINKED ISSUE #1171: Show git commmit as version number (when not running a release)
Status Quo: At the moment running latest master shows the last release as `version`, which is confusing at best.

What we want: set the  "version" field in (e.g. `xucli getinfo`) to the latest git commit hash and only to a release name if it's indeed the release

How: some kind of script that fetches the hash of the latest git commit and writes it to a TypeScript file. Then execute this script before `npm run compile`. Like this, the commit hash is set at compile time.

michael1011:
Closes #1171 

I created a new script (`parseGitCommit.js`) that will be executed before the `compile` NPM script. It parses the latest git commit hash (if that commit doesn't also tag a release) and whether there are uncommitted changes in the working directory and writes a corresponding string into `lib/Version.ts`.

For example the `version` in `xucli getinfo -j` could look like: `1.0.0-testnet.1-9a03442c-dirty`

### DIFF
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -12,6 +12,7 @@ import SwapClientManager from '../swaps/SwapClientManager';
 import { OrderSidesArrays } from '../orderbook/TradingPair';
 import { SwapSuccess, SwapFailure, ResolveRequest } from '../swaps/types';
 import { errors as swapsErrors } from '../swaps/errors';
+import commitHash from '../Version';
 
 /**
  * The components required by the API service layer.
@@ -269,7 +270,7 @@ class Service {
       nodePubKey,
       uris,
       numPairs,
-      version: this.version,
+      version: `${this.version}${commitHash}`,
       numPeers: this.pool.peerCount,
       orders: {
         peer: peerOrdersCount,
--- a/parseGitCommit.js
+++ b/parseGitCommit.js
@@ -1,0 +1,33 @@
+const fs = require('fs');
+const childProcess = require('child_process');
+
+const executeCommand = (command) => {
+  return childProcess.execSync(command, { encoding: 'utf8' });
+};
+
+const getCommitHash = () => {
+  const result = executeCommand('git reflog --decorate -1');
+
+  // Do not show the commit hash if that commit is also a tag
+  const tag = result.match(/tag\:\s[a-zA-Z0-9\-\.]+\,/g);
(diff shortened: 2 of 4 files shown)


parseGitCommit.js
```diff
@@ -0,0 +1,36 @@
...
+  // Do not show the commit hash if that commit is also a tag
+  const tag = result.match(/tag\:\s[a-zA-Z0-9\-\.]+\,/g);
+  if (tag && tag.length > 0) {
+    return '';
+  } else {
+    return result.match(/[a-z0-9]+\s\(HEAD/g)[0].slice(0, -6);
+  }
+};
+
+const isDirty = () => {
```
kilrau: Nice!

lib/service/Service.ts
```diff
@@ -269,7 +270,7 @@ class Service {
       nodePubKey,
       uris,
       numPairs,
-      version: this.version,
+      version: `${this.version}${commitHash}`,
```
kilrau: I would go for `${this.version}${commitHash}`, since imo the version name should only be on the official release.

`version` in `xucli getinfo -j` should look like: `master-9a03442c-dirty`

michael1011: Are you sure about that? With the current approach if the commit is also the tag it will just show the version from `package.json`. If not it will give you the `package.json` version and the commit hash. This is enough information to know whether you are running the release I think.

Also it is really nice to know on what release the branch you are running is based on.

kilrau: > Also it is really nice to know on what release the branch you are running is based on.

Ok, I can see that. Fine then.

michael1011:
I just addressed your comments @sangaman! Thanks for the review!

I also just tried tagging a release and it worked fine for me. Could you give it another try?

parseGitCommit.js
```diff
@@ -0,0 +1,36 @@
...
+  const result = executeCommand('git status --short');
+  return result.length > 0;  
+};
+
+const versionFilePath = `${__dirname}/lib/Version.ts`;
+
+try {
+  // Delete the version file is it exists
+  if (fs.access(versionFilePath)) {
```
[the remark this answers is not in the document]
sangaman: Duh, I forgot about `unlinkSync`. Definitely the simplest way to do it.

PR #1203, opened 2019-09-05
fix(cli): naming inconsistencies

kilrau:
* nodePubKey -> node_key or node key
* node -> local xud node, remote node, peer
* other minor inconsistencies


lib/cli/commands/ban.ts
```diff
@@ -4,11 +4,11 @@ import { BanRequest } from '../../proto/xudrpc_pb';
 
 export const command = 'ban <node_pub_key>';
 
-export const describe = 'ban an xu node';
+export const describe = 'ban a peer';
 
 export const builder = {
   node_pub_key: {
-    description: 'nodePubKey of the node to ban',
+    description: 'node_pub_key of the peer to ban',
```
michael1011: Should we really use `node_pub_key` instead of `public key of the node` in the descriptions of the arguments?

kilrau: yeah, we can stick with `public key` in all textual descriptions

kilrau:
> Also there is a linting error. Please run `npm run lint` and make sure to fix it for the build to pass.

Fails. Can you help me check what's wrong? https://paste.ubuntu.com/p/m5dhDbMGJC/

michael1011:
> Fails. Can you help me check what's wrong? paste.ubuntu.com/p/m5dhDbMGJC

That's a weird error. For me it fails with:
```
michael@dell:~/Git/JavaScript/xud$ npm run lint

> xud@1.0.0-testnet.1 lint /home/michael/Git/JavaScript/xud
> tslint --fix --project tsconfig.json && tslint --fix --config tslint-alt.json 'bin/*' 'test/**/*.ts' 'tasks/**/*.ts'


ERROR: /home/michael/Git/JavaScript/xud/lib/cli/commands/create.ts:11:1 - Exceeds maximum line length of 150
```

kilrau:
`cli/commands/create.ts:11:1 - Exceeds maximum line length of 150`

I didn't see that one, but anyhow shortened the description to sth much shorter. Linting still failing locally.

lib/cli/commands/unlock.ts
```diff
@@ -4,11 +4,11 @@ import { UnlockNodeRequest } from '../../proto/xudrpc_pb';
 
 export const command = 'unlock <password>';
 
-export const describe = 'unlock an xud node';
+export const describe = 'unlock your xud node';
```
sangaman: This seems a bit off to me because we're not using second person (you/your) elsewhere and it's not really what I'm used to with other command line tools` help, for example `lncli unlock` says "unlock an encrypted wallet." This could just say "unlock xud" perhaps.

kilrau: Alright, maybe a bit too much "product", fixed.

lib/cli/commands/openchannel.ts
```diff
@@ -3,13 +3,13 @@ import { callback, loadXudClient } from '../command';
...
 import { coinsToSats } from '../utils';
 
-export const command = 'openchannel <nodePubKey> <currency> <amount>';
+export const command = 'openchannel <node_pub_key> <currency> <amount>';
 
-export const describe = 'open a payment channel';
+export const describe = 'open a payment channel with a peer';
 
 export const builder = {
   nodePubKey: {
```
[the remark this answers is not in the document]
kilrau: all -> `node_key`

lib/cli/commands/ban.ts
```diff
@@ -4,11 +4,11 @@ import { BanRequest } from '../../proto/xudrpc_pb';
 
 export const command = 'ban <node_pub_key>';
 
-export const describe = 'ban an xu node';
+export const describe = 'ban a peer';
 
 export const builder = {
   node_pub_key: {
-    description: 'nodePubKey of the node to ban',
+    description: 'Replace <node_pub_key> with the public key of the peer to ban',
```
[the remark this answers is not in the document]
kilrau: Agree, fixed.

lib/cli/commands/ban.ts
```diff
@@ -4,11 +4,11 @@ import { BanRequest } from '../../proto/xudrpc_pb';
 
 export const command = 'ban <node_pub_key>';
 
-export const describe = 'ban an xu node';
+export const describe = 'ban a peer';
```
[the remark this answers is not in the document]
kilrau: Right. While factually correct, I don't think most users think much about this. Why I'd want to stick with 'peer': I tried to consistently call other nodes in the network 'peer', no matter if connected or not. And the local xud 'xud node'.

kilrau:
Addressed your review and straightened out all descriptions to start with lower case. Please have another look, thanks @sangaman

lib/cli/commands/openchannel.ts
```diff
@@ -3,22 +3,22 @@ import { callback, loadXudClient } from '../command';
...
-export const command = 'openchannel <nodePubKey> <currency> <amount>';
+export const command = 'openchannel <node_key> <currency> <amount>';
 
-export const describe = 'open a payment channel';
+export const describe = 'open a payment channel with a peer';
 
 export const builder = {
-  nodePubKey: {
-    description: 'The node pub key of the peer to open the channel with',
+  node_key: {
```
erkarl: Think we also need to change:
`request.setNodePubKey(argv.nodePubKey);`
in this file.

lib/cli/commands/unban.ts
```diff
@@ -2,17 +2,17 @@ import { callback, loadXudClient } from '../command';
...
-export const command = 'unban <node_pub_key> [reconnect]';
+export const command = 'unban <node_key> [reconnect]';
 
-export const describe = 'unban an xud node';
+export const describe = 'unban a previously banned peer';
 
 export const builder = {
-  node_pub_key: {
-    description: 'NodePubKey of the node to unban',
+  node_key: {
```
erkarl: Same here. Think we also need to change:
`request.setNodePubKey(argv.node_pub_key);`

lib/cli/commands/getinfo.ts
```diff
@@ -121,7 +121,7 @@ const displayGetInfo = (response: GetInfoResponse.AsObject) => {
 
 export const command = 'getinfo';
 
-export const describe = 'get general info from the xud node';
+export const describe = 'get general info from your xud node';
```
sangaman: "your" usage here as well.

lib/cli/commands/discovernodes.ts
```diff
@@ -2,13 +2,13 @@ import { callback, loadXudClient } from '../command';
...
 
-export const command = 'discovernodes <peer_pub_key>';
+export const command = 'discovernodes <node_key>';
 
 export const describe = 'discover nodes from a specific peer';
 
 export const builder = {
-  peer_pub_key: {
-    description: 'nodePubKey of the peer to discover nodes from',
+  node_key: {
```
sangaman: I'd be a bit hesitant renaming to `node_key` in that it may be confusing and is inconsistent with lnd's naming for their similar concept, which is almost always referred to as "pubkey" or "public key" from what I can tell. It would also be inconsistent with our API naming.

At a minimum I think we should mention "public" in the description if we keep `node_key` as the arg name.

lib/cli/commands/streamorders.ts
```diff
@@ -9,7 +9,7 @@ export const describe = 'stream order added, removed, and swapped events (DEMO)'
 
 export const builder = {
   existing: {
-    description: 'should return existing orders',
+    description: 'returns existing orders',
```
sangaman: Maybe "whether to return existing orders" ?

kilrau: Overlooked that it's a boolean. Fixed.

lib/cli/commands/unban.ts
kilrau: Fixed.

lib/cli/commands/openchannel.ts
kilrau: Fixed.

lib/cli/commands/getinfo.ts
kilrau: Changed back to `get general info from the xud node`

lib/cli/commands/discovernodes.ts
kilrau: I'm a bit confused. Above we agreed to move everything to "node key", especially in the descriptions. So what exactly what do you want to change? Call all "node key" "node public key" in the `description:` fields?

sangaman: Sorry I just realized I gave conflicting feedback by suggesting "node key" myself as an alternative further up. I suppose it's fine, although I still slightly prefer "node pubkey" and "node_pubkey".

sangaman:
My last sticking point is use of "peer" vs "node", maybe we can discuss on the call today so we can move this PR along and get it merged soon. Thanks

lib/cli/commands/discovernodes.ts
kilrau: Did stick with "node key" and "node_key".

kilrau:
Please give this another look @sangaman

lib/cli/commands/discovernodes.ts
erkarl: We also need to change line 18 in this file `peer_pub_key->node_key`.

kilrau:
Damn... good catch. Went through again making sure all others commands use `(argv.node_key);` too.

Issue #1184, opened 2019-08-26
Swap failed


offerm:
xud-simnet-native.

using the script xud-simnet-check-swaps. 
5 swapAccepted received from peer but only 2 invoices created. This cause the other swap not to progress and to fail with timeout.

check for rHash 49d3729d1f0e75bac9810a34f6e4a446a706fbf7e03845c14f582672712dfb1b

The taker log:
```
26/08/2019 23:54:25.494 [SWAPS] debug: New deal: {"takerCltvDelta":100,"rHash":"49d3729d1f0e75bac9810a34f6e4a446a706fbf7e03845c14f582672712dfb1b","orderId":"c25cc980-c815-11e9-8eeb-b1a5f3e667ad","pairId":
"BTC/DAI","proposedQuantity":10000000,"rPreimage":"d93f21e6580551ad99f525c6c93940996f134a7329723072e21ec93c76306d2c","takerCurrency":"DAI","makerCurrency":"BTC","takerAmount":99980000000,"makerAmount":100
00000,"takerUnits":999800000000000000000,"makerUnits":10000000,"destination":"03a0bdf210dcecebb2d6f092b47fb7753af8114261888a365dfcb0d4548bbbdd2a","peerPubKey":"02b66438730d1fcdf4a4ae5d3d73e847a272f160fee2
938e132b52cab0a0d9cfc6","localId":"b01b8b70-c843-11e9-8b25-6d96ce5f243f","price":9998,"isBuy":true,"phase":0,"state":0,"role":0,"createTime":1566852865494}
26/08/2019 23:54:25.495 [SWAPS] debug: New deal: {"takerCltvDelta":40,"rHash":"e3318438b5d6c48981a19f5edc570034ea104a4bab7732fc14dd24db9817595a","orderId":"c26973b0-c815-11e9-8eeb-b1a5f3e667ad","pairId":"
LTC/BTC","proposedQuantity":10000000,"rPreimage":"9225920be1a537fae5073b2e365b6477bc3367546ebaf986a0b19b137b90a64b","takerCurrency":"BTC","makerCurrency":"LTC","takerAmount":77000,"makerAmount":10000000,"
takerUnits":77000,"makerUnits":10000000,"destination":"0270e8254e07649cdde230c9e09de6ff63c28a6d275c30b0a6863028d9db0e7c6f","peerPubKey":"02b66438730d1fcdf4a4ae5d3d73e847a272f160fee2938e132b52cab0a0d9cfc6"
[...]

PR #1207, opened 2019-09-07
fix(swaps): add resolve request validation slippage

ghost:
In this commit we're relaxing raiden's resolve request validation by 3 blocks in order to prevent an issue where new blocks are mined during the payment.

This is a quick fix to the issue described in: https://github.com/ExchangeUnion/xud/issues/1184#issuecomment-529089822

As a follow-up I'd like to add a simulation test scenario to make sure maker still has sufficient buffer.

### DIFF
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -739,8 +739,17 @@ class Swaps extends EventEmitter {
         source = 'Taker';
         destination = 'Maker';
         const lockExpirationDelta = expiration - chain_height;
-        if (deal.makerCltvDelta! > lockExpirationDelta) {
-          this.logger.error(`cltvDelta of ${lockExpirationDelta} does not meet ${deal.makerCltvDelta!} minimum`);
+        // We relax the validation by LOCK_EXPIRATION_SLIPPAGE blocks because
+        // new blocks could be mined during the time it takes from taker's
+        // payment to reach the maker for validation.
+        // This usually happens in simulated environments with fast mining enabled.
+        const LOCK_EXPIRATION_SLIPPAGE = 3;
+        if (deal.makerCltvDelta! - LOCK_EXPIRATION_SLIPPAGE > lockExpirationDelta) {
+          this.logger.error(`
+            lockExpirationDelta of ${lockExpirationDelta} does not meet
+            makerCltvDelta ${deal.makerCltvDelta!} - LOCK_EXPIRATION_SLIPPAGE ${LOCK_EXPIRATION_SLIPPAGE}
+            = ${deal.makerCltvDelta! - LOCK_EXPIRATION_SLIPPAGE} minimum
+          `);
           this.failDeal(deal, SwapFailureReason.InvalidResolveRequest, 'Insufficient CLTV received on first leg');
           return false;
         }


lib/swaps/Swaps.ts
```diff
@@ -739,7 +739,8 @@ class Swaps extends EventEmitter {
         source = 'Taker';
         destination = 'Maker';
         const lockExpirationDelta = expiration - chain_height;
-        if (deal.makerCltvDelta! > lockExpirationDelta) {
+        const LOCK_EXPIRATION_SLIPPAGE = 3;
+        if (deal.makerCltvDelta! - LOCK_EXPIRATION_SLIPPAGE > lockExpirationDelta) {
```
offerm: consider to add the LOCK_EXPIRATION_SLIPPAGE to the maker's final_cltv_delta  and here just check without the LOCK_EXPIRATION_SLIPPAGE.

erkarl: I don't understand. In the case of raiden there's no concept of `final_cltv_delta`. The way I understand it is that we only want to add the slippage for raiden payments from taker to maker.

offerm:
There can also be a slippage on the LN payment. Same reason.

The maker, when calculating the lock_duration is using the route to the
taker + ~24 hours. This is converted to the maker blocks. To this we should
add 3 blocks before we give to the taker.

When getting the payment we should make a normal check (as it is now,
without the 3 blocks)

erkarl:
> There can also be a slippage on the LN payment. Same reason.

The `cltv_expiry` for LN payment is relative. How can there be slippage?

sangaman (approved):
Looks good to me, I'd only recommend a line or two of comments explaining what the slippage is and why we must account for it, so in the future it's not a mystery to someone reading the code that's unfamiliar with the concept.

As for applying this to the lnd invoices, I'm also under the impression that it's not necessary since the values are relative. Ultimately the check that we have sufficient cltv delay happens within lnd code as I understand it, not xud (we just pass lnd the relative value when adding an invoice).

kilrau:
> Looks good to me, I'd only recommend a line or two of comments explaining what the slippage is and why we must account for it, so in the future it's not a mystery to someone reading the code that's unfamiliar with the concept.

Agree, these comments would be very nice :)

> As for applying this to the lnd invoices, I'm also under the impression that it's not necessary since the values are relative. Ultimately the check that we have sufficient cltv delay happens within lnd code as I understand it, not xud (we just pass lnd the relative value when adding an invoice).

That's also my understanding, could you confirm or explain what you mean so that this can be merged asap? @offerm

kilrau:
I just tested this branch with a native simnet setup. 1. leg lightning, 2. leg raiden orders (`xucli buy 1 btc/dai 10000`) still not working. Error same: `09/09/2019 12:44:46.856 [SWAPS] debug: deal 92521de004cd68ff57f8854d895b25f62b6775bf101d70ba135de4d502bd57bc failed due to SwapTimedOut: undefined`

Isn't this supposed to fix this or am I missing something?

kilrau:
Oh, the xud1.simnet.exchangeunion.com would need to be updated too I guess.

kilrau:
There are some other weird entries with the node key being replaced by `undefined`. Do we know why is that?

sangaman:
Those logs show `09/09/2019 10:44:18.393 [SWAPS] error: cltvDelta of 7359 does not meet 7360 minimum` which indicates that it's failing due to the bug that's fixed here, so we're all good on that front.

I'll look into that node key being replaced by undefined issue, definitely separate from this though.

sangaman (approved):

PR #1211, opened 2019-09-09
fix(peer): logging label/node pub key

sangaman:
This fixes a bug with how peers were labeled in the logs. The ternary operator that determined whether to log the pub key or the address was not evaluating as expected. This also adds test cases to prevent regression bugs in the future.

@kilrau This fixes the `undefined` you were seeing when peer pub keys were being logged.

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -124,8 +124,9 @@ class Peer extends EventEmitter {
 
   public get label(): string {
     return this.nodePubKey ||
-      this.expectedNodePubKey ? `${this.expectedNodePubKey}@${addressUtils.toString(this.address)}` :
-      addressUtils.toString(this.address);
+      (this.expectedNodePubKey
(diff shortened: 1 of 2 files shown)


lib/p2p/Peer.ts
```diff
@@ -124,8 +124,8 @@ class Peer extends EventEmitter {
 
   public get label(): string {
     return this.nodePubKey ||
-      this.expectedNodePubKey ? `${this.expectedNodePubKey}@${addressUtils.toString(this.address)}` :
-      addressUtils.toString(this.address);
+      (this.expectedNodePubKey ? `${this.expectedNodePubKey}@${addressUtils.toString(this.address)}` :
```
erkarl: A small style nit:
```
      (this.expectedNodePubKey
        ? `${this.expectedNodePubKey}@${addressUtils.toString(this.address)}`
        : addressUtils.toString(this.address));
```
I find this more readable when we separate the if/else clauses on new lines.

PR #1212, opened 2019-09-09
feat(cli): enhance create output

### LINKED ISSUE #1210: Master password prompts -> xucli
Then ping @reliveyy to adjust https://github.com/ExchangeUnion/xud-docker/pull/134

sangaman:
Closes #1210.

This improves the output of the `xucli create` command to provide clearer information to the user about how they should proceed and the outcome of the command.

Below is sample output:

```console
$ ./xucli create
You are creating an xud key and underlying wallets secured by a single password.
Enter a password: 
Re-enter password: 

 1. able        2. desk        3. online      4. that      
 5. skill       6. clap        7. color       8. this      
 9. figure     10. cage       11. torch      12. main      
13. screen     14. angle      15. hour       16. sail      
17. zoo        18. off        19. push       20. ancient   
21. bubble     22. pitch      23. involve    24. endless   

The following lnd wallets were initialized: BTC

Please write down your 24 word mnemonic. It will allow you to recover your xud
key and funds for the initialized wallets listed above should you forget your
password or lose your device. Keep it somewhere safe, it is your ONLY backup.
```

### DIFF
--- a/lib/cli/commands/create.ts
+++ b/lib/cli/commands/create.ts
@@ -1,7 +1,7 @@
-import { Arguments } from 'yargs';
-import { callback, loadXudInitClient } from '../command';
-import { CreateNodeRequest } from '../../proto/xudrpc_pb';
 import readline from 'readline';
+import { Arguments } from 'yargs';
+import { CreateNodeRequest, CreateNodeResponse } from '../../proto/xudrpc_pb';
+import { callback, loadXudInitClient } from '../command';
 
 export const command = 'create';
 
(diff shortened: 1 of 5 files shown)


lib/cli/commands/create.ts
```diff
@@ -1,31 +1,59 @@
...
-import readline from 'readline';
 
 export const command = 'create';
 
 export const describe = 'create an xud node';
 
 export const builder = {};
 
+const formatOutput = (response: CreateNodeResponse.AsObject) => {
+  if (response.seedMnemonicList.length > 0) {
```
erkarl:
```suggestion
  if (response.seedMnemonicList.length === 24) {
```

lib/cli/commands/create.ts
```diff
@@ -1,31 +1,59 @@
...
+    const numberedMnemonic = response.seedMnemonicList.map((value, index) => {
+      return `${index >= 9 ? '' : ' '}${index + 1}. ${value.padEnd(10)}`;
+    });
+    console.log('----------------------BEGIN XUD SEED---------------------');
+    for (let n = 0; n < response.seedMnemonicList.length / 4; n += 1) {
```
erkarl: Lots of magic numbers here, but at the same time I don't see a better way of doing it. Perhaps we could assign `const WORDS_PER_ROW = 4`?

sangaman: Sure I can do WORDS_PER_ROW

kilrau:
Very nice, liking the
`----------------------BEGIN XUD SEED---------------------`
`-----------------------END XUD SEED----------------------`
style

PR #1213, opened 2019-09-10
feat(config): exit on invalid config file

sangaman:
This commit makes `xud` exit during initialization if it encounters a config file that cannot be parsed properly as TOML. Previously, `xud` would silently error due to any syntax error in the config file and start up without any of settings from the file, potentially leading to confusion and unintended actions.

### DIFF
--- a/lib/Config.ts
+++ b/lib/Config.ts
@@ -131,7 +131,11 @@ class Config {
     };
   }
 
-  public load = async (args?: { [argName: string]: any }): Promise<Config> => {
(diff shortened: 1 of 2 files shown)


sangaman:
> Nice! In the future we could even add more validations. At the moment it is still possible to add invalid values with valid syntax (for example `port = "hello"`).

Yup, I have thought that too. Just having a set of config checks up front that ensure the correct types and valid values (so not `port = -1`).

PR #1215, opened 2019-09-10
Raiden addresses should be in EIP55 format

offerm:


erkarl:
Created https://github.com/ExchangeUnion/xud/issues/1216 to prevent this from happening with manually added currencies.

PR #1221, opened 2019-09-12
feat: shutdown gracefully on SIGTERM

ghost:
In this commit we add a listener to shut down xud gracefully in the
event that the process receives SIGTERM signal.


sangaman (approved):
Should have handled SIGTERM all along, thanks!

PR #1223, opened 2019-09-12
feat(swapclient): update capacity on channelbalance

sangaman:
This modifies the logic around updating the maximum outbound capacity for a swap client to occur every time the channel balance is fetched, regardless of whether `channelbalance` was called by the timer to periodically update capacity or via an `xud` gRPC request. This will give us slightly more up-to-date cached balance values without needing to make any additional calls to the swap clients.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -143,12 +143,7 @@ class LndClient extends SwapClient {
   }
 
   protected updateCapacity = async () => {
-    try {
-      this.maximumOutboundAmount = (await this.channelBalance()).balance;
-    } catch (e) {
-      // TODO: Mark client as disconnected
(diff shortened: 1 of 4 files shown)


sangaman:
Currently the `maximumOutboundCapacity` will automatically refresh periodically (with #1224 reducing the time between refreshes significantly), so manually checking the balance in step 3 above shouldn't be necessary.

We can still use your `hasSufficientOutboundBalance` approach - but I'm not sure it's worth it. It would be similar to checking the balance on every order placement.

erkarl:
> with #1224 reducing the time between refreshes significantly

I'd leave it as is for now since the check interval has been reduced to 3 seconds.

kilrau:
Nice feat! Needs another review or can be merged? @sangaman

PR #1243, opened 2019-09-24
Listtrades Enhancements #1232

### LINKED ISSUE #1232: Enhancement of `listtrades`
* `Completed at`: Timestamp (formatted, json can be unix)
* `Order ID`: UUID of the original order, can be the OrderID of multiple trades in the event of partial fills. abbrev. in formatted output, full in json output.
* `Trade ID`: rhash , abbreviated in formatted output, full in json output
* `Price per ETH`: `0.02 BTC`
* `Buy (Received)`: `50 ETH`
* `Sell (Sent)`: `1 BTC`
* `Add. Network Fees`: `0.0000004 BTC` 
* `Role`: Taker/Maker
* `Peer`: Alias

And probably we can close https://github.com/ExchangeUnion/xud/pull/625, I see a huge overlap and would want to avoid confusion by having two very similar calls listing trade/swap history.

rsercano:
resolves #1232 

This's in progress, I need your opinions on this, I just get a bit confused on where to get order information `(buy/sell, pubkey, price etc.)`. So I followed the previous way; if there's a maker order in trade then it's maker, otherwise take the info from taker. 

Please let me know if that's the right thing to do. 

image

### DIFF
--- a/lib/cli/commands/listtrades.ts
+++ b/lib/cli/commands/listtrades.ts
@@ -4,31 +4,46 @@ import { ListTradesRequest, ListTradesResponse, Trade } from '../../proto/xudrpc
 import Table, { HorizontalTable } from 'cli-table3';
 import colors from 'colors/safe';
 import { satsToCoinsStr } from '../utils';
+import { OrderSide } from '../../constants/enums';
 
 const HEADERS = [
   colors.blue('Trading Pair'),
-  colors.blue('Trade Quantity'),
+  colors.blue('Amount'),
+  colors.blue('Buy/Sell'),
   colors.blue('Price'),
-  colors.blue('Order Type'),
+  colors.blue('Peer Pubkey'),
+  colors.blue('Secret'),
+  colors.blue('Role'),
 ];
+
+const trimPubKey = (key: string) => {
+  if (key.length <= 0) {
+    return '';
+  }
+  return `${key.slice(0, 10)}...${key.slice(key.length - 10)}`;
+};
 
 const displayTrades = (trades: ListTradesResponse.AsObject) => {
   const table = new Table({ head: HEADERS }) as HorizontalTable;
   trades.tradesList.forEach((trade: Trade.AsObject) => {
-    const type = trade.makerOrder ? 'maker' : 'taker';
-    let price = 0;
-    if (trade.makerOrder) {
-      price = trade.makerOrder.price;
-    } else if (trade.takerOrder) {
-      price = trade.takerOrder.price;
-    }
-
-    table.push([
+    const details = [
       trade.pairId,
       satsToCoinsStr(trade.quantity),
-      parseFloat(price.toFixed(5)),
-      type,
-    ]);
+    ];
+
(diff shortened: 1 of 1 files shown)


kilrau:
I missed this a lot when testing on the weekend. It's also a main feature a market maker in an unattended  arby bot setup would want: see which trades happened.

Changes to current state of the PR:
- `listtrades` -> `tradehistory`. To some it might not be clear that order = still open and trade = filled orders. Also to differentiate from `listorders` which is not conform to what exchanges currently do since it includes orders from others. But open to ideas!
- `peerPubKey` -> `Partner Alias`
- `Secret` -> `Trade/Order ID`

Up to pick this up? @sangaman

kilrau:
And since https://github.com/ExchangeUnion/xud/pull/913 is quite a mess, close - wdyt? @sangaman

kilrau:
Only thing left is how to show partial fills. I would indicate it with a `(p) in the formatted and (`partially filled`) in the json output of `listorders`. The filled parts appear as regular trades in `tradehistory`. Wdyt?

raladev:
My suggestion how it should look in the end + another suggestion for partial fills:

-tradehistory command

Fields:

- Trade Id 
- **Order Id** (which looks like  UUID, not local id) - it is my suggestion for partial fills indication and matching with orders - trade id is unique for each trade, but order id for partial fills would be the same
- Date - would be good for filter

- Trading Pair
- Amount
- Side
{
- Price
- **Routing fee** - if someone took it, i want to now about that amount 
- **Total payment**
}

- Partner Alias
- Role (Taker/Maker)
- **Order Type (Limit / Market)**

sangaman:
> Changes to current state of the PR:
> 
>     Secret -> Trade ID (different from order id)

In its current state "Secret" is just the rHash, which is not even secret. We can certainly refer to this as `Trade Id` in the cli, although there won't be such a value for internal matches. I think "Payment Hash" or "Swap Hash" or just "Hash" might be ok as well.

> Fields:
> 
>     * **Routing fee** - if someone took it, i want to now about that amount
> 
>     * **Order Type (Limit / Market)**

Routing fee isn't something we currently store for previously executed swaps, although I agree it would be important. I'll open a separate issue to track this.

I'm not sure there's much value in displaying limit vs market - I think maker vs taker is more meaningful and there's not much space in the terminal for this. If anyone else agrees that limit/market would make sense here then we can add it, but it seems a bit much to me.

sangaman:
I'm closing this because there are a lot of changes from the current state of this PR to get it where we want it, I'll reopen an updated one shortly.

kilrau:
> I'm not sure there's much value in displaying limit vs market - I think maker vs taker is more meaningful and there's not much space in the terminal for this. If anyone else agrees that limit/market would make sense here then we can add it, but it seems a bit much to me.

That's fine, we can leave that out.

kilrau:
So new PR(s) should take care of:
- https://github.com/ExchangeUnion/xud/issues/1232
- https://github.com/ExchangeUnion/xud/issues/1511
- https://github.com/ExchangeUnion/xud/issues/1565

PR #1245, opened 2019-09-24
chore: run jest script in jest dir

sangaman:
This prevents the jest script from running any nested versions of xud such as those that may be installed in the simulation test temp dir.

It also separates the seedutil tests from running with the rest of the jest tests ~~and removes its dependencies from the travis config~~.


sangaman:
Yeah it's intentional, I figured that the seedutil code is separate and mostly static such that we don't gain much from compiling and running it every build. Paricularly with the pending Raiden changes which makes the build take much longer.

michael1011:
@sangaman I wouldn't feel to comfortable not running some tests at all in the CI.

What about using a build matrix similar to what [LND is doing](https://github.com/lightningnetwork/lnd/blob/master/.travis.yml#L17)? This would allow that the seedutil is compiled and tested in parallel to the normal CI script? (If we agree on setting it up it doesn't have to be implemented in this PR though)

sangaman:
Hmm ok I will put it back in for now, I think longer term it's worth changing though. I'll look into a build matrix or another approach to split it off from our main suite of tests.

PR #1246, opened 2019-09-24
feat(cli): improve create call response wording

kilrau:


lib/cli/commands/create.ts
```diff
@@ -25,13 +25,13 @@ const formatOutput = (response: CreateNodeResponse.AsObject) => {
...
+      console.log('The wallet for raiden was initialized.');
     }
 
     console.log(`
-Please write down your 24 word mnemonic. It will allow you to recover your xud
-key and funds for the initialized wallets listed above should you forget your
-password or lose your device. Keep it somewhere safe, it is your ONLY backup.
+Please write down your 24 word mnemonic. It will allow you to recover your xud node key and
+on-chain funds for the initialized wallets listed above, should you forget your password or
+lose your device. Keep it somewhere safe, it is your ONLY backup. Channels are not included.
```
sangaman: Maybe instead of "Channels are not included" say "Channel states can not be recovered from this seed" or something like that? Maybe specify that separate backups must be used to prevent loss of funds in channels? I just think "included" isn't very clear.

Or it could be left off altogether, I think specifying "on-chain" funds might be good enough.

lib/cli/commands/create.ts
```diff
@@ -25,13 +25,13 @@ const formatOutput = (response: CreateNodeResponse.AsObject) => {
...
-      console.log('The wallet for raiden was initialized');
+      console.log('The wallet for raiden was initialized.');
     }
 
     console.log(`
-Please write down your 24 word mnemonic. It will allow you to recover your xud
-key and funds for the initialized wallets listed above should you forget your
-password or lose your device. Keep it somewhere safe, it is your ONLY backup.
+Please write down your 24 word mnemonic. It will allow you to recover your xud node key and
+on-chain funds for the initialized wallets listed above, should you forget your password or
```
sangaman: Minor grammar nit but I don't think the comma after "above" is necessary.

sangaman:
Another minor thing is for the commit message this probably shouldn't be a `fix` since nothing is realy wrong or broken per se, maybe best a `feat` since this is more an enhancement to the existing wording.

lib/cli/commands/create.ts
kilrau: Changed wording. Please check again.

lib/cli/commands/create.ts
kilrau: Removed.

kilrau:
> Good details to add to the output. I'd also suggest keeping the line length no more than 80 characters (common default terminal width) for these log statements.

Fixed.

> Another minor thing is for the commit message this probably shouldn't be a fix since nothing is realy wrong or broken per se, maybe best a feat since this is more an enhancement to the existing wording.

Ok, true :)

Issue #1030, opened 2019-06-11
[XUD Master Password] Support aezeed passphrase encryption


kilrau:
Continuation of #912:
 
[Support (optional) aezeed passphrase encryption](https://github.com/ExchangeUnion/xud/pull/1006#issuecomment-499826994)
* optional
* requires this password in addition to mnemonic to restore seed

PR #1247, opened 2019-09-24
feat: create eth keystore with master seed

### LINKED ISSUE #1242: Initialize ETH keystore file/wallet from master seed
Currently the master seed does NOT restore the geth seed

Idea to include seedutil (go): when installing from npm: `npm install -g xud` - bundle 2 executables together

Also: this includes a PR against xud-docker to include the same functionality there

sangaman:
Closes #1242.

This commit uses the seed generated during the `CreateNode` call to create the raiden/ethereuem keystore along with the lnd wallets. This is the final step to having a single recovery seed and a single encryption password for all lnd and raiden wallets.

A new configuration option `raiden.keystorepath` specifies where xud should create the keystore directory. 

@reliveyy You should pass in this new option to `xud` when creating the wallets in docker. We will still need to figure out the best approach to ensuring the seedutil binary is available through non-docker installations. For dev/testing purposes if you have Go installed you can just use `npm run compile:seedutil`.

### DIFF
--- a/lib/Config.ts
+++ b/lib/Config.ts
@@ -47,24 +47,28 @@ class Config {
 
   constructor() {
     const platform = os.platform();
-    let lndDefaultDatadir;
+    let lndDefaultDatadir: string;
+    let raidenDefaultKeystorePath: string;
     switch (platform) {
       case 'win32': { // windows
         const homeDir = process.env.LOCALAPPDATA!;
         this.xudir = path.join(homeDir, 'Xud');
         lndDefaultDatadir = path.join(homeDir, 'Lnd');
+        raidenDefaultKeystorePath = path.join(homeDir, 'Ethereum');
         break;
       }
       case 'darwin': { // mac
         const homeDir = process.env.HOME!;
         this.xudir = path.join(homeDir, '.xud');
         lndDefaultDatadir = path.join(homeDir, 'Library', 'Application Support', 'Lnd');
+        raidenDefaultKeystorePath = path.join(homeDir, 'Library', 'Ethereum');
         break;
       }
       default: { // linux
         const homeDir = process.env.HOME!;
         this.xudir = path.join(homeDir, '.xud');
         lndDefaultDatadir = path.join(homeDir, '.lnd');
+        raidenDefaultKeystorePath = path.join(homeDir, '.ethereum');
         break;
       }
     }
@@ -126,6 +130,7 @@ class Config {
       disable: false,
       host: 'localhost',
       port: 5001,
+      keystorepath: raidenDefaultKeystorePath,
     };
   }
 
--- a/lib/cli/commands/create.ts
+++ b/lib/cli/commands/create.ts
@@ -25,14 +25,15 @@ const formatOutput = (response: CreateNodeResponse.AsObject) => {
       console.log(`The following lnd wallets were initialized: ${response.initializedLndsList.join(', ')}`);
     }
     if (response.initializedRaiden) {
-      console.log('The wallet for raiden was initialized.');
+      console.log('The keystore for raiden was initialized.');
     }
 
     console.log(`
 Please write down your 24 word mnemonic. It will allow you to recover your xud
 node key and on-chain funds for the initialized wallets listed above should you
-forget your password or lose your device. Off-chain funds are NOT recovered with
-it and are secured separately. Keep it somewhere safe, it is your ONLY backup.
+forget your password or lose your device. Off-chain funds in channels can NOT
+be recovered with it and must be backed up and recovered separately. Keep it
+somewhere safe, it is your ONLY backup in case of data loss.
     `);
   } else {
     console.log('xud was initialized without a seed because no wallets could be initialized.');
@@ -45,7 +46,10 @@ export const handler = (argv: Arguments) => {
     terminal: true,
   });
 
-  console.log('You are creating an xud node key and underlying wallets. All secured by a single password.');
+  console.log(`
+You are creating an xud node key and underlying wallets. All will be secured by
+a single password provided below.
+  `);
   process.stdout.write('Enter a password: ');
   rl.question('', (password1) => {
     process.stdout.write('\nRe-enter password: ');
--- a/lib/swaps/SwapClientManager.ts
+++ b/lib/swaps/SwapClientManager.ts
@@ -11,6 +11,7 @@ import { SwapClientType } from '../constants/enums';
 import { EventEmitter } from 'events';
 import Peer from '../p2p/Peer';
 import { UnitConverter } from '../utils/UnitConverter';
+import seedutil from '../utils/seedutil';
 
 export function isRaidenClient(swapClient: SwapClient): swapClient is RaidenClient {
   return (swapClient.type === SwapClientType.Raiden);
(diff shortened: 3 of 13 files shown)


lib/cli/commands/create.ts
```diff
@@ -25,7 +25,7 @@ const formatOutput = (response: CreateNodeResponse.AsObject) => {
       console.log(`The following lnd wallets were initialized: ${response.initializedLndsList.join(', ')}`);
     }
     if (response.initializedRaiden) {
-      console.log('The wallet for raiden was initialized');
+      console.log('The keystore for raiden was initialized');
```
erkarl:
```suggestion
      console.log('The keystore for raiden was initialized.');
```

lib/Config.ts
```diff
@@ -47,24 +47,28 @@ class Config {
...
         this.xudir = path.join(homeDir, '.xud');
         lndDefaultDatadir = path.join(homeDir, 'Library', 'Application Support', 'Lnd');
+        raidenDefaultKeystorePath = path.join(homeDir, 'Library', 'Ethereum');
         break;
       }
       default: { // linux
         const homeDir = process.env.HOME!;
         this.xudir = path.join(homeDir, '.xud');
         lndDefaultDatadir = path.join(homeDir, '.lnd');
+        raidenDefaultKeystorePath = path.join(homeDir, '.ethereum');
```
erkarl: Not a big deal, but I'm wondering if the default path to the generated keystore file should be xud's data directory instead. When I was testing the feature without reading the code I assumed it would be there after reading the output from xucli: `The keystore for raiden was initialized`.

sangaman: I just tried to match the default location that raiden uses for the keystore, but if the xud data dir makes more sense then that's fine. What do you think @michael1011  @kilrau ?

kilrau:
> I'm also wondering if we should be using xud's master password that we use to encrypt the private keys on disk as the password for the seed (or let user specify an extra seed password). With the current implementation we can recover the wallet only with 24 words (no password required). This can be done in a separate PR if we choose to go that route.

We discussed this before and it has an issue: https://github.com/ExchangeUnion/xud/issues/1030

We decided to push this to `post-1.0.0`, since it's "advanced" functionality which needs proper UX. The "lost device" case is covered, but once you forget your password you are screwed.

lib/Config.ts
kilrau: I think what Daniel did is right. [Geth](https://github.com/ethereum/go-ethereum/wiki/Backup-&-restore), as well as [raiden](https://medium.com/raiden-network/introducing-the-raiden-wizard-6c7c61c5b695) place the keystore file in

    Mac: `~/Library/Ethereum/keystore`
    Linux: `~/.ethereum/keystore`
    Windows: `%APPDATA%\Ethereum\keystore`

by default. What we can do, is adjust the output from xucli: `The keystore for raiden was initialized in [folder].`

kilrau: @reliveyy Docker should place the generated keystore file into the geth folder (we already have a keystore file there which should be replaced (correct me if I'm wrong @sangaman @erkarl @michael1011 )
```
root@K-Yoga:/home/kilrau/.xud-docker/mainnet/data/geth/keystore# ls -la
total 20
drwx------ 2 root root 4096 Sep 25 10:40 .
drwxr-xr-x 4 root root 4096 Sep 25 10:40 ..
-rw------- 1 root root  491 Sep 25 10:40 UTC--2019-09-25T08-40-13.140840485Z--ee9c0a1ed60057e3d0fcd5052aec084edad4b89e
```

Until https://github.com/raiden-network/raiden/issues/4298 is implemented in raiden, the keystore file is used by geth.

erkarl: What if the existing ethereum keystore path/folder already has 10+ keystore files in it. How do you know which one to use?

erkarl:
> We discussed this before and yes we want this at some point. It already has an issue: #1030 - but definitely not in this PR.

Thanks for reminding me. I'll bring this up during our next call to discuss if we can implement this without a breaking change.

lib/Config.ts
kilrau: Good point. `xud` should definitely know which one to use in that case. How I see it, we have two options and this needs to be added in this PR:
1. name the created keystore file sth like `xud_keystore` and make `xud` look for this file in the ethereum default path (quick n easy)
2. create a `keystore` config paramter for `[raiden]` in `xud.conf` which saves the full path to the keystore file to be used. This would also allow the user to easily switch to other (existing) keystore files, which will definitely be a requirement in future (better)

Not sure if there's a 3rd option. Let us know what you think @sangaman

erkarl: Just to clarify - unless I'm missing something xud doesn't need to know about raiden's keystore path after `xucli create`. The question  (with the current implementation) is how does raiden know which one to use if it saves it to a path with existing keystore files?

kilrau: >  xud doesn't need to know about raiden's keystore path

Right. But shouldn't `xud` somehow know which of the ethereum 'accounts' it created? To know which one to try to unlock upon `unlock`? Of course we could just try all in path and whichever successfully unlock can be used by raiden/geth.

> The question (with the current implementation) is how does raiden know which one to use if it saves it to a path with existing keystore files?

I asked the raiden team.

lib/swaps/SwapClientManager.ts
```diff
@@ -132,23 +133,36 @@ class SwapClientManager extends EventEmitter {
...
           swapClient.logger.debug(`could not initialize wallet: ${err.message}`);
         });
         initWalletPromises.push(initWalletPromise);
       }
     }
 
-    await Promise.all(initWalletPromises);
+    if (!this.raidenClient.isDisabled()) {
+      const { keystorepath } = this.config.raiden;
+      const keystorePromise = seedutil(seedMnemonic, walletPassword, keystorepath).then(() => {
```
erkarl: In order to programmatically unlock raiden we'll need to store the password of the keystore file in plain text. If I'm reading the code correctly it's using the xud master password to protect the keystore. Can we temporarily set it to empty string until we have a way of unlocking raiden via the REST API?

sangaman: Good thinking, I will do this on a temporary basis until we get a much needed call to unlock the keystore.

lib/Config.ts
kilrau: > > The question (with the current implementation) is how does raiden know which one to use if it saves it to a path with existing keystore files?
> 
> I asked the raiden team.

You *have* to specify `--address` and `--pasword-file` which identifies the keystore file.

sangaman: I was thinking the same config parameter here used to tell xud where to create the keystore would also be used to unlock the keystore once we have that functionality available.

I'm still not sure what the best approach here is in terms of the default value but I still figure it's OK to use the default eth folder. In a case where there are other keystores, we maybe should just expect users to set a path here in the config. It seems to me that either way, they'll need to be configuring either raiden OR xud *unless* we use the default eth keystore pass.

erkarl: Let's leave it as is.

lib/swaps/SwapClientManager.ts
sangaman: Done, I had to make a few changes to the seedutil tool to make the encryption password optional.

PR #1248, opened 2019-09-25
feat(cli): wait for client on create/unlock

sangaman:
This adds logic to wait for the xud client to be ready before attempting the `CreateNode` or `UnlockNode` calls. These calls may be made shortly after starting xud - for example as part of a bash script - and this wait of up to 3 seconds may prevent transient failures or race conditions.


sangaman:
Hmm I can add additional logic (ideally for the create call only) that waits up to a few seconds for the cert file to be created, I was thinking that would be in done withing docker. It wouldn't hurt to have it in both places, though, it's just one more quick check.

sangaman:
@kilrau @erkarl I added logic to wait for the cert file in a separate commit, you should be able to try reproducing without running into that ENOENT error now (it works in my tests).

lib/cli/command.ts
```diff
@@ -1,41 +1,23 @@
...
-}
+import { getDefaultCertPath } from './utils';
 
 /**
  * A generic function to instantiate an XU client.
  * @param argv the command line arguments
  */
 export const loadXudClient = (argv: Arguments) => {
-  const certPath = argv.tlscertpath ? argv.tlscertpath : path.join(getXudDir(), 'tls.cert');
+  const certPath = argv.tlscertpath ? argv.tlscertpath : getDefaultCertPath();
```
michael1011:
```suggestion
  const certPath = argv.tlscertpath || getDefaultCertPath();
```

Can be simplified a litte

lib/cli/command.ts
```diff
@@ -1,41 +1,23 @@
...
   const cert = fs.readFileSync(certPath);
   const credentials = grpc.credentials.createSsl(cert);
 
   return new XudClient(`${argv.rpchost}:${argv.rpcport}`, credentials);
 };
 
 export const loadXudInitClient = (argv: Arguments) => {
-  const certPath = argv.tlscertpath ? argv.tlscertpath : path.join(getXudDir(), 'tls.cert');
+  const certPath = argv.tlscertpath ? argv.tlscertpath : getDefaultCertPath();
```
michael1011:
```suggestion
  const certPath = argv.tlscertpath || getDefaultCertPath();
```

Same here

lib/cli/commands/create.ts
```diff
@@ -49,15 +82,30 @@ export const handler = (argv: Arguments) => {
...
-    rl.question('', (password2) => {
+    rl.question('', async (password2) => {
       process.stdout.write('\n\n');
       rl.close();
       if (password1 === password2) {
         const request = new CreateNodeRequest();
         request.setPassword(password1);
-        loadXudInitClient(argv).createNode(request, callback(argv, formatOutput));
+
+        const certPath = argv.tlscertpath ? argv.tlscertpath : getDefaultCertPath();
```
michael1011:
```suggestion
        const certPath = argv.tlscertpath || getDefaultCertPath();
```
Same here

lib/cli/commands/create.ts
```diff
@@ -49,15 +82,30 @@ export const handler = (argv: Arguments) => {
...
+          await waitForCert(certPath);
+        } catch (err) {
+          console.error(err);
+          process.exitCode = 1;
+          return;
+        }
+
+        const client = loadXudInitClient(argv);
+        // wait up to 3 seconds for rpc server to listen before call in case xud was just started
+        client.waitForReady(Date.now() + 3000, () => {
```
michael1011: Why are waiting 3 seconds for the gRPC server but 5 for the certificate to get created?

sangaman: I just tried to pick numbers I thought were reasonable. I made the maximum wait for the cert slightly longer thinking that the first run of xud might have a slightly longer initialization time, with creating the database and certificates. I'm open to suggestions for change tho.

PR #1261, opened 2019-10-01
Adds aliases to list orders

### LINKED ISSUE #643: Listorders CLI output - peerpubkey

rsercano:
resolves #643 

This's blocked by #486


sangaman:
If we go with putting aliases in the `listorders` output, I think it would be better to keep the current rpc calls as is and instead have the cli make separate calls to get node info for the pubkeys it gets back from the `ListOrders` call.

rsercano:
I think this should be fully clarified @sangaman

kilrau:
In general I would not want to have the CLI calls differ much from the underlying gRPC calls but in the case of `alias`, I agree since it's purely for the user.

So yes, whenever we want to show an alias to the user, we can use other rpc calls to get it.

kilrau:
How it is:
```
Trading pair: ETH/BTC
┌────────────────────────────────────────────────────┬────────────────────────────────────────────────────┐
│ Buy                                                │ Sell                                               │
├───────────────────┬───────────────────┬────────────┼───────────────────┬───────────────────┬────────────┤
│ Quantity          │ Price             │ Own Order  │ Quantity          │ Price             │ Own Order  │
├───────────────────┼───────────────────┼────────────┼───────────────────┼───────────────────┼────────────┤
│ 0.001             │ 2                 │            │ 0.001             │ 2                 │            │
├───────────────────┼───────────────────┼────────────┼───────────────────┼───────────────────┼────────────┤
│                   │                   │            │ 0.001             │ 2                 │            │
└───────────────────┴───────────────────┴────────────┴───────────────────┴───────────────────┴────────────┘
```

How we want it:
```
Trading pair: ETH/BTC
┌────────────────────────────────────────────────────┬────────────────────────────────────────────────────┐
│ Buy                                                │ Sell                                               │
├───────────────────┬───────────────────┬────────────┼───────────────────┬───────────────────┬────────────┤
│ Quantity          │ Price             │ Own Order  │ Quantity          │ Price             │ Own Order  │
├───────────────────┼───────────────────┼────────────┼───────────────────┼───────────────────┼────────────┤
│ 0.001             │ 2                 │            │ 0.001             │ 2                 │            │
├───────────────────┼───────────────────┼────────────┼───────────────────┼───────────────────┼────────────┤
│                   │                   │            │ 0.001             │ 2                 │            │
└───────────────────┴───────────────────┴────────────┴───────────────────┴───────────────────┴────────────┘
```

sangaman:
Closing this PR since it's incomplete and somewhat dated, I'll open another one to resolve the same issue.

PR #1262, opened 2019-10-01
feat(swaps): raiden claims recovered swap payments

### LINKED ISSUE #1251: Recover Raiden Swaps
Enhancement of #1081:

Currently #1081 only recovers swaps for lnd. This is about recovering swaps for raiden.

Dependency: https://github.com/ExchangeUnion/raiden/issues/12 by @offerm

sangaman:
Closes #1251.

This adds logic to track recovered swap preimages for claiming raiden payments. A proposed change to raiden has it continuously attempt to resolve a hash until it gets a response - if xud crashes during or before the first attempt then raiden will continue trying until xud comes back online. When it does, xud will check if it has recovered the preimage and return it to raiden if so, effectively claiming the payment and preventing loss of funds.

### DIFF
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -900,6 +900,16 @@ class Swaps extends EventEmitter {
 
     this.logger.debug(`handleResolveRequest starting with hash ${rHash}`);
 
+    // first check if we have recovered this deal from a previous swap attempt
+    const recoveredSwap = this.swapRecovery.recoveredPreimageSwaps.get(rHash);
+    if (recoveredSwap && recoveredSwap.rPreimage) {
+      recoveredSwap.state = SwapState.Recovered;
+      recoveredSwap.save().catch(this.logger.error);
+      this.swapRecovery.recoveredPreimageSwaps.delete(rHash);
+      this.logger.info(`handleResolveRequest returning recovered preimage ${recoveredSwap.rPreimage} for hash ${rHash}`);
+      return recoveredSwap.rPreimage;
+    }
+
     const deal = this.getDeal(rHash);
 
     if (deal) {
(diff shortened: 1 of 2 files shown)


lib/swaps/Swaps.ts
```diff
@@ -900,6 +900,16 @@ class Swaps extends EventEmitter {
 
     this.logger.debug(`handleResolveRequest starting with hash ${rHash}`);
 
+    // first check if we have recovered this deal from a previous swap attempt
+    const recoveredSwap = this.swapRecovery.recoveredPreimageSwaps.get(rHash);
+    if (recoveredSwap && recoveredSwap.rPreimage) {
+      recoveredSwap.state = SwapState.Recovered;
+      recoveredSwap.save().catch(this.logger.error);
+      this.swapRecovery.recoveredPreimageSwaps.delete(rHash);
```
erkarl: I'm wondering if this is the correct place to delete swap from memory and mark it as recovered? Should we wait for confirmation that the recovery was successful?

What happens when:
1. swap client requests the preimage
2. xud sends the preimage and deletes it from memory
3. swap client crashes before it manages to claim the payment
4. swap client comes back online and requests the preimage again
5. xud no longer has the preimage in memory (we still have the preimage in the database, but the swap state is marked as recovered)

sangaman: That's a valid concern, it ventures into how xud deals with swap clients crashing/malfunctioning. Technically that can already happen, as we treat a swap deal as "completed" once we send the preimage to raiden and then remove it from memory (we do persist to the database though). It doesn't necessarily have to be a swap that xud has recovered upon startup.

I think it's slightly out of scope to address in this PR, but we could make it so that on all resolve requests if we don't recognize the hash we check the database for completed swaps and return the preimage if we find a match.

erkarl: > I think it's slightly out of scope to address in this PR, but we could make it so that on all resolve requests if we don't recognize the hash we check the database for completed swaps and return the preimage if we find a match.

That works.

erkarl: Created an issue: https://github.com/ExchangeUnion/xud/issues/1264

PR #1283, opened 2019-10-10
refactor(cli): hide executeswap command

### LINKED ISSUE #1231: Remove `executeswap` call from CLI

rsercano:
resolves #1231 

This completely removes `executeSwap` command from CLI. If you would prefer to comment it out instead, let me know. 

On the other hand I couldn't test it properly  `npm run compile` command doesn't generate `dist` folder properly? Even if I remove it, it doesn't create it. So my compiled `JS` files still have this command but according to the changed `TS` files it's impossible.


rsercano:
Alright, removed the description of `executeswap`, therefore it won't be shown on `help` output

Before;
image

After;
image

PR #1285, opened 2019-10-10
feat: add xud status

### LINKED ISSUE #1176: Add xud "status"
Dependent on https://github.com/ExchangeUnion/xud/issues/1174, which introduces `Status` fields for L2 clients.

We would want xud to have a `Status` field too, with some basic checks:
* check if L2 clients all have status `Ready`, if one does not show `Ready`, show error string starting with swap client e.g. `LND-BTC: lnd is NotInitialized`
* check if `listpeers` has an entry, if not show `Not connected to any peers`
* If checks above succeed, show `Ready`

rsercano:
resolves #1176 

Please see attached screenshots for demonstration.

### DIFF
--- a/lib/cli/commands/getinfo.ts
+++ b/lib/cli/commands/getinfo.ts
@@ -29,13 +29,38 @@ ${info.urisList[0].substring(info.urisList[0].indexOf('@'))}` : '';
   console.log(basicInfotable.toString(), '\n');
 };
 
+const determineXudStatus = (lndMap: [string, LndInfo.AsObject][], raiden: RaidenInfo.AsObject | undefined) => {
+  let status = '';
+
+  /*const peers = this.listPeers();
+  if (!peers || peers.length === 0) {
+    status = 'Not connected to any peers';
+    status += '\n';
+  }*/
+
+  lndMap.forEach((asset) => {
+    if (asset[1].status !== 'Ready') {
+      status += `LND-${asset[0]}: ${asset[1].status}`;
+      status += '\n';
+    }
+  });
+
+  if (!raiden) {
+    status += 'Raiden: Could not fetch status information';
+  } else if (raiden.status !== 'Ready') {
+    status += `Raiden: ${raiden.status}`;
+  }
+
+  return status.substring(0, status.length - 1) || 'Ready';
+};
+
 const displayGeneral = (info: GetInfoResponse.AsObject) => {
   const table = new Table() as VerticalTable;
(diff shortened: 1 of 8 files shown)


rsercano:
Alright then I'll remove it from the proto and as you said will move determining logic to the CLI side.

rsercano:
Okay, I moved the logic of determining status field to the `CLI` side, but this comes with a problem which makes `getinfo` calls to depend on `listpeers` call, since there's no field to carry this information from `grpc` to `CLI`.

So to show peer information I first need to call `listpeers` and on it's callback I need to call `getinfo`, which seems not okay to me. 

What do you think ? @sangaman @kilrau

kilrau:
> I am also not sure it makes sense to repeat the raiden/lnd status in the xud status, but that's not a big deal if it makes sense to other people.

Looking at https://github.com/ExchangeUnion/xud/issues/1176, what do you recommend to show instead? My thinking is that if one of the sub-systems has a problem, xud has a problem and the user should see it at a glance. It would be somewhat confusing if xud showed `Ready`, but e.g. everything else is down. Open to ideas! @sangaman 

Can you show a screenshot if multiple (e.g. both LND's) are not ready? @rsercano
There should definitely be a line break if the status field has multiple entries.

> p.s. I didn't implement this logic it's a few lines of change in case you want it, but the depending on listpeers call while using getinfo call situtation is not okay to me.

Don't have a better idea. @sangaman ?

rsercano:
There's a demonstration in the first comment Kilian @kilrau

kilrau:
Overlooked that - LGTM

rsercano:
Great, then Daniel could you please confirm that it's okay to call `listpeers` and depend on it while calling `getinfo` ?

lib/cli/commands/getinfo.ts
```diff
@@ -29,12 +29,38 @@ ${info.urisList[0].substring(info.urisList[0].indexOf('@'))}` : '';
   console.log(basicInfotable.toString(), '\n');
 };
 
+const determineXudStatus = (lndMap: [string, LndInfo.AsObject][], raiden: RaidenInfo.AsObject | undefined) => {
+  let status = '';
+
+  /*const peers = this.listPeers();
```
michael1011: Why call `this.listPeers()` if we already have the number of peers in the response of `GetInfo` as `info.numPeers`?

kilrau: Probably because I wrote it that way in the issue...

Agree - just check if this number is >0 does the trick!

PR #1288, opened 2019-10-11
fix: simplify wallet creation output

### LINKED ISSUE #1263: Simplify wallet created output
How it is:
```
The following lnd wallets were initialized: BTC, LTC
The keystore for raiden was initialized.
```

How it should be:
```
The following wallets were initialized: BTC, LTC, ERC20(ETH).
```

michael1011:
Closes #1263

The new output will look like this: 
```
The following wallets were initialized: LTC, BTC, ERC20(ETH)
```

### DIFF
--- a/lib/cli/commands/create.ts
+++ b/lib/cli/commands/create.ts
@@ -54,12 +54,21 @@ const formatOutput = (response: CreateNodeResponse.AsObject) => {
     }
     console.log('-----------------------END XUD SEED----------------------\n');
 
+    let walletInitializedMessage = 'The following wallets were initialized: ';
+
     if (response.initializedLndsList.length) {
-      console.log(`The following lnd wallets were initialized: ${response.initializedLndsList.join(', ')}`);
+      walletInitializedMessage += response.initializedLndsList.join(', ');
     }
+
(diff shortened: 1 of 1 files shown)


lib/cli/commands/create.ts
```diff
@@ -54,13 +54,22 @@ const formatOutput = (response: CreateNodeResponse.AsObject) => {
     }
     console.log('-----------------------END XUD SEED----------------------\n');
 
+    let walletInitializedMessage = 'The following wallets were initialized: ';
+
     if (response.initializedLndsList.length) {
-      console.log(`The following lnd wallets were initialized: ${response.initializedLndsList.join(', ')}`);
+      walletInitializedMessage += response.initializedLndsList.join(', ');
```
erkarl: What if Raiden is not initialized (only lnds)? The message will end with ", ".

michael1011: Not it won't because the [`join` method just puts the separator between the elements of the array](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/join). The message will end with the last entry in `response.initializedLndsList`

PR #1293, opened 2019-10-14
Adds configuration value type checks before startup

### LINKED ISSUE #1222: Configuration value checks on startup
We currently load configuration values from the config file without validating them. It's therefore possible to submit values like `port="notanumber"` or `port=-1`. We should add checks up front to ensure that the configuration values we read are of the expected type and within known boundaries of valid values. If we encounter any invalid values, exit xud and print an error message.

rsercano:
resolves #1222 

This's what I can think of to validate configuration values, by recursively checking the properties that comes from config file and comparing them with the ones in `Config.ts`

### DIFF
--- a/lib/Config.ts
+++ b/lib/Config.ts
@@ -173,6 +173,8 @@ class Config {
     const configProps = await Config.readConfigProps(configPath);
 
     if (configProps) {
+      this.validateConfigProps(configProps);
+
       // set the network and xudir props up front because they influence default config values
       if (configProps.network && (!args || !args.network)) {
         this.network = configProps.network;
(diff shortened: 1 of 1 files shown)


lib/Config.ts
```diff
@@ -295,6 +297,36 @@ class Config {
...
     return process.env.NODE_ENV === 'production' ? Level.Info : Level.Debug;
   }
+
+  private validateConfigProps(configProps: any) {
+    for (const prop in configProps) {
+      this.checkTypes(configProps, prop, (this as any)[prop]);
+    }
+  }
+
+  private checkTypes(configProps: any, prop: string, expectedVal: any) {
```
michael1011: I would call this function `checkType` because it checks just a single type.

lib/Config.ts
```diff
@@ -295,6 +297,36 @@ class Config {
...
+    if (typeof configProps[prop] !== 'object') {
+      const expectedType = Object.prototype.toString.call(expectedVal);
+      const typeInConfigFile = Object.prototype.toString.call(configProps[prop]);
```
michael1011: `typeInConfigFile` is a little weird for a variable name. What about `actualType`? Would be a great contrast to `expectedType`?

lib/Config.ts
```diff
@@ -295,6 +297,36 @@ class Config {
...
+        const typeInConfigFile = Object.prototype.toString.call(fileValue);
+        if (expectedType !== typeInConfigFile) {
+          console.log(`${prop}[${fileKey}] in config file should have a type of ${expectedType}`);
+          process.exit(-1);
+        }
+
+        if (expectedType === '[object Object]') {
+          this.checkTypes(configProps[prop], fileKey, expectedVal[fileKey]);
+        }
+      }
```
michael1011: Why start with the recursion in the second nested `object`? The whole logic in the loop could be replaced by a recursive call.

Maybe also add an optional parameter to the function so that the whole path of the misconfigured value is shown in the error message.

sangaman:
@rsercano Will you get a chance to respond to michael's feedback above? I can give another thorough review after his concerns are addressed.

rsercano:
I'll try my best Daniel, nowadays I just started to a new job so it's getting hard for me to contribute

sangaman:
> I'll try my best Daniel, nowadays I just started to a new job so it's getting hard for me to contribute

No worries, just let us know.

sangaman:
@rsercano Should I take this over or do you think you'll have time for this in the coming weeks? Thank you!

sangaman:
I'm closing this PR and will open another one because my changes are significant. The next PR will do some sanity checking on the actual values passed in and will check not just the config file properties but also the command arguments (since it's possible to pass invalid values from the command line).

PR #1296, opened 2019-10-15
feat(swaps): add flexibility to second leg route

### LINKED ISSUE #1165: Add flexibility on second leg's `lockDuration`
Status Quo: currently the maker calls `QueryRoutes` to get a route with a specific total cltv and specific cltv_limit. Chances are high that `SendPayment` is limited to this one specific route (and maybe some more routes with shorter cltv values).

Idea: add an equivalent of e.g. 40 BTC blocks to the total cltv of the returned route to allow `SendPayment` to use routes with slightly higher cltv values and increase routing success. 

TBD: Raiden?

sangaman:
This adds a buffer to the `takerMaxTimeLock` equal to our `finalLock` value, which by default is 40 blocks on BTC, which may open the possibility of using alternate routes with slightly longer lock times should payment using the first route we found fail.

Closes #1165.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -459,9 +459,8 @@ class LndClient extends SwapClient {
         destination: deal.takerPubKey!,
         amount: deal.takerAmount,
         finalCltvDelta: deal.takerCltvDelta,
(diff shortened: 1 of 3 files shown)


kilrau:
How about Raiden?

PR #1297, opened 2019-10-15
feat(resolver): retry flag in error message

### LINKED ISSUE #1282: achilles heel
https://github.com/raiden-network/raiden/pull/5025 was merged to Raiden.

IMHO, this open the door for a "not atomic swap".

consider raiden as first leg and lnd as second leg. Alice and Bob agreed on a swap and swap started. Alice issue first leg to Bob, Bob's xud called from raiden and issued the second leg. Now, there was some kind of a problem on LND and bob's resolver returned HTTP 500 to raiden.

Outcome - alice got paid (second leg). Bob got the secret from Alice. Bob didn't get his payment from Alice. resolving the payment on-chain by Bob is very hard and practically impossible. 

I haven't done full analysis of xud to check if there is a scenario that can cause it. Even if it is hard to find one now, we can't be sure that we will not add one in the future.

The resolution IMHO is that the resolver will only send back HTTP codes 200 (with secret)  and 404 (not found). We can consider other codes if these represents situations that we understand. But 500 and other generic codes should never be  used.

sangaman:
Closes #1282.

This adds a `retry` flag to the resolver error message to indicate whether the resolver request should be retried. This is set in cases where xud is not certain that the swap will never be resolved, such as in cases where a call to send payment is interrupted or does not respond normally. It also sets a 503 status code when a retry is indicated.


lib/http/HttpServer.ts
```diff
@@ -54,35 +54,43 @@ class HttpServer {
...
+        try {
+          const resJson = await this.httpService.resolveHashRaiden(reqJson);
+          res.writeHead(200, { 'Content-Type': 'application/json' });
+          res.end(JSON.stringify(resJson));
+        } catch (err) {
+          switch (err.code) {
+            case swapErrorCodes.INVALID_RESOLVE_REQUEST:
+            case errorCodes.INVALID_ARGUMENT:
+              res.writeHead(400);
+              res.end({ message: err.message, retry: false });
```
offerm: What is the called expected to do?

lib/http/HttpServer.ts
```diff
@@ -54,35 +54,43 @@ class HttpServer {
...
+              break;
+            case swapErrorCodes.PAYMENT_HASH_NOT_FOUND:
+              res.writeHead(404);
+              res.end({ message: err.message, retry: false });
+              break;
+            case swapErrorCodes.FINAL_PAYMENT_ERROR:
+            case swapErrorCodes.PAYMENT_REJECTED:
+              res.writeHead(500);
+              res.end({ message: err.message, retry: false });
```
offerm: What is the called expected to do?  
Are we sure that second leg payment was not done?

lib/http/HttpServer.ts
```diff
@@ -54,35 +54,43 @@ class HttpServer {
...
+              break;
+            default:
+              res.writeHead(503);
+              res.end({ err, retry: true });
+              break;
+          }
+        }
+        break;
+      default:
+        res.writeHead(404);
```
offerm: Isn't is safer to return 400 here?

lib/lndclient/LndClient.ts
```diff
@@ -506,20 +506,28 @@ class LndClient extends SwapClient {
   private executeSendRequest = async (
     request: lndrpc.SendRequest,
   ): Promise<string> => {
+    if (!this.isConnected()) {
+      throw swapErrors.FINAL_PAYMENT_ERROR(errors.UNAVAILABLE(this.currency, this.status).message);
```
offerm: Why is this a final error? 
We are clearly in a swap but LND is temporarily not connected. Shouldn't we retry?

lib/lndclient/LndClient.ts
```diff
@@ -506,20 +506,28 @@ class LndClient extends SwapClient {
...
+
     this.logger.trace(`sending payment with ${JSON.stringify(request.toObject())}`);
     let sendPaymentResponse: lndrpc.SendResponse;
     try {
       sendPaymentResponse = await this.sendPaymentSync(request);
     } catch (err) {
       this.logger.error('got exception from sendPaymentSync', err);
-      throw swapErrors.PAYMENT_ERROR(err.message);
+      if (typeof err.message === 'string' && err.message.includes('chain backend is still syncing')) {
+        throw swapErrors.FINAL_PAYMENT_ERROR(err.message);
```
offerm: Same here. Shouldn't we retry until LND is operational again?

lib/lndclient/LndClient.ts
```diff
@@ -506,20 +506,28 @@ class LndClient extends SwapClient {
...
+      } else {
+        throw swapErrors.UNKNOWN_PAYMENT_ERROR(err.message);
```
offerm: This is the worst case. We have no idea if the payment was done or not.

lib/http/HttpServer.ts
sangaman: I refactored this a bit but the functionality is the same. When the resolve request is invalid (e.g. it does not specify a `secrethash` or is otherwise malformed) we respond with `400 BAD REQUEST`.

lib/http/HttpServer.ts
sangaman: Yes, although this relies on the logic behind `FINAL_PAYMENT_ERROR` and `PAYMENT_REJECTED` to only be used when we're sure the payment has not and will not go through.

lib/http/HttpServer.ts
sangaman: I think 404 is the standard HTTP status code when a request targets a path that does not exist or is not recognized. But part of what I'm going for here is that the requester (raiden) should not rely on the status code to determine whether a resolve request failure is permanent or not.

lib/lndclient/LndClient.ts
sangaman: My thinking here is that if lnd is down, we should fail the swap immediately rather than wait in case lnd comes back. The swap could then be retried with a new swap deal. This covers an edge case where lnd goes down in the middle of a swap.

This is the current behavior - there's currently no mechanism to put a swap on hold while we wait to see if lnd comes back online - so I'm just trying to categorize this as an error case where we know that the payment won't go through (because no send payment call has been made, in this case).

lib/lndclient/LndClient.ts
sangaman: In this case lnd is operational, but it is won't accept send payment calls because it is out of sync. This is also an edge case as we wait for lnd to sync to chain before we'll attempt any swaps with it, but I just put this here to cover the edge case where it goes back out of sync somehow.

lib/lndclient/LndClient.ts
sangaman: Yes, so when we encounter an unknown payment error we fail the swap but we *don't* respond to the resolve request with `retry=false`.

test/jest/HttpServer.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+import Service from '../../lib/service/Service';
+import errors from '../../lib/swaps/errors';
+import { ResolveRequest } from '../../lib/swaps/types';
+import { getUnusedPort } from '../utils';
+
+jest.mock('../../lib/Logger');
+jest.mock('../../lib/service/Service');
+const mockedService = <jest.Mock<Service>><any>Service;
+
+describe('HttpServer - resolveraiden', () => {
```
erkarl: Nice tests :).

sangaman:
Rebased!

sangaman:
@erkarl Really rebased now if you want to take one more look.

Issue #1216, opened 2019-09-10
Convert Raiden contract addresses to EIP55 format


erkarl:
Raiden requires contract addresses to be in [EIP55](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md) format. We should convert the provided address on `addcurrency` request.

PR #1298, opened 2019-10-16
feat(grpc): Convert Raiden contract addresses to EIP55 format

krrprr:
Closes https://github.com/ExchangeUnion/xud/issues/1216


lib/service/Service.ts
```diff
@@ -96,11 +98,27 @@ class Service {
     argChecks.VALID_SWAP_CLIENT(args);
     const { currency, swapClient, tokenAddress, decimalPlaces } = args;
 
+    let address = tokenAddress;
+    if (args.swapClient === SwapClientType.Raiden && !!address) {
+      address = address.toLowerCase().replace('0x', '');
```
erkarl: Could we move the conversion logic to to `utils.ts` and use `toChecksumAddress` in this function?

krrprr: Sure, seems like a good idea. Will do so.

lib/utils/utils.ts
```diff
@@ -5,6 +5,8 @@ import { createHash, randomBytes as cryptoRandomBytes } from 'crypto';
 import { promisify } from 'util';
 import moment from 'moment';
 
+const createKeccakHash = require('keccak');
```
sangaman: Rather than use `require` here, I think it'd be better to try and stay consistent and use `import` instead. I know that it'll give us an error because no type declarations are available for the module, but you can use `@ts-ignore` like so:

```typescript
// @ts-ignore
import createKeccakHash from 'keccak';
```

lib/utils/utils.ts
```diff
@@ -208,3 +210,23 @@ export const base64ToHex = (b64: string) => {
 export const hexToUint8Array = (hex: string) => {
   return Uint8Array.from(Buffer.from(hex, 'hex'));
 };
+
+/** Converts input to EIP55 format.
+ * Prints the ith digit in uppercase if it's a letter and the 4*ith bit of the hash of the lowercase hexadecimal address is 1
+ * otherwise prints it in lowercase.
+ * Example: '0xfb6916095ca1df60bb79ce92ce3ea74c37c5d359' is converted to '0xfB6916095ca1df60bB79Ce92cE3Ea74c37c5d359'
+ */
+export const toChecksumAddress = (address: string) => {
```
sangaman: I'm thinking it might be more clear to call this `toEip55Address`?

lib/utils/utils.ts
```diff
@@ -208,3 +210,23 @@ export const base64ToHex = (b64: string) => {
 export const hexToUint8Array = (hex: string) => {
   return Uint8Array.from(Buffer.from(hex, 'hex'));
 };
+
+/** Converts input to EIP55 format.
```
sangaman: Minor style nit but I think for the comments to be properly parsed the text should start on the line after `/**` when it's a multi-line comment block, you can find examples elsewhere if you're unclear.

lib/service/Service.ts
```diff
@@ -96,11 +96,16 @@ class Service {
     argChecks.VALID_SWAP_CLIENT(args);
     const { currency, swapClient, tokenAddress, decimalPlaces } = args;
 
+    let address = tokenAddress;
+    if (args.swapClient === SwapClientType.Raiden && !!address) {
```
sangaman: Very minor but the if condition logic here should work the same without the `!!`.

lib/utils/utils.ts
krrprr: I agree.

krrprr:
Thanks for the comments @sangaman ! Implemented the changes.

sangaman:
The build checks passed with LTS node so I'm going to merge this despite the node v13 failure since that still does not work with grpc.

kilrau:
> The build checks passed with LTS node so I'm going to merge this despite the node v13 failure since that still does not work with grpc.

Didn't we decide to only support LTS on travis?

PR #1303, opened 2019-10-29
chore: disable Travis builds with latest node

michael1011:
Since the update to Node `v13.0.1` on the `node` version on Travis, our builds have been failing because that node version is incompatible with a gRPC dependency.

This PR causes Travis to just execute on the LTS version of node. Maybe we should document somewhere that we are just working with/testing on the latest Node LTS?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: node_js
 
 node_js:
-  - "node"
   - "lts/*"
 
 cache:


sangaman:
Thanks - I tried the workaround (building from source) reported in the issue I opened (https://github.com/grpc/grpc-node/issues/1086) but it seems so far there is no solution.

In the past the latest release of node caused the grpc dependency to break but it was quickly fixed, within a day or so, but now it has been close to a week. I think this is also the 3rd time in a row the latest release of node caused builds to fail due to grpc - even when it's fixed it's still an inconvenience. So I am fine with only building LTS in travis going forward (not just as a temporary workaround).

PR #1316, opened 2019-11-02
chore: update grpc dependency for node v13

sangaman:
This updates the `grpc` dependency to add compatibility with Node.js v13. It also updates the `grpc-dynamic-gateway` dependency to align with the grpc version used by xud.


sangaman:
> LGTM! Should we enable the Travis builds on Node v13 again?

I don't think so because I don't want to have to worry about this again when node v14 comes out. When it does, I'll try and upgrade this module yet again so that xud is compatible but I don't want to have Travis builds start failing again.

Issue #1130, opened 2019-08-08
getinfo doesn't show inactive lnd channels, channelbalance does


kilrau:
### Background

Inactive lnd channels don't show up in `xucli getinfo` under "inactive" but in `xucli channelbalance`. Tested on native simnet with both, `lndbtc` & `lndltc`.

### Your environment

* version of `xud`: latest master (`HEAD 3c46327`)
* which operating system (`uname -a` on *Nix): ubuntu 18.04.3
* version of `btcd`, `bitcoind`, `ltcd`, `litecoind`, `geth`, `parity`, `raiden`, `lnd` or other chain backend: native simnet
* any other relevant environment details:

### Steps to reproduce

* install native simnet and wait for 1 btc channel, 1 ltc channel, 2 raiden channels to be open. Take peer offline to get channel marked as inactive
* outcome:
```
kilrau@K-Yoga:~$ xucli channelbalance
13.4548 BTC
50000 DAI
1000.99875999 LTC
500 WETH
kilrau@K-Yoga:~$ lndbtc-lncli listchannels
{
    "channels": [
        {
            "active": false,
            "remote_pubkey": "036f9b004f047283b33a640bc4afdd26e2eb82e73938f8ac0796641ad20dc5bdd4",
            "channel_point": "fde8e01e6c2b3374003efbbefdd1d948994e8d63631f111d7192b717b008b7b9:1",
            "chan_id": "54641329854021633",
            "capacity": "2500000000",
            "local_balance": "1345480000",
            "remote_balance": "1154510950",
            "commit_fee": "9050",
            "commit_weight": "724",
            "fee_per_kw": "12500",
            "unsettled_balance": "0",
            "total_satoshis_sent": "0",
            "total_satoshis_received": "95480000",
            "num_updates": "18",
            "pending_htlcs": [
            ],
            "csv_delay": 1173,
            "private": false,
            "initiator": false,
            "chan_status_flags": "ChanStatusDefault"
        }
    ]
}
kilrau@K-Yoga:~$ xucli getinfo
{
  "version": "1.0.0-testnet.1",
  "nodePubKey": "032eb6b343e2e9685a2339516c2fbf8c31d0ad45e6d32344060b51ddd88ed7e62b",
  "urisList": [],
  "numPeers": 3,
  "numPairs": 4,
  "orders": {
    "peer": 1125,
    "own": 0
  },
  "lndMap": [
    [
      "BTC",
      {
        "error": "",
        "channels": {
[...]

PR #1317, opened 2019-11-03
fix(rpc/cli): separate inactive channel balance in getbalance

krrprr:
Closes https://github.com/ExchangeUnion/xud/issues/1130


lib/swaps/SwapClient.ts
```diff
@@ -217,6 +219,8 @@ abstract class SwapClient extends EventEmitter {
     },
   ): Promise<void>;
 
+  public abstract listChannels(activeOnly: boolean): Promise<any>;
```
sangaman: I don't think we actually need this as a public generic method, at least for now. If you drop this you can also remove the stub implementation for the raiden class.

lib/raidenclient/RaidenClient.ts
```diff
@@ -511,6 +511,10 @@ class RaidenClient extends SwapClient {
     return body.channel_address;
   }
 
+  public listChannels = (_activeOnly: boolean): Promise<Channel[]> => {
```
sangaman: FWIW, Raiden does have a method to [list channels](https://raiden-network.readthedocs.io/en/stable/rest_api.html#querying-information-about-channels-and-tokens), however I'm not seeing how to determine whether the partner for a channel is online & active or not, so we couldn't calculate inactive balance as far as I can see anyhow.

test/jest/Service.spec.ts
```diff
@@ -90,4 +93,94 @@ describe('Service', () => {
...
+      expect(btcBalance.inactiveChannelBalance).toEqual(18817);
+      expect(btcBalance.walletBalance).toEqual(10000);
+      expect(btcBalance.unconfirmedWalletBalance).toEqual(0);
+      expect(btcBalance.totalBalance).toEqual(289008);
+    });
+
+    test('throws in case of invalid currency', async () => {
+      setup();
+      try {
+        await service.getBalance({ currency: 'A' });
```
sangaman: As written I don't think this actually fails if `getBalance` doesn't throw. I believe you should use `await expect(...).rejects.toMatchSnapshot()` here and in the test case below.

krrprr: Indeed, good point!

PR #1320, opened 2019-11-06
fix(p2p): handshake race conditions

### LINKED ISSUE #1309: Intermittent failed Pool test
There's a test case that fails intermittently in the Pool jest test suite, giving the following error:

```
FAIL  test/jest/Pool.spec.ts
 ● Console

   console.log node_modules/winston/lib/winston/transports/console.js:79
     2019-10-31 11:23:38.3838 [P2P] warn: could not open connection to peer (127.0.0.1:59962): could not authenticate peer (027ba09e90b1085bc651d40ac98b18be27b2e3188d3beea2ff203ed24e014c9a79): invalid target (0388d4d4a1bb551c247bf1ea87fe19da84c5e1e03df3d315250fe43a7397a7aa06)
     
   console.log node_modules/winston/lib/winston/transports/console.js:79
     2019-10-31 11:23:38.3838 [P2P] warn: could not open connection to peer (02ef10d41a52343e4406181e21272cda5eb7ff3793c1e6ee5b15c41ba346f26632): Peer 02ef10d41a52343e4406181e21272cda5eb7ff3793c1e6ee5b15c41ba346f26632 was destroyed
     
   console.log node_modules/winston/lib/winston/transports/console.js:79
     2019-10-31 11:23:38.3838 [P2P] warn: could not open connection to peer (033fc92de8d75951076786720887570f1dee12cf32c4c8fb336bbdb360b3cbb15f): node 033fc92de8d75951076786720887570f1dee12cf32c4c8fb336bbdb360b3cbb15f at 127.0.0.1:57918 already connected
     
   console.log node_modules/winston/lib/winston/transports/console.js:79
     2019-10-31 11:23:38.3838 [P2P] warn: could not open connection to peer (02e249349b309ce6b2aea65a965d806c7ce54b04e59d5d5a74c481e1379cf48fad@localhost:33953): Peer 02e249349b309ce6b2aea65a965d806c7ce54b04e59d5d5a74c481e1379cf48fad@localhost:33953 disconnected from us due to AlreadyConnected
     

 ● P2P Pool › it rejects when already connected

   expect(received).toEqual(expected) // deep equality

   Expected: 1
   Received: 0

     184 |       );
     185 |     }
   > 186 |     expect(pool['peers'].size).toEqual(1);
         |                                ^
     187 |     expect(pool2['peers'].size).toEqual(1);
     188 |   });
     189 | 

     at Pool.spec.ts:186:32
         at Generator.throw (<anonymous>)
     at rejected (Pool.spec.ts:6:65)
```

sangaman:
Closes #1309.

This commit makes several changes to address potential race conditions in the case where two peers attempt connections to each other simultaneously.

The key fix is that the p2p pool no longer prunes recently closed peers from the list of active peers by node pub key alone. This could result in scenarios where a pending connection that is closed causes an active connection to be discarded. Instead, only peers that are active get removed from the list when they are closed. Additionally, pending outbound and inbound peers are no longer removed from their respective lists upon the peers being closed. Rather, the `addOutbound` and `addInbound` methods - which currently add pending peers to their lists - now also remove peers from the list if they are no longer pending.

The `openPeer` method now resolves immediately after the target peer becomes active rather than waiting for followup tasks such as updating the database and sending a `GetNodes` packet. These tasks are now called and awaited separately.

Another race condition is prevented whereby handshakes may be completed on two simultaneous socket connections without being detected in the `validatePeer` method that is called after a handshake begins and the node pub key for the inbound peer is established but before it completes. In this case, both peers cannot close the "duplicate" connection or else they may wind up closing both of them. The approach implemented here arbitrarily picks the peer with the higher node pubkey to close the socket when it encounters a duplicate, while the other peer waits briefly to see if the already established connection is closed.

Lastly there are some fixes/enhancements to the Pool tests, ensuring that we wait for inbound connections to complete before we check them.

### DIFF
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -1,9 +1,9 @@
 import { EventEmitter } from 'events';
-import P2PRepository from './P2PRepository';
-import { NodeInstance, NodeFactory, ReputationEventInstance } from '../db/types';
-import { Address } from './types';
-import addressUtils from '../utils/addressUtils';
 import { ReputationEvent } from '../constants/enums';
+import { NodeFactory, NodeInstance, ReputationEventInstance } from '../db/types';
+import addressUtils from '../utils/addressUtils';
+import P2PRepository from './P2PRepository';
+import { Address } from './types';
 
 export const reputationEventWeight = {
   [ReputationEvent.ManualBan]: Number.NEGATIVE_INFINITY,
@@ -95,12 +95,13 @@ class NodeList extends EventEmitter {
   }
 
   /**
-   * Create a Node in the database.
+   * Persists a node to the database and adds it to the node list.
    */
-  public createNode = async (nodeFactory: NodeFactory): Promise<NodeInstance> => {
-    const node = await this.repository.addNode(nodeFactory);
-    this.nodes.set(node.nodePubKey, node);
-    return node;
+  public createNode = async (nodeFactory: NodeFactory) => {
+    const node = await this.repository.addNodeIfNotExists(nodeFactory);
+    if (node) {
+      this.nodes.set(node.nodePubKey, node);
+    }
   }
 
   /**
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -1,23 +1,23 @@
+import assert from 'assert';
+import { EventEmitter } from 'events';
 import net, { Server, Socket } from 'net';
 import semver from 'semver';
-import { EventEmitter } from 'events';
-import errors, { errorCodes } from './errors';
-import Peer, { PeerInfo } from './Peer';
-import NodeList, { reputationEventWeight } from './NodeList';
-import P2PRepository from './P2PRepository';
-import * as packets from './packets/types';
-import { Packet, PacketType } from './packets';
-import { OutgoingOrder, OrderPortion, IncomingOrder } from '../orderbook/types';
+import { DisconnectionReason, ReputationEvent, XuNetwork } from '../constants/enums';
 import { Models } from '../db/DB';
+import { ReputationEventInstance } from '../db/types';
 import Logger from '../Logger';
-import { NodeState, Address, NodeConnectionInfo, PoolConfig } from './types';
+import NodeKey from '../nodekey/NodeKey';
+import { IncomingOrder, OrderPortion, OutgoingOrder } from '../orderbook/types';
 import addressUtils from '../utils/addressUtils';
 import { getExternalIp } from '../utils/utils';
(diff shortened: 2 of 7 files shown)


lib/p2p/NodeList.ts
```diff
@@ -95,12 +95,17 @@ class NodeList extends EventEmitter {
...
-  public createNode = async (nodeFactory: NodeFactory): Promise<NodeInstance> => {
-    const node = await this.repository.addNode(nodeFactory);
-    this.nodes.set(node.nodePubKey, node);
-    return node;
+  public createNode = async (nodeFactory: NodeFactory) => {
+    try {
+      const node = await this.repository.addNode(nodeFactory);
+      this.nodes.set(node.nodePubKey, node);
+    } catch (err) {
+      if (err.name !== 'SequelizeUniqueConstraintError') {
```
erkarl: It feels weird that we're expecting this to occasionally error :thinking:. Don't have a better solution/recommendation - just pointing this out.

lib/p2p/Pool.ts
```diff
@@ -501,11 +507,44 @@ class Pool extends EventEmitter {
...
+      if (this.nodePubKey > peerPubKey) {
+        await peer.close(DisconnectionReason.AlreadyConnected);
+        throw errors.NODE_ALREADY_CONNECTED(peerPubKey, peer.address);
+      } else {
+        const stillConnected = await new Promise<boolean>((resolve) => {
+          this.peers.get(peerPubKey)!.once('close', () => {
+            resolve(false);
+            clearTimeout(timeout);
+          });
+          const timeout = setTimeout(() => resolve(true), Peer.STALL_INTERVAL);
```
erkarl: Can we move the timeout declaration above? Reading from top down I got confused because `clearTimeout(timeout)`, but didn't see it declared.

lib/p2p/NodeList.ts
sangaman: You're right it is weird, I think it really only comes up in tests where p2p latency is ~zero. But the end result is that we have the node in the database and in the nodelist which is the desired state, and catching the uniqueness error is simpler than tracking a list of pending nodes that are currently being written to the db. I will explain here with a comment.

lib/p2p/Pool.ts
sangaman: No problem.

lib/p2p/NodeList.ts
sangaman: Actually, I'm going to go ahead and move this code relating to sequelize and the database over to the `P2PRepository` class since that's a more appropriate place for it.

PR #1327, opened 2019-11-09
feat(rpc): add only_own_orders parameter for listorders

### LINKED ISSUE #1323: `only_own_orders` parameter for listorders
By @offerm :
> there is no easy way to extract from xud only my own orders
I'd suggest an `only_own_orders` parameter for https://api.exchangeunion.com/#listorders since we already have `include_own_orders`

Goes without saying: all parameters should be available as `xucli` parameters.

Something for you @krrprr ?

krrprr:
This commit adds an optional `only_own_orders` parameter for the `listorders` command.
The default value for this is `false`.

The commit also fixes a bug that caused a situation where sell orders were displayed instead of buy orders when the limit was set.

Closes #1323

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -492,6 +492,7 @@
 | ----- | ---- | ----- | ----------- |
 | pair_id | [string](#string) |  | The trading pair for which to retrieve orders. |
 | include_own_orders | [bool](#bool) |  | Whether own orders should be included in result or not. |
+| only_own_orders | [bool](#bool) |  | Whether to return only own orders. |
 | limit | [uint32](#uint32) |  | The maximum number of orders to return from each side of the order book. |
 
 
--- a/lib/cli/commands/listorders.ts
+++ b/lib/cli/commands/listorders.ts
@@ -78,7 +78,7 @@ const displayTables = (orders: ListOrdersResponse.AsObject) => {
   formatOrders(orders).forEach(displayOrdersTable);
 };
 
-export const command = 'listorders [pair_id] [include_own_orders] [limit]';
+export const command = 'listorders [pair_id] [include_own_orders] [only_own_orders] [limit]';
 
 export const describe = 'list orders from the order book';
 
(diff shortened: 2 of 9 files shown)


kilrau:
> The commit also fixes a bug that caused a situation where sell orders were displayed instead of buy orders when the limit was set.

Good job catching this!

kilrau:
> Looks good functionally, but I find it a little awkward to have both a flag to include own orders as well as a flag to include _only_ own orders. How about creating an enum `owner` field with values `BOTH`, `OWN`, and `PEER` where `BOTH` equals 0 and is therefore the default? That could replace both fields. On the CLI layer we could take values of `own` or `peer` for an optional `owner` parameter.

Gotcha and agree it not being super nice wording, but wouldn't a simple renaming ` [include_own_orders] & [only_own_orders]` -> `[own_and_peer]` & `[own]` do?

sangaman:
> Gotcha and agree it not being super nice wording, but wouldn't a simple renaming ` [include_own_orders] & [only_own_orders]` -> `[own_and_peer]` & `[own]` do?

What if we want only peer orders though? I suppose we could have separate `own` and `peer` flags, but that seems like a more complicated approach to me.

kilrau:
> I suppose we could have separate own and peer flags, but that seems like a more complicated approach to me.

That's what you proposed [above](https://github.com/ExchangeUnion/xud/pull/1327#pullrequestreview-321078232), and honestly, both fine. @krrprr please decide for one and let's get this closed.

sangaman:
> That's what you proposed above

I was actually thinking something slightly different. My preference stated originally is something like `xucli listorders --owner own` where `own` could be `peer` or `both` as well. Another way to do it would be `xucli listorders --own` or `xucli listorders --peer` where `both` is used as the default if neither is specified.

kilrau:
> I was actually thinking something slightly different. My preference stated originally is something like `xucli listorders --owner own` where `own` could be `peer` or `both` as well. Another way to do it would be `xucli listorders --own` or `xucli listorders --peer` where `both` is used as the default if neither is specified.

Both fine! Let's leave it up to @krrprr .

krrprr:
Sorry for leaving this alone for so long time. As I wasn't able to push to xud repo anymore, I created a new pull request: https://github.com/ExchangeUnion/xud/pull/1392

Issue #1313, opened 2019-10-31
Replace Order


kilrau:
Traditional exchanges provide "cancel-replace" order or "order-cancel-order" (OCO) functionality, we should provide something similar. Reasons:
* market makers continuously need to replace orders since the price on the exchanges they are connected to is continuously moving. As of now this requires two API calls:  https://api.exchangeunion.com/#removeorder & https://api.exchangeunion.com/#placeorder. This is not what market makers are used to.
* Also, as of now, two packets need to be broadcasted to the network: invalidating the old order and broadcasting the new order. Since time is of the essence and prices can move fast, we'd want to reduce this to one `replace_order` packet only.

Obviously, we need to take great care for the sender to sign and recipient to verify signature, so that a node can only replace its own orders (compare https://github.com/ExchangeUnion/xud/issues/242).

kilrau:
Implementation-wise, @reliveyy recommends to **not** change order price or quantity, but to actually replace it with a new order (and new order id). Simpliefies xud's matching engine since it can assume an order doesn't change.

LePremierHomme:
> Obviously, we need to take great care for the sender to sign and recipient to verify signature, so that a node can only replace its own orders (compare #242).

As discussed in #242, signing isn't required here, since authentication is done during the handshake, and communication is limited to the secured socket. Correct?

> Implementation-wise, @reliveyy recommends to not change order price or quantity, but to actually replace it with a new order (and new order id). Simpliefies xud's matching engine since it can assume an order doesn't change.

Sounds like a good suggestion. So `replace_order` would be similar to `order`, only that it will carry an additional field for the order to cancel?

kilrau:
> As discussed in #242, signing isn't required here, since authentication is done during the handshake, and communication is limited to the secured socket. Correct?

As of now correct. With upcoming p2p protocol changes this will be required since this packet will not be limited to the secure socket anymore. But I guess it makes sense to wait until we get there. So I agree, not need in the course of this PR.

> Sounds like a good suggestion. So replace_order would be similar to order, only that it will carry an additional field for the order to cancel?

an additional field for another order to cancel yeah. That what you had in mind? @reliveyy Other opinions? @sangaman @erkarl @offerm

LePremierHomme:
It looks like that creating a new API call `ReplaceOrder` which will combine both `RemoveOrder` and `PlaceOrder` is somehow straightforward, but making them atomic in the P2P level isn’t.

Steps:
[...]

PR #1333, opened 2019-11-17
PlaceOrder to support replacing a previous order

LePremierHomme:
Closing #1313.

`replace_order_id` was added to `PlaceOrderRequest`, to allow an optional removal of a previous order, before placing the new one. 

This was the simplest possible implementation. If this approach is accepted, I can add more tests.


kilrau:
Restarted build, please keep an eye on why tests are failing

lib/orderbook/OrderBook.ts
```diff
@@ -620,11 +620,13 @@ class OrderBook extends EventEmitter {
   /**
    * Removes all or part of an order from the order book by its local id. Throws an error if the
    * specified pairId is not supported or if the order to cancel could not be found.
+   * @param allowEventualRemoval whether to allow an eventual async removal of the order in case
+   * some quantity of the order is on hold and cannot not be immediately removed.
```
sangaman: remove extra "not"

lib/orderbook/OrderBook.ts
```diff
@@ -620,11 +620,13 @@ class OrderBook extends EventEmitter {
   /**
    * Removes all or part of an order from the order book by its local id. Throws an error if the
    * specified pairId is not supported or if the order to cancel could not be found.
+   * @param allowEventualRemoval whether to allow an eventual async removal of the order in case
```
sangaman: Maybe specify here that an error is thrown if this is `false` but the order is on hold. Also it may be clearer to say `allowAsyncRemoval`.

proto/xudrpc.proto
```diff
@@ -592,6 +592,8 @@ message PlaceOrderRequest {
   string order_id = 4 [json_name = "order_id"];
   // Whether the order is a buy or sell.
   OrderSide side = 5 [json_name = "side"];
+  // The local id of a previous order to be replaced (optional).
```
sangaman: I think the "optional" here is implied. And maybe say "existing" order instead of previous?

On second thought, I think even better would be to clarify here that *if* `replace_order_id` is provided, the order must be successfully found and removed before the new order is placed otherwise an error is returned.

LePremierHomme:
Thanks for the good feedback.

LePremierHomme:
Looks like the simulation test is failing on `installing raiden-contracts`. @erkarl are you familiar with this?

sangaman:
Try rebasing on master, it might help with a recent PR.

PR #1334, opened 2019-11-21
feat(rpc): RestoreNode call to restore from seed

### LINKED ISSUE #1017: Allow user to provide own seed
Continuation of https://github.com/ExchangeUnion/xud/issues/912:

[Allow user to provide own seed when creating xud node](https://github.com/ExchangeUnion/xud/pull/1006#issuecomment-499826994)

### LINKED ISSUE #1020: [XUD Master Password] Recover xud key from mnemonic
Continuation of #912:

[Recover xud key from mneominc (extend seedutil functionality?)](https://github.com/ExchangeUnion/xud/pull/1006#issuecomment-499826994)

sangaman:
Closes #1017. Closes #1020.

This adds a new `RestoreNode` call that accepts a 24 word seed mnemonic
to restore an xud key, lnd wallets, and a raiden keystore from the seed.
This call can be extended to also restore lnd channel backups and xud
db backups with additional parameters.

The seedutil tool is extended to support an `encipher` subcommand that
converts a mnemonic to the enciphered seed bytes, which is necessary
for deriving an xud key from aezeed.

Sample output from the cli below:

```
./bin/xucli restore

You are restoring an xud node key and underlying wallets. All will be secured by
a single password provided below.
  
Enter your 24 word mnemonic: abstract swear air swamp carpet that retire pool produce food join inform giraffe local region anchor march advice blanket quick farm mandate shell lens
Enter a password: 
Re-enter password: 

The following wallets were restored: BTC
```


lib/service/InitService.ts
```diff
@@ -85,6 +86,54 @@ class InitService extends EventEmitter {
...
+    }
+    if (this.nodeKeyExists) {
+      throw errors.UNIMPLEMENTED;
+    }
+    if (this.swapClientManager.misconfiguredClientLabels.length > 0) {
+      throw swapErrors.SWAP_CLIENT_MISCONFIGURED(this.swapClientManager.misconfiguredClientLabels);
+    }
+    if (this.pendingCall) {
+      throw errors.PENDING_CALL_CONFLICT;
+    }
```
michael1011: These checks are the same as the ones of the `createNode` method (apart from the the one that checks the length of the mnemonic). Maybe put them in a private method to ensure that the checks and the error messages are consistent between the `create` and `restore` call.

lib/utils/seedutil.ts
```diff
@@ -5,14 +5,14 @@ import { promisify } from 'util';
...
- * according to a given mnemonic and password at the specified path
+ * Executes the seedutil tool to generate an ethereum keystore from the given
+ * mnemonic and password at the specified path.
  * @param mnemonic the 24 seed recovery mnemonic
  * @param password the password to protect the keystore
  * @param path the path in which to create the keystore directory
  */
-const seedutil = async (mnemonic: string[], password: string, path: string) => {
-  const { stdout, stderr } = await exec(`./seedutil/seedutil -pass=${password} -path=${path} ${mnemonic.join(' ')}`);
+async function keystore(mnemonic: string[], password: string, path: string) {
```
michael1011: Why aren't we using arrow functions in files like this one again? I think Moshe told me a while ago but I forgot already :pensive:

lib/grpc/GrpcInitService.ts
```diff
@@ -74,6 +74,28 @@ class GrpcInitService {
...
+    if (!this.isReady(this.initService, callback)) {
+      return;
+    }
+    try {
+      const { restoredLndWallets, restoredRaiden } = await this.initService.restoreNode(call.request.toObject());
+      const response = new xudrpc.RestoreNodeResponse();
+      if (restoredLndWallets) {
+        response.setRestoredLndsList(restoredLndWallets);
+      }
+      response.setRestoredRaiden(restoredRaiden);
```
erkarl: Should there also be a conditional to prevent setting `restoredRaiden` when it's undefined (analogically to `restoredLndWallets`?

seedutil/README.md
```diff
@@ -10,10 +10,18 @@ This utility is used to derive an Ethereum keystore file from [aezeed](https://g
 
 It is recommended to use this tool on the command line ONLY for development purposes.
 
-`seedutil [-pass=encryption password] [-path=optional/keystore/path] [-aezeedpass=optional_seed_pass] <twenty four recovery words separated by spaces>`
+### Generate Etherem Keystore
```
erkarl:
```suggestion
### Generate Ethereum Keystore
```

lib/service/InitService.ts
sangaman: I had thought about this but wasn't sure how to do it best given the three methods all make a slightly different set of checks, but I'll give this another pass.

lib/utils/seedutil.ts
sangaman: I think the main difference is that it's not a member method of a class, where usage of `this` can complicate things. In this case it's just a plain function that makes no use of `this`, whereas within class methods we frequently use `this` and always want it to refer to an instance of that class. @LePremierHomme feel free to chime in if you have any thoughts.

lib/grpc/GrpcInitService.ts
sangaman: Actually in double checking this code, even the first check shouldn't be necessary, as these values will never be `undefined`. I also simplified the code a bit in InitService.ts to remove some unnecessary `let` statements.

sangaman:
I made changes according to the feedback. Another thing I refactored on my own was to set the `pendingCall` value back to false within the `InitService` class itself via a `finally` block in each method. It felt a bit clumsy doing it in a separate class.

lib/service/InitService.ts
```diff
@@ -13,65 +14,109 @@ interface InitService {
...
-      initializedLndWallets,
-      initializedRaiden,
-      mnemonic: seed ? seed.cipherSeedMnemonicList : undefined,
-    };
   }
 
-  public unlockNode = async (args: { password: string }) => {
-    const { password } = args;
+  private prepareCall = async () => {
     if (!this.nodeKeyExists) {
```
erkarl: This breaks `xucli create` because it was previously:
```
    if (this.nodeKeyExists) {
      throw errors.UNIMPLEMENTED;
    }
```

sangaman: Ah sorry thanks for catching this, I overlooked that unlock and create/restore have opposite conditions for this check when I was refactoring. Fixing now.

kilrau:
One more manual test by @erkarl

lib/cli/commands/restore.ts
```diff
@@ -0,0 +1,86 @@
...
+    input: process.stdin,
+    output: process.stdout,
+    terminal: true,
+  });
+
+  console.log(`
+You are restoring an xud node key and underlying wallets. All will be secured by
+a single password provided below.
+  `);
+  rl.question('Enter your 24 word mnemonic: ', (mnemonicStr) => {
```
erkarl: Should we clarify here that the words need to be separated by space?

erkarl:
Also needs a rebase before merging.

lib/cli/commands/restore.ts
sangaman: Not a bad idea, I'll do that.

Issue #1331, opened 2019-11-12
TradingLimits (GetBalance enhancements)


kilrau:
Off-shoot from https://github.com/ExchangeUnion/xud/issues/1329 by @offerm :

> I need a way to know how much I can trade. for this I need the max out going and incoming channel for each asset. The current information can't help with this. Incoming is not provided at all and out going (channel balance ) my be over multiple channels so it can't be used in a single swap.

> #2 - what I need is a reliable way to ensure that an order I'm about to issue can be placed.
So, if I'm about to place a buy 1 BTC/DIA 8800 order, I need to know that I can send 8800 DAI and can receive 1 BTC. Ideally it should take into account my other orders of that pair in that direction.
For example, if I already sent buy 1 BTC/DIA 8800 and I'm now sending `buy 2 BTC/DAI 8600' I want to ensure that I can send 28600+18800 and receive 3 BTC.

erkarl:
What Offer said makes a lot of sense and I came across the same concerns while implementing overview of balances for Arby. I'd like to have output for the maximum amounts I can send/receive via a single channel.

Considering active orders would be a nice to have, but not required at this point.

kilrau:
Implementation idea: Loop through channels and find max local balance and max remote balance

sangaman:
As discussed today, I think it makes more sense to return these limits in a separate call from `getbalance`. Maybe something like `channellimits` or `tradinglimits`?

We also should take `local_chan_reserve_sat` and `remote_chan_reserve_sat` into account by subtracting them from the available balance.

We can also assign the outbound max to the `maximumOutboundAmount` property on each `LndClient` (and perhaps create a `maximumInboundAmount` field with corresponding logic) but that can be done separately.

erkarl:
Sounds good - I'd prefer `tradinglimits` for the name.

offerm:
Need both in and out.
Consider to provide the information by pairs
Best would be to provide the Max sell and Max buy per pair.

kilrau:
> We also should take local_chan_reserve_sat and remote_chan_reserve_sat into account by subtracting them from the available balance.
:+1: 
> Sounds good - I'd prefer `tradinglimits` for the name.
:+1: 
> We can also assign the outbound max to the maximumOutboundAmount property on each LndClient (and perhaps create a maximumInboundAmount field with corresponding logic) but that can be done separately.
I think the below is better (assign limits to pair, not swap client). Let's also not forget about Raiden(!)
[...]

PR #1341, opened 2019-11-28
feat(rpc): TradingLimits API call

LePremierHomme:
Closing #1331.

* haven't manually tested Raiden currencies yet
* can I rename the existing `maximumOutboundCapacity` to `totalOutboundCapacity` throughout the code?  @sangaman 
* the max channel outbound/inbound capacities are cached, but they are not used. This is why I haven't added it to the periodic `updateCapacity()` calls. It should be added there only if we'll use it for some validation (like with the existing `maximumOutboundCapacity`), and not just as response to API calls (which updates it anyway).


kilrau:
> It should be added there only if we'll use it for some validation (like with the existing maximumOutboundCapacity), and not just as response to API calls (which updates it anyway).

I'd say it's fine for now and we only add these periodic calls if we have usage for the cached value.

lib/lndclient/LndClient.ts
```diff
@@ -51,6 +51,9 @@ class LndClient extends SwapClient {
   private channelBackupSubscription?: ClientReadableStream<lndrpc.ChanBackupSnapshot>;
   private invoiceSubscriptions = new Map<string, ClientReadableStream<lndrpc.Invoice>>();
   private maximumOutboundAmount = 0;
+  private maximumChannelOutboundAmount = 0;
```
sangaman: I think it would make sense to call thse `maxChannelOutboundAmount` and `maxChannelInboundAmount` and then rename the existing `maximumOutboundAmount` to `channelBalance` or `totalBalance`, which is a more accurate description I think.

lib/lndclient/LndClient.ts
```diff
@@ -177,6 +180,14 @@ class LndClient extends SwapClient {
     return this.maximumOutboundAmount;
   }
 
+  public maximumChannelOutboundCapacity = () => {
```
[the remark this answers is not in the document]
sangaman: See what you think of my naming suggestions above, I think it's a little bit less verbose/more direct.

I do like the suggestion of making the private field start with an underscore and then using the same name minus the underscore for the getter, it's a bit clearer to me than alternating usage of "amount" and "capacity" (which don't necessarily mean the same thing.

sangaman: This can be a true getter instead of a method with `public get maxOutboundAmount() {...` (assuming we use my suggested naming). I think it's a bit neater/more consistent that way.

lib/lndclient/LndClient.ts
```diff
@@ -177,6 +180,14 @@ class LndClient extends SwapClient {
...
+    return this.maximumChannelOutboundAmount;
+  }
+
+  public maximumChannelInboundCapacity = () => {
```
sangaman: This can also be made into a real getter.

lib/raidenclient/RaidenClient.ts
```diff
@@ -472,6 +482,47 @@ class RaidenClient extends SwapClient {
...
+      if (maxOutbound < outbound) {
+        maxOutbound = outbound;
+      }
+
+      const inbound = channel.total_deposit - channel.balance;
+      if (maxInbound < inbound) {
+        maxInbound = inbound;
+      }
+    });
+
```
sangaman: I think an important step here would be to convert the max channel balances we've fetched from the base "units" for each token, which is 10^-18 for the ERC20 tokens we're using so far, to the "amounts" that we use throughout the codebase for orders which are represented with satoshis/10^-8.

lib/service/Service.ts
```diff
@@ -181,6 +182,36 @@ class Service {
...
+  public tradingLimits = async (args: { currency: string }) => {
+    const { currency } = args;
+    const tradingLimits = new Map<string, TradingLimits>();
+
+    if (currency) {
+      argChecks.VALID_CURRENCY(args);
+
+      const swapClient = this.swapClientManager.get(currency.toUpperCase());
+      if (swapClient) {
+        const tl = await swapClient.tradingLimits(currency);
```
sangaman: I'd prefer `tradingLimit` to `tl` if that's alright with you, just a bit clearer that way. Same for where it's used a few lines below.

kilrau:
> The only meaningful change I see would be to convert from the base units in raiden to satoshi units. Some tests as Karl suggested would be good too.

This is a must change, all others are "beautifying" @LePremierHomme

LePremierHomme:
#### Renaming

* I ended up not using getters instead of getters-like functions, due to incompatibility with the inherited types from `SwapClient` abstract class.
* I renamed `maximumOutboundCapacity` as `totalOutboundAmount`. Using `channelBalance` wasn't applicable because we are already using the name for the fetcher method (following LND api). 
* We have many usages of "Capacity" throughout the code and documentation, but I wasn't sure if we want to abolish all of them or not. I think using "Amount" for the low-level representation, while using "Capacity" for the application-level usages (including logs) is fine. If you want me to unify the entire codebase/docs usages of "Capacity" to "Amount", let me know.

Issue #1264, opened 2019-10-02
Resolve unknown hashes from database


erkarl:
Offshoot from: https://github.com/ExchangeUnion/xud/pull/1262#discussion_r330678588

kilrau:
could you explain @erkarl

erkarl:
I think the comment above explains it pretty clearly. Xud holds the preimage in memory and the swap client requests it, but we don't actually wait for confirmation that the swap client has processed it. If the swap client requests the preimage again xud will respond with unknown preimage because it's no longer in the memory (but we still have it in the database).

The solution would be to also query the database for the preimage if we cannot find it in memory.

PR #1342, opened 2019-11-28
feat(swaps): Resolve unknown hashes from database

LePremierHomme:
Closing https://github.com/ExchangeUnion/xud/issues/1264.

I'm not sure if this is the desired behaviour - please review.

### DIFF
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -1,4 +1,4 @@
-import { SwapPhase, SwapRole, SwapState, SwapFailureReason, ReputationEvent } from '../constants/enums';
+import { ReputationEvent, SwapFailureReason, SwapPhase, SwapRole, SwapState } from '../constants/enums';
 import Peer from '../p2p/Peer';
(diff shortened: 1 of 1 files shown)


lib/swaps/Swaps.ts
```diff
@@ -926,6 +926,10 @@ class Swaps extends EventEmitter {
     } else if (this.getPendingSwapHashes().includes(rHash)) {
       throw errors.PAYMENT_PENDING(rHash);
     } else {
+      const dealInstance = await this.repository.getSwapDeal(rHash);
+      if (dealInstance && dealInstance.rPreimage && (dealInstance.state === SwapState.Completed || dealInstance.state === SwapState.Recovered)) {
```
sangaman: I think it's probably safer (and removes some unnecessary checks) to return the preimage as long as we have it, regardless of the swapstate. I think one of the objectives is to return the preimage to as many requests as possible to ensure we can settle any inbound payments, even if there's some data inconsistency in the database otherwise.

PR #1344, opened 2019-12-03
feat(orderbook): immediate-or-cancel orders

### LINKED ISSUE #622: Limit Order (IOC)

sangaman:
Closes #622.

This adds an "immediate-or-cancel" order type that prevents an order from entering the order book after matching is complete. It combines the immediate nature of a market order with the price requirement of a limit order.


kilrau:
Review by @erkarl

lib/cli/placeorder.ts
```diff
@@ -23,12 +23,19 @@ export const placeOrderBuilder = (argv: Argv, side: OrderSide) => {
...
     describe: 'whether to execute in streaming mode',
+    alias: 's',
     default: false,
   })
   .option('replace_order_id', {
     type: 'string',
+    alias: 'r',
     describe: 'the local order id of a previous order to be replaced',
   })
+  .option('ioc', {
```
erkarl: We're defining the ioc option here, but this value is never passed to the `PlaceOrderRequest`.

```
➜  xud git:(ioc) ✗ ./bin/xucli buy 0.0001 LTC/BTC 1 --ioc
no matches found
remaining 0.0001 LTC entered the order book as fd3184a0-1ce9-11ea-a368-ebe1579e30d5
➜  xud git:(ioc) ✗ ./bin/xucli orderbook

Trading pair: LTC/BTC
│ Quantity          │ Price             │ Price             │ Quantity          │
│ 0.0001            │ 1                 │                   │                   │
└───────────────────┴───────────────────┴───────────────────┴───────────────────┘
```

sangaman: Good catch, fixed this.

erkarl: Thanks. The functionality seems to work now (the order is not entered to the orderbook), but the output from the CLI is misleading:

```
➜  xud git:(ioc) ✗ ./bin/xucli buy 0.0001 LTC/BTC 1 --ioc
no matches found
remaining 0.0001 LTC entered the order book as e7892df0-1cef-11ea-9026-6359673420ce
➜  xud git:(ioc) ✗ ./bin/xucli orderbook

Trading pair: LTC/BTC
│ Quantity          │ Price             │ Price             │ Quantity          │
└───────────────────┴───────────────────┴───────────────────┴───────────────────┘
```

Should we also remove the `remaining X XYZ entered the order book...` for `ioc` orders?

kilrau: YES WE SHOULD! :+1:

sangaman: Ahhh good catch, I totally didn't think about how that was misleading. I'll update the cli logic not to display this for `ioc` orders.

sangaman: This should be fixed now as well.

PR #1345, opened 2019-12-04
chore: use go 1.13

sangaman:
This updates the go modules and instructions for the simulation tests and seedutil tool for the latest golang version, 1.13. This prepares for use with newer versions of lnd which require a minimum of 1.13. Go 1.12 is still supported and can be used to compile seedutil and run simulation tests with this commit.


kilrau:
`The command "npm run test:sim" exited with 1.`

Restarted the build.

test/simulation/go.mod
```diff
@@ -2,28 +2,33 @@ module github.com/ExchangeUnion/xud-simulation
...
+	github.com/btcsuite/go-flags v0.0.0-20150116065318-6c288d648c1c // indirect
+	github.com/codahale/chacha20 v0.0.0-20151107025005-ec07b4f69a3f // indirect
+	github.com/codahale/chacha20poly1305 v0.0.0-20151127064032-f8a5c4830182 // indirect
 	github.com/go-errors/errors v1.0.1
 	github.com/golang/protobuf v1.3.2
 	github.com/google/go-cmp v0.3.0 // indirect
+	github.com/howeyc/gopass v0.0.0-20190910152052-7cb4b85ec19c // indirect
 	github.com/kkdai/bstream v1.0.0 // indirect
-	github.com/lightningnetwork/lnd v0.6.1-beta
+	github.com/lightningnetwork/lnd v0.8.1-beta
```
erkarl: I'm wondering about this diff here because the `LND_TAG` is set to `v0.7.1-beta` in https://github.com/ExchangeUnion/xud/blob/go1.13/test/simulation/.env#L17

sangaman: I think it's just the latest version that it grabbed, I don't think it conflicts with the fact that we're running 0.7.1 for the sim tests. Also I'm thinking of upgrading to 0.8.x soon anyhow.

PR #1347, opened 2019-12-05
 feat: disable raiden currencies w/out direct chan 

### LINKED ISSUE #1027: Direct Channel Checks (Updated)

sangaman:
Closes #1027.

This PR attempts to ensure that we have a direct channel with peers for raiden currencies before exchanging orders with that peer for any pairs involving that currency.

When a peer advertises a trading pair, either via initial handshake or on a node state update, we check that we have a direct channel for any raiden currencies before activating advertised pairs. If the trading pair is not activated, then we do not send or accept orders for that pair with the peer.

An existing timer with an interval of one minute is used for each peer to periodically check if inactive currencies and pairs can be activated by performing the same checks described above, including the direct channel check for raiden currencies.

It also reworks the error handling logic for swaps to identify a particular currency that caused the error when applicable. This allows us to deactivate that currency, such that if a swap fails due to `NoRouteFound` (meaning no direct channel) with a raiden currency we can disable that currency right away.

Some examples of how this logic may work:

1. Handshake with peer advertises BTC/WETH
2. A direct channel is not found for WETH
3. BTC/WETH is disabled, no orders are broadcasted to the peer or accepted from that peer to the order book.
4. Later, a direct channel for WETH is opened and the periodic check detects it.
5. BTC/WETH is enabled, a `GetOrders` packet is sent for BTC/WETH, and new orders are transmitted as they arrive.

or

1. Handshake with peer advertises BTC/WETH.
2. A direct channel is found for WETH
3. A `GetOrders` packet is sent for BTC/WETH, and new orders are transmitted as they arrive.
4. A WETH payment fails due to `NoRouteFound` (meaning no direct channel).
5. BTC/WETH is disabled for that peer, all existing orders from that peer are removed from the order book and new orders are no longer broadcast or accepted.

### DIFF
--- a/lib/constants/enums.ts
+++ b/lib/constants/enums.ts
@@ -135,6 +135,8 @@ export enum SwapFailureReason {
   RemoteError = 13,
   /** The swap failed because of a system or xud crash while the swap was being executed. */
   Crash = 14,
+  /** A swap was attempted between an invalid matching of orders */
+  InvalidOrders = 15,
 }
 
 export enum DisconnectionReason {
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -132,6 +132,7 @@ class OrderBook extends EventEmitter {
     const { minQuantity } = this.thresholds;
     return order.quantity >= minQuantity;
   }
+
   /**
    * Checks that a currency advertised by a peer is known to us, has a swap client identifier,
    * and that their token identifier matches ours.
@@ -583,7 +584,8 @@ class OrderBook extends EventEmitter {
 
     this.emit('ownOrder.added', order);
 
-    this.broadcastOrder(order);
+    const outgoingOrder = OrderBook.createOutgoingOrder(order);
+    this.pool.broadcastOrder(outgoingOrder);
     return true;
   }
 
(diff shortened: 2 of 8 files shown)


kilrau:
Recap: the original reason for the raiden direct channel check is we simply cannot guarantee the first routing attempt of raiden to succeed. If it needs a second, the swap will fail (because the preimage changes).

Our goal for the coming months should be a close-to-100% swap success rate for anyone using the "default setup". It's ok to make some temporary compromises on decentralization for that (compare phoenix).

Thoughts:
- maintaining (open, continuously rebalance) direct raiden channels for more than 2-3 peers and more than 2-3 pairs is unrealistic; even for a market maker
- we have a list of bootstrap nodes, Requirement currently is only a 24/7 uptime to share other nodes and make them find each other.
- I think we can built on that, adding the requirement for a bootstrap node to also have fat, direct, continously rebalanced channels with all other bootstrap nodes
- additionally only bootstrap nodes should be offering an "open channel service" to xud users
- the default setup gets channels from/to one of the bootstrap nodes when the user deposits something into her xud node

This would result in "as long as I have good channels with one of the bootstrap nodes, I can swap reliably" scenario.

The maximum hops are 3. UserA -> bootstrapA -> boostrapB -> UserB. For the time being we only have one bootstrap node, which simplifies the scenario to UserA -> xud1.exchangeunion.com -> UserB.

Summary:
- checks shouldn't be `direct channel for raiden`, but `direct channel for <currency> OR direct channel with minimum one bootstrap node`
- depends on open channel service to be available

erkarl:
Not considering @kilrau's comment above code looks good. Haven't tested this manually, yet.

sangaman:
> Recap: the original reason for the raiden direct channel check is we simply cannot guarantee the first routing attempt of raiden to succeed. If it needs a second, the swap will fail (because the preimage changes).
> 
> But considering #1361, we could achieve a similar high routing success on first attempt. Meaning:
> 
>     * checks shouldn't be `direct channel for raiden`, but `direct channel for <currency> OR direct channel with minimum one bootstrap node`
> 
>     * obviously depends on open channel service to be available

This makes sense, but in revisting this PR I think we could better control for this through tweaking the existing `directChannelChecks` config option logic which currently defaults to true. Even with the changes in this PR, setting that option to `false` will not disable raiden currencies that don't have direct channels. I have two approaches in mind:

1. Have the user manually set this config option to fault if they intend to use a routing service for raiden.

2. Get rid of the config option and treat it as if it's always `true` - but then internally count a channel with a hardcoded routing service node as satisfying the direct channel requirement (so for any given currency/peer, we need either a channel with that peer or a channel with a routing service for the currency).

I'd lean towards #2, but in that case the logic can also be done in a separate PR and we can use the manual approach from #1 for now if we want to relax this requirement.

erkarl:
> > Recap: the original reason for the raiden direct channel check is we simply cannot guarantee the first routing attempt of raiden to succeed. If it needs a second, the swap will fail (because the preimage changes).
> > But considering #1361, we could achieve a similar high routing success on first attempt. Meaning:
> > ```
> > * checks shouldn't be `direct channel for raiden`, but `direct channel for <currency> OR direct channel with minimum one bootstrap node`
> > 
> > * obviously depends on open channel service to be available
> > ```
> 
> This makes sense, but in revisiting this PR I think we could better control for this through tweaking the existing `directChannelChecks` config option logic which currently defaults to true. Even with the changes in this PR, setting that option to `false` will not disable raiden currencies that don't have direct channels. I have two approaches in mind:
> 
>     1. Have the user manually set this config option to fault if they intend to use a routing service for raiden.
> 
>     2. Get rid of the config option and treat it as if it's always `true` - but then internally count a channel with a hardcoded routing service node as satisfying the direct channel requirement (so for any given currency/peer, we need either a channel with that peer or a channel with a routing service for the currency).
> 
> 
> I'd lean towards #2, but in that case the logic can also be done in a separate PR and we can use the manual approach from #1 for now if we want to relax this requirement.

I'm leaning towards 2. as well in order to reduce the manual actions required by the user. For testing a manual override with 1. seems good.

kilrau:
I think #2 is the better way too. Close and new, separate PR? @sangaman

sangaman:
As we discussed, I think this PR still makes sense to merge (once I sort out why sim tests are failing) because it doesn't change the behavior requiring direct raiden channels by default, instead it only makes it so orders aren't exchanged between peers that require direct channels. Currently they will exchange orders and then fail any attempts to swap them if they lack a direct channel. I'll open a separate PR to track the second approach above.

sangaman:
@erkarl I finally got the tests to pass on this one. I know this is now somewhat irrelevant with deemphasizing raiden, but I still think this makes some good code organizational changes re: disabling currencies for which we don't have any routes to a peer.

PR #1349, opened 2019-12-12
feat(config): add maxlimits options

### LINKED ISSUE #1304: Add config option to raise mainnet limits to ln/raiden max values
* Add config option to raise xud mainnet trading limits to lndbtc/lndltc/raiden max values. Maybe `network_max_limits`.
* BTC, LTC Values: https://github.com/lightningnetwork/lnd/pull/1246/files. Our Raiden contract max values: @offerm

sangaman:
This adds a new `maxlimits` config option that overrides the hardcoded mainnet limits when set to true. In place of the mainnet limits, the maximum channel size for a currency's respective network is used as the maximum order size. These higher limits are now also enforced on non-mainnet networks regardless of the `maxlimits` value.

Closes #1304.

### DIFF
--- a/lib/constants/limits.ts
+++ b/lib/constants/limits.ts
@@ -1,9 +1,15 @@
-/** Order size limits denominated in satoshis. */
-const limits: { [currency: string]: number } = {
+/** Mainnet order size limits denominated in satoshis. */
+export const limits: { [currency: string]: number }  = {
   BTC: 100000,
   LTC: 15000000,
   WETH: 5000000,
   DAI: 1000000000,
 };
 
-export default limits;
+/** The maximum network channel sizes for each currency denominated in satoshis. */
+export const maxLimits: { [currency: string]: number } = {
+  BTC: 16777216, // 4294967 sat payment limit lifted as of 2019
+  LTC: 16777216 * 60, // lnd uses fixed 60 to 1 btc conversion rate
+  WETH: 11579208923731619542357098500868790785326998466564056403945758400791,
+  DAI: 11579208923731619542357098500868790785326998466564056403945758400791,
+};
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -11,7 +11,7 @@ import Logger from '../Logger';
 import { derivePairId, ms, setTimeoutPromise } from '../utils/utils';
 import { Models } from '../db/DB';
 import Swaps from '../swaps/Swaps';
-import limits from '../constants/limits';
+import { limits, maxLimits } from '../constants/limits';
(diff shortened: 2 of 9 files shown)


lib/constants/limits.ts
```diff
@@ -1,9 +1,15 @@
...
   BTC: 100000,
   LTC: 15000000,
   WETH: 5000000,
   DAI: 1000000000,
 };
 
-export default limits;
+/** The maximum network channel sizes for each currency denominated in satoshis. */
+export const maxLimits: { [currency: string]: number } = {
+  BTC: 16777216, // 4294967 sat payment limit lifted as of 2019
```
kilrau: I actually didn't know that, great!

kilrau:
Could you check on the tests though?

lib/orderbook/OrderBook.ts
```diff
@@ -404,17 +407,26 @@ class OrderBook extends EventEmitter {
...
+      // otherwise use the maximum channel sizes as order size limits
+      outboundCurrencyLimit = maxLimits[outboundCurrency];
+      inboundCurrencyLimit = maxLimits[inboundCurrency];
+    }
+
+    if (outboundCurrencyLimit && outboundAmount > outboundCurrencyLimit) {
+      throw errors.EXCEEDING_LIMIT(outboundCurrency, outboundAmount, outboundCurrencyLimit);
+    }
+    if (inboundCurrencyLimit && inboundAmount > inboundCurrencyLimit) {
+      throw errors.EXCEEDING_LIMIT(inboundCurrency, inboundAmount, inboundCurrencyLimit);
```
LePremierHomme: `errors.EXCEEDING_LIMIT` msg is misleading - it's referring the outbound amount, while here it's used for the inbound amount.

lib/constants/limits.ts
sangaman: Yep, I found that out while working on this PR. https://twitter.com/Stadicus3000/status/1088391443067080704

lib/orderbook/OrderBook.ts
sangaman: Good call, I'll change the message to make it apply either way.

lib/constants/limits.ts
kilrau: Reading through the thread, it seems for direct channels only, routed payments (which actually use HTLCs) are still subject to the 0.04 BTC limit. Anyhow, it's ok to lift it to 0.16 for our purpose since we expect to have a good amount of direct channels.

sangaman:
Tests should be fixed now too, some of the tests had a backwards price (100 BTC per LTC instead of the other way around) causing some of the limits to be exceeded.

lib/constants/limits.ts
sangaman: Ahh gotcha, but yeah it makes sense for an upper bound which is what we're aiming for here.

PR #1350, opened 2019-12-12
refactor(cli): handle stream error event

sangaman:
This sets a handler for the error event with streaming gRPC calls to explicitly set the exit code to 1 and log a tidier message, similar to what we do with regular request/response calls.


erkarl:
Looks good, but needs a rebase.

PR #1352, opened 2019-12-18
feat: restore from backup

kilrau:

PR #1353, opened 2019-12-18
test: add test case for backup feature

kilrau:
Tests for https://github.com/ExchangeUnion/xud/pull/1032. Some dependencies might have to be updated to make Travis pass for this.

### DIFF
--- a/test/jest/Backup.spec.ts
+++ b/test/jest/Backup.spec.ts
@@ -1,0 +1,173 @@
+import fs from 'fs';
+import path from 'path';
+import Backup from '../../lib/backup/Backup';
+
+const removeDir = (dir: string) => {
+  if (fs.existsSync(dir)) {
+    fs.readdirSync(dir).forEach((file) => {
+      fs.unlinkSync(`${dir}/${file}`);
+    });
+  }
+
+  fs.rmdirSync(dir);
+};
+
+const wait = (ms: number) => {
+  return new Promise(resolve => setTimeout(resolve, ms));
+};
+
+const backupdir = 'backup-test';
+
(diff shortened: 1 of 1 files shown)


test/jest/Backup.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+  if (fs.existsSync(dir)) {
+    fs.readdirSync(dir).forEach((file) => {
+      fs.unlinkSync(`${dir}/${file}`);
+    });
+  }
+
+  fs.rmdirSync(dir);
+};
+
+const wait = (ms: number) => {
```
sangaman: There's a utils `setTimeoutPromise` for this you can import and use in the same way instead of this function.

test/jest/Backup.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+  });
+
+  test('should write Raiden backups on new event', async () => {
+    fs.writeFileSync(
+      raidenDatabase,
+      backups.raiden.event,
+    );
+
+    // Wait 100ms to make sure the file watcher handled the new file
+    await wait(100);
```
sangaman: It would be nice here to be able to actually be notified when the new file update has been handled, instead of waiting a fixed amount of time. I figure the easiest way to do that would be to make the `Backup` class extend `EventEmitter`, then add an event for when a new backup is written and emit it after or at the end of the `writeBackup` call along with an identifier (either client or dbpath). Then you can listen for that here and elsewhere `wait` is being used.

sangaman:
Combining this PR with #1352 because I want to ensure this test case passes with the changes from that PR.

PR #1359, opened 2020-01-13
Backup related fixes

ghost:
This PR fixes 2 issues:
- Re-subscribe to lnd channel backups when lnd is restarted while the backup daemon is running

- Properly auto unlock lnd when it crashes and immediately restarts

Previously, it was stuck in a loop of failed `channelbalance` requests while receiving `UNIMPLEMENTED` gRPC response

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -190,8 +190,29 @@ class LndClient extends SwapClient {
     return this._maxChannelInboundAmount;
   }
 
+  /** Lnd specific procedure to mark the client as locked. */
+  private lock = async () => {
+    this.walletUnlocker = new WalletUnlockerClient(this.uri, this.credentials);
+    if (this.lightning) {
+      this.lightning.close();
+    }
+    this.lightning = undefined;
+
+    if (!this.isWaitingUnlock()) {
+      await this.setStatus(ClientStatus.WaitingUnlock);
+      this.emit('locked');
+    }
+  }
+
   protected updateCapacity = async () => {
-    await this.channelBalance().catch(err => this.logger.error('failed to update total outbound capacity', err));
+    await this.channelBalance().catch(async (err) => {
+      if (err.code === grpc.status.UNIMPLEMENTED) {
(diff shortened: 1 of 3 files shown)


lib/lndclient/LndClient.ts
```diff
@@ -189,7 +191,14 @@ class LndClient extends SwapClient {
   }
 
   protected updateCapacity = async () => {
-    await this.channelBalance().catch(err => this.logger.error('failed to update total outbound capacity', err));
+    await this.channelBalance().catch(async (err) => {
+      if (err.code === grpc.status.UNIMPLEMENTED) {
+        // if ChannelBalance is unimplemented, it means this lnd instance is online but locked
+        await this.setStatus(ClientStatus.Disconnected);
```
sangaman: Should we set the status to locked here instead of disconnected?

lib/lndclient/LndClient.ts
```diff
@@ -949,8 +958,12 @@ class LndClient extends SwapClient {
       return;
     }
 
-    const deleteChannelBackupSubscription = () => {
+    const deleteChannelBackupSubscription = async () => {
       this.channelBackupSubscription = undefined;
+      // setting status as disconnected because it's reasonable to assume
+      // that the swap client ended the subscription because it's disconnected
```
sangaman: I'm thinking it might be best to scrap this method and in its place use the `disconnect` method we already have. Then we'd also add a check there to close `channelBackupSubscription` and set it to undefined (unless it's already undefined). It's a bit counterintuitive to me that `deleteChannelBackupSubscription` also disconnects the client, even tho this comment does help to clear it up.

sangaman (commented):
See my comments.

One other thing I'm thinking is that where we call `await this.startLndSubscriptions();` we should probably instead do `this.startLndSubscriptions().catch(...` so that we don't wait on it. Right now, it's possible to get stuck waiting for lnd and then we won't reach the code to start the raiden database file watcher.

sangaman (approved):
Looks good, I think some of the LndClient code around dealing with the `UNIMPLEMENTED` error can be refactored  some more but I'll open a PR for that separately.

PR #1360, opened 2020-01-15
WIP: retry misconfigured lnd clients

ghost:


erkarl:
https://github.com/ExchangeUnion/xud/pull/1363 is a better solution.

PR #1376, opened 2020-01-29
feat: use seedutil to generate mnemonic

### LINKED ISSUE #1253: Use seedutil to generate aezeed mnemonic
Currently we depend on lnd to generate the master recovery aezeed mnemonic. When lnd is not available, we don't generate a mnemonic at all. Instead, if it's unavailable we should use the seedutil tool, see: https://github.com/ExchangeUnion/xud/blob/master/lib/swaps/SwapClientManager.ts#L124

This issue tracks the enhancement to the seedutil tool to generate a seed from scratch and to use it in xud.

sangaman:
This changes adds the functionality to seedutil to generate a random 24 word mnemonic, which then replaces lnd for the purposes of generating the seed and mnemonic for a new xud node. This removes a dependency and allows xud to create a node without the help of lnd.

Closes #1253.

### DIFF
--- a/lib/service/InitService.ts
+++ b/lib/service/InitService.ts
@@ -3,7 +3,7 @@ import { promises as fs } from 'fs';
 import NodeKey from '../nodekey/NodeKey';
 import swapErrors from '../swaps/errors';
 import SwapClientManager from '../swaps/SwapClientManager';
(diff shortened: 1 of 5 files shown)


kilrau:
Restarted Travis builds, please watch out for the result - might need fixing.

PR #1380, opened 2020-01-29
feat(cli): better messages when unlock/create fail

### LINKED ISSUE #1370: Improve error message when xucli unlock fails
When `xucli unlock` fails because no node exists, it currently prints a not very informative message of `Error: 12 UNIMPLEMENTED: call is not supported by the current state of xud`. A better message would inform the caller that xud can't be unlocked because the node has not been created yet, and the cli should recommend `xucli create` or `xucli restore` instead.

Another issue is that when you try to unlock an already unlocked wallet, it prints a successful output when it should print that xud was already unlocked.

sangaman:
This displays better error messages when `xucli unlock`, `xucli create`, or `xucli restore` calls fail due to an xud node not existing or already existing. Instead of the generic `call is not supported by the current state of xud` message being printed, a more helpful message is printed informing the user of the state of xud and which call should be tried instead.

This also returns an error when a wallet that is already unlocked is attempted to be unlocked again.

Closes #1370.

### DIFF
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -137,7 +137,7 @@ class Xud extends EventEmitter {
         throw new Error('rpc server cannot be disabled when xud is locked');
       }
       if (this.rpcServer) {
-        this.rpcServer.grpcInitService.disabled = true;
+        this.rpcServer.grpcInitService.disable();
       }
 
       this.logger.info(`Local nodePubKey is ${nodeKey.pubKey}`);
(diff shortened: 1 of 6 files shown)


sangaman:
> LGTM. The only thing I'd improve in the future is the error message when entering an invalid password (outside the scope of this PR):
> 
> ```
> ➜  xud git:(cli/unlock-messages) ./bin/xucli unlock
> Enter master xud password:
> Error: 2 UNKNOWN: error:06065064:digital envelope routines:EVP_DecryptFinal_ex:bad decrypt
> ```

Yup I had this same thought, see #1381.

Issue #1343, opened 2019-11-29
Add minimum price precision for orders


erkarl:
At the moment it's possible to create orders with many decimal places.

```
➜  xud git:(feat/change-orderbook-default-precision) ./bin/xucli buy 0.00000001 LTC/BTC 0.00000000000000000000000000000000000000001
no matches found
remaining 0.00000001 LTC entered the order book as c736a520-12b1-11ea-979c-9b01abcb2b70
➜  xud git:(feat/change-orderbook-default-precision) ./bin/xucli orderbook

Trading pair: LTC/BTC
┌───────────────────────────────────────┬───────────────────────────────────────┐
│ Buy                                   │ Sell                                  │
├───────────────────┬───────────────────┼───────────────────┬───────────────────┤
│ Quantity          │ Price             │ Price             │ Quantity          │
└───────────────────┴───────────────────┴───────────────────┴───────────────────┘
➜  xud git:(feat/change-orderbook-default-precision) ./bin/xucli listorders

Trading pair: LTC/BTC
┌────────────────────────────────────────────────────┬────────────────────────────────────────────────────┐
│ Buy                                                │ Sell                                               │
├───────────────────┬───────────────────┬────────────┼───────────────────┬───────────────────┬────────────┤
│ Quantity          │ Price             │ Own Order  │ Quantity          │ Price             │ Own Order  │
├───────────────────┼───────────────────┼────────────┼───────────────────┼───────────────────┼────────────┤
│ 0.00000001        │ 1e-41             │ X          │                   │                   │            │
└───────────────────┴───────────────────┴────────────┴───────────────────┴───────────────────┴────────────┘
➜  xud git:(feat/change-orderbook-default-precision) ./bin/xucli orderbook --precision 80

Trading pair: LTC/BTC
┌───────────────────────────────────────┬───────────────────────────────────────┐
│ Buy                                   │ Sell                                  │
├───────────────────┬───────────────────┼───────────────────┬───────────────────┤
│ Quantity          │ Price             │ Price             │ Quantity          │
├───────────────────┼───────────────────┼───────────────────┼───────────────────┤
│ 0.00000001        │ 1e-41             │                   │                   │
└───────────────────┴───────────────────┴───────────────────┴───────────────────┘
```

engwarrior:
Hi, I want to work on this issue and others and eventually join your team, I spoke to Killian and he advised me  to start work on the issues. There should be estimation on how much I get paid on completion, can you give me one?

[...]

PR #1385, opened 2020-02-04
feat(cli): add price max precision checks for orders

engwarrior:
https://github.com/ExchangeUnion/xud/issues/1343. We decided to go with 10 ** 12 max precision. The number is easily configurable any time. Actually was quite a challenge to setup xud with lnd, btcd, ltcd and still shows some swapClient when executing order, but precision checks work fine.
Screenshot 2020-02-04 at 15 12 24

### DIFF
--- a/lib/cli/placeorder.ts
+++ b/lib/cli/placeorder.ts
@@ -2,11 +2,13 @@ import { Arguments, Argv } from 'yargs';
 import { Order, OrderSide, PlaceOrderEvent, PlaceOrderRequest, PlaceOrderResponse, SwapFailure, SwapSuccess } from '../proto/xudrpc_pb';
 import { callback, loadXudClient } from './command';
 import { coinsToSats, satsToCoinsStr } from './utils';
+import { checkDecimalPlaces } from '../utils/utils';
 
 export const placeOrderBuilder = (argv: Argv, side: OrderSide) => {
   const command = side === OrderSide.BUY ? 'buy' : 'sell';
   argv.option('quantity', {
     type: 'number',
+    describe: 'the quantity to trade',
   })
   .option('pair_id', {
     type: 'string',
@@ -59,6 +61,12 @@ export const placeOrderHandler = (argv: Arguments<any>, side: OrderSide) => {
     }
   } else if (priceStr !== ('mkt') && priceStr !== ('market')) {
     console.log('price must be numeric for limit orders or "mkt"/"market" for market orders');
+    return;
+  }
+
+  if (checkDecimalPlaces(numericPrice)) {
+    process.exitCode = 1;
+    console.error('price cannot have more than 12 decimal places');
     return;
   }
 
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -11,7 +11,7 @@ import { TradingLimits } from '../swaps/SwapClient';
 import Swaps from '../swaps/Swaps';
 import { ResolveRequest, SwapFailure, SwapSuccess } from '../swaps/types';
 import { parseUri, toUri, UriParts } from '../utils/uriUtils';
-import { sortOrders, toEip55Address } from '../utils/utils';
+import { checkDecimalPlaces, sortOrders, toEip55Address } from '../utils/utils';
 import commitHash from '../Version';
 import errors from './errors';
 
@@ -58,6 +58,9 @@ const argChecks = {
   POSITIVE_AMOUNT: ({ amount }: { amount: number }) => { if (amount <= 0) throw errors.INVALID_ARGUMENT('amount must be greater than 0'); },
   POSITIVE_QUANTITY: ({ quantity }: { quantity: number }) => { if (quantity <= 0) throw errors.INVALID_ARGUMENT('quantity must be greater than 0'); },
   PRICE_NON_NEGATIVE: ({ price }: { price: number }) => { if (price < 0) throw errors.INVALID_ARGUMENT('price cannot be negative'); },
+  PRICE_MAX_DECIMAL_PLACES: ({ price }: { price: number }) => {
+    if (checkDecimalPlaces(price)) throw errors.INVALID_ARGUMENT('price cannot have more than 12 decimal places');
+  },
   VALID_CURRENCY: ({ currency }: { currency: string }) => {
     if (currency.length < 2 || currency.length > 5 || !currency.match(/^[A-Z0-9]+$/)) {
       throw errors.INVALID_ARGUMENT('currency must consist of 2 to 5 upper case English letters or numbers');
@@ -447,6 +450,7 @@ class Service {
     const { pairId, price, quantity, orderId, side, replaceOrderId, immediateOrCancel } = args;
     argChecks.PRICE_NON_NEGATIVE(args);
     argChecks.POSITIVE_QUANTITY(args);
+    argChecks.PRICE_MAX_DECIMAL_PLACES(args);
     argChecks.HAS_PAIR_ID(args);
 
     if (replaceOrderId) {
--- a/lib/utils/utils.ts
+++ b/lib/utils/utils.ts
@@ -9,6 +9,8 @@ import { promisify } from 'util';
 import { Order, Pair } from '../orderbook/types';
 import p2pErrors from '../p2p/errors';
 
+const MAX_DECIMAL_PLACES = 12;
+
 /**
  * Gets the external IP of the node.
  */
@@ -103,6 +105,12 @@ export const groupBy = (arr: object[], keyGetter: (item: any) => string | number
     }
   });
   return ret;
+};
+
+/** Returns true if number has more than MAX_DECIMAL_PLACES, false otherwise. */
+export const checkDecimalPlaces = (digits: number) => {
+  const fixed = Number(digits.toFixed(MAX_DECIMAL_PLACES));
+  return fixed < digits;
 };
 
 /**


engwarrior:
CI build failed but it's not related to tests or lint or built: its setting up simulation which fails on 2020/02/04 12:49:29 cannot start xud network: context deadline exceeded

lib/cli/placeorder.ts
```diff
@@ -1,20 +1,21 @@
...
 import { Order, OrderSide, PlaceOrderEvent, PlaceOrderRequest, PlaceOrderResponse, SwapFailure, SwapSuccess } from '../proto/xudrpc_pb';
 import { callback, loadXudClient } from './command';
-import { coinsToSats, satsToCoinsStr } from './utils';
+import { checkPrecision, coinsToSats, satsToCoinsStr } from './utils';
 
 export const placeOrderBuilder = (argv: Argv, side: OrderSide) => {
   const command = side === OrderSide.BUY ? 'buy' : 'sell';
   argv.option('quantity', {
-    type: 'number',
+    type: 'string',
```
sangaman: I think it's better to keep this as a number, since that's what it is, and make the string conversion (if necessary) later in the code.

lib/cli/utils.ts
```diff
@@ -4,6 +4,7 @@ import os from 'os';
 import path from 'path';
 
 const SATOSHIS_PER_COIN = 10 ** 8;
+const MAX_PRECISION = 12;
```
sangaman: Maybe call this `MAX_PRICE_DECIMAL_PLACES`. "Decimal place" is more accurate in this context I believe because "precision" can refer to the number of significant digits either to the left or right of the decimal point.

lib/cli/utils.ts
```diff
@@ -33,6 +34,18 @@ export const coinsToSats = (coinsQuantity: number) => {
   return Math.round(coinsQuantity * SATOSHIS_PER_COIN);
 };
 
+/** Checks order price precision is not more than MAX_PRECISION. */
+export const checkPrecision = (price: string) => {
```
sangaman: `checkDecimalPlaces` maybe for same reasoning as above?

lib/cli/placeorder.ts
```diff
@@ -1,20 +1,21 @@
...
+    describe: 'the quantity for limit/market orders, max 12 digits on the right after a decimal point',
   })
   .option('pair_id', {
     type: 'string',
     describe: 'the trading pair ticker id for the order',
   })
   .option('price', {
     type: 'string',
-    describe: 'the price for limit orders, or "mkt"/"market" for market orders',
+    describe: 'the price for limit orders, or "mkt"/"market" for market orders, max 12 digits on the right after a decimal point',
```
sangaman: "max 12 decimal places" ? just a bit shorter that way and there's not much room in terminals

lib/cli/placeorder.ts
engwarrior: It's not that simple unfortunately, because if the user sets the quantity to 1e-24 for example it will be converted with .toString() to "1e-24".

engwarrior:
I will add to RPC too, its very fast. But you probably mean gRPC right? I see only gRPC https://github.com/ExchangeUnion/xud/blob/master/lib/grpc/GrpcService.ts#L15

lib/cli/placeorder.ts
engwarrior: in case if we want to use toFixed(decimals), we will have to specify large number of decimals like 100 for potential user case when people will specify 10 ** 100 precision and check all digits after decimal point if they are not 0.

engwarrior:
RPC logic is very trivial after testing new method implementation. I changed that to fit number rather than string as requested and being cleaner.

package.json
```diff
@@ -129,6 +129,7 @@
   ],
   "dependencies": {
     "@exchangeunion/grpc-dynamic-gateway": "^0.3.10",
+    "bignumber.js": "^9.0.0",
```
erkarl: I'm not sure about adding this dependency. Is this really necessary?

engwarrior: I didn't find more elegant and correct way without checking out corner cases to check if Number decimal places are at most 12.

engwarrior:
@erkarl @sangaman @kilrau are going to review this guys?

lib/service/Service.ts
```diff
@@ -58,6 +59,8 @@ const argChecks = {
   POSITIVE_AMOUNT: ({ amount }: { amount: number }) => { if (amount <= 0) throw errors.INVALID_ARGUMENT('amount must be greater than 0'); },
   POSITIVE_QUANTITY: ({ quantity }: { quantity: number }) => { if (quantity <= 0) throw errors.INVALID_ARGUMENT('quantity must be greater than 0'); },
   PRICE_NON_NEGATIVE: ({ price }: { price: number }) => { if (price < 0) throw errors.INVALID_ARGUMENT('price cannot be negative'); },
+  QUANTITY_DECIMAL_PLACES: ({ quantity }: { quantity: number }) => { if (!checkDecimalPlaces(quantity)) throw errors.INVALID_ARGUMENT('quantity mush have max 12 decimal places'); },
+  PRICE_DECIMAL_PLACES: ({ price }: { price: number }) => { if (!checkDecimalPlaces(price)) throw errors.INVALID_ARGUMENT('price mush have max 12 decimal places'); },
```
michael1011:
```suggestion
  PRICE_DECIMAL_PLACES: ({ price }: { price: number }) => { if (!checkDecimalPlaces(price)) throw errors.INVALID_ARGUMENT('price must have less than 13 decimals); },
```

Maybe reword the error so that it is more clear?

lib/service/Service.ts
```diff
@@ -58,6 +59,8 @@ const argChecks = {
   POSITIVE_AMOUNT: ({ amount }: { amount: number }) => { if (amount <= 0) throw errors.INVALID_ARGUMENT('amount must be greater than 0'); },
   POSITIVE_QUANTITY: ({ quantity }: { quantity: number }) => { if (quantity <= 0) throw errors.INVALID_ARGUMENT('quantity must be greater than 0'); },
   PRICE_NON_NEGATIVE: ({ price }: { price: number }) => { if (price < 0) throw errors.INVALID_ARGUMENT('price cannot be negative'); },
+  QUANTITY_DECIMAL_PLACES: ({ quantity }: { quantity: number }) => { if (!checkDecimalPlaces(quantity)) throw errors.INVALID_ARGUMENT('quantity mush have max 12 decimal places'); },
```
michael1011: Same here

lib/cli/placeorder.ts
```diff
@@ -62,6 +63,11 @@ export const placeOrderHandler = (argv: Arguments<any>, side: OrderSide) => {
     return;
   }
 
+  if (!checkDecimalPlaces(numericPrice) || !checkDecimalPlaces(argv.quantity)) {
+    console.log('price and quantity must have maximum 12 decimal places');
```
michael1011: Same here

lib/cli/placeorder.ts
sangaman: You can do `toFixed` with 12 decimal places, then compare the price to the fixed string.

```typescript
const fixedPrice = price.toFixed(12);
if (price > fixedPrice) // we have numbers after the first 12 decimal places, invalid argument
```

package.json
sangaman: I agree that this is not necessary, see my comment above about a simpler way to check if we have more than 12 decimal places.

lib/cli/placeorder.ts
engwarrior: @sangaman I hope JS compares doubles correctly.

sangaman: What is your concern? All numbers are doubles in JS.

engwarrior: @sangaman its better to avoid comparison of doubles with high precision

engwarrior:
Removed bignumber.js

erkarl:
Looks like this PR now deletes `npm-shrinkwrap.json`.

lib/cli/placeorder.ts
```diff
@@ -1,20 +1,21 @@
...
 import { Order, OrderSide, PlaceOrderEvent, PlaceOrderRequest, PlaceOrderResponse, SwapFailure, SwapSuccess } from '../proto/xudrpc_pb';
 import { callback, loadXudClient } from './command';
-import { coinsToSats, satsToCoinsStr } from './utils';
+import { checkDecimals, coinsToSats, satsToCoinsStr } from './utils';
 
 export const placeOrderBuilder = (argv: Argv, side: OrderSide) => {
   const command = side === OrderSide.BUY ? 'buy' : 'sell';
   argv.option('quantity', {
     type: 'number',
+    describe: 'the quantity for limit/market orders, less than 13 decimals',
```
sangaman: We aren't applying this 12 max decimal places limit to quantity, that is already limited to 8 (satoshis). We can leave this alone.

lib/cli/placeorder.ts
```diff
@@ -1,20 +1,21 @@
...
   })
   .option('pair_id', {
     type: 'string',
     describe: 'the trading pair ticker id for the order',
   })
   .option('price', {
     type: 'string',
-    describe: 'the price for limit orders, or "mkt"/"market" for market orders',
+    describe: 'the price for limit orders, or "mkt"/"market" for market orders, less than 13 decimals',
```
sangaman: I personally think "max 12 decimal places" is more clear - although I also think this description can also be left alone as in most cases people aren't going to be trying to enter prices with that many decimal places anyway. Maybe better not to confuse anyone, if somebody tries a really weird price then they will get a more descriptive error message.

lib/cli/placeorder.ts
```diff
@@ -62,6 +63,11 @@ export const placeOrderHandler = (argv: Arguments<any>, side: OrderSide) => {
     return;
   }
 
+  if (checkDecimals(numericPrice) || checkDecimals(argv.quantity)) {
+    console.log('price and quantity must have less than 13 decimals');
```
sangaman: Again there shouldn't be a check on quantity here.

lib/cli/utils.ts
```diff
@@ -4,6 +4,7 @@ import os from 'os';
 import path from 'path';
 
 const SATOSHIS_PER_COIN = 10 ** 8;
+const MAX_DECIMALS = 12;
```
sangaman: I'd prefer what I suggested originally so it is clear what this is for, `MAX_PRICE_DECIMAL_PLACES`

lib/service/Service.ts
```diff
@@ -58,6 +59,8 @@ const argChecks = {
   POSITIVE_AMOUNT: ({ amount }: { amount: number }) => { if (amount <= 0) throw errors.INVALID_ARGUMENT('amount must be greater than 0'); },
   POSITIVE_QUANTITY: ({ quantity }: { quantity: number }) => { if (quantity <= 0) throw errors.INVALID_ARGUMENT('quantity must be greater than 0'); },
   PRICE_NON_NEGATIVE: ({ price }: { price: number }) => { if (price < 0) throw errors.INVALID_ARGUMENT('price cannot be negative'); },
+  QUANTITY_CORRECT_PRECISION: ({ quantity }: { quantity: number }) => { if (checkDecimals(quantity)) throw errors.INVALID_ARGUMENT('quantity must have less than 13 decimal'); },
```
sangaman: We don't need this for quantity. Also on the grpc layer quantity will always be an integer, as it represents whole numbers of satoshis, so it would never fail this check.

lib/cli/utils.ts
```diff
@@ -33,6 +34,12 @@ export const coinsToSats = (coinsQuantity: number) => {
   return Math.round(coinsQuantity * SATOSHIS_PER_COIN);
 };
 
+/** Checks max decimals. */
```
sangaman: It's not clear what this does based on the description, it looks like earlier this returned `true` when the check passed, but now it returns `true` when the check fails. So better to specify here that the function returns true when the decimal places are exceeded.

lib/service/Service.ts
```diff
@@ -58,6 +59,8 @@ const argChecks = {
...
+  PRICE_CORRECT_PRECISION: ({ price }: { price: number }) => { if (checkDecimals(price)) throw errors.INVALID_ARGUMENT('price must have less than 13 decimal'); },
```
sangaman: Let's go with `PRICE_MAX_DECIMAL_PLACES` for the check name and `price cannot have more than 12 decimal places` for the message, if you don't mind.

flicker-harmony:
travis fails because of private key https://travis-ci.org/ExchangeUnion/xud/builds/646934061#L854 locally tests are fine

lib/cli/placeorder.ts
```diff
@@ -2,11 +2,13 @@ import { Arguments, Argv } from 'yargs';
 import { Order, OrderSide, PlaceOrderEvent, PlaceOrderRequest, PlaceOrderResponse, SwapFailure, SwapSuccess } from '../proto/xudrpc_pb';
 import { callback, loadXudClient } from './command';
 import { coinsToSats, satsToCoinsStr } from './utils';
+import { checkDecimalPlaces } from '../utils/utils';
 
 export const placeOrderBuilder = (argv: Argv, side: OrderSide) => {
   const command = side === OrderSide.BUY ? 'buy' : 'sell';
   argv.option('quantity', {
     type: 'number',
+    describe: 'the quantity for limit/market orders',
```
sangaman: We have some order besides limit and market, so I'd either leave this line out (not needed for this PR anyhow) or just say something more generic like "the quantity to trade".

lib/cli/placeorder.ts
```diff
@@ -62,6 +64,11 @@ export const placeOrderHandler = (argv: Arguments<any>, side: OrderSide) => {
     return;
   }
 
+  if (checkDecimalPlaces(numericPrice)) {
+    console.log('price cannot have more than 12 decimal places');
```
sangaman: We should log this to `console.error` and also set the status code to a non-zero value to indicate the call has failed with `process.exitCode = 1`.

lib/cli/placeorder.ts
flicker-harmony: Done

lib/cli/placeorder.ts
flicker-harmony: Done

erkarl:
Looks good :+1:.

As a final nit let's clean up the commit message and PR title to exclude `and quantity`.

engwarrior:
Fails integration on travis...

erkarl:
> Fails integration on travis...

We're experiencing major flakiness in our simulation tests - I restarted your build. Things should improve a lot once https://github.com/ExchangeUnion/xud/pull/1357 gets merged.

engwarrior:
Still

engwarrior:
Doesn't help. Is something broken? I don't understand test that is failing. And it passed before.

kilrau:
Restarted again. It's annoying, I know but improvement is on the way (https://github.com/ExchangeUnion/xud/pull/1357).

engwarrior:
Maybe my PR broke integration test? But I can't check it locally and it was okay before.

erkarl:
> Maybe my PR broke integration test? But I can't check it locally and it was okay before.

All good on your side. This will need to get merged before the test suite will pass: https://github.com/ExchangeUnion/xud/pull/1386

engwarrior:
Will you merge tests fix PR and merge mine today?

kilrau:
Just merged the fix and restarted tests. Thanks for your patience @engwarrior

flicker-harmony:
@kilrau merge.

kilrau:
> @kilrau merge.

Aye, sir ;)

PR #1386, opened 2020-02-08
chore: disable raiden simulation tests

ghost:
This PR disables/removes the raiden simulation tests. Right now our simulation tests are not passing at all since we closed the matrix server.


sangaman (approved):
Looks good to me, I didn't actually realize these depended on an external server. All the more reason to remove it. Thank you.

Issue #1154, opened 2019-08-16
Add tor configuration option


erkarl:
Now that `xud-docker` is using Tor by default we'll have lots of nodes with `hash.onion` advertised addresses. The nodes that are not using Tor will be unable to connect to those addresses.

I propose to have a `tor = false` configuration option that for now will prevent `xud` from trying to connect to those addresses.

hatmer:
There are proxy services that allow regular internet nodes to connect to Tor hidden services, they are pretty sketchy though. Eventually we could build one to allow full connectivity from non-tor to tor nodes.

erkarl:
>  they are pretty sketchy though.

Agreed. For now I'd go with a simple solution and just not connect to onion addresses if Tor is disabled.

sangaman:
Seems like a good approach to me, and fairly standard from what I've seen in other services where tor is optional. What I'm not sure about is whether tor should be enabled or disabled by default. Maybe it can be disabled by default but if we detect an advertised tor address then we enable it?

engwarrior:
Pls provide estimate for this task.

flicker-harmony:
@kilrau @erkarl @sangaman how it can be determined in current xud code if Node has hash.onion advertised address? For now I can't find anything related to it in current Pool Config, Peer info and Xud configs.

engwarrior:
@kilrau @erkarl @sangaman how it can be determined in current xud code if Node has hash.onion advertised address? For now I can't find anything related to it in current Pool Config, Peer info and Xud configs.

erkarl:
@engwarrior https://github.com/ExchangeUnion/xud/blob/master/sample-xud.conf#L67

PR #1387, opened 2020-02-11
feat(p2p): add p2p tor config option

engwarrior:
EDIT by @kilrau: Closes https://github.com/ExchangeUnion/xud/issues/1154

### DIFF
--- a/bin/xud
+++ b/bin/xud
@@ -148,6 +148,12 @@ const { argv } = require('yargs')
       type: 'number',
       alias: 'p',
     },
+    'p2p.tor': {
+      describe: 'Allow connections to tor nodes',
+      type: 'boolean',
+      default: undefined,
+      alias: 't',
+    },
     'raiden.disable': {
       describe: 'Disable raiden integration',
       type: 'boolean',
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -408,6 +408,11 @@ class Pool extends EventEmitter {
       throw errors.ATTEMPTED_CONNECTION_TO_SELF;
     }
 
+    // check if we allow connections to tor addresses
+    if (!this.config.tor && address.host.indexOf('.onion') !== -1) {
+      throw errors.NODE_TOR_ADDRESS(nodePubKey, address);
+    }
+
     if (this.nodes.isBanned(nodePubKey)) {
       throw errors.NODE_IS_BANNED(nodePubKey);
     }
--- a/lib/p2p/errors.ts
+++ b/lib/p2p/errors.ts
@@ -30,6 +30,7 @@ const errorCodes = {
   FRAMER_INCOMPATIBLE_MSG_ORIGIN_NETWORK: codesPrefix.concat('.22'),
   FRAMER_INVALID_MSG_LENGTH: codesPrefix.concat('.23'),
   POOL_CLOSED: codesPrefix.concat('.24'),
+  NODE_TOR_ADDRESS: codesPrefix.concat('.25'),
 };
 
 const errors = {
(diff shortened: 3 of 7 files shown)


bin/xud
```diff
@@ -148,6 +148,12 @@ const { argv } = require('yargs')
       type: 'number',
       alias: 'p',
     },
+    'p2p.tor': {
+      describe: 'Enable/disable connections to tor nodes',
+      type: 'boolean',
+      default: false,
```
sangaman: For all booleans, we actually want the default to be `undefined`, the reason being is that if it's not specified via command line we don't want it to override the setting in the config file. If we have a default of false, it be false even if it's true in the config file.

lib/p2p/types.ts
```diff
@@ -44,6 +44,8 @@ export type PoolConfig = {
   /** Which port to listen on. If 0, a random unused port will be used. */
   port: number;
 
+  /** Enable/disable connections to tor nodes. */
```
sangaman: How about "Whether to allow connections to tor nodes"

lib/p2p/Pool.ts
```diff
@@ -358,6 +358,8 @@ class Pool extends EventEmitter {
 
     for (const address of sortedAddresses) {
       if (node.lastAddress && addressUtils.areEqual(address, node.lastAddress)) continue;
+      // check if its tor node address
```
sangaman: How about "check if we allow connections to tor addresses"

bin/xud
```diff
@@ -148,6 +148,12 @@ const { argv } = require('yargs')
       type: 'number',
       alias: 'p',
     },
+    'p2p.tor': {
+      describe: 'Enable/disable connections to tor nodes',
```
sangaman: I think we can just say "Allow connections to tor nodes"

lib/p2p/Pool.ts
```diff
@@ -358,6 +358,8 @@ class Pool extends EventEmitter {
...
+      if (!this.config.tor && address.host.indexOf('.onion') !== -1) continue;
```
sangaman: This logic looks good, but the check is only happening for when we connect to peer's advertised addresses. It doesn't happen when we connect to the last good address, or when we use the Connect rpc command.

I think this logic should go in the `addOutbound` method which is called on every connection. In there it should throw an error if we don't allow tor addresses but the address is for tor. You'll have to create a new error in the p2p errors.ts file to be thrown. That error will then also need to be added to the switch statement in grpc/getGrpcError.ts - I figure it makes most sense to be added as an `ILLEGAL_ARGUMENT` error. That's so that if someone attempts to connect to a tor address over rpc when doing so is disabled, they'll get an intelligible error message.

Let me know if you have any questions.

engwarrior: Yeah, you really think a lot about all the functions, errors and comments texts - it takes even more time than implementing something)

engwarrior: status.ILLEGAL_ARGUMENT error doesn't exist, maybe you meant INVALID_ARGUMENT?

lib/p2p/errors.ts
```diff
@@ -41,6 +42,10 @@ const errors = {
     message: `node (${nodePubKey}) is not connected`,
     code: errorCodes.NOT_CONNECTED,
   }),
+  NODE_TOR_ADDRESS: (nodePubKey: string, address: Address) => ({
+    message: `node (${nodePubKey}) at ${addressUtils.toString(address)} is tor node`,
```
sangaman: How about a message like "can't connect to node {node} at tor address {address} because tor is disabled"

test/integration/Pool.spec.ts
```diff
@@ -88,6 +90,12 @@ describe('P2P Pool Tests', async () => {
     expect(nodeReputationPromise.banned).to.be.true;
   });
 
+  it('should throw error when connecting to tor node', async () => {
```
sangaman: 'should throw error when connecting to tor node with tor disabled'

test/integration/Pool.spec.ts
```diff
@@ -88,6 +90,12 @@ describe('P2P Pool Tests', async () => {
...
+    const address = addressUtils.fromString('.onion');
```
sangaman: I think it makes sense to use a real onion address for completeness sake, you can plug in 3g2upl4pq6kufc4m.onion which belongs to DuckDuckGo.

engwarrior: thanks)

PR #1390, opened 2020-02-12
chore: generate slate on new version

sangaman:
This automatically generates new slate API documentation when a new version is tagged.


erkarl:
Simtests failed due to xud cleanup. Restarted the job. It's so nice that we can now restart individual jobs.

PR #1391, opened 2020-02-12
chore: ignore simulation temp directory

sangaman:
The simulation temp directory is still used but not ignored after #1357. This makes the folder ignored once again to prevent unintentionally committed files.

Note that this is a quick fix, however when examining the issue I was also thinking that the temp folder might not be necessary and can be removed altogether by simply using the dist folder. I'll open a separate PR for that to discuss that approach.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,9 @@ docker/xud/xud.db
 seedutil/go
 seedutil/seedutil
 
+# temp simulation test directory
+test/simulation/temp
+
 # Version file generated in build process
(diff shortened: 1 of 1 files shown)


sangaman:
I'm also realizing that the instability and security simulation tests are not dockerized (yet?) so they would create `go` and `temp` folders on their own that would currently not be ignored. The best approach would probably be to dockerize those as well, but we could re-ignore those folders as a temporary fix if needed.

PR #1392, opened 2020-02-14
feat(rpc): add owner parameter for listorders

### LINKED ISSUE #1323: `only_own_orders` parameter for listorders

krrprr:
This commit adds an optional `owner` parameter for the `listorders` command while removing `include_own_orders_parmeter` and replacing its functionality.
The default value for the new parameter is `Both`, other options are `Own` and `Peer`.

The commit also fixes a bug that caused a situation where sell orders were displayed instead of buy orders when the limit was set.

Closes #1323

### DIFF
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -1,4 +1,4 @@
-import { OrderSide, SwapClientType, SwapRole } from '../constants/enums';
+import { OrderSide, SwapClientType, SwapRole, Owner } from '../constants/enums';
 import { LndInfo } from '../lndclient/types';
 import OrderBook from '../orderbook/OrderBook';
 import { OrderSidesArrays } from '../orderbook/TradingPair';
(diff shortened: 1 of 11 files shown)


lib/service/Service.ts
```diff
@@ -358,8 +358,12 @@ class Service {
...
    * Get a map between pair ids and its orders from the order book.
    */
-  public listOrders = (args: { pairId: string, includeOwnOrders: boolean, limit: number }): Map<string, OrderSidesArrays<any>> => {
-    const { pairId, includeOwnOrders, limit } = args;
+  public listOrders = (
+    args: { pairId: string, owner: Owner | number, limit: number },
+    ): Map<string, OrderSidesArrays<any>> => {
+    const { pairId, owner, limit } = args;
+    const includeOwnOrders = owner !== Owner.Peer;
+    const includePeerOrders = owner !== Owner.Own;
```
sangaman: I don't know if we'll ever add or change the options of Own/Peer/Both, but just in case it might be good to have this logic check "if Peer or Both" for includePeerOrders and "if Own or Both" for includeOwnOrders.

krrprr: Alright, sounds reasonable, made the changes :)

PR #1394, opened 2020-02-17
fix(lnd): auto unlock consistently

sangaman:
This fixes the logic around automatically unlocking lnd if it is restarted and comes back online in a locked state while xud is running. Previously, it would be possible for lnd to go down, come back online, get unlocked, then quickly go down again leaving it in a `WAITING_UNLOCK` status. If lnd comes back online while still in this status, it would not be automatically unlocked. This fix ensures that lnd would be unlocked even in this case. Although it is somewhat of an edge case, it can manifest when lnd is unable to communicate with its backing node as its behavior is to immediately shut down after unlock when that is the case.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -224,8 +224,9 @@ class LndClient extends SwapClient {
 
     if (!this.isWaitingUnlock()) {
       await this.setStatus(ClientStatus.WaitingUnlock);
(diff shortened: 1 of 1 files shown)


erkarl:
Nice! I believe I ran into this edge case on Friday, but wasn't quite sure on how to reproduce it. This makes a lot of sense.

PR #1395, opened 2020-02-18
feat(peer): add native tor support

ghost:
This commit adds support for Tor's SOCKS5 proxy. This allows us to remove `proxychains4` dependency within xud-docker that is causing the thread to block whenever an outbound connection attempt is made.

A new dependency [socks](https://github.com/JoshGlazebrook/socks) is added along with a new configuration option `p2p.torport` to specify the port that Tor's exposed SOCKS5 proxy is listening on.

I consciously did not add support for `torhost` since I'm assuming that the proxy will be running on localhost.

### DIFF
--- a/bin/xud
+++ b/bin/xud
@@ -154,6 +154,10 @@ const { argv } = require('yargs')
       default: undefined,
       alias: 't',
     },
+    'p2p.torport': {
+      describe: 'Port that Tor\'s exposed SOCKS5 proxy is listening on.',
+      type: 'number',
+    },
     'raiden.disable': {
       describe: 'Disable raiden integration',
       type: 'boolean',
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -18,6 +18,7 @@ import { isPacketType, isPacketTypeArray, ResponseType } from './packets/Packet'
 import * as packets from './packets/types';
 import Parser from './Parser';
 import { Address, NodeConnectionInfo, NodeState } from './types';
+import { SocksClient, SocksClientOptions } from 'socks';
 
 /** Key info about a peer for display purposes */
 type PeerInfo = {
@@ -242,7 +243,15 @@ class Peer extends EventEmitter {
    * Prepares a peer for use by establishing a socket connection and beginning the handshake.
    * @returns the session init packet from beginning the handshake
    */
-  public beginOpen = async ({ ownNodeState, ownNodeKey, ownVersion, expectedNodePubKey, retryConnecting = false }:
+  public beginOpen = async (
+    {
+      ownNodeState,
+      ownNodeKey,
+      ownVersion,
+      expectedNodePubKey,
+      retryConnecting = false,
+      torport,
+    }:
     {
       /** Our node state data to send to the peer. */
       ownNodeState: NodeState,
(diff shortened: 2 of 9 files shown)


bin/xud
```diff
@@ -154,6 +154,11 @@ const { argv } = require('yargs')
       default: undefined,
       alias: 't',
     },
+    'p2p.torSocks': {
+      describe: 'Port that Tor\'s exposed SOCKS5 proxy is listening on.',
+      type: 'number',
+      default: undefined,
```
sangaman: I don't think it hurts, but I don't think the default is needed for numbers. Undefined is the built in default.

bin/xud
```diff
@@ -154,6 +154,11 @@ const { argv } = require('yargs')
       default: undefined,
       alias: 't',
     },
+    'p2p.torSocks': {
```
sangaman: What about `socksport`? It could also be `torsocksport` or `torport`. Either way I think the "port" qualifier is helpful.

Even if you'd rather keep the name, we should make it all lower caps (`torsocks`) just to be consistent with the other arg names.

lib/p2p/Peer.ts
```diff
@@ -173,7 +174,7 @@ class Peer extends EventEmitter {
   /**
    * @param address The socket address for the connection to this peer.
    */
-  constructor(private logger: Logger, public address: Address, network: Network) {
+  constructor(private logger: Logger, public address: Address, network: Network, private torSocks?: number) {
```
sangaman: I don't think we need to make `torSocks` a member field for each peer, right? I think it makes more sense to be passed into the `beginOpen` method, then onto the `initConnection` method where it's used.

lib/p2p/Peer.ts
```diff
@@ -464,14 +465,46 @@ class Peer extends EventEmitter {
       const startTime = Date.now();
       let retryDelay = Peer.CONNECTION_RETRIES_MIN_DELAY;
       let retries = 0;
-
-      this.socket = net.connect(this.address.port, this.address.host);
       this.inbound = false;
       this.connectionRetriesRevoked = false;
 
+      const connectViaProxy = () => {
+        const proxyOptions = {
```
sangaman: I think it's a bit better to do `const proxyOptions: SocksClientOptions = {` here to check that our types are correct rather than `as SocksClientOptions` further below. It'll require to add the `!` operator to `torSocks` or to make an upfront check to make sure `torSocks` has a value, but I think it'd be more robust going forward that way.

sangaman (changes requested):
Looking good, I made some minor comments/suggestions.

kilrau:
In your tests closing https://github.com/ExchangeUnion/xud-docker/issues/241 ? If so, let's add that to the PR description.

bin/xud
erkarl: I ended up going with `torport`.

lib/p2p/Peer.ts
erkarl: You're right, good thinking. Changed.

erkarl:
> In your tests closing ExchangeUnion/xud-docker#241 ? If so, let's add that to the PR description.

That will require a separate PR to remove proxychains4 integration. It's trivial to do, will throw up the PR as soon as this is merged.

erkarl:
Ready for round 2.

erkarl:
Tests are failing. Will fix shortly.

sangaman (approved):
LGTM, thank you!

PR #1398, opened 2020-02-20
refactor: only verify peer currencies if connected

sangaman:
This modifies the logic around verifying whether peer currencies are tradable to skip over any currencies for which our local swap client is currently disconnected. This prevents foreseeable failures when we try to send or receive with a swap client that is offline.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -793,7 +793,7 @@ class OrderBook extends EventEmitter {
 
   /**
    * Verifies the advertised trading pairs of a peer. Checks that the peer has advertised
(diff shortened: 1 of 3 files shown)


sangaman:
> Do you think in the future it makes sense to block these calls on a `SwapClient` level with the exception of the call we use to verify the connection?

I was thinking about this, but what I got kind of stuck on is what we do with the blocked calls in that case. If we throw an exception, that's not much different than the current behavior where making a call on a connected client results in an exception due to connection error or similar. Returning undefined or something along those lines would be a separate case that needs handling, as typically these calls are expecting a result.

PR #1399, opened 2020-02-20
refactor: organize/clarify canRouteToPeer logic

sangaman:
This commit attempts to better organize and clarify the `hasRouteToPeer` logic used to determine whether any possible route exists to pay a peer for a given currency before enabling trading pairs involving that currency for that peer.

Previously, lnd and raiden specific logic was located inside the SwapClientManager class. This has been refactored into the respective LndClient and RaidenClient classes.

The naming of the method is also changed from "hasRoute" to "canRoute" to indicate that it only tests whether such a route is possible, not whether one necessarily exists or is functional. Methods have been commented to better indicate what they do.

### DIFF
--- a/lib/raidenclient/RaidenClient.ts
+++ b/lib/raidenclient/RaidenClient.ts
@@ -331,6 +331,11 @@ class RaidenClient extends SwapClient {
     }
   }
 
+  public canRouteToNode = async (destination: string, currency: string) => {
+    const route = await this.getRoute(1, destination, currency);
+    return route !== undefined;
+  }
+
   public getHeight = async () => {
     return 1; // raiden's API does not tell us the height
   }
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -818,6 +818,14 @@ class LndClient extends SwapClient {
     return route;
   }
 
+  public canRouteToNode = async (_destination: string) => {
+    // lnd doesn't currently have a way to see if any route exists, regardless of balance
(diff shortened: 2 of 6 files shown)


lib/raidenclient/RaidenClient.ts
```diff
@@ -331,6 +331,11 @@ class RaidenClient extends SwapClient {
     }
   }
 
+  public canRouteToNode = async (destination: string, currency: string) => {
+    const route = await this.getRoute(1, destination, currency);
```
erkarl: I think you missed `peer.raidenAddress && swapClient.isConnected()` from previous logic here.

sangaman: The check for `peer.raidenAddress` is done in `canRouteToPeer` when we call `peer.getIdentifier` and return right away if the result is undefined. `getIdentifer` will return the raiden address when it's called with a raiden swapClient.

As for checking `isConnected`, I worked on this PR right after #1398, which ensures we are connected before even attempting this call. You're right that it's missing in this PR and I should have kept it or at least made note of it here, in my mind I was taking the changes for granted. But now that the other one is merged I don't think it's necessary.

However, in thinking about this some more there's still a possibility for this call to throw an unhandled exception even with the other PR's changes. We can suddenly become disconnected from the client and make this call before the status updates, or we can hit some other unexpected exception. So what I'm going to do is wrap the `canRouteToNode` call with a try/catch that returns `false` in case of an exception - the logic being that if we can't make this check then we can't route payments in the current state.

Issue #1356, opened 2019-12-22
Simulation tests dev roadmap - beta.4


LePremierHomme:
This issue is to track the simulation tests dev roadmap and discussions. Tasks are ordered by priority:

- [x] dockerize the simulation tests
- [x] support `npm run test:sim:logs`
- [x] support `npm run test:sim:run:security` 
- [x] support `npm run test:sim:run:instability`
- [x] support faster test code builds
- [x] update README.md and provide npm scripts for test developer 
- [x] before allowing RPC calls, verify that xud has finished the initialization phase (not returning "xud is starting" error)
- [x] when custom-xud is being downloaded/installed, extend the test context lifetime 
- [x] cache custom-xud downloads
- [x] debugging: export lnd logs
- [x] security/coverage: https://github.com/ExchangeUnion/xud/issues/1126
- [ ] coverage: https://github.com/ExchangeUnion/xud/issues/1016
---
- [x] fix `TestSecurityUnsettledChannels/maker_shutdown_after_1st_htlc`
- [x] fix `TestInstability/maker_crashed_after_send_payment_with_delayed_settlement`
- [x] run above two on travis: https://github.com/ExchangeUnion/xud/issues/1220
---
- [ ] test infra stability: https://github.com/ExchangeUnion/xud/issues/1315

erkarl:
For dockerizing the test environment a good first step would be to create a `geth` container from these scripts:
https://github.com/ExchangeUnion/xud/blob/master/test/simulation/install.sh#L6-L11

https://github.com/ExchangeUnion/xud/blob/master/test/simulation/install.sh#L9 should go into a separate `raiden` container.

kilrau:
Regarding our discussion whether to integrate xud-docker into the dockerized similuation tests: let's hold off on this. While it will be somehow possible, I'm afraid that it'll set us back for quite some time. Getting e.g. geth and contracts dockerized and tests stable is prio #1292 

Could you open a WIP PR so we can track status? @LePremierHomme

kilrau:
I want to add one requirement though: it has to be easily possible to run the tests not only with different xud versions, but also the same way with different raiden & lnd versions. The raiden team should be able to use our tests for developing raiden the same way we use our tests for developing xud.

kilrau:
As discussed on discord, we should also consider using bitcoind & litecoind regtest to replace btcd and especially ltcd. You can add this to the requirement list above @LePremierHomme

sangaman:
[...]

PR #1403, opened 2020-02-23
test: fix simtest infra 

LePremierHomme:
### Changes
* enable security tests run, via `npm run test:sim:run:security`
* enable instability tests run, via `npm run test:sim:run:instability`
* enable logs fetch (to `test/simulation/temp/logs`), print (via `npm run test:sim:logs`) and cleanup (part of `npm run test:sim:clean`)
* add some assertion to the test code

### Open issues
2 tests are currently failing:
* TestSecurityUnsettledChannels/maker_shutdown_after_1st_htlc
* TestInstability/maker_crashed_after_send_payment

They were not part of the CI recently, so i'm not sure when this break occurred.
Investigating the issues was too slow, so before continuing with it, I rather finish on an optimization that would allow faster builds: separating the node env/dependencies installation (inside docker) from the test code copy (to docker), by using a separate image/container. 

Since this PR is about fixing the test infra, I don't think that the tests failure is blocking.

--- 

Part of the #1356 epic.


test/simulation/xud_test.go
```diff
@@ -377,6 +377,9 @@ func launchNetwork(noBalanceChecks bool) (*xudtest.NetworkHarness, func()) {
 		log.Fatalf("cannot start xud network: %v", err)
 	}
 
+	// Wait a bit to avoid calls to xud nodes while still initializing.
+	time.Sleep(3 * time.Second)
```
sangaman: It would be good to handle this with a check & retry approach rather than a single wait. Depending on the system 3 seconds could be too much and slow down the tests, but in some cases it could be not enough and cause flake. It's probably not a big deal and doesn't have to be changed in this PR necessarily, but it'd be nice to at least have a TODO here or maybe add to the list of sim test improvements.

test/simulation/xud_test.go
```diff
@@ -137,7 +137,7 @@ func TestInstability(t *testing.T) {
 
 	for _, testCase := range instabilityTestCases {
 		success := t.Run(testCase.name, func(t1 *testing.T) {
-			ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
sangaman: How did you settle on 5 minutes here?

test/simulation/xud_test.go
LePremierHomme: I actually encountered this just once (xud returning `xud is starting` error), so I wasn't thinking of doing a more robust solution. 

I'll add it as a task to the roadmap.

LePremierHomme: I actually encountered this just once (xud returning "xud is starting" error), so I wasn't thinking of doing a more robust solution.

I'll add it as a task to the roadmap.

test/simulation/xud_test.go
LePremierHomme: I added it as a temporary step - 
The reason is that whenever a custom xud is being downloaded and installed, the test context lifetime is being exhausted, so the first RPC call will get a timeout. Best solution would be to extend the context whenever we're downloading/installing a custom xud, but I didn't want to deal with it in this PR.

Added to the roadmap.

LePremierHomme:
> I worked on some of the security/stability tests so I might be able to help with the tests that are currently failing.

That would be great. Let's wait first for me to do the refactoring for faster builds, so this kind of task won't be too annoying, as it currently is. Gonna work on it tomorrow.

test/simulation/xud_test.go
LePremierHomme: We can also define the required branches for the test suit before the test execution starts, but this might be a bit clumsy. I'll try it out when i'll work on it.

PR #1409, opened 2020-02-25
refactor(swapclient): status and initialization

sangaman:
This commit makes several related changes to the `SwapClient` class, especially in regards to managing the status of a client, with the goal of making future changes and adding new `SwapClient` implementaions or statuses simpler. Key changes include:

- Removing the logic that would enable and disable timers such as the 
`reconnectionTimer` and `updateCapacityTimer` based on the status. Instead, the timers are always running, but only take action if the status matches a specified condition.

- Making all status change calls synchronous. When a particular status involves an asynchronous component, a method that wraps that status change with the asynchronous logic can be used instead (see next point).

- A new `setConnected` method that handles all steps related to marking a client as connected, including updating its capacity.

- An inherited `init` method that handles generic steps around initializing a swap client. It calls an abstract `initSpecific` method that is implemented by each class extending `SwapClient` to contain the client-specific logic for initialization.

- Making the raiden client optional within the swap client manager. When disabled, the raiden client will not be initialized at all.

### DIFF
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -102,10 +102,9 @@ class Xud extends EventEmitter {
 
       const nodeKeyPath = NodeKey.getPath(this.config.xudir, this.config.instanceid);
       const nodeKeyExists = await fs.access(nodeKeyPath).then(() => true).catch(() => false);
-      const awaitingCreate = !nodeKeyExists && !this.config.noencrypt;
 
       this.swapClientManager = new SwapClientManager(this.config, loggers, this.unitConverter);
-      await this.swapClientManager.init(this.db.models, awaitingCreate);
+      await this.swapClientManager.init(this.db.models);
 
       let nodeKey: NodeKey | undefined;
       if (this.config.noencrypt) {
@@ -184,7 +183,7 @@ class Xud extends EventEmitter {
         shutdown: this.beginShutdown,
       });
 
-      if (this.swapClientManager.raidenClient.isOperational()) {
+      if (this.swapClientManager.raidenClient?.isOperational()) {
         this.httpServer = new HttpServer(loggers.http, this.service);
         await this.httpServer.listen(
           this.config.http.port,
@@ -217,7 +216,7 @@ class Xud extends EventEmitter {
     const closePromises: Promise<void>[] = [];
 
     if (this.swapClientManager) {
-      closePromises.push(this.swapClientManager.close());
+      this.swapClientManager.close();
     }
     if (this.httpServer) {
       closePromises.push(this.httpServer.close());
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -43,6 +43,7 @@ class LndClient extends SwapClient {
   private lightning?: LightningClient;
   private walletUnlocker?: WalletUnlockerClient;
   private invoices?: InvoicesClient;
+  /** The path to the lnd admin macaroon, will be undefined if `nomacaroons` is enabled */
   private macaroonpath?: string;
   private meta = new grpc.Metadata();
   private uri!: string;
@@ -75,7 +76,7 @@ class LndClient extends SwapClient {
     { config, logger, currency }:
     { config: LndClientConfig, logger: Logger, currency: string },
   ) {
-    super(logger);
+    super(logger, config.disable);
     this.config = config;
     this.currency = currency;
     this.finalLock = config.cltvdelta;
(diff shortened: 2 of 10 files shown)


lib/Xud.ts
```diff
@@ -184,7 +183,7 @@ class Xud extends EventEmitter {
         shutdown: this.beginShutdown,
       });
 
-      if (this.swapClientManager.raidenClient.isOperational()) {
+      if (this.swapClientManager.raidenClient?.isOperational()) {
```
erkarl: Really like this new optional chaining syntax :).

lib/swaps/SwapClient.ts
```diff
@@ -80,7 +80,8 @@ abstract class SwapClient extends EventEmitter {
   public abstract readonly finalLock: number;
   public abstract readonly type: SwapClientType;
   /** Time in milliseconds between attempts to recheck connectivity to the client. */
-  public static readonly RECONNECT_TIME_LIMIT = 5000;
+  public static readonly RECONNECT_INTERVAL = 5000;
+  // protected disable: boolean;
```
erkarl:
```suggestion
```

lib/lndclient/LndClient.ts
```diff
@@ -155,8 +146,8 @@ class LndClient extends SwapClient {
       this.credentials = grpc.credentials.createSsl(lndCert);
     } else {
       this.logger.error(`could not load tls cert from ${certpath}, is lnd installed?`);
-      await this.setStatus(ClientStatus.Misconfigured);
-      this.initRetryTimeout = global.setTimeout(this.init, 5000, awaitingCreate);
+      this.setStatus(ClientStatus.Misconfigured);
+      this.initRetryTimeout = setTimeout(this.init, 5000);
```
erkarl: Would it make sense to use `SwapClient.RECONNECT_INTERVAL` here?

lib/swaps/SwapClient.ts
```diff
@@ -142,43 +143,72 @@ abstract class SwapClient extends EventEmitter {
...
 
-  protected setStatus = async (status: ClientStatus): Promise<void> => {
+  public init = async () => {
+    // up front checks before initializing client
+    if (!this.isNotInitialized() && !this.isMisconfigured()) {
+      // we only initialize from NotInitialized or Misconfigured status
+      this.logger.warn(`can not init in ${this.status} status`);
+      return;
+    }
+    if (this.disable) {
```
erkarl: Can this be moved before the up front checks?

lib/swaps/SwapClient.ts
```diff
@@ -142,43 +143,72 @@ abstract class SwapClient extends EventEmitter {
...
+    // final steps to complete initialization
+    this.setStatus(ClientStatus.Initialized);
+    this.setTimers();
+    this.emit('initialized');
+    await this.verifyConnectionWithTimeout();
+  }
+
+  protected abstract async initSpecific(): Promise<void>;
+
+  protected setConnected = async (newPubKey?: string, newUris?: string[]) => {
```
erkarl: `newPubKey` seems to always be present when calling `setConnected`. Should we also rename it to `newIdentifier`?

lib/swaps/SwapClient.ts
```diff
@@ -142,43 +143,72 @@ abstract class SwapClient extends EventEmitter {
...
+    if (this.isConnected()) {
+      await this.updateCapacity();
+    }
+  }
+
+  private reconnectionTimerCallback = async () => {
+    if (this.status === ClientStatus.Disconnected || this.status === ClientStatus.OutOfSync || this.status === ClientStatus.WaitingUnlock) {
+      try {
+        await this.verifyConnection();
+      } catch {}
```
erkarl: Are we sure about silencing all errors here? Maybe `.catch(this.logger.error)` or `this.logger.trace`?

lib/swaps/SwapClient.ts
sangaman: Yes, it should be `newIdentifier`, but it won't always have a value. When the identifier doesn't change, it will be undefined. That logic hasn't changed in this PR.

lib/swaps/SwapClient.ts
sangaman: I did think about this some and wasn't too sure, I can put in a debug line. I think any errors are already already being logged elsewhere so I didn't want to pollute the logs.

lib/swaps/SwapClient.ts
sangaman: I can, does it make a difference though?

lib/lndclient/LndClient.ts
sangaman: It sure would, good thinking.

lib/swaps/SwapClient.ts
erkarl: Shouldn't make a difference, but from semantically it feels better.

lib/swaps/SwapClient.ts
erkarl: You're right. That logic didn't change :). Sorry.

PR #1413, opened 2020-02-26
test(sim): decouple nvm/xud installation from the test files image

LePremierHomme:
Since working on the sim tests mostly involves refactoring test code, binding its deployment to docker with nvm/xud installation caused long docker build times. 

This PR decouples nvm/xud installation to a separate image, so that the test code container can consume it via volumes.

As previously, after changing test code one must run (from within the `test/simulation` dir):
```
$ docker-compose build test
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,6 +76,7 @@ seedutil/seedutil
 
 # temp simulation test directory
 test/simulation/temp
+test/docker-xud/xud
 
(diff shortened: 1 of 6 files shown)


LePremierHomme:
Yes, it's is required after any change to the test code (not just the docker setup code). 
I will add it as an npm script and update the README in the next PR.

PR #1421, opened 2020-03-01
test(sim): update npm scripts & README.md

LePremierHomme:
Simple update to npm scripts & `README.md`.

* I haven't included instructions to install docker/docker-compose. Do you think it's necessary?
* @sangaman you previously asked to change the script `test:sim:run` to `test:sim`, but I found it again a bit misleading, because if one will run it before doing a build, it will fail (not because of the missing docker builds, which will be built anyway, but because the missing xud files under `temp/xud`, which are copied by `test:sim:build` script). That's why I'd rather have the `build`/`run` phases more explicit.


test/simulation/README.md
```diff
@@ -2,24 +2,42 @@
 
 This test suite simulates real usage of `xud` and the second-layer payment channel network clients that it interacts with. It creates chains, generates blocks, and establishes payment channels. All commands to `xud` are issued via the gRPC API. This helps ensure that `xud` integrates with other nodes and software components properly and allows for more complete and realistic testing of `xud`.
 
-## Install dependencies
+--- 
 
-The simulation tests are currently only tested and supported on Ubuntu 18.04, 19.04 & 19.10.
+The simulation tests only dependencies are Docker & Docker Compose.
```
kilrau: [Docker](https://docs.docker.com/install/) & [Docker Compose](https://docs.docker.com/compose/install/).

kilrau:
Docker links look good. Please check the travis build if it fails again.

test/simulation/README.md
```diff
@@ -2,24 +2,42 @@
 
 This test suite simulates real usage of `xud` and the second-layer payment channel network clients that it interacts with. It creates chains, generates blocks, and establishes payment channels. All commands to `xud` are issued via the gRPC API. This helps ensure that `xud` integrates with other nodes and software components properly and allows for more complete and realistic testing of `xud`.
 
-## Install dependencies
+--- 
 
-The simulation tests are currently only tested and supported on Ubuntu 18.04, 19.04 & 19.10.
+The simulation tests only dependencies are [Docker](https://docs.docker.com/install/) & [Docker Compose](https://docs.docker.com/compose/install/).
```
sangaman: grammar nit: "tests`" with apostrophe at end.

sangaman:
@kilrau We'll need you to approve the changes before this can be merged.

LePremierHomme:
Thanks for the feedback.

I added `test:sim` to combine both build and run as you suggested.

I think that dropping the `test` prefix would be a bit confusing, because we also have the concept of simnet, as contrary to simtest. So I rather leave it as is.

PR #1424, opened 2020-03-03
Connext

pedrouid:


lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,513 @@
...
+    currencyInstances,
+  }: {
+    unitConverter: UnitConverter;
+    config: ConnextClientConfig;
+    currencyInstances: CurrencyInstance[],
+    logger: Logger;
+  }) {
+    super(logger, config.disable);
+
+    this.network = config.network || '';
```
erkarl: Think the 3 values below should always be defined.

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,513 @@
...
+    if (!this.wallet) {
+      throw errors.CONNEXT_WALLET_NOT_INITIATED;
+    }
+    const wallet = this.wallet;
+    const channel = await connext.connect(this.network, {
+      ethProviderUrl: this.ethProviderUrl,
+      nodeUrl: this.nodeUrl,
+      xpub: wallet.xpub,
+      keyGen: (index: string) => wallet.keyGen(index),
+      store: new ConnextStore(new FileStorage()),
```
erkarl: If I'm reading the code correctly it will create a data directory to `./connext-store` in the runtime directory of xud? We could provide a path to xud's data directory here from config.

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,513 @@
...
+    });
+    this.channel = channel;
+    return channel;
+  }
+  /**
+   * Associate connext with currencies that have a token address
+   */
+  private setTokenAddresses = (currencyInstances: CurrencyInstance[]) => {
+    currencyInstances.forEach((currency) => {
+      if (currency.tokenAddress) {
```
erkarl: If we decide to go with 1 swap client per currency then this will also need a check for `SwapClient.Connext`.

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,513 @@
...
+  public settleInvoice = async () => {
+    // not implemented, connext does not use invoices
+  }
+
+  public removeInvoice = async () => {
+    // not implemented, connext does not use invoices
+  }
+
+  public lookupPayment = async (
+    rHash: string,
```
erkarl: `rHash` is unused.

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,513 @@
...
+    // TODO: ndeed to expose from node api
+    const res = this.channel?.isAvailable();
+    return (res as any);
+  }
+
+  private getPaymentEvents = async (
+    currency?: string,
+    destination?: string,
+  ) => {
+    let endpoint = 'payments';
```
erkarl: Lots of unused `endpoint` variables in this file. I'm assuming it's a copy paste leftover from `RaidenClient` and won't comment on it further.

lib/constants/enums.ts
```diff
@@ -2,6 +2,7 @@
 export enum SwapClientType {
   Lnd,
   Raiden,
+  Connext,
```
erkarl: Now that we have more than 1 swap client we'll need to modify `SwapRecovery` L83 to also save the recovered preimages to memory. This is used when the payment succeeded, but xud managed to crash before using the preimage to claim the second payment.

lib/db/seeds/mainnet.ts
```diff
@@ -14,22 +14,29 @@ const nodes = [
...
+  {
+    id: 'ETH',
+    swapClient: SwapClientType.Connext,
+    decimalPlaces: 18,
+    tokenAddress: '0x0000000000000000000000000000000000000000',
+  },
   {
     id: 'WETH',
-    swapClient: SwapClientType.Raiden,
+    swapClient: SwapClientType.Raiden, // needs to be configurable to support Connext
```
erkarl: I'm not quite sure if we want to have multiple swap clients per currency. Can you remind me what was the primary use case of having such functionality?

lib/swaps/SwapClientManager.ts
```diff
@@ -213,6 +226,11 @@ class SwapClientManager extends EventEmitter {
       initWalletPromises.push(keystorePromise);
     }
 
+    if (this.connextClient) {
+      const initWalletPromise = this.connextClient.initWallet(seedMnemonic);
```
erkarl: Just to be clear - this code is only called the very first time the user creates xud seed/wallet. If I'm reading the `ConnextClient` code correctly then it needs to call `initWallet` for every start. Is that correct?

If that's the case then we'll need to do changes our end to provide a seed for `ConnextClient`.

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,513 @@
...
+  TokenPaymentRequest,
+  TokenPaymentResponse,
+} from './types';
+
+const MAX_AMOUNT = Number.MAX_SAFE_INTEGER;
+
+/**
+ * A class representing a client to interact with connext.
+ */
+class ConnextClient extends SwapClient {
```
erkarl: I think this is still missing the code where we listen for incoming payments with unknown preimages to request them from xud?

ghost (commented):
First review round complete. In general I think we're headed in the right direction :+1:. As discussed earlier it would be better to keep `@connext/client` separate from xud codebase, behind a wrapper that xud's `SwapClient.Connext` will hit.

As a minor nit please try to avoid making style changes in the future as it makes the PR harder to review.

lib/db/seeds/mainnet.ts
pedrouid: The case here is that Connext and Raiden have an overlap of assets that they can swap. Meaning that if a user wants to swap either DAI or WETH it can be done either with Connext or Raiden. 

Alternatively you could duplicate the assets with different swapClients. Or just replace the Raiden client for DAI and  WETH

lib/connextclient/ConnextClient.ts
pedrouid: This will be revised once we have the HTLC functionality on our client

lib/connextclient/ConnextClient.ts
pedrouid: Correct, this can be configured into the FileStorage class. Where would you like to assign the directory?

erkarl: If we go with the REST API wrapper approach then just having it configurable via runtime arguments or config file would do.

lib/connextclient/ConnextClient.ts
kilrau: > if we decide to go with 1 swap client per currency

definitely one swap client per currency, see below

lib/db/seeds/mainnet.ts
kilrau: > Or just replace the Raiden client for DAI and WETH

That is the ultimate goal. And definitely only one swap client per asset. Until then we'd like to keep raiden in. Proposal how to do that: Leave WETH & DAI to Raiden. Give ETH & USDT to connext. Once everything is integrated, working and tested we'll also switch WETH & DAI to connext.

lib/swaps/SwapClientManager.ts
kilrau: We actually just discussed this internally. You probably explained this before, but could you remind us again:
a) which keys exactly are derived from this seed, if more than one
b) how is the key/keys used
c) why does it need a new seed on every start

lib/Config.ts
```diff
@@ -50,12 +51,13 @@ class Config {
   public logpath: string;
   public logdateformat: string;
   public network: XuNetwork;
-  public rpc: { disable: boolean, host: string, port: number };
-  public http: { host: string, port: number };
+  public rpc: { disable: boolean; host: string; port: number };
```
sangaman: As a matter of style, I'd prefer to keep things as is with regards to member delimiters here. It actually seems that [semicolons are preferred in typescript](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-delimiter-style.md) but I'd rather keep styling consistent and not change this as part of the PR.

Enforcing this as a linting rule looks like it'd require migrating from tslint to typescript-eslint, which I started working on but is not a trivial effort, so for now it'd be good to just leave these lines as is.

lib/Config.ts
```diff
@@ -147,7 +157,14 @@ class Config {
     this.lnd.BTC = {
       disable: false,
       certpath: path.join(lndDefaultDatadir, 'tls.cert'),
-      macaroonpath: path.join(lndDefaultDatadir, 'data', 'chain', 'bitcoin', this.network, 'admin.macaroon'),
+      macaroonpath: path.join(
```
sangaman: This sort of style change is less significant as it doesn't break consistency, but as @erkarl said it makes the PR harder to review and I'd prefer to limit code churn and keep any changes in this PR relevant to its purpose.

lib/Config.ts
```diff
@@ -168,20 +192,28 @@ class Config {
...
+      nodeUrl: '',
+    };
   }
 
   private static readConfigProps = async (configPath: string) => {
     let configText: string | undefined;
     try {
       configText = await fs.readFile(configPath, 'utf8');
-    } catch (err) {}
+    } catch (err) { }
```
sangaman: Same here for unnecessary style changes.

lib/Config.ts
```diff
@@ -168,20 +192,28 @@ class Config {
...
 
     let configProps: any;
     if (configText) {
       try {
         configProps = toml.parse(configText);
       } catch (e) {
-        throw new Error(`Error parsing config file at ${configPath} on line ${e.line}, column ${e.column}: ${e.message}`);
+        throw new Error(
```
sangaman: And here.

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,491 @@
...
+
+    return {
+      balance,
+      pendingOpenBalance,
+      inactiveBalance,
+    };
+  }
+
+  public tradingLimits = async (): Promise<TradingLimits> => {
+    // assume MAX_AMOUNT since Connext will re-collaterize accordingly
```
sangaman: Can you explain a bit here what you mean by re-collaterizing? Eventually we'll want a way to know what our capacity is for sending and receiving payments with connext.

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,491 @@
...
+} from './types';
+
+const MAX_AMOUNT = Number.MAX_SAFE_INTEGER;
+
+/**
+ * A class representing a client to interact with connext.
+ */
+class ConnextClient extends SwapClient {
+  public readonly type = SwapClientType.Connext;
+  public readonly finalLock = 100;
```
sangaman: Is this just a placeholder value or is this fixed with connext?

lib/swaps/SwapClientManager.ts
sangaman: Am I correct in understanding that there'll be a mnemonic & seed associated with each channel in Connext? We definitely wouldn't want to use our "master" seed which is tied to on-chain funds as is currently being done.

Would it make sense to store connext channel seeds in the xud database and pass them to connext each time xud starts up?

Karl is right that the current implementation would not be able to initialize connext, as it would only call `initWallet` on the very first time we start xud.

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,491 @@
...
+      this.setStatus(ClientStatus.Initialized);
+      this.emit('initialized');
+      await this.verifyConnectionWithTimeout();
+    }
+  }
+  /**
+   * Initiates wallet for the Connext client
+   */
+  public initWallet = async (seedMnemonic: string[]) => {
+    this.wallet = new ConnextWallet(seedMnemonic.join(' '));
```
sangaman: Following on the discussion about the mnemonic and wallet initialization, we wouldn't want to create the `ConnextWallet` here. As it's used with lnd, `initWallet` is only called the very first time we create a wallet within lnd. Rather, I'm thinking it makes sense to have the seed as part of the `ConnextClient` constructor.

lib/connextclient/ConnextClient.ts
sangaman: On a side note (and not needed to be done in this PR) it would be good to have the same logic in RaidenClient, or actually better yet within SwapClientManager. I don't think this needs to be done as part of this PR though, I can take a look at implementing it separately.

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,491 @@
...
+    return MAX_AMOUNT;
+  }
+
+  protected updateCapacity = async () => {
+    try {
+      const channelBalancePromises = [];
+      for (const [currency] of this.tokenAddresses) {
+        channelBalancePromises.push(this.channelBalance(currency));
+      }
+      await Promise.all(channelBalancePromises);
```
sangaman: It looks like this is currently pulling back all the channel balances, but it doesn't do anything with the data it receives. Would it make sense to set this as our max inbound/outbound amounts?

lib/connextclient/errors.ts
```diff
@@ -0,0 +1,77 @@
...
+  TIMEOUT: codesPrefix.concat('.3'),
+  INVALID: codesPrefix.concat('.4'),
+  SERVER_ERROR: codesPrefix.concat('.5'),
+  UNEXPECTED: codesPrefix.concat('.6'),
+  TOKEN_ADDRESS_NOT_FOUND: codesPrefix.concat('.7'),
+  INVALID_TOKEN_PAYMENT_RESPONSE: codesPrefix.concat('.8'),
+  CONNEXT_IS_UNAVAILABLE: codesPrefix.concat('.9'),
+  CONNEXT_HAS_NO_ACTIVE_CHANNELS: codesPrefix.concat('.10'),
+  CONNEXT_CLIENT_NOT_INITIATED: codesPrefix.concat('.11'),
+  CONNEXT_WALLET_NOT_INITIATED: codesPrefix.concat('.12'),
```
sangaman: "Initialized" instead of "Initiated"?

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,491 @@
...
+
+  protected verifyConnection = async () => {
+    this.logger.info('trying to verify connection to connext');
+    try {
+      const client = this.getConnextClient();
+
+      await client.isAvailable();
+
+      this.emit('connectionVerified', { newIdentifier: client.publicIdentifier });
+      this.setStatus(ClientStatus.ConnectionVerified);
```
sangaman: This is a change I made recently, but I made a `setConnected` method that handles emitting the connected event and updating the status at once - in the future if there are any additional steps involved with updating status to connected we can just update that method. Can you use that here instead of these two lines?

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,491 @@
...
+  }
+
+  protected verifyConnection = async () => {
+    this.logger.info('trying to verify connection to connext');
+    try {
+      const client = this.getConnextClient();
+
+      await client.isAvailable();
+
+      this.emit('connectionVerified', { newIdentifier: client.publicIdentifier });
```
sangaman: Apologies if this wasn't clear, `newIdentifier` would only need to be set if it actually changed from its previous value. So it would be good to store the client identifier to a temp variable before verifying the connection, then only setting a value for `newIdentifier` if the new value doesn't match the old. You can use similar logic as is currently in LndClient.ts.

lib/connextclient/types.ts
```diff
@@ -0,0 +1,107 @@
...
+ * The response for tokenPayment call.
+ */
+export type TokenPaymentResponse = TokenPaymentRequest & {
+  secret?: string;
+};
+
+/**
+ * The payload for tokenPayment call.
+ */
+export type TokenPaymentRequest = {
```
sangaman: This type and its members (with the `snake_case` variable names) are specific to Raiden and there's no reason they need to be reused for Connext. I'd say it's best that Connext types and properties/names align with the names used natively by the Connext client and/or standard javascript/typescript naming conventions (which usually doesn't use snake_case).

Edit: Same thing applies to other types in this class.

lib/connextclient/errors.ts
```diff
@@ -0,0 +1,77 @@
...
+  TIMEOUT: {
+    message: 'request timed out',
+    code: errorCodes.TIMEOUT,
+  },
+  /**
+   * For HTTP 409 responses from Connext. For token payments 409 means "the address
+   * or the amount is invalid, or there is no path to the target, or the identifier
+   * is already in use for a different payment."
+   */
+  INVALID: (message: string) => ({
```
sangaman: It seems like most (or all?) of the errors from this point down aren't being used, but are rather carryovers from the Raiden REST HTTP error codes. No need to have them here, errors should be Connext specific and only included if they're actually being used.

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,491 @@
...
+      throw errors.INVALID_TOKEN_PAYMENT_RESPONSE;
+    }
+  }
+
+  /**
+   * Deposits more of a token to an existing client.
+   * @param channel_address the address of the client to deposit to
+   * @param balance the amount to deposit to the client
+   */
+  public depositToChannel = async (
```
sangaman: If there are no plans to use this method or no corresponding call within Connext, I think we can just remove it.

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,491 @@
...
+    });
+
+    const response: TokenPaymentResponse = {
+      ...payload,
+      secret: transfer.paymentId,
+    };
+
+    if (response.secret) {
+      // remove '0x'
+      return response.secret.slice(2);
```
sangaman: Is the '0x' included in the preimage response we'll get from Connext?

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,491 @@
...
+      maxSell: MAX_AMOUNT,
+      maxBuy: MAX_AMOUNT,
+    };
+  }
+
+  /**
+   * Returns the balances available in wallet for a specified currency.
+   */
+  public walletBalance = async (_currency?: string): Promise<WalletBalance> => {
+    return { totalBalance: 0, confirmedBalance: 0, unconfirmedBalance: 0 };
```
sangaman: Do you expect that this method won't be relevant to Connext?

lib/connextclient/ConnextClient.ts
```diff
@@ -0,0 +1,491 @@
...
+  }
+  /**
+   * Creates a payment client.
+   */
+  public openChannel = async (): Promise<void> => {
+    if (!this.client) {
+      await this.initConnextClient();
```
sangaman: I'm thinking rather than call this initialization logic, this may just not be relevant to connext (as with the closeChannel method) and can do nothing or throw an exception.

lib/db/seeds/mainnet.ts
sangaman: I'll take a look at better handling the option of more than 1 client for a given currency in a separate PR.

lib/p2p/Peer.ts
```diff
@@ -216,6 +216,9 @@ class Peer extends EventEmitter {
     if (clientType === SwapClientType.Raiden) {
       return this.nodeState.raidenAddress;
     }
+    if (clientType === SwapClientType.Connext) {
```
sangaman: Might be nice to combine this and the if statements above into a switch statement, now that there are more than 2 possibilities.

lib/p2p/Peer.ts
```diff
@@ -252,20 +255,20 @@ class Peer extends EventEmitter {
...
+        ownNodeKey: NodeKey,
+        /** The version of xud we are running. */
+        ownVersion: string
+        /** The expected nodePubKey of the node we are opening a connection with. */
+        expectedNodePubKey?: string,
+        /** Whether to retry to connect upon failure. */
+        retryConnecting?: boolean,
+        /** Port that Tor's exposed SOCKS5 proxy is listening on. */
+        torport: number,
+      }): Promise<packets.SessionInitPacket> => {
```
sangaman: More style changes that I'd prefer to leave as-is.

sangaman (changes requested):
Thanks, there's a lot to process here but I think this is a good start.

Main takeaways:

- Keep style changes to existing code to a minimum to help further reviews and limit code churn. I'm looking at updating our linting setup - there are a good number of lines where this catches superfluous double whitespaces that I'd actually like to see fixed.

- There seems to be a decent amount of raiden carryover in types/errors/naming that isn't necessary.

- Wallet init and the mnemonic needs a different approach.

- We need a cleaner way to choose Connext over Raiden for tokens that are potentially supported by both, I'll look into handling this in a separate PR.

lib/connextclient/ConnextClient.ts
pedrouid: Correct, this is no longer relevant 👍

lib/connextclient/ConnextClient.ts
pedrouid: As per the other comment, WETH and DAI will use Raiden and ETH and USDT will use Connext

lib/constants/enums.ts
pedrouid: Is this still relevant if each asset only has one SwapClient?

lib/db/seeds/mainnet.ts
pedrouid: Added USDT and assigned it SwapClientType.Connext

lib/Config.ts
pedrouid: Totally, these were made before I disabled Prettier for this workspace... I will revert them

lib/connextclient/errors.ts
pedrouid: Correct, however they will be useful for using the new REST API client that Im working on

lib/connextclient/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Installing Connext node for local development
+```bash
+git clone https://github.com/ConnextProject/indra.git
+cd indra
+make
+```
+### Running Connext node
+```bash
+make start
```
sangaman, line 11: I needed to run `docker swarm init` at before this line would work.

lib/connextclient/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Installing Connext REST API client
+```bash
+git clone https://github.com/ConnextProject/rest-api-client
+cd rest-api-client
+npm install
+```
+
+### Running Connext REST API client
+Create environment variables file and point to the Etheruem and Connext node
```
sangaman, line 27: Maybe specify to create a `.env` file here as it wasn't immediately clear to me what to do here.

lib/connextclient/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+```bash
+CONNEXT_ETH_PROVIDER_URL="http://0.0.0.0:8545"
+CONNEXT_NODE_URL="http://0.0.0.0:8080"
```
sangaman, line 31: Will this conflict with the grpc web proxy which also uses this port by default?

sangaman (commented):
It looks like I have Connext up and running but I'm not seeing any balance, I'll reach out to you tomorrow @erkarl to make sure I'm not doing anything wrong with my setup.

I still need to review all the code changes, but I noticed that `bin/xud` needs to add the new connext config options so that we can pass them in as command line args.

Also this could use a rebase or merge with master, conflicts are minimal but it would be good to test with the latest xud code.

lib/connextclient/README.md
erkarl: Interesting... added it to the readme, just in case.

lib/db/seeds/simnet.ts
```diff
@@ -51,6 +51,7 @@ const pairs = [
   { baseCurrency: 'WETH', quoteCurrency: 'BTC' },
```
kilrau, line 51:
```suggestion
  { baseCurrency: 'WETH', quoteCurrency: 'BTC' },{ baseCurrency: 'ETH', quoteCurrency: 'BTC' }
```

raladev:
Mentions of the raiden in (which should not be if we remove it):
1. xucli help (xucli restore [backup_directory]   [raiden_database_path])
2. xucli getinfo (General XUD info - Raiden: Could not fetch status informatio)

Also:
1 Connecxt container is running instead of Ready in docker setup for status command (getinfo returns Ready state)
2. xud container too in Container is running state instead of Ready/Waiting, even after restart. logs does not contain errors or something that might attract attention(getinfo returns Not connected any peers in status field)

erkarl:
Closing in favour of: https://github.com/ExchangeUnion/xud/pull/1424

PR #1430, opened 2020-03-10
feat(swapclient): add Locked status

### LINKED ISSUE #1411: Add more sophisticated lnd status' between "unlocked" and "connection verified"
Reason: we all too often unlock lnd successfully, but then it still takes a while until lnd's grpc actually becomes responsive or stops throwing `transient` & similar errors. Or it never becomes responsive because of some other errors.

We want to let our users know by having state's expessing if lnd is in a "normal" startup procedure or if something indeed is wrong.

sangaman:
This adds a status to the swap clients for after a client has been unlocked but before we have verified its connectivity. There can be additional initialization steps after a client has been unlocked and a possibility that it will encounter an error and shutdown, so we can not assume that the client will be available and set the status to `ConnectionVerified`. However, leaving the status as `WaitingUnlock` can be misleading or lead to additional unlock attempt that will fail. The new `Unlocked` status prevents further unlock attempts.

Closes #1411.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -210,6 +210,22 @@ class LndClient extends SwapClient {
     this.emit('locked');
   }
 
+  /** Lnd specific procedure to mark the client as unlocked. */
+  private setUnlocked = () => {
+    // we should close and unreference the wallet unlocker service when we set the status to Unlocked
+    if (this.walletUnlocker) {
+      this.walletUnlocker.close();
+      this.walletUnlocker = undefined;
+    }
+
+    if (this.isWaitingUnlock()) {
+      this.setStatus(ClientStatus.Unlocked);
+    } else {
+      // we should not be calling this method we were in the WaitingUnlock status
+      this.logger.warn(`tried to set client status to WaitingUnlock from status ${this.status}`);
+    }
+  }
+
   protected updateCapacity = async () => {
     await this.channelBalance().catch(async (err) => {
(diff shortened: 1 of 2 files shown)


sangaman:
Oddly the simulation tests are passing for me locally, but not on Travis even after some restarts. I'm currently not sure what the issue could be, the code changes are relatively minimal.

lib/lndclient/LndClient.ts
```diff
@@ -210,6 +210,22 @@ class LndClient extends SwapClient {
     this.emit('locked');
   }
 
+  /** Lnd specific procedure to mark the client as unlocked. */
+  private setUnlocked = () => {
+    // we should close and unreference the wallet unlocker service when we set the status to Unlocked
```
erkarl: I'm glad we won't have to deal with this soon when Lnd moves to single service for all calls.

erkarl:
I'm also experiencing tests getting stuck with https://github.com/ExchangeUnion/xud/pull/1431 with similar output. Think the issue might not be in the changes in this PR.

lib/lndclient/LndClient.ts
sangaman: Oh yeah that would be good, where have you been following that progress? We might want to do the same as well, I had based the separate unlock service off lnd but then the more I've had to deal with it (and after discussing issues in the grpc repos) I'm realizing having grpc service per port is the way to go.

erkarl: https://github.com/lightningnetwork/lnd/issues/3982

PR #1431, opened 2020-03-11
feat: derive child seed from node key

ghost:
This PR adds support to generate child seed for the upcoming Connext SwapClient. The child seed is derived by using the hash of our node's private key + swap client enum as entropy.

### DIFF
--- a/lib/nodekey/NodeKey.ts
+++ b/lib/nodekey/NodeKey.ts
@@ -2,6 +2,8 @@ import secp256k1 from 'secp256k1';
 import { randomBytes } from '../utils/utils';
 import { promises as fs } from 'fs';
 import { createCipheriv, createDecipheriv, createHash } from 'crypto';
+import { entropyToMnemonic } from 'bip39';
+import { SwapClientType } from '../constants/enums';
 
 /**
  * A class representing an ECDSA public/private key pair that identifies an XU node on the network
@@ -112,6 +114,19 @@ class NodeKey {
     }
     await fs.writeFile(path, buf);
   }
+
+  /**
+   * Derives a child seed from the private key for the swap client
+   * @param swapClient the swap client to create the seed for
+   */
+  public childSeed = (swapClient: SwapClientType) => {
+    const privKeyHex = this.privKey.toString('hex');
+    const childSeedEntropy = createHash('sha256')
+      .update(`${privKeyHex}-${swapClient}`)
+      .digest();
+    return entropyToMnemonic(childSeedEntropy);
+  }
+
 }
 
 export default NodeKey;
--- a/lib/swaps/SwapClientManager.ts
+++ b/lib/swaps/SwapClientManager.ts
@@ -225,7 +225,10 @@ class SwapClientManager extends EventEmitter {
    * Unlocks wallets with a password.
    * @returns an array of currencies for each lnd client that was unlocked
    */
-  public unlockWallets = async (walletPassword: string) => {
+  public unlockWallets = async ({ walletPassword }: {
+    walletPassword: string,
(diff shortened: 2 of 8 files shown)


lib/swaps/SwapClientManager.ts
```diff
@@ -213,6 +214,9 @@ class SwapClientManager extends EventEmitter {
       initWalletPromises.push(keystorePromise);
     }
 
+    // TODO: pass seed to Connext SwapClient
+    console.log(`ConnextClient.initWallet: ${connextSeed}`);
```
erkarl: This temporary console.log is intentional as I wasn't able to use `_connextSeed` to silence it when using object destructuring.

sangaman: You could just leave it off the type declaration for the parameter in the method signature, right? It can still be passed in when calling the method for now, and then can be added in when it's actually being used.

Not a big deal, though.

lib/nodekey/NodeKey.ts
```diff
@@ -112,6 +114,19 @@ class NodeKey {
...
+
+  /**
+   * Derives a child seed from the private key for the swap client
+   * @param swapClient the swap client to create the seed for
+   */
+  public childSeed = (swapClient: SwapClientType) => {
+    const privKeyHex = this.privKey.toString('hex');
+    const privKeyHexHash = createHash('sha256')
+      .update(`${privKeyHex}-${swapClient}`)
+      .digest('hex');
```
sangaman: Is there any need to digest to hex? I think we could just do `digest()` and pass a `Buffer` to `entropyToMnemonic` below.

test/jest/NodeKey.spec.ts
```diff
@@ -60,4 +61,21 @@ describe('NodeKey', () => {
     const nodeKey = NodeKey.fromBytes(await randomBytes(22));
     validateNodeKey(nodeKey);
   });
+
+  test('it should derive a child seed from private key', async () => {
+    const expectedChildSeed = 'crouch cause bicycle tomato choice city produce ocean canyon rare laugh sudden close angle cage item special jeans either heavy skill bracket lift butter';
+    const nodeKey = NodeKey.fromBytes(
+      Buffer.from([
+        0x60, 0x6c, 0x9b, 0xfe,
```
sangaman: Where are these bytes from? If it's an example test published somewhere maybe we can link to it in the comments. If it's just arbitrary maybe we can use a snapshot? It's just not very clear what this is testing.

sangaman (commented):
Nice approach. Made a few comments below but it looks like this should work.

test/jest/NodeKey.spec.ts
erkarl: It's arbitrary bytes to ensure that we get the same child seed given this static master private key. I changed the assertion below to use a snapshot.

erkarl: This also ensures that we get a failing test if we change the order of `SwapClientType` enum.

erkarl:
@sangaman updated with your feedback. I removed the parts where we provided the seed for `initWallet` calls since the REST API will require a seed only for `unlockWallet` call.

package.json
```diff
@@ -132,6 +132,7 @@
   ],
   "dependencies": {
     "@exchangeunion/grpc-dynamic-gateway": "^0.3.10",
+    "bip39": "3.0.2",
```
erkarl: This is locked down to a specific version now.

sangaman (approved):
Looks excellent now, just gotta figure out the Travis issues.

michael1011 (approved):

Issue #1433, opened 2020-03-16
Add swap client per currency


kilrau:
Also enables different swap clients for ETH-ERC20:
Raiden: WETH, DAI
Connext: ETH, USDT

Related: https://github.com/ExchangeUnion/xud/issues/1406

sangaman:
#1435 is related but doesn't totally resolve this issue. I started working on an approach that initializes the db currency table according to the swap client configuration on first boot, but it was becoming complex and unnecessarily coupling parts of the code together. I also considered allowing multiple rows in the database per currency, with 1 swap client for each row and a flag that allows us to specify which one is enabled, however I also moved away from that approach.

Instead here's what I propose for handling situations where more than 1 swap client can support a given currency.

- Use our db/seeds files to specify the default swap client and token address for each currency. I expect that, especially early on, we'll want to try to stick to one per currency for the whole network and so encouraging usage of the defaults makes sense.

- Use AddCurrency/RemoveCurrency or maybe a new rpc call to change the default swap clients used, this is a little clunky perhaps but it shouldn't have to be done very often.

- We can also hardcode known token addresses for alternate swap clients in the seeds files, but not write them to the db automatically. I'm thinking this could save users from having to input token addresses manually via the cli.

Another thing I was looking at is that we may need to modify our p2p protocol and `NodeState` messages to include which swap clients we are using, so we can ensure that we're using the same one as our peers. At least this is necessary if there's any overlap between token addresses across different clients.

kilrau:
> #1435 is related but doesn't totally resolve this issue. 

Gotcha, we can definitely do this separately.

> Use our db/seeds files to specify the default swap client and token address for each currency.

Agreed.

> Use AddCurrency/RemoveCurrency or maybe a new rpc call to change the default swap clients used, this is a little clunky perhaps but it shouldn't have to be done very often.

Don't think we need this rn. If a user really wants to change that, there is always the option to change the db..

> We can also hardcode known token addresses for alternate swap clients in the seeds files, but not write them to the db automatically. I'm thinking this could save users from having to input token addresses manually via the cli.

Great idea! So the users would punch in the ticker symbol? Couldn't find an API that provides token contract addresses for the TOP ETH ERC20 tokens, so I guess we manually add e.g. the TOP100: https://etherscan.io/tokens

kilrau:
Any updates here?

sangaman:
> Don't think we need this rn. If a user really wants to change that, there is always the option to change the db..

[...]

PR #1435, opened 2020-03-17
refactor(swapclients): currency:client type check

sangaman:
EDIT: Closes https://github.com/ExchangeUnion/xud/issues/1433

This adds a check of the swap client type value for each currency before associating it with a given swap client. Currently this only applies to Raiden, which previously was being automatically associated with any currency having a token address.


kilrau:
Should we also add Raiden to the default currencies WETH & DAI?

sangaman:
> Should we also add Raiden to the default currencies WETH & DAI?

What do you mean by this? Raiden is already the default for these I believe.

PR #1440, opened 2020-03-20
test(sim): export lnd instances logs

LePremierHomme:
Export test run bitcoin/litecoin lnd instances logs under `test/simulation/temp/logs` (available for print via `npm run test:sim:logs`).

---

Part of https://github.com/ExchangeUnion/xud/issues/1356

### DIFF
--- a/test/simulation/lntest/node.go
+++ b/test/simulation/lntest/node.go
@@ -34,7 +34,7 @@ var (
 
 	// logOutput is a flag that can be set to append the output from the
 	// seed nodes to log files.
-	logOutput = flag.Bool("logoutput", false,
+	logOutput = flag.Bool("logoutput", true,
 		"log output from node n to file outputn.log")
 
 	// logPubKeyBytes is the number of bytes of the node's PubKey that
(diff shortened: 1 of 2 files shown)


LePremierHomme:
Yes, it's the same as before - it just export everything, and it's up to you to cleanup the folder whenever you want.
I thought that printing is just for cloud debugging (travis), since it's much less comfortable than manually picking the files you want to look at (instead of seeing them all at once). 
If you think otherwise, and you want filtered printing, let me know which indexes you're interested in (node name, etc.). If not, we can leave it as is.

sangaman:
> Yes, it's the same as before - it just export everything, and it's up to you to cleanup the folder whenever you want.
> I thought that printing is just for cloud debugging (travis), since it's much less comfortable than manually picking the files you want to look at (instead of seeing them all at once).
> If you think otherwise, and you want filtered printing, let me know which indexes you're interested in (node name, etc.). If not, we can leave it as is.

I think it's fine, I was mostly just wondering. I think the one thing that might be nice to do if not too much trouble is to only print the logs from the most recent run on this command.

LePremierHomme:
This will require to persist somewhere the last run id.. it will require some additional code. Let me know if you want this.

sangaman:
> This will require to persist somewhere the last run id.. it will require some additional code. Let me know if you want this.

It's OK for now, like I said not worth it if it's much work to implement.

PR #1443, opened 2020-03-27
chore: update proto definitions

ghost:
This PR updates the proto definitions so that there's no additional diff
when new changes are made. In addition we also remove the `README.md`
from `/test/simulation/xudrpc/README.md` as it was causing
confusion/duplication and using different commit versions.


sangaman (commented):
Would it make sense to keep the README and refer to the main proto/README.md file, but still keep the instructions for running gen_protos.sh? On the other hand I'm not certain that script is needed at all any longer with the dockerized setup, in which case we can just remove the script entirely, maybe in a separate commit/PR. I will investigate.

sangaman (approved):
I confirmed we do still need the generated protobufs for the simulation tests. However, to keep the tests in sync I think it's better to do it automatically rather than manually as the current README instructs, I'll open a separate PR for that.

PR #1445, opened 2020-03-27
chore: automatically generate sim test protos

sangaman:
This automatically generates the Go protobufs used by the simulation tests every time `npm run proto` is run, ensuring that the protobuf code used by xud and by its simulation tests remain in sync without manual intervention.

I ensured that the changes to xudrpc.pb.go here are identical whether running `npm run proto` or the now obsolete `gen_protos.sh` script.


sangaman:
Something is wrong with travis and it's not detecting that the tests have passed: https://travis-ci.org/github/ExchangeUnion/xud/builds/667846463

PR #1446, opened 2020-03-27
chore: update protoc-gen-swagger

sangaman:
This updates the version of `protoc-gen-swagger` to the version used by lnd. This will reduce conflicts when developing both lnd and xud in the same environment.


sangaman:
Something is wrong with github not detecting the travis build, it passed: https://travis-ci.org/github/ExchangeUnion/xud/builds/667846463

PR #1452, opened 2020-04-01
fix: update xud version on every compile

### LINKED ISSUE #1449: xud version clean
### Expected behaviour

How it should be: `1.0.0-beta.2` for the release tag
How it should be: `1.0.0-beta.2-f75f94c` for a non-release commit
How it should be `1.0.0-beta.2-f75f94c-dirty` only if xud was modified locally

### Actual behaviour

In all above scenarios: `1.0.0-beta.2-f75f94c-dirty`

sangaman:
This ensures that the version displayed by xud on `GetInfo` calls is updated every time it is compiled, to ensure that the version matches the code that is running. Previously the version was only updated on installation via npm.

Fixes #1449.


michael1011:
> Can we ignore this file?

No. What you can do locally is `git stash` to undo the unstaged changes

sangaman:
> Can we ignore this file?

Yeah we can't ignore it, but you should be able to `git stash` or `git reset --hard` if you want to get rid of the changes permanently and try again, I'm not sure why that file changed for you.

kilrau:
> > Can we ignore this file?
> 
> Yeah we can't ignore it, but you should be able to `git stash` or `git reset --hard` if you want to get rid of the changes permanently and try again, I'm not sure why that file changed for you.

It always changes for me...never figured out why.

sangaman:
> It always changes for me...never figured out why.

That's very strange, what version of npm are you using?

Does reverting those changes make the first test case pass for you? If git is reporting uncommitted changes, the `dirty` tag is to be expected.

PR #1462, opened 2020-04-08
feat(cli): automatically determine rpc host/port

### LINKED ISSUE #1451: Let xucli automatically determine gRPC port
Now that we have

mainnet: 8886
testnet: 18886
simnet: 28886

one has to specify `-p` for `xucli` when running a non-mainnet network. Let's check how we can have xucli read from the xud config file (or another way) to determine what port it's running on?

sangaman:
This enhances the cli to automatically attempt to find and read from the xud configuration file to determine the port and host for the xud gRPC service based on the configured network - which determines default port numbers - or configured port if specified. This prevents the user from having to specify the rpc port on `xucli` commands when using a non-mainnet xud instance or a non-default gRPC port via the config file.

Closes #1451.

### DIFF
--- a/lib/cli/command.ts
+++ b/lib/cli/command.ts
@@ -1,25 +1,54 @@
 import fs from 'fs';
 import grpc, { status } from 'grpc';
+import path from 'path';
 import { Arguments } from 'yargs';
+import Config from '../Config';
 import { XudClient, XudInitClient } from '../proto/xudrpc_grpc_pb';
-import { getDefaultCertPath } from './utils';
 
 /**
  * A generic function to instantiate an XU client.
  * @param argv the command line arguments
  */
-export const loadXudClient = (argv: Arguments<any>) => {
-  const certPath = argv.tlscertpath || getDefaultCertPath();
+export const loadXudClient = async (argv: Arguments<any>) => {
+  const config = new Config();
+  await config.load({
+    xudir: argv.xudir,
+    rpc: {
+      port: argv.rpcport,
+      host: argv.rpchost,
+    },
+  });
+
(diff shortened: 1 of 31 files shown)


sangaman:
> The only minor nit I have is the error message when xud is not running:
> 
> ```
> ➜  xud git:(cli/detect-network) ✗ ./bin/xucli getinfo -j | jq
> could not connect to xud at undefined:undefined, is xud running?
> ```

Good catch, I put in a quick fix to preserve the port and host we determined to use. Thanks.

erkarl:
I'm now getting:
```
➜  xud git:(cli/detect-network) ✗ ./bin/xucli getinfo
Error: 14 UNAVAILABLE: DNS resolution failed
```
using this branch (when xud is running & also closed).

lib/cli/command.ts
```diff
@@ -1,27 +1,56 @@
...
   const cert = fs.readFileSync(certPath);
   const credentials = grpc.credentials.createSsl(cert);
 
-  return new XudClient(`${argv.rpchost}:${argv.rpcport}`, credentials);
+  // in case port and host args were not set, we update them to the values we
+  // determined by loading the config
+  argv.rpcport = config.rpc.port;
+  argv.rpchost = config.rpc.host;
+
+  return new XudClient(`${argv.rpchost}}:${argv.rpcport}`, credentials);
```
sangaman, line 31: Looks like I typoed an extra `}` here after `rpchost` which is why the DNS resolution failures were happening, it was trying to resolve `localhost}` instead of `localhost`. Took me a while to figure out what was going on because I could've sworn it was working for me when I had tested it then I was getting the dns error too out of the blue. Thanks for catching this @erkarl .

PR #1467, opened 2020-04-09
feat(rpc): push amount in OpenChannel call

### LINKED ISSUE #1464: Expose push_amt for openchannel rpc call
> As discussed, we'd additionally want to expose lnd's `--push_amt` flag via grpc & xucli

https://github.com/ExchangeUnion/xud/pull/1463#pullrequestreview-390675869

This tracks the ability to open channels through xud with some balance in the channel pushed to the remote node.

sangaman:
This adds a push amount parameter for the `OpenChannel` call which is passed to lnd as the `push_sats` parameter to specify a starting balance for the remote side in a channel.

Closes #1464.


proto/xudrpc.proto
```diff
@@ -507,8 +507,10 @@ message OpenChannelRequest {
   string node_identifier = 1 [json_name = "node_identifier"];
   // The ticker symbol of the currency to open the channel for.
   string currency = 2 [json_name = "currency"];
-  // The amount of the channel denominated in satoshis.
+  // The amount to be deposited into the channel denominated in satoshis.
   int64 amount = 3 [json_name = "amount"];
+  // The balance amount to be pushed to the remote side of the channel denominated in satoshis.
+  int64 push_amount = 4 [json_name = "push_amount"];
```
michael1011, line 513: Why are we using `int64` for the amount again? `uint32` or `uint64` would be better I think since the amount can never be negative

sangaman: You're absolutely right, thanks.

Issue #1428, opened 2020-03-04
Testnet | Channel is not getting opened - as it shows an error - connect PeerAddreses failed


fagunbhavsar:
**Describe the bug**
On testnet, unable to open channel, as it throws an error - connect PeerAddreses failed

**To Reproduce**
1. Connect to AWS EC2 instance
2. Run the installation
3. Select testnet
4. Run listpeers
5. Run connect <node_id> <currency> <quantity>
6. Validate result

**Expected behavior**
Step 6 - Channel should get opened

**Actual Behaviour**
Step 6 - It throws an error - "connectPeer Addreses failed"

**Screenshots**
![Screenshot 2020-03-03 at 4 30 00 PM](https://user-images.githubusercontent.com/18718151/75861600-9e3ebf00-5e23-11ea-9853-3654763caa16.png)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

fagunbhavsar:
@kilrau Here it is!

[testnet.zip](https://github.com/ExchangeUnion/xud/files/4286892/testnet.zip)

kilrau:
Reproduced the same today on simnet: https://github.com/ExchangeUnion/xud/issues/1448 (closed as duplicate). Bumping prio.

kilrau:
Fixed by https://github.com/ExchangeUnion/xud/pull/1468

PR #1468, opened 2020-04-09
feat(lnd): attempt openchannel when connect fails

sangaman:
We currently attempt to connect to a peer's lnd node using their advertised listening addresses before opening a channel over lnd. If the connection attempt fails for any reason other than that we are already connected to the peer, then we fail the open channel attempt. However, this prematurely fails cases where a peer does not advertise listening addresses, such that connection attempts won't even be made and the `OpenChannel` call will fail right away, but is already connected to us. This could happen if a peer had previously connected to us.

Instead, we first attempt to connect to the peer's lnd node but then proceed to attempt to open a channel regardless of where we were able to connect or not. If we aren't connected to that peer, the lnd `OpenChannel` call will fail right away and the end result will be the same. On occasions where we are already connected, the call will succeed.

This also performs some minor refactoring to rename `lndUris` to `uris` so that it is more generic for supporting future swap clients that also advertise listening addresses for nodes.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -718,15 +718,14 @@ class LndClient extends SwapClient {
    * Opens a channel given peerPubKey and amount.
    */
   public openChannel = async (
-    { peerIdentifier: peerPubKey, units, lndUris, pushUnits = 0 }:
-    { peerIdentifier: string, units: number, lndUris: string[], pushUnits?: number },
+    { peerIdentifier: peerPubKey, units, uris, pushUnits = 0 }:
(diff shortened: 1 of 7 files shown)


kilrau:
> However, this prematurely fails cases where a peer does not advertise listening addresses, such that connection attempts won't even be made and the OpenChannel call will fail right away, but is already connected to us.

Do your reckon this being the reason for https://github.com/ExchangeUnion/xud/issues/1428 meaning this should close it?

sangaman:
> Do your reckon this being the reason for #1428 meaning this should close it?

It might but I don't think so, at least not in all cases. I haven't tried to reproduce that issue yet, but I did want to get this in before in case it was related. There may need to be further refinements/fixes.

PR #1469, opened 2020-04-09
test(sim): cache custom-xud downloads

LePremierHomme:
Test infra: add volume to persist custom-xud downloads, since they don't change frequently, and take a long time to fetch on every test iteration. 

Also adding some more specific build scripts and info on `README.md`.


package.json
```diff
@@ -38,11 +38,14 @@
     "test:p2p:watch": "mocha --timeout 10000 -r ts-node/register test/p2p/*  --watch --watch-extensions ts",
     "test:crypto": "mocha --timeout 10000 -r ts-node/register test/crypto/*",
     "test:sim": "(npm run test:sim:build && npm run test:sim:run)",
-    "test:sim:build": "(npm run compile:dev && mkdir -p test/simulation/docker-xud/xud && cp -a dist test/simulation/docker-xud/xud && cp -a bin test/simulation/docker-xud/xud && cp package.json test/simulation/docker-xud/xud && cd test/simulation && ./docker-build.sh)",
+    "test:sim:build": "(npm run test:sim:update-xud && cd test/simulation && ./docker-build.sh)",
     "test:sim:build:test": "(cd test/simulation && ./docker-build.sh test)",
+    "test:sim:build:xud": "(npm run test:sim:update-xud && cd test/simulation && ./docker-build.sh xud)",
```
sangaman, line 43: When should we run this script? It's not used by other scripts and I don't see it mentioned in the readme.

test/simulation/README.md
```diff
@@ -8,38 +8,63 @@ The simulation tests’ only dependencies are [Docker](https://docs.docker.com/i
...
 
 ```bash
 $ npm run test:sim:build
 $ npm run test:sim:run
 ```
 
-Whenever test code is being refactored, one must build the `test` image before re-running the tests so that changes could take effect:
+#### Specific builds
+
+For taking into effect changes in test code, one must build the `test` image before re-running the tests:
```
sangaman, line 20: This line is slightly unclear to me. How about "When making changes to the simulation test code, one must build the `test` image before rerunning the tests"

test/simulation/README.md
```diff
@@ -8,38 +8,63 @@ The simulation tests’ only dependencies are [Docker](https://docs.docker.com/i
...
 $ npm run test:sim:run
 ```
---- 
 
-The `test:sim:run` script mentioned above runs only the Integration tests suite.
+For taking into effect changes in `xud` (non-test, production code), one must build the `xud` image before re-running the tests, in addition to cleaning its volume:
+
+```bash
+$ docker volume rm simulation_xud-vol
+$ npm run test:sim:build:test
```
sangaman, line 31: Isn't this line supposed to be `npm run test:sim:build`? Or it this where `npm run test:sim:build:xud` would be appropriate?

test/simulation/README.md
```diff
@@ -8,38 +8,63 @@ The simulation tests’ only dependencies are [Docker](https://docs.docker.com/i
...
+For taking into effect changes in test code, one must build the `test` image before re-running the tests:
 
 ```bash
 $ npm run test:sim:build:test
 $ npm run test:sim:run
 ```
---- 
 
-The `test:sim:run` script mentioned above runs only the Integration tests suite.
+For taking into effect changes in `xud` (non-test, production code), one must build the `xud` image before re-running the tests, in addition to cleaning its volume:
```
sangaman, line 27: Here also I'd say "When making changes to non-test `xud` code, one must clean and build the `xud` image before rerunning the tests."

test/simulation/README.md
LePremierHomme: > Or it this where npm run test:sim:build:xud would be appropriate?

Yes, my bad.

test/simulation/README.md
LePremierHomme: The "clean" is being referred to the volume, not the image (which always get overriden, unlike the volume). This is why it's distinct.

package.json
LePremierHomme: Like you said, this is script is needed whenever non-test code is changed. 
It is included inside the `test:sim:build` procedure.

test/simulation/README.md
sangaman: I see, how about "one must build the `xud` image and clean its volume before rerunning..." ?

LePremierHomme:
> How do the cached instability and security images of xud get updated when their respective branches get updated?

If you are referring to the to the docker volume, then yes, they need to be cleared. I’ll add it as an npm script and to the `README.md` instructions as well. (we don't use images for the custom branches, but rather plain persisted disk volume, so after the first download, it won't be repeated).

> Should we have npm scripts to clean lnd and xud rather than using docker commands direclty, like test:sim:clean:xud and test:sim:clean:lnd maybe

Yes, I added that as well.

Issue #1476, opened 2020-04-13
LND/connext: uncooperative `closechannel`


kilrau:
A concern while working on the `closechannel` functionality  is that we rely on peers to tell us what their lnd pub keys are and we don't persist them anywhere. That means if they're not online or not cooperative, we won't know their pub key to close their channel with. there's also a concern that they give us someone else's pub key, and we unintentionally close a channel with someone else. So we implemented a naive version of closechannel for now. This issue is about the enhanced version, persisting the channel ids (channelid <-> node key relation) created through xud with peers so we can close them later even if the peer goes offline (which is a condition where we'd likely want to close channels).

rsercano:
Are we talking about creating a new unique id per channel or using tx id?  @kilrau @sangaman

rsercano:
I took a closer look at it, first of all keeping channels in the db requires a lot of effort to make it sync since opening and closing can happen outside of xud.

Apart from that, I guess with below sentence:
> In a second step we'd need to adjust closechannel to lookup channel data from the db and don't request any info from the peer 

You mean to take only channel id parameter for `closechannel` command? If so, why don't we implement a new grpc call `listchannels` which would print, **channel id**, **remote pub key** and **alias** and send request directly to lnd so that we get what's recent always?

In other case we would still need such a call to show channel ids to user so he/she can know what to close. 

### Therefore what I suggest for this issue:

- Implement `listchannels` and print **channel_id, remote_pub_key, alias** at least to output.
- Change `closechannel` command to take `channel_id` instead of `pub_key` and call `closechannel` from LND after finding necessary information from the list of channels by id.
- Make these both changes as much as generic so that we can use them for connext soon too. 

### So that:

- With this way peer doesn't have to be "cooperative or online" because we'll get channel ids from lnd directly and will show alias/pubkey to user so he/she will know which channel he/she is closing. Only concern can be about alias I'm not sure if we can reach it if the peer is offline @sangaman 
- I don't think somebody would unintentionally close another channel with this way, because we'll print what he/she has in terms of channels.
- We can still check peer's offline status and throw force close error. 

If that's not the intention, and I misunderstood, please let me know.

erkarl:
@rsercano thanks for doing the research on this one. 

One thing to consider is that remote peer's lnd pubkey could also change over time when running xud without docker. 

As for closing the channels I think the only thing we need is the channel point (funding transaction id + output index). Could we persist those + link it to **xud** public key?

[...]

Issue #1472, opened 2020-04-10
Connext: open/closechannel calls


kilrau:
Equivalent of https://github.com/ExchangeUnion/xud/pull/1044 / https://github.com/ExchangeUnion/xud/issues/1471

PR #1478, opened 2020-04-14
feat(rpc): CloseChannel call

### LINKED ISSUE #1471: LND: closechannel <alias>/<nodekey> <currency> [force]
for lnd only

sangaman:
This introduces the ability to close channels through xud. Currently this is only implemented for lnd, and is implemented in a naive way that relies on the peer we are trying to close channels with being online and cooperative.

Closes #1471. Follow-ups are #1472 and #1476.

I've tested this fairly extensively to ensure we are handling the responses from lnd correctly and it's working well.


lib/cli/commands/closechannel.ts
```diff
@@ -0,0 +1,30 @@
+import { Arguments } from 'yargs';
+import { CloseChannelRequest } from '../../proto/xudrpc_pb';
+import { callback, loadXudClient } from '../command';
+
+export const command = 'closechannel <node_identifier> <currency>';
```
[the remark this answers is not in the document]
sangaman: This was actually intentional since it's a flag, I'm envisioning it's usage as just adding `--force` to the command. 

It looks like I can put `closechannel <node_identifier> <currency> [--force]` here and then it'll work either by putting `true` at the end of the command or `--force` anywhere in the command, so that's what I'll go with.

kilrau:
Would yo ustill want this for the bot? @michael1011

michael1011:
> Would yo ustill want this for the bot for beta.3?

It would be nice to have. I will give this PR another review

lib/lndclient/LndClient.ts
```diff
@@ -1004,30 +1004,46 @@ class LndClient extends SwapClient {
...
+      channelPoint.setOutputIndex(outputIndex);
+      request.setChannelPoint(channelPoint);
+      request.setForce(force);
+
+      this.lightning.closeChannel(request, this.meta)
+        .on('data', (message: lndrpc.CloseStatusUpdate) => {
+          if (message.hasClosePending()) {
+            const txId = message.getClosePending()?.getTxid_asB64();
+            if (txId) {
+              this.logger.info(`channel closed with tx id ${txId}`);
```
michael1011, line 1024: Have you tested this code? Does it output the transaction id in the usual hex format? I have had some mixed experiences with getting transactions ids from the LND gRPC

sangaman: No this would actually print in base64, which is not ideal I realize as far as looking up a transaction based on what's in the logs. I'll make this use hex and also give it a quick test to ensure the output is good.

sangaman: I confirmed that this is outputting hex now.

PR #1479, opened 2020-04-14
fix: string message grpc error handling

sangaman:
This fixes an internal bug where passing a string to the `getGrpcError` method - or possibly any object without `code`, `message`, or `name` properties - would result in a gRPC exception of `TypeError: Incorrectly typed arguments to startBatch` that is unhandled and crashes xud. This ensures that the gRPC errors returned by xud are well-formed.

### DIFF
--- a/lib/grpc/getGrpcError.ts
+++ b/lib/grpc/getGrpcError.ts
@@ -75,9 +75,9 @@ const getGrpcError = (err: any) => {
 
   // return a grpc error with the code if we've assigned one, otherwise pass along the caught error as UNKNOWN
   const grpcError: grpc.ServiceError = {
-    code: code || status.UNKNOWN,
-    message: err.message,
-    name: err.name,
+    code: code ?? status.UNKNOWN,
+    message: err.message ?? (typeof err === 'string' ? err : ''),
+    name: err.name ?? 'UnknownError',
   };
 
   return grpcError;


lib/grpc/getGrpcError.ts
```diff
@@ -75,9 +75,9 @@ const getGrpcError = (err: any) => {
 
   // return a grpc error with the code if we've assigned one, otherwise pass along the caught error as UNKNOWN
   const grpcError: grpc.ServiceError = {
-    code: code || status.UNKNOWN,
-    message: err.message,
-    name: err.name,
+    code: code ?? status.UNKNOWN,
```
michael1011, line 78: The `??` operator is new in TypeScript and I am not too familiar with it yet. What's the difference between those two statements?
- `code: code ?? status.UNKNOWN,`
- `code: code || status.UNKNOWN,`

sangaman: `??` only uses the right side of the expression if the left side is actually `null` or `undefined` and not just falsey. So 

```typescript
const code = 0;
console.log(code || 'unknown'); // "unknown"
console.log(code ?? 'unknown'); // 0
```

I'm aiming to use any of those if they have a value, even if it's a 0 (which happens to be grpc.status `OK`) or an empty string, but if they're missing altogether then I set those defaults.

PR #1482, opened 2020-04-16
fix(cli): use localhost when rpc config 0.0.0.0 

sangaman:
This fixes an issue where xucli could read an rpc host of `0.0.0.0` from the config file, indicating that xud is listening on all interfaces, and then attempt to connect to it using that ip address which fails. Instead, if the xucli tool detects that xud is configured to listen on all interfaces, it should default to attempting to connect on localhost.

This above behavior would fix the issue on xud1-simnet, which listens to 0.0.0.0 for rpc commands, so that we don't need to explicitly pass `localhost` as the rpc host to xucli commands.

This also adds some error handling to proceed in case loading the xud config fails, so that we go on to attempt the command anyhow.

### DIFF
--- a/lib/cli/command.ts
+++ b/lib/cli/command.ts
@@ -13,13 +13,19 @@ import { XudClient, XudInitClient } from '../proto/xudrpc_grpc_pb';
  */
 const loadXudConfig = async (argv: Arguments<any>) => {
   const config = new Config();
-  await config.load({
-    xudir: argv.xudir,
-    rpc: {
-      port: argv.rpcport,
-      host: argv.rpchost,
-    },
-  });
+  try {
+    await config.load({
+      xudir: argv.xudir,
+      rpc: {
+        port: argv.rpcport,
+        host: argv.rpchost,
+      },
+    });
+  } catch (err) {
+    // if we can't load the config file, we should alert the user but continue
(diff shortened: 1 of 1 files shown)


lib/cli/command.ts
```diff
@@ -11,45 +11,67 @@ import { XudClient, XudInitClient } from '../proto/xudrpc_grpc_pb';
...
 
-  // in case port and host args were not set, we update them to the values we
-  // determined by loading the config
-  argv.rpcport = config.rpc.port;
-  argv.rpchost = config.rpc.host;
-
   return new XudClient(`${argv.rpchost}:${argv.rpcport}`, credentials);
 };
 
 export const loadXudInitClient = async (argv: Arguments<any>) => {
```
michael1011, line 45: At first glance `loadXudClient` and `loadXudInitClient` look really similar to me apart from the return statement. Could we put the parsing of CLI arguments and the config into a separate method to avoid the redundant code or am I missing something?

sangaman: You're right that they're almost identical. I'd just been leaving them like that because eventually I'd like to merge the two services into one, at which point one of these methods can be thrown away. But I looked this over again and I came up with a pretty clean/simple way to refactor and remove most duplicate code, so I'm going with that.

PR #1487, opened 2020-04-20
fix(cli): openchannel push amount

ghost:
This commit fixes a bug with the openchannel CLI command where the
`push_amount` value was disregarded and amount value was used instead.
As a result the `openchannel` calls would always fail because the
`push_amount` was too high.


kilrau:
good catch..

michael1011 (approved):

sangaman (approved):
Thanks for catching this!

Issue #1502, opened 2020-04-23
Extract simnet connext channel opening logic to a separate project


erkarl:
The logic for automatically creating simnet channels for xud currently resides in xud codebase. This issue is about extracting it to run in a separate project.

PR #1503, opened 2020-04-23
feat(simnet): automatic channels for Connext

ghost:
This commit adds logic to automatically requests ETH funds from our
simnet wallet and open a channel to Connext node. This logic is only
initialized when xud is running in simnet mode and Connext SwapClient is
configured.

The code is setup in a way that it doesn't use internal code/functions of xud. All the communication with xud happens via the gRPC calls. I've created a followup issue to extract this from xud codebase into a separate project: https://github.com/ExchangeUnion/xud/issues/1502

### DIFF
--- a/lib/utils/simnet-connext-channels.ts
+++ b/lib/utils/simnet-connext-channels.ts
@@ -1,0 +1,187 @@
+import { from, of, empty, Observable, combineLatest, throwError } from 'rxjs';
+import { mergeMap, concat, map, catchError, delay, repeat, retryWhen, take } from 'rxjs/operators';
+import { XudClient } from '../proto/xudrpc_grpc_pb';
+import grpc from 'grpc';
+import http from 'http';
+import { loadXudClient } from '../cli/command';
+import {
+  GetBalanceRequest,
+  GetBalanceResponse,
+  GetInfoRequest,
+  GetInfoResponse,
+  OpenChannelRequest,
+  OpenChannelResponse,
+ } from '../proto/xudrpc_pb';
+
+type Balances = {
+  channelBalance: number;
+  walletBalance: number;
+};
+
+const processResponse = (resolve: Function, reject: Function) => {
+  return (error: grpc.ServiceError | null, response: any) => {
+    if (error) {
+      reject(error);
+    } else {
+      resolve(response);
+    }
+  };
+};
+
+const getBalance = async (client: XudClient, currency?: string): Promise<GetBalanceResponse> => {
+  const request = new GetBalanceRequest();
+  if (currency) {
+    request.setCurrency(currency.toUpperCase());
+  }
+  const balances = await new Promise((resolve, reject) => {
+    client.getBalance(request, processResponse(resolve, reject));
+  });
+  return balances as GetBalanceResponse;
+};
+
+const openConnextChannel = async (client: XudClient, currency: string, amount: number): Promise<OpenChannelResponse> => {
+  const request = new OpenChannelRequest();
+  request.setCurrency(currency.toUpperCase());
+  request.setAmount(amount);
+  const openChannelResponse = await new Promise((resolve, reject) => {
+    client.openChannel(request, processResponse(resolve, reject));
+  });
+  return openChannelResponse as OpenChannelResponse;
+};
+
+const checkBalanceObservable = (client: XudClient, currency: string, minimumWalletBalance: number): Observable<Balances> => {
+  return from(
+    getBalance(client, currency),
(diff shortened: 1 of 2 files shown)


lib/utils/simnet-connext-channels.ts
```diff
@@ -0,0 +1,196 @@
...
+    req.end();
+  });
+};
+
+const createSimnetChannel = (
+  { currency, minChannelAmount, minWalletAmount, retryInterval }:
+  { currency: string, minChannelAmount: number, minWalletAmount: number, retryInterval: number },
+) => {
+  const client$ = from(loadXudClient({}));
+  const balances$ = client$.pipe(
```
sangaman, line 162: Is this line (along with the one above) equivalent to doing 

```
const client = await loadXudClient();
const balances$ = checkBalanceObservable(client, currency, minWalletAmount);
```

The `await` approach seems more readable to me but maybe I'm just not used to rxjs enough or whether it's ok to mix & match `await` statements like this. Or maybe the downside would be having to make `createSimnetChannel` async, and this is just analagous to using promises `then` function?

lib/utils/simnet-connext-channels.ts
```diff
@@ -0,0 +1,196 @@
...
+    mergeMap((balances) => {
+      // if the walletBalance is below the minimum threshold we error in
+      // order to trigger the faucet request
+      if (balances.walletBalance < minimumWalletBalance) {
+        return throwError(`balance under the threshold of ${minimumWalletBalance}`);
+      } else {
+        return of(balances);
+      }
+    }),
+    catchError(() => {
```
sangaman, line 78: Will this catch errors that are not related to being below balance? Why not put this faucet request directly in the `if (balances.walletBalance < minimumWalletBalance)` block?

lib/utils/simnet-connext-channels.ts
```diff
@@ -0,0 +1,196 @@
...
+
+const getBalance = async (client: XudClient, currency?: string): Promise<GetBalanceResponse.AsObject> => {
+  const request = new GetBalanceRequest();
+  if (currency) {
+    request.setCurrency(currency.toUpperCase());
+  }
+  const balances = await new Promise((resolve, reject) => {
+    client.getBalance(request, processResponse(resolve, reject));
+  });
+  return balances as GetBalanceResponse.AsObject;
```
sangaman, line 44: In general it feels a little cleaner to me to use the native grpc response/request objects rather than converting them to javascript objects. Internally I know we use `toObject` and `AsObject` in a few places but I think mainly that's where we're trying to keep grpc-related code separate from internal code that shouldn't be concerned with how our rpc layer is implemented. But since this tool is explicitly using our grpc interface I think it's appropriate here.

I don't think it applies here but for some of the the data types like `bytes` you have more control over the format/encoding of the fields using the grpc objects vs when you use `AsObject` you get less clear types like `string | Buffer`.

sangaman (commented):
Looks good mostly, just made a few comments and yes I agree that longer term it definitely makes sense for this to be separate from the main xud codebase.

lib/utils/simnet-connext-channels.ts
erkarl: > The await approach seems more readable to me but maybe I'm just not used to rxjs enough or whether it's ok to mix & match await statements like this.

I agree it looks cleaner and you certainly can convert static values to observables that will next and complete immediately.

````
const client = await loadXudClient();
of(client) // creates a static observable to be mixed with other observables
```

>  the downside would be having to make createSimnetChannel async

That was mainly the reason, yes. In addition it's considered a good practice to not use higher order function's variables for inputs. It is recommended to use the stream's input as an output for the inner observables. That makes it easier to extract the observables into smaller functions down the road. I'm partially guilty of ignoring some of these in this PR.

I'll leave it as is for now.

lib/utils/simnet-connext-channels.ts
erkarl: Very good point. Thanks. I moved this block directly under the `walletBalance` check.

lib/utils/simnet-connext-channels.ts
erkarl: > In general it feels a little cleaner to me to use the native grpc response/request objects rather than converting them to javascript objects.

For sure. It's much cleaner now with native grpc objects.

erkarl:
@sangaman thanks a lot for the review. Feedback applied with the exception of `client$` comment. I'm going to merge this so that I can provide an updated docker image for others to test.

PR #1501, opened 2020-04-22
test(sim): fix security tests

LePremierHomme:
Fixing 2 different issues which caused the security test to fail:

1. **production code**: `swaps.handleSwapTimeout` to support a possibility that the swap peer isn’t connected. Previously it caused `xud` to panic and crash, hence was an attack vector.

2. **test code**: when asserting wallet/channels balance, take into consideration the amount that was pushed when the channels were opened.

Edit by karl:
Fixes https://github.com/ExchangeUnion/xud/issues/1455

PR #1484, opened 2020-04-17
test(sim): maker lnd to crash before order settlement

LePremierHomme:
Closing https://github.com/ExchangeUnion/xud/issues/1126.

This test relies on this commit to the `instability` custom-xud branch: https://github.com/ExchangeUnion/xud/commit/34a73540a48f4428152f7b948e42b4cf4c49637a

It doesn’t run on travis yet, and in order to test it locally, one must to disable the other instability tests, because they are currently failing.

Issue #1220, opened 2019-09-12
Integrate security and instability simulations tests into travis


sangaman:
Currently, we have `security` and `instability` simulation tests that rely on the `adversarial/breakswap` and `instability` branches, respectively. These tests are not run in travis currently, and the branches they rely on need to be updated manually to keep up with `master`. They also may need to use commits from a feature branch to test new changes, which may cause issues should we have two pending PRs simultaneously that impact these tests.

Another concern is that, if we don't run these tests with every PR that could potentially cause them to break, we won't know which commits caused the failure if several were merged since the last known successful run.

This issue tracks how we can resolve these issues without excessively lengthening the time of travis builds or introducing code into the `master` branch that cause side effects or require much maintenance.

One potential solution may be to dynamically apply the changes for the modified branches when the tests are run, perhaps by attempting to cherry-pick the changes on top of the current code. Another may be to automatically update the `adversarial/breakswap` and `instability` branches whenever `master` is updated via travis script. I'd need to do some research/poking around to see how feasible these solutions might be.

LePremierHomme:
I think that applying a git patch, which will be defined in the test code, can work. It might get outdated due to the line numbers, so it will require some maintenance.

This will also spare us from downloading the custom xud client, since the canonical one will be used instead (after being patched).

What do you say?

kilrau:
fine with me. wdyt? @sangaman

sangaman:
Yeah that sounds like a good idea actually - it's the same amount of maintenance as keeping the separate branches up to date. So when we run the the other tests, we recreate the branch off the latest code and then apply the patch automatically?

LePremierHomme:
Yes.

It would be possible to create a patch per test case and to remove the reliance on cmd args for the differentiation, but I think it will introduce much higher maintenance over head, so i'll probably use one patch for all tests (same as one custom branch).

LePremierHomme:
Closed by #1507 & #1534

PR #1507, opened 2020-04-27
test(sim): create custom-xud via git patch

LePremierHomme:
Closing https://github.com/ExchangeUnion/xud/issues/1220 (except the final integration into travis).

This PR changes how `custom-xud` is created, so that instead of relying on branches and git clone, it is being created on top of the current working tree prior to every test docker image build via a git patch. For more info please see changes in `README.md`.

This will allow future travis integration, after https://github.com/ExchangeUnion/xud/pull/1501 & https://github.com/ExchangeUnion/xud/pull/1484 will be merged, and tests could pass.
For now, `travis.yml` just changed to explicitly run the integration tests suite, because the default run behaviour changed to include all the test suites (Integration, Security, Instability). 

After this PR will be merged, we can delete `adversarial/breakswap` and `instability` branches, since both of them were used in creating the patch file, in order to create one unified `custom-xud`, which will be used for all test scenarios (while using env vars for the differentiation).


package.json
```diff
@@ -38,14 +38,19 @@
...
+    "test:sim:build:xud": "(npm run test:sim:compile:xud && npm run test:sim:compile:custom-xud && cd test/simulation && ./docker-build.sh xud)",
     "test:sim:build:lnd": "(cd test/simulation && ./docker-build.sh lnd)",
-    "test:sim:run": "(cd test/simulation && ./docker-run.sh TestIntegration)",
+    "test:sim:run": "(cd test/simulation && ./docker-run.sh)",
+    "test:sim:run:integration": "(cd test/simulation && ./docker-run.sh TestIntegration)",
     "test:sim:run:instability": "(cd test/simulation && ./docker-run.sh TestInstability)",
     "test:sim:run:security": "(cd test/simulation && ./docker-run.sh TestSecurity)",
-    "test:sim:update-xud": "(npm run compile:dev && mkdir -p test/simulation/docker-xud/xud && cp -a dist test/simulation/docker-xud/xud && cp -a bin test/simulation/docker-xud/xud && cp package.json test/simulation/docker-xud/xud)",
+    "test:sim:compile:xud": "(npm run compile && mkdir -p test/simulation/docker-xud/xud && cp -a dist test/simulation/docker-xud/xud && cp -a bin test/simulation/docker-xud/xud && cp package.json test/simulation/docker-xud/xud)",
```
sangaman, line 49: Will there ever be a reason to run `compile:xud` and `compile:custom-xud` on their own? It seems to me like it's always going to in connection with `build:xud`, so I'm wondering if it makes sense to get rid of these scripts and just incorporate these commands direclty into `build:xud`. It would make the `build:xud` script quite long, but I think as it's a little confusing to have both `test:sim:compile:xud` and `test:sim:build:xud` since those are nearly identical in naming, and the README instructions don't reference any of the `test:sim:compile` scripts so they're currently undocumented.

LePremierHomme: No, there's no real reason to run them on their own. But they are part of both `test:sim:build` and `test:sim:build:xud`, so writing them inline will make things unreadable/maintainable IMO (specially because `test:sim:compile:custom-xud` is quite long). 
I would have moved them to shell scripts under `test/simulation`, but they have dependency on other npm scripts.
I would have added them to README, but they are not supposed to be used directly.

So in the overall, I leaving them as-is is the least bad option for now.
Best solution IMO would be to document the npm scripts.

PR #1508, opened 2020-04-27
Add XUC to simnet

ghost:
This PR adds XUC to simnet seeds and extends channel opener logic to automatically open XUC channel when running in simnet mode.


kilrau:
Can we remove all Raiden simnet currencies/pairs here too?

erkarl:
> Can we remove all Raiden simnet currencies/pairs here too?

Yes, removed.

sangaman (approved):
Looks good, nice approach. I do think if we're going to remove the raiden currencies, we can also go ahead and disable raiden by default in the config.

kilrau (approved):

PR #1509, opened 2020-04-28
Add aliases to list orders

### LINKED ISSUE #643: Listorders CLI output - peerpubkey

sangaman:
This resolves #643 and is a continuation of #1256.

I still need to test this but I'm opening as a draft for any conceptual review. Eventually I think it will make sense to also include alias in `SwapSuccess` and `SwapFailure` messages so we can see aliases when we perform swaps. It should be easy to reuse the same code I'm introducing here for those calls.


kilrau:
Agree, those alias additions something for you? @hatmer

raladev:
First notes:
1. isOwnOrder flag is not working; my orders marked as not mine;
2. I think we should extend Alias column in xucli;

sangaman:
> First notes:
> 
>     1. isOwnOrder flag is not working; my orders marked as not mine;
>        [Screenshot](https://user-images.githubusercontent.com/29906866/80625165-794a7f00-8a55-11ea-88ef-5d2c16958fe1.png)
> 
>     2. I think we should extend Alias column a little bit in xucli; Most of aliases have the same length; it will be good to see whole alias instead of 'AugustSu..' ; I think we can use augustSupply as max len alias;
>        ComicHalf
>        IssueMad
>        EraseChef
>        AugustSupply
>        DangerAdapt
>        [Screenshot](https://user-images.githubusercontent.com/29906866/80625531-055ca680-8a56-11ea-8a38-d39e9ab5fe80.png)

Thanks a lot for testing, could you make sure you're using the `5225c4d` commit I pushed a few hours ago when switching this PR over from draft status? I fixed a bug where all orders were being marked as not own, even if they were own. I also made the alias column 16 characters wide (from 12 previously) which is enough for AugustSupply (12 chars) + a space on each side with two characters leftover for longer aliases. It might make sense to make that even a little longer since it looks like the longest aliases would be 16 characters long before counting the spacing.

raladev:
Yeap, I missed last commit because of force-push -_-
Checked with 5225c4d for mine and peers orders; looks great.

sangaman:
I rebased and widened the `Alias` column by two more characters.

PR #1515, opened 2020-04-29
fix(swapclient): don't initialize if misconfigured

sangaman:
This ensures that we don't attempt to use an lnd client that was marked as `Misconfigured` because we were unable to load the tls.cert file for making rpc calls.

I also added logic to try and detect and prevent similar unintentional status transitions from happening in the future by defining allowed status transitions and preventing disallowed ones.

### DIFF
--- a/lib/swaps/SwapClient.ts
+++ b/lib/swaps/SwapClient.ts
@@ -183,13 +183,42 @@ abstract class SwapClient extends EventEmitter {
 
   }
 
-  protected setStatus = (status: ClientStatus): void => {
-    if (this.status === status) {
+  protected setStatus = (newStatus: ClientStatus): void => {
+    if (this.status === newStatus) {
       return;
     }
 
-    this.logger.info(`${this.constructor.name} status: ${ClientStatus[status]}`);
-    this.status = status;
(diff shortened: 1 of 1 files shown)


lib/swaps/SwapClient.ts
```diff
@@ -180,13 +183,42 @@ abstract class SwapClient extends EventEmitter {
...
+        // this is the starting status and cannot be reassigned
+        validStatusTransition = false;
+        break;
+    }
+
+    if (validStatusTransition) {
+      this.logger.info(`new status: ${ClientStatus[newStatus]}`);
+      this.status = newStatus;
+    } else {
+      this.logger.error(`cannot set status to ${ClientStatus[newStatus]} from ${ClientStatus[this.status]}`);
```
erkarl, line 220: I really like this addition. If get to this point it'll serve as a warning that there's most likely something wrong in the code.

erkarl:
Looks like the simtests failing might be a legit failure.

sangaman:
> Looks like the simtests failing might be a legit failure.

Yes I straightened it out, needed to change the `IsOperational` check in the init method to `IsNotInitialized`.

Issue #1521, opened 2020-05-05
[Discussion] List of interface enhancements 


raladev:
### Background

I worked with xud during last weeks and as a result I collected inconveniences that I had to face. There is a little improvements and conceptual that should be discussed.

### List

- **Done** ~~change addpair interface in cli from 'addpiar BTC LTC' to 'addpair BTC/LTC' because this format is used in other places.~~
- add all pairs in error message when u delete currency with existed pair (if u have 2 pairs that block deletion of the currency, error message should contains 2 pairs);  
- make discover output more useful - now it returns number of discovered peers but does not return number of autoconnected peers after discover (it can be different because of tor using);
- add plain text output in cli for removeorder/discover commands - now they have json output;
- change output of removeorder. As a user I want to know status of the operation, but instead of that i got info about amount in hold (it is additional information, not main) - It seems removed qty should be returned in response + remained qty if it was partial removal + qty in hold if it exists.
- **Not Required**  ~~add option to build orderbook using orders only from peers that have direct channels with me - as a user, I dont want to get No route errors when fill order because of channel states of another(third) user.~~
- **Boltz**  ~~Add sort of mechanism to channel opening from second side. I found a very big minus for me in this system - for a successful trade I need remote balance for one of the currencies. There are two ways to get it: 1. push amount 2. liquidity maker should open channel with me for one currency. 
First is good only if I want to get profit in arbitrage and I have balances in both currencies and **I completely trust the second side**(because he can close channel forcely and I will lose push amount).
Second is better but we have no clear mechanism to say to second part something like "I opened 3 BTC channel with u and want buy LTC periodically from u, can u please open channel with me". And second part after recivivng can see that someone have active channel with 3 BTC balance and he asked about LTC channel with equivalent balance.~~

kilrau:
>     * change addpair interface in cli from 'addpiar BTC LTC' to 'addpair BTC/LTC' because this format is used in other places.

We should allow `addpair ltc btc`, `addpair ltc/btc` both (ignore caps as usual too). Should be a quick one right? @sangaman Please open an issue for beta.5 and assign to @sangaman 



> ### Background
> 
> I worked with xud during last weeks and as a result I collected inconveniences that I had to face. There is a little improvements and conceptual that should be discussed.
> ### List
> 
>     * change addpair interface in cli from 'addpiar BTC LTC' to 'addpair BTC/LTC' because this format is used in other places.
> 
[...]

PR #1522, opened 2020-05-06
feat(cli): use ticker format for addpair

sangaman:
This modifies the `addpair` command to use the trading pair ticker format such as LTC/BTC used in different commands, while keeping the ability to specify currencies separately.

Related issue #1521.

### DIFF
--- a/lib/cli/commands/addpair.ts
+++ b/lib/cli/commands/addpair.ts
@@ -1,25 +1,34 @@
-import { Arguments } from 'yargs';
-import { callback, loadXudClient } from '../command';
+import { Arguments, Argv } from 'yargs';
 import { AddPairRequest } from '../../proto/xudrpc_pb';
+import { callback, loadXudClient } from '../command';
 
-export const command = 'addpair <base_currency> <quote_currency>';
+export const command = 'addpair <pair_id|base_currency> [quote_currency]';
 
 export const describe = 'add a trading pair';
 
-export const builder = {
-  base_currency: {
-    description: 'the currency bought and sold for this trading pair',
+export const builder = (argv: Argv) => argv
+  .positional('pair_id', {
+    description: 'the pair ticker id or base currency',
     type: 'string',
-  },
-  quote_currency: {
+  })
+  .positional('quote_currency', {
     description: 'the currency used to quote a price',
     type: 'string',
-  },
-};
+  })
+  .example('$0 addpair LTC/BTC', 'add the LTC/BTC trading pair by ticker id')
+  .example('$0 addpair LTC BTC', 'add the LTC/BTC trading pair by currencies');
 
 export const handler = async (argv: Arguments<any>) => {
   const request = new AddPairRequest();
-  request.setBaseCurrency(argv.base_currency.toUpperCase());
-  request.setQuoteCurrency(argv.quote_currency.toUpperCase());
+  let baseCurrency: string;
+  let quoteCurrency: string;
+  if (argv.base_currency && argv.quote_currency) {
+    baseCurrency = argv.base_currency;
+    quoteCurrency = argv.quote_currency;
+  } else {
+    [baseCurrency, quoteCurrency] = argv.pair_id.split('/');
+  }
+  request.setBaseCurrency(baseCurrency.toUpperCase());
+  request.setQuoteCurrency(quoteCurrency.toUpperCase());
   (await loadXudClient(argv)).addPair(request, callback(argv));
 };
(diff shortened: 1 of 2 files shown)


raladev:
@sangaman 
1. addpair BTC/LTC works
2. I cant specify currencies separately. I think that one of the following commands should work :) :
./xucli -p 28886 addpair BTC LTC
./xucli -p 28886 addpair --base_currency BTC --quote_currency LTC
./xucli -p 28886 addpair -b=BTC -q=LTC
./xucli -p 28886 addpair -b BTC -q LTC
[bad_attempts.log](https://github.com/ExchangeUnion/xud/files/4586273/bad_attempts.log)
3. also I tried ./xucli -p 28886 addpair --pair=BTC/LTC but it does not work too and I dont understand why I cant use --pair option as help saids (I think its connected with previous point).

@kilrau 
1.  [U wrote](https://github.com/ExchangeUnion/xud/issues/1521#issuecomment-624189165) that **addpair BTC LTC** should work too. **addpair -b BTC -q LTC** is enough or should it be **addpair BTC LTC**? Two interfaces by default (without flags) for one command looks a bit strange.

lib/cli/commands/addpair.ts
```diff
@@ -1,25 +1,43 @@
-import { Arguments } from 'yargs';
-import { callback, loadXudClient } from '../command';
+import { Arguments, CommandBuilder } from 'yargs';
 import { AddPairRequest } from '../../proto/xudrpc_pb';
+import { callback, loadXudClient } from '../command';
 
-export const command = 'addpair <base_currency> <quote_currency>';
+export const command = 'addpair <pair>';
 
 export const describe = 'add a trading pair';
```
kilrau, line 7:
```suggestion
export const describe = 'add a trading pair such as LTC/BTC';
```

kilrau:
Let's keep it simple. This is not worth catching 10 different ways of specifiying a pair. Only the following should work:
`addpair LTC/BTC`
`addpair `lTc/bTc`

But `xucli addpair --help` should make that clear. Please paste a screenshot of what you find confusing in this description.

> I cant use --pair option as help saids (I think its connected with previous point).
Help doesn't say that, it says `addpair <pair>`, which means `addpair ltc/btc`, it does **not** mean `addpair --pair ltc/btc` (which doesn't make much sense either).

kilrau:
Looking at the help output, you are right @raladev , these should work
```
addpair LTC BTC
addpair --base_currency LTC --quote_currency BTC
addpair -b LTC -q BTC
```

raladev:
> Help doesn't say that, it says `addpair <pair>`, which means `addpair ltc/btc`, it does **not** mean `addpair --pair ltc/btc` (which doesn't make much sense either).

it says 
 --pair              the trading pair in ticker format such as LTC/BTC        [string]
that means that --pair can be used as a flag and addpair --pair ltc/btc should work

kilrau:
Gotcha. Since it's in the help it should work, but honestly if it takes longer than 5 minutes to figure it out, I am fine with removing all of these:
```
--pair the trading pair in ticker format such as LTC/BTC [string]
--base_currency    the currency bought and sold for this trading pair [string]
--quote_currency   the currency used to quote a price                 [string]
```
and just allow the `ltc/btc` format.

sangaman:
Digging into this a bit made me realize that the approach we're using across almost every cli command is somewhat misleading in that it always prints the required, positional command arguments as "Options" in the help output. If you were to actually specify the option by passing it in as a flag instead of a positional argument, such as `xucli connect --node_uri 038395febbce...` instead of `xucli connect 038395febbce...` you'd get an error.

I can clarify those other commands in a separate PR. For now I changed this so it only uses positional arguments and that both `addpair LTC/BTC` and `addpair LTC BTC` work and the help output is clear with example commands and no misleading option descriptions.

```bash
$ ./bin/xucli addpair --help
xucli addpair <pair_id|base_currency> [quote_currency]

add a trading pair

Positionals:
  pair_id, base_currency  the pair ticker id or base currency[string] [required]
  quote_currency          the currency used to quote a price            [string]

Options:
  --help             Show help                                         [boolean]
  --version          Show version number                               [boolean]
  --rpcport, -p      RPC service port                                   [number]
  --rpchost, -h      RPC service hostname                               [string]
  --tlscertpath, -c  Path to the TLS certificate of xud                 [string]
  --json, -j         Display output in json format    [boolean] [default: false]
  --xudir, -x        Data directory for xud                             [string]

Examples:
  xucli addpair LTC/BTC  add the LTC/BTC trading pair by ticker id
  xucli addpair LTC BTC  add the LTC/BTC trading pair by currencies
```

PR #1529, opened 2020-05-07
fix: add connextUpdate listener when not operational

ghost:
This commit fixes a bug where `connextUpdate` events were not pushed to the pool. Connext client is currently dependent on NodeKey to derive the child seed and has to be initialized later. As a result the `connectionVerified` listener was not attached because the SwapClient was not operational, yet.

This is a quick fix - longer term solution would be to not depend on lnds to be operational when we create/load the NodeKey. The longer term fix would also allow us to initialize lnd wallets during runtime and decouple it from `xucli create` call.

### DIFF
--- a/lib/swaps/SwapClientManager.ts
+++ b/lib/swaps/SwapClientManager.ts
@@ -606,7 +606,7 @@ class SwapClientManager extends EventEmitter {
         }
       });
     }
-    if (this.connextClient?.isOperational()) {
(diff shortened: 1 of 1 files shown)


sangaman (approved):
This looks like the simplest/cleanest way to deal with the issue short of a way to refactor the code so we can initialize connext before the Pool and everything else.

Issue #1514, opened 2020-04-29
Print peer order match (and more)


kilrau:
## How it is:
In case of own order matches, a `match` is printed:
```
matched 0.001 ETH @ 3 with own order 29e3eeb0-8a28-11ea-90cf-439fb244cf44
```
For peer orders not:
```
simnet > buy 1 ltc/btc 0.05
swapped 1 LTC with peer order 79d2cd30-8a26-11ea-90cf-439fb244cf44
```
Which is not consistent and not very good UX since between pressing enter and the result up to 45 seconds can pass without the trader knowing what's going on under the hood.

## How it should be:
```
simnet > buy 1 ltc/btc 0.05
matched 1 LTC @ 0.05 with peer order 79d2cd30-8a26-11ea-90cf-439fb244cf44 (NodeAlias), attempting swap...
successfully swapped 1 LTC with peer order 79d2cd30-8a26-11ea-90cf-439fb244cf44 (NodeAlias)
```

How it should be (partial success, partial fail)
```
simnet > buy 1 ltc/btc 0.05
matched 0.5 LTC @ 0.05 with peer order 79d2cd30-8a26-11ea-90cf-439fb244cf44 (NodeAlias), attempting swap...
matched 0.5 LTC @ 0.05 with peer order 72166790-8a27-11ea-9b5f-0b3481cf9ca3 (NodeAlias), attempting swap...
successfully swapped 0.5 LTC with peer order 79d2cd30-8a26-11ea-90cf-439fb244cf44 (NodeAlias)
failed to swap with peer order 72166790-8a27-11ea-9b5f-0b3481cf9ca3 (NodeAlias)
remaining 0.5 LTC entered the order book as 66a79330-8a26-11ea-9b5f-0b3481cf9ca3
```

kilrau:
Another scenario to cover:
```
matched with X, swapping
swap failed due to X
continue matching route
matched with Y, swapping
success
```

Which happened today:
```
07/05/2020 10:50:06.371 [RPC] debug: received call /xudrpc.Xud/PlaceOrderSync
07/05/2020 10:50:06.373 [ORDERBOOK] debug: reduced order 3a222f80-904a-11ea-81e2-fbacbf25335a by 5000000 quantity while matching order 835a0640-9050-11ea-8db1-8fc1982a7fbc
[...]

PR #1536, opened 2020-05-07
feat(cli): use streaming PlaceOrder by default

sangaman:
This is an initial step towards resolving #1514.

Once this is in I'll work towards printing peer order matches (but not swaps) for streaming placeorder calls.

### DIFF
--- a/lib/cli/placeorder.ts
+++ b/lib/cli/placeorder.ts
@@ -134,7 +134,7 @@ const formatInternalMatch = (order: Order.AsObject) => {
 
 const formatSwapSuccess = (swapSuccess: SwapSuccess.AsObject) => {
   const baseCurrency = getBaseCurrency(swapSuccess.pairId);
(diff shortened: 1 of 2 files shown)


sangaman:
> when place order with the same local id, not well formatted error printed in xud.log
> "08/05/2020 07:58:41.756 [RPC] error: call /xudrpc.Xud/PlaceOrder errored with code undefined: undefined"
> instead of
> 08/05/2020 14:56:50.978 [RPC] error: call /xudrpc.Xud/PlaceOrderSync errored with code 6: order with local id 1 already exists

Thanks for catching this, that's very observant. It looks like the grpc error is different for the streaming version of the call, so I added some logic into how we log it to handle both cases. It's working now for me:

`12/05/2020 02:07:28.543 [RPC] error: call /xudrpc.Xud/PlaceOrder errored with code 6: order with local id 1 already exists`

PR #1541, opened 2020-05-07
feat(cli): clarify positional args & add examples

sangaman:
This clarifies the usage and help output of each `xucli` command. Previously, all arguments were described as "Options" - even the positional arguments that were actually required. Specifying these required arguments as flags (for example `xucli ban --node_identifier CheeseMonkey`) would not work.

This commit attempts to properly use yargs' `positional` method wherever applicable for arguments with a fixed position in the command syntax. It also provides examples for the various ways commands can be used.

### DIFF
--- a/lib/cli/commands/addcurrency.ts
+++ b/lib/cli/commands/addcurrency.ts
@@ -1,32 +1,33 @@
-import { Arguments } from 'yargs';
-import { callback, loadXudClient } from '../command';
-import { Currency } from '../../proto/xudrpc_pb';
+import { Arguments, Argv } from 'yargs';
 import { SwapClientType } from '../../constants/enums';
+import { Currency } from '../../proto/xudrpc_pb';
+import { callback, loadXudClient } from '../command';
 
 export const command = 'addcurrency <currency> <swap_client> [decimal_places] [token_address]';
 
 export const describe = 'add a currency';
 
-export const builder = {
-  currency: {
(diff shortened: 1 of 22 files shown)


sangaman:
> incorrect help for withdraw command (xucli withdraw [amount] [currency] [destination] [fee]), params marker as optional ([ ]) but It seems they are mandatory (<>).

For withdraw I had to accommodate the command to the fact that `--all` is an option. So I either had to change the order of the args in the command, which I ended up not doing because "withdraw X btc to address Y" is more natural than putting the amount as the last arg (where it could be optional), or I could make them all optional and allow them to either be passed as positional args or as flags. So even though `currency` and `destination` are required, they can be passed in as either positional args or with `--currency` and `--destination` flags.

> selected part of options belongs to xucli help, but not to unban help. It would be nice to print this part only for (./xucli or ./xucli help or ./xucli <bad_name_of_function>).

I think this is just tied into how yargs works, and it also does mostly make sense to me since they do apply to that command, just not that specific command, also this is more or less how I'm used to seeing the help output from other command line tools. More importantly though it's not something that can be changed easily and it's not something introduced in this PR.

raladev:
> I could make them all optional and allow them to either be passed as positional args or as flags. So even though `currency` and `destination` are required, they can be passed in as either positional args or with `--currency` and `--destination` flags.

looks good

PR #1542, opened 2020-05-07
fix(cli): cleaner message if decimal_places NaN

### LINKED ISSUE #1539: cli: Error with traceroute misuse addcurrency command
### Description

When string used as [decimal_places] parameter in addcurrency parameter, cli output returns Assertion Error

### Steps to reproduce

1. addcurrency ADH Connext  0xe69a353b3152dd7b706ff7dd40fe1d18b7802d31

### Expected behaviour

only help printed

### Actual behaviour

help + error with traceroute are printed
Screenshot

sangaman:
Fixes #1539.

### DIFF
--- a/lib/cli/commands/addcurrency.ts
+++ b/lib/cli/commands/addcurrency.ts
@@ -29,6 +29,9 @@ export const builder = {
 };
 
 export const handler = async (argv: Arguments<any>) => {
+  if (isNaN(argv.decimal_places) || argv.decimal_places >= 100 || argv.decimal_places < 0) {
+    throw 'decimal_places must be a number between 0 and 100';
+  }
   const request = new Currency();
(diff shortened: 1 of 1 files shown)


sangaman:
> The error still appears in same case without changes.
> It seems that decimal is not nan because of **0x**e69a353b3152dd7b706ff7dd40fe1d18b7802d31

Hm good catch... I had tested this with a `0x` followed by gibberish including non-hex letters, which fails, but only using 0-9 and a-f doesn't. What makes that complicated is that the hex string gets converted to a regular number before I can access the value, so I can't check whether it's a number or not. What I could do is check if `decimal_places` is over some high number like 100, because any hex string with more than 2 digits is going to exceed that. I'll make that change.

PR #1546, opened 2020-05-08
Connext minor fixes

ghost:
This PR addresses some of the comments in: https://github.com/ExchangeUnion/xud-docker/pull/429#issuecomment-625391648

- Lower the logging level for unexpected incoming transfers for Connext. This happens because `CREATED` and `UNLOCKED` events are fired from both sides (sender and receiver). Keeping the logging for debugging purposes.

- Change the `getinfo` response to return `connext simnet` instead of `connext-simnet`.

- Do not display error for incoming transfers that have the expected hash, but different units, timelock or tokenAddress. This happens when Connext is used as a swap client for both currencies.

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -140,39 +140,30 @@ class ConnextClient extends SwapClient {
 
   private onTransferReceived = (transferReceivedRequest: TransferReceivedEvent) => {
     const {
-        tokenAddress,
-        units,
-        timelock,
-        rHash,
-      } = transferReceivedRequest;
+      tokenAddress,
+      units,
+      timelock,
+      rHash,
+    } = transferReceivedRequest;
     const expectedIncomingTransfer = this.expectedIncomingTransfers.get(rHash);
     if (!expectedIncomingTransfer) {
-      this.logger.warn(`received unexpected incoming transfer with rHash ${rHash}`);
+      this.logger.warn(`received unexpected incoming transfer created event with rHash ${rHash}`);
       return;
     }
     const {
-        rHash: expectedHash,
-        units: expectedUnits,
-        expiry: expectedTimelock,
-        tokenAddress: expectedTokenAddress,
-      } = expectedIncomingTransfer;
+      units: expectedUnits,
+      expiry: expectedTimelock,
+      tokenAddress: expectedTokenAddress,
+    } = expectedIncomingTransfer;
     const currency = this.getCurrencyByTokenaddress(tokenAddress);
     if (
-        units === expectedUnits &&
-        rHash === expectedHash &&
-        timelock === expectedTimelock &&
-        tokenAddress === expectedTokenAddress
-      ) {
+      tokenAddress === expectedTokenAddress &&
+      units === expectedUnits &&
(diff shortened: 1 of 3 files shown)


lib/connextclient/ConnextClient.ts
```diff
@@ -157,17 +157,22 @@ class ConnextClient extends SwapClient {
...
-      this.logger.debug(`accepting incoming transfer with rHash: ${rHash}, units: ${units}, timelock ${timelock} and currency ${currency}`);
-      this.emit('htlcAccepted', rHash, units, currency);
-      this.expectedIncomingTransfers.delete(rHash);
+        this.logger.debug(`accepting incoming transfer with rHash: ${rHash}, units: ${units}, timelock ${timelock} and currency ${currency}`);
+        this.emit('htlcAccepted', rHash, units, currency);
+        this.expectedIncomingTransfers.delete(rHash);
+      }
     } else {
-      this.logger.error(`ignoring received pending transfer because it does not meet the requirements -
+      this.logger.warn(`ignoring received pending transfer because it does not meet the requirements -
```
sangaman, line 175: We're really only going to be logging this if `rHash !== expectedHash` so I don't think the wording of "does not meet the requirements" applies. It also seems like we won't ever be reaching this code, since if the hashes do not match then above where we do `expectedIncomingTransfers.get(rHash)` we'll get back `undefined` and return from the method. So maybe we should return the above to a warn saying we don't recognize the has, and then here we can assume `rHash === expectedHash` without explicitly checking for it as we're doing above.

Regarding the `CREATED` and `UNLOCKED` events, is there a way we can check which event we're dealing with and log accordingly? Basically if it's an expected condition, I don't even think we'd need to log a trace line. But if we're getting requests for unexpected hashes, a warning is probaly more appropriate.

lib/connextclient/ConnextClient.ts
```diff
@@ -324,7 +329,7 @@ class ConnextClient extends SwapClient {
         tokenAddress = this.tokenAddresses.get(deal.takerCurrency)!;
         const executeTransfer = this.executeHashLockTransfer({
           assetId: tokenAddress,
-          amount: `${amount}`,
+          amount: BigInt(amount).toString(),
```
sangaman, line 332: I think it makes sense to do this BigInt conversion where we initialize a value for amount a few lines above, and declare amount as string.

lib/connextclient/ConnextClient.ts
erkarl: > if the hashes do not match then above where we do expectedIncomingTransfers.get(rHash) we'll get back undefined and return from the method.

You're right. I ended up removing the `"does not meet the requirements"` conditional and the reduntant `rHash === expectedHash` check.

> Regarding the CREATED and UNLOCKED events, is there a way we can check which event we're dealing with and log accordingly?

The `onTransferReceived` function only deals with `CREATED` events. I've clarified the warn logging accordingly.

lib/connextclient/ConnextClient.ts
erkarl: Good idea, changed.

sangaman (approved):
LGTM

PR #1549, opened 2020-05-08
refactor: remove SwapCompleted phase & packet

### LINKED ISSUE #1483: Remove SwapCompleted packets & status
This issue tracks removing the redundant `SwapCompleted` and the corresponding swap status. This packet has long been a formality - it is sent only after the swap is actually settled, and recently it has caused some headaches and complexity for development purposes. Furthermore setting the complete status crates an unnecessary depency on a peer cooperating and sending us this packet.

Instead, each node should consider a swap to be completed once it has received the funds from its side of the swap. At that point, all obligations related to the swap should be considered complete for that node. We can consider the existing `PaymentReceived` swap status to mean that a swap is complete.

Note that this introduces some potentially breaking changes on the p2p communication layer, as nodes still relying on SwapCompleted packets may experience issues.

sangaman:
This removes code related to the unnecessary "SwapComplete" packet and swap phase. This brings us in line with treating a swap as "complete" once we receive our funds, `PaymentReceived` is now treated as the final phase indicating a successful swap.

Closes #1483.


erkarl:
Is this still a draft @sangaman?

sangaman:
> Is this still a draft @sangaman?

I think it's ready for review now, just not a priority.

lib/swaps/Swaps.ts
```diff
@@ -1169,13 +1163,9 @@ class Swaps extends EventEmitter {
...
       case SwapPhase.PaymentReceived:
         assert(deal.phase === SwapPhase.SendingPayment, 'PaymentReceived can be only be set after SendingPayment');
-        this.logger.debug(`Payment received for deal with payment hash ${deal.rPreimage}`);
-        break;
-      case SwapPhase.SwapCompleted:
-        assert(deal.phase === SwapPhase.PaymentReceived, 'SwapCompleted can be only be set after PaymentReceived');
         deal.completeTime = Date.now();
         deal.state = SwapState.Completed;
-        this.logger.debug(`Swap completed. preimage = ${deal.rPreimage}`);
+        this.logger.debug(`Payment received for deal with payment hash ${deal.rPreimage}`);
```
erkarl, line 1168: Should we log both the payment hash and the preimage here?

sangaman: Yes that makes sense, this is just an oversight.

Issue #1473, opened 2020-04-10
lnd on-chain deposit/withdraw renaming + connext equivalent


kilrau:
Thought about it some more: we shouldn't use lnd specific terminology, especially since `newaddress`/`sendcoins` it not *great* naming and not generic enough for e.g. connext token withdrawals

- [ ] rename [existing lnd `deposit`/`withdraw`](https://github.com/ExchangeUnion/xud/issues/1062) -> `walletdeposit`/`walletwithdraw`
- [ ] connext equivalent `walletdeposit`/`walletwithdraw` which simply returns the local accounts 0x address/sends out eth/tokens

PR #1557, opened 2020-05-14
[WIP]: fix(orderbook): fully remove matched own orders 

### LINKED ISSUE #1556: ALREADY_EXISTS error for id of maker order after self-match
### Steps to reproduce

1. Place order buy 0.1 BTC/LTC 0.1 11
2. Place order sell 0.1 BTC/LTC 0.1 12 -> full self-match
3. Place order buy 0.1 BTC/LTC 0.1 11

### Expected behaviour

Order placed to orderbook

### Actual behaviour

Error: 6 ALREADY_EXISTS: order with local id 11 already exists

### Screens

Screenshot with xud logs
Screenshot with cli output

sangaman:
This ensures that orders that are internally matched get fully removed from the order book. Previously, internally matched orders would be removed from the trading pair queues but not from the global order book mapping between local order ids and global order ids.

The `TradingPair` class now acts as an event emitter to notify its parent `OrderBook` when orders are matched and whether they were matched fully, with no remaining quantity or hold. Previously the `OrderBook` received a list of matches when calling the `match` method but wasn't aware when orders were fully matched.

Fixes #1556.

Some follow-ups that came to mind while investigating/working on this include:

- Change the `hold` vs `quantity` convention so that `quantity` only refers to the quantity available at that exact moment, excluding holds. In other words adding to `hold` would involve subtracting from `quantity`. There are a few parts where this would make the code and matching routine simpler, and it seems more intuitive to me after thinking through it some.

- Further investigate a potential bug where if the best available order is completely on hold, matching will immediately stop and not go to the next best orders. This is due to encountering no "matching quantity" with this best order, and is partly related to the bullet point above.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -146,6 +146,17 @@ class OrderBook extends EventEmitter {
     return ourTokenIdentifier === peerTokenIdentifier;
   }
 
+  private bindTradingPair = (tp: TradingPair) => {
+    tp.on('ownOrder.filled', (order, fullyRemoved) => {
+      this.pool.broadcastOrderInvalidation(order);
+      this.emit('ownOrder.filled', order);
+
+      if (fullyRemoved) {
+        this.localIdMap.delete(order.localId);
+      }
+    });
+  }
+
   private bindPool = () => {
     this.pool.on('packet.order', this.addPeerOrder);
     this.pool.on('packet.orderInvalidation', this.handleOrderInvalidation);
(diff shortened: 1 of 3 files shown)


kilrau:
>  since the bug is relatively minor (and doesn't affect swapping with peers)

I agree and think we can leave this in draft state for now. We have more urgent things to get done:
https://github.com/ExchangeUnion/xud/issues/1472
https://github.com/ExchangeUnion/xud/issues/1473

sangaman:
@raladev Done, but the sim test are still failing so this will need some more attention after other priorities are taken care of.

sangaman:
This needs some rebasing and isn't a priority but I still think is relevant, so I'm moving it to draft status for now.

sangaman:
I'm closing this since it is old and no longer worth rebasing but open a new PR soon that attempts to solve the same underlying issue.

Issue #1565, opened 2020-05-19
Track & display routing fees for executed swaps


sangaman:
We currently do not preserve how much we paid in routing fees for completed swaps. This may be relevant information to market makers, so it is something we should save to the database when completing swaps and retrieve for any relevant rpc/cli calls.

PR #1567, opened 2020-05-20
feat(rpc): TradeHistory

### LINKED ISSUE #1232: Enhancement of `listtrades`

sangaman:
This replaces the partially implemented `ListTrades` call with the more comprehensive and descriptive `TradeHistory`. This displays key info about every trade we have made.

Closes #1232.

Example output is below. I modified this a bit from earlier suggestion and I'm open to critiques on changing formatting or renaming/reordering columns. I tried to keep the data concise since we have width limitations inside a console.

```
Trades:
┌───────────────┬──────────┬───────┬─────────────┬─────────────┬───────────┬──────────────┬────────────────────────┐
│ Execution     │ Price    │ Role  │ Maker Order │ Taker Order │ Swap Hash │ Counterparty │ Executed At            │
├───────────────┼──────────┼───────┼─────────────┼─────────────┼───────────┼──────────────┼────────────────────────┤
│ Buy 0.001 LTC │ 0.01 BTC │ Taker │ b38d9b80... │ bdfb8ef0... │ c3968f... │ OnionToilet  │ 5/18/2020, 11:46:43 AM │
├───────────────┼──────────┼───────┼─────────────┼─────────────┼───────────┼──────────────┼────────────────────────┤
│ Buy 0.001 LTC │ 0.01 BTC │ Taker │ b38d9b80... │ 76af5430... │ 4cfe12... │ OnionToilet  │ 5/18/2020, 11:23:29 AM │
├───────────────┼──────────┼───────┼─────────────┼─────────────┼───────────┼──────────────┼────────────────────────┤
│ Buy 0.001 LTC │ 0.01 BTC │ Taker │ b38d9b80... │ 96b58240... │ e4c356... │ OnionToilet  │ 5/18/2020, 9:00:55 AM  │
└───────────────┴──────────┴───────┴─────────────┴─────────────┴───────────┴──────────────┴────────────────────────┘

```

There are several follow-ups that come to mind:

- Displaying network fees for swaps, which also needs to start getting tracked. This is covered by #1565.
- Retrieving only trades that were swapped with peers (no internal orders, which might make sense as the default for the cli), or only with a specific peer.
- Retrieving only trades for a specific pair, or only where we were maker or taker.
- Specifying a date range rather than a hard limit of orders. I figure the date range can outright replace the fixed limit.

I can open issues for the other items above if they make sense.


kilrau:
Nice! Actually like how you display Execution + Price better. Short and neat.

Question:
* `Executed At` is the timestamp when the trade gets marked completed in the local db, correct?




>     * Retrieving only trades that were swapped with peers (no internal orders, which might make sense as the default for the cli), or only with a specific peer.

Makes sense, please open an issue.

>     * Retrieving only trades for a specific pair, or only where we were maker or taker.

Or with one specific counterparty. A general "filter". Makes sense, please open an issue.

>     * Specifying a date range rather than a hard limit of orders. I figure the date range can outright replace the fixed limit.

Maybe, not sure since this is inconvenient to punch in via the CLI.

+ I thought about something like a `Final Balances: 1 BTC 0.1 LTC` to make it traceable how the balance in the channels changed over time. Especially if combined with

+ Filter by pair (pair only exists in JSON output)

Since this PR gives a pretty good overview alr I'd categorize all of the above as P3 and not assigned to any project yet.

sangaman:
> `Executed At` is the timestamp when the trade gets marked completed in the local db, correct?

Yes, in practice this may be some milliseconds after the swap is paid.

> Maybe, not sure since this is inconvenient to punch in via the CLI.

Good point, I hadn't really thought about how that would look like from the CLI. Particularly complicated because of variation in date formats - 1/2/2020 is either Jan 2nd or Feb 1st depending on where you are. We can leave that alone unless it comes up on its own down the road.

> I thought about something like a Final Balances: 1 BTC 0.1 LTC to make it traceable how the balance in the channels changed over time.

That would be a new thing we'd have to start tracking as well, so it would be quite a bit harder to implement than the other items which just filter existing data. But still maybe a nice feature down the road.

raladev:
@kilrau @sangaman 
question about this thing - Role  │ Maker Order │ Taker Order:

I cannot find the reason for such structure, I cant understand its advantages, it seems to me a little redundant.

If I understood correctly it is the solution for partially fills monitoring, but using of it requires additional logic for my trading bot or extra effort for visual comparation to find id of my order.

At first I need to check Role field to understand which order_id is mine (Taker or Maker) and only after that i can get order_id of my order.

In my opinion it would be better  to have following structure to remove step with Role checking:
Role  │ Own Order │ Counterparty Order

WDYT?

sangaman:
> @kilrau @sangaman
> question about this thing - Role │ Maker Order │ Taker Order:
> 
> I cant find the reason for such structure and cant understand its advantages, it seems to me a little redundant.
> 
> If I understood correctly it is the solution for partially fills monitoring, but using of it requires additional logic for my trading bot or extra effort for visual comparation to find id of my order.
> 
> At first I need to check Role field to understand which order_id is mine (Taker or Maker) and only after that i can get order_id of my order.
> 
> In my opinion it would be better to have following structure to remove step with Role checking:
> Role │ Own Order │ Counterparty Order
> 
> WDYT?

This makes sense to me, but it's a bit more complicated when dealing with internal orders. Both orders are "Own" in that case. I like your suggestion for when there *is* a counterparty, do you have a recommendation for how to display a mix of internal trades and swaps?

It might make sense to display internal trades and swapped trades in separate tables, perhaps, with separate headers? That would also make sense for when we implement filters - if there are no internal trades to show then we just don't show that table.

kilrau:
Good feedback. I like:

`Role │ Own OrderID │ Counterparty OrderID`

For internal swaps role can be `both` and then the `Own OrderID` `Counterparty OrderID` are simply the same. The field `Counterparty` should show `me`/`self`/`internal` instead of an Alias (allows external services to query/filter by this keyword instead of hardcoding an Alias if one wants filter for internal orders only).

Does that solve everthing without getting all too complicated?

sangaman:
> Good feedback. I like:
> 
> `Role │ Own OrderID │ Counterparty OrderID`
> 
> For internal swaps role can be `both` and then the `Own OrderID` `Counterparty OrderID` are simply the same. The field `Counterparty` should show `me`/`self`/`internal` instead of an Alias (allows external services to query/filter by this keyword instead of hardcoding an Alias if one wants filter for internal orders only).
> 
> Does that solve everthing without getting all too complicated?

Yeah good suggestion, I'm on board. I'll make this change, make an internal trade, then show the `tradehistory` output again.

sangaman:
How does this look? I felt having "Counterparty" and "Counterparty Order Id" widened the table too much.

```
┌───────────────┬────────────┬──────────┬─────────────┬───────────────────────────────┬───────────┬────────────────────────┐
│ Execution     │ Price      │ Role     │ Order Id    │ Order Id (Counterparty)       │ Swap Hash │ Executed At            │
├───────────────┼────────────┼──────────┼─────────────┼───────────────────────────────┼───────────┼────────────────────────┤
│ Sell 0.03 LTC │ 0.0091 BTC │ Internal │ 3aa4e8c0... │ 34dfdbc0... (Self)            │           │ 5/22/2020, 1:08:09 AM  │
├───────────────┼────────────┼──────────┼─────────────┼───────────────────────────────┼───────────┼────────────────────────┤
│ Buy 0.02 LTC  │ 0.01 BTC   │ Taker    │ 02e5d250... │ 12433120... (HarvestSolution) │ 6a742c... │ 5/22/2020, 1:06:52 AM  │
├───────────────┼────────────┼──────────┼─────────────┼───────────────────────────────┼───────────┼────────────────────────┤
│ Buy 0.001 LTC │ 0.01 BTC   │ Taker    │ bdfb8ef0... │ b38d9b80... (OnionToilet)     │ c3968f... │ 5/18/2020, 11:46:43 AM │
├───────────────┼────────────┼──────────┼─────────────┼───────────────────────────────┼───────────┼────────────────────────┤
│ Buy 0.001 LTC │ 0.01 BTC   │ Taker    │ 76af5430... │ b38d9b80... (OnionToilet)     │ 4cfe12... │ 5/18/2020, 11:23:29 AM │
├───────────────┼────────────┼──────────┼─────────────┼───────────────────────────────┼───────────┼────────────────────────┤
│ Buy 0.001 LTC │ 0.01 BTC   │ Taker    │ 96b58240... │ b38d9b80... (OnionToilet)     │ e4c356... │ 5/18/2020, 9:00:55 AM  │
└───────────────┴────────────┴──────────┴─────────────┴───────────────────────────────┴───────────┴────────────────────────┘
```

kilrau:
@raladev you can merge this straight after it passed your review

sangaman:
@raladev Your issue should be fixed now, it had to do with listing trades where we were the maker. One thing to note is that we won't have a counterparty orderId in those cases, because when we accept a swap as maker we don't actually see the order the peer place that matched ours.

kilrau:
> because when we accept a swap as maker we don't actually see the order the peer place that matched ours

That makes sense. Maybe just default to display `local` instead of the Order Id in that case. Better than nothing imo, because that leaves the user wondering why this is empty. `local (OnionTrader)` gives an idea that the order id is only known locally to the counterparty.

raladev:
> @raladev Your issue should be fixed now, it had to do with listing trades where we were the maker. One thing to note is that we won't have a counterparty orderId in those cases, because when we accept a swap as maker we don't actually see the order the peer place that matched ours.

tested, yeap, it is fixed, but it seems there is one more.

- [ ] tradehistory returns 0 in quantity for json output; Seems it should return filled qty (qty from Executed field).
[Screenshot from 2020-05-27 18-47-15](https://user-images.githubusercontent.com/29906866/83042796-bb61e300-a04a-11ea-889e-c0bb4f245224.png)

raladev:
> @raladev Your issue should be fixed now, it had to do with listing trades where we were the maker. One thing to note is that we won't have a counterparty orderId in those cases, because when we accept a swap as maker we don't actually see the order the peer place that matched ours.

Tested, its fixed, but i dont understand why qty for makerOrder and takerOrder is always 0 and price for taker order too.

sangaman:
> tradehistory returns 0 in quantity for json output; Seems it should return filled qty (qty from Executed field) or there is should be another way to get filled qty from json.

Also fixed this, I agree it makes sense to set the quantity value for the orders. It was 0 because I was only setting quantity on the `Trade` message, but the `Order` message also has a quantity field and that defaults to 0 unless it's set explicitly.

> is it correct that sometimes makerOrder and takerOrder can have different prices? Cant understand why it can be true because in my head trade means equal prices for maker and taker orders.

Yes and to me it makes sense this way. The two orders don't need to have the same price to match. One order can be market order meaning it has no "price" at all. Also, an order to buy LTC/BTC for 0.02 will match with an order to sell LTC/BTC for 0.01, the price used for the trade will be whichever was the maker order.

raladev:
> Also fixed this, I agree it makes sense to set the quantity value for the orders. It was 0 because I was only setting quantity on the `Trade` message, but the `Order` message also has a quantity field and that defaults to 0 unless it's set explicitly.

Tested, fixed, but I still dont see this one:

>That makes sense. Maybe just default to display local instead of the Order Id in that case. Better than nothing imo, because that leaves the user wondering why this is empty. local (OnionTrader) gives an idea that the order id is only known locally to the counterparty.

[Screen](https://user-images.githubusercontent.com/29906866/83128815-9fab1b00-a0e4-11ea-8ea2-6eb3a37b592c.png)

And also my suggestion about that:
''local" word does not give an idea that the order id is only known locally to the counterparty at least for me.
Can we add sentence about this behavior to tradehistory --help ?

sangaman:
I hadn't implemented kilian's suggestion yet, I went ahead and put "N/A" in there instead of local since I felt that made a bit more sense. I struggled to come up with a succinct way to explain the field in the `--help` output but I did add a couple lines to the `taker_order` field comment in the API proto definition.

PR #1571, opened 2020-05-21
fix(swaps): failed payment for completed swap

### LINKED ISSUE #1569: taker xud crashes after payment received
Based on: https://github.com/ExchangeUnion/xud-docker/pull/464#issuecomment-630986211

1. taker sent BTC payment
2. taker received ERC20 payment (revealing the preimage)
3. maker was unable to claim the payment
4. after some time taker's xud crashed

sangaman:
This handles an edge case described below that causes xud to crash due to an assertion error when a payment fails as taker for a swap that has already completed, as described below.

1. Taker sends payment.
2. Maker sees payment, and sends payment to Taker.
3. Taker receives payment, and marks the swap as complete.
4. Maker fails to claim payment, so the attempt to send payment (from step 1) fails.
5. Error handling logic when Taker's payment attempt doesn't go through causes it to "fail" the deal.
6. But the swap actually succeeded - from Taker's perspective - in step 3, and Taker hits an assertion error trying to fail a deal that has completed.

Fixes #1569.

### DIFF
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -759,6 +759,13 @@ class Swaps extends EventEmitter {
     try {
       await makerSwapClient.sendPayment(deal);
     } catch (err) {
+      // first we must handle the edge case where the maker has paid us but failed to claim our payment
+      // in this case, we've already marked the swap as having been paid and completed
+      if (deal.state !== SwapState.Completed) {
+        this.logger.warn(`maker was unable to claim payment for ${deal.rHash} but has already paid us`);
+        return;
+      }
+
       if (err.code === errorCodes.PAYMENT_REJECTED) {
         // if the maker rejected our payment, the swap failed due to an error on their side
         // and we don't need to send them a SwapFailedPacket
@@ -776,7 +783,6 @@ class Swaps extends EventEmitter {
           errorMessage: err.message,
         });
       }
-      return;
     }
   }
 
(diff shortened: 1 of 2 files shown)


kilrau:
image
:heart:

lib/swaps/Swaps.ts
```diff
@@ -759,6 +759,13 @@ class Swaps extends EventEmitter {
     try {
```
erkarl, line 759: nit: I'm wondering if it's worth extracting this try/catch block into `sendMakerPayment(swapClient, deal)` and unit test the catch block behaviour?

erkarl:
Simtests were failing - restarted the job.

lib/swaps/Swaps.ts
sangaman: It's a fine idea, but I just took a stab at this and it involved quite a bit of mocking. Perhaps in another PR? Or if you have an approach in mind maybe you can take a stab at it. It would be good to have a test for the `PAYMENT_REJECTED` error as well.

sangaman:
> Simtests were failing - restarted the job.

They were failing because the custom-xud patch needed to be updated. I just took care of that, so they should pass now.

raladev:
I dont think that this thing can be tested manually for now.
Can we add sim tests for that and similar cases like performing actions with completed swaps, filled/removed orders?
@kilrau @LePremierHomme

kilrau:
> I dont think that this thing can be tested manually for now.
> Can we add sim tests for that and similar cases like performing actions with completed swaps, filled/removed orders?
> @kilrau @LePremierHomme

Yes, opened: https://github.com/ExchangeUnion/xud/issues/1573

Issue #1520, opened 2020-05-05
Deposit/Withdrawal Swaps


kilrau:
`xud` part of https://github.com/ExchangeUnion/xud/issues/1361

Example usage `deposit`:

- `deposit btc`, throws `You will receive the deposit in a balanced lightning channel and the fee for this service is 0.25%. For more information, visit https://boltz.exchange/. Do you want to continue? [Y/n]` `Please deposit between x and y BTC to bc1... in the next z minutes.`

Example usage `withdraw`:
- `withdraw 1 btc bc1...` throws `You will receive the withdrawal on-chain and the fee for this service is 0.25%. For more information, visit https://boltz.exchange/. Do you want to continue? [Y/n]`
- `Withdrawing...done: blockstream.info link`

Existing on-chain deposit withdrawal renamed to `walletdeposit`/`walletwithdraw`: https://github.com/ExchangeUnion/xud/issues/1473 .

PR #1577, opened 2020-05-25
feat(rpc/connext): deposit & openchannel calls

### LINKED ISSUE #1472: Connext: open/closechannel calls

### LINKED ISSUE #1473: lnd on-chain deposit/withdraw renaming + connext equivalent

sangaman:
Closes #1472. Closes #1473.

This refactors the `Deposit` and `OpenChannel` methods in `SwapClient` to make them generic and applicable to Connext. It implements this functionality for Connext, namely retrieiving an address for sending funds on-chain to the Connext wallet and depositing those funds into Connext channels.

It also refactors the `CloseChannel` call similarly, however the Connext REST client does not currently implement a `/withdraw` endpoint for taking funds out of Connext channels. This adds dummy code to construct a withdraw payload and make a request. Edit: withdraw endpoint is being added here https://github.com/connext/rest-api-client/pull/29

Withdrawing coins from the on-chain Connext wallet is not currently supported by the Connext client and so the `Withdraw` rpc call remains specific to lnd.

I didn't rename the `deposit` and `withdraw` cli calls since they won't conflict with the current approach to a separate tool for the deposit/withdrawal swaps described in #1520. I can still rename those if we'd like, I just wasn't sure whether it still made sense.

~Also it does look like it's possible to make calls directly to `ethProvider` from the connext rest-api-client, that would allow us to send on-chain coins directly from the connext wallet (rather than a connext channel), so that's a possibility if we ever need that functionality.~ Edit: this is being implemented by https://github.com/connext/rest-api-client/pull/20 so after it's merged we can integrate it to xud.

I still need to do more testing with this PR but it's ready for review.


kilrau:
> I didn't rename the deposit and withdraw cli calls since they won't conflict with the current approach to a separate tool for the deposit/withdrawal swaps described in #1520. I can still rename those if we'd like, I just wasn't sure whether it still made sense

Thanks for bringing this up, I would still rename the existing lnd wallet deposit/withdraw -> walletdeposit/walletwithdraw simply because we have two different ways to get from an external on-chain wallet to funds in an active channel
1. `deposit`: btc/ltc: boltz submarine/channel creation swap, connext: direct deposit into indra channel (needs https://github.com/connext/indra/issues/1166). Default option.
2. `walletdeposit`: throw an address from the local wallet, send on-chain from local wallet, then manually call `openchannel` once funds are confirmed. Manual re-balancing for inbound liquidity needed on btc/ltc.

As discussed today, these two currently unimplemented calls need to be added to https://github.com/ExchangeUnion/rest-api-client to complete above:
- eth/token withdrawal
- indra channel withdrawal

sangaman:
> Thanks for bringing this up, I would still rename the existing lnd wallet deposit/withdraw -> walletdeposit/walletwithdraw simply because we have two different ways to get from an external on-chain wallet to funds in an active channel and I believe one has to be renamed if we want to keep both (which we want):

The boltz submarine channel creation swap won't be happening through xucli though, right? So it wouldn't conflict with the current `deposit`/`withdraw` naming scheme, unless I'm misunderstanding something.

kilrau:
Correct, technically not through xucli, but though the xud-docker cli which is xucli + a bunch of other commands. There, the user should be able to type `deposit btc` to trigger a boltz submarine swap. I can't think of a good way how this could coexist with the current `deposit` (which calls lnd's newaddress) apart from renaming the latter.

A clear mental model would be:
1. Type `deposit` -> deposit into channel, default
2. Type `walletdeposit` -> deposit into wallet

Of course we could leave xucli untouched and just wrap the current `xucli deposit` in `walletdeposit` in the xud-docker cli, just that this leaves `deposit` in xud-docker doing something different than `deposit` in xud native. But if you feel strongly about this, that is fine too.

raladev:
Testing results:
- [ ] closechannel does not work for connext. 1. because of peer is required 2. because of BigInt error
[Screen](https://user-images.githubusercontent.com/29906866/82898814-909c5f80-9f62-11ea-8bc2-faf8c54a91cb.png)
- [ ] deposit returns empty address for any currency (not only for connext)
[Screen](https://user-images.githubusercontent.com/29906866/82898967-da854580-9f62-11ea-811b-cce902db2738.png)
- [ ] openchannel to amount bigger then exists in wallet = connext server error
[Screen](https://user-images.githubusercontent.com/29906866/82899049-fb4d9b00-9f62-11ea-8f05-278e2f600d30.png)
Question: is it correct that there is no apportunity to open Connext channel to not our Connext node?

kilrau:
To answer your question: yes, that is correct

sangaman:
> Correct, technically not through xucli, but though the xud-docker cli which is xucli + a bunch of other commands. There, the user should be able to type `deposit btc` to trigger a boltz submarine swap. I can't think of a good way how this could coexist with the current `deposit` (which calls lnd's newaddress) apart from renaming the latter.
> 
> A clear mental model would be:
> 
>     1. Type `deposit` -> deposit into channel, default
> 
>     2. Type `walletdeposit` -> deposit into wallet
> 
> 
> Of course we could leave xucli untouched and just wrap the current `xucli deposit` in `walletdeposit` in the xud-docker cli, just that this leaves `deposit` in xud-docker doing something different than `deposit` in xud native. But if you feel strongly about this, that is fine with me too.

Makes sense to me, I'll rename it.

kilrau:
6.6.2 (https://github.com/ExchangeUnion/xud-docker/pull/478) contains https://github.com/ExchangeUnion/xud-docker/pull/478 - so connext on-chain deposit and withdrawal should be testable now @raladev

sangaman:
Thanks a lot for testing these.

> Testing results:
> 
>     * [ ]  closechannel does not work for connext. 1. because of peer is required 2. because of BigInt error

Should be fixed now.

>     * [ ]  deposit returns empty address for any currency (not only for connext)

Also fixed, I confirmed this.

>     * [ ]  openchannel to amount bigger then exists in wallet = connext server error

This one I'm still getting, and in fact I'm getting it even when I try to query my balance. I think this is likely an underlying issue with connext that we'd need to address via modifying the connext client, not something I can address in this PR.

raladev:
- [ ] deposit still is not working, but now because of syntax error; @sangaman 
[Screen](https://user-images.githubusercontent.com/29906866/83133787-097af300-a0ec-11ea-8153-8ad75b8c09a2.png)

- [ ] Note/question about closechannel command. Daniel solves 'peer is required' problem by moving peer to optional param, but now command 'closechannel BTC' (without peer) does nothing and returns 'success'. How it should work for now? @kilrau @sangaman 

I see 2 solutions:
1. Close all btc channels for all peers
2. throw error 'peer is required for lnd channles'
First is more preferable.

- [ ] closechannel command stuck for connext; tested with connext 6.6.2; @erkarl  
1. closechannel ETH -> connext logs contains
```
Executing with params: {"multisigAddress":"0xe37A..d684","assetId":"0x0000..0000"}
[1590662664150] INFO  (71 on connext): received request
    reqId: 24549
    url: "/withdraw"
    id: 24549
2020-05-28T10:44:24.152Z [WithdrawalController] withdraw started: {
  "recipient": "",
  "amount": "900000000"
}
```
2. socket hang up after ~10 sec
3. also it tooks closing commission but there is no closing -_-
Screens:
[Screen](https://user-images.githubusercontent.com/29906866/83133459-96717c80-a0eb-11ea-8f1d-8296b6a716c3.png)
[Screen](https://user-images.githubusercontent.com/29906866/83133462-983b4000-a0eb-11ea-8bdf-70bda680bfbd.png)

Logs:
todo

Note:
I can create separate issue if it is required, but it seems as blocker for that issue, so I just wrote about that here.

- [ ] https://github.com/ExchangeUnion/xud/pull/1577#issuecomment-633985762 also last bug in this comment as a bonus @erkarl;

sangaman:
>     * [ ]  deposit still is not working, but now because of syntax error; @sangaman

Sorry, I should've mentioned I went ahead with kilian's suggestion and renamed to `walletdeposit` and `walletwithdraw`. Can you retry with `walletdeposit`?

> 
>     * [ ]  Note/question about closechannel command. Daniel solves 'peer is required' problem by moving peer to optional param, but now command 'closechannel BTC' (without peer) does nothing and returns 'success'. How it should work for now? @kilrau @sangaman
>       [Screen](https://user-images.githubusercontent.com/29906866/83134981-d76a9080-a0ed-11ea-9858-a855c1835a39.png)
> 
> 
> I see 2 solutions:
> 
>     1. Close all btc channels for all peers
> 
>     2. throw error 'peer is required for lnd channles'
>        First is more preferable.

Thanks for catching this, I would think we'd actually want to stick with the second option, since that is the current behavior, plus it would be unfortunate if you forgot to type in a peer and accidentally closed all your BTC channels! So I can make sure the call returns an error if we expect a peer but don't get one. What do you think @kilrau?

> 
>     * [ ]  closechannel command stuck for connext; tested with connext 6.6.2; @erkarl

I think this one is probably connext implementation specific, rather than a problem with xud's API or cli, so let's follow up on this separately after this is merged.

kilrau:
> Thanks for catching this, I would think we'd actually want to stick with the second option, since that is the current behavior, plus it would be unfortunate if you forgot to type in a peer and accidentally closed all your BTC channels! So I can make sure the call returns an error if we expect a peer but don't get one. What do you think @kilrau?

Agreed!

raladev:
- walletdeposit/walletwithraw work
- [ ] to do: throw error for closechannel <lndcurrency> 
- [ ] one more bug. Incorrect convertation of ETH to  satoshiETH between xucli andwithdraw call.
withdraw --amount 2 -> only 0.00000002 withdraw insteadof 2 ETH. 
[Screenshot](https://user-images.githubusercontent.com/29906866/83265648-72389d00-a1ca-11ea-8d4f-40e953a080b5.png)
- [ ] note:strange balance  behavior. max amount increased after first withdrawal operation.(can be because of incorrect precision of eth in getbalance - should be 18 instead of 8)

sangaman:
>     * walletdeposit/walletwithraw works
> 
>     * [ ]  to do: throw error for closechannel 'lndcurrency'
> 
>     * [ ]  one more bug. Incorrect convertation of ETH to  satoshiETH between xucli and withdraw call.
>       withdraw --amount 2 -> only 0.00000002 withdraw instead of 2 ETH.
>       [Screenshot](https://user-images.githubusercontent.com/29906866/83265648-72389d00-a1ca-11ea-8d4f-40e953a080b5.png)
> 
>     * [ ]  note:strange balance  behavior. max amount increased after first withdrawal operation.(can be because of incorrect precision of eth in getbalance - should be 18 instead of 8)

I fixed the first two issues (closechannel was also missing the coins to satoshis conversion for the `amount` option), but I'm not sure I understand the last one. Are you able to reproduce or give some more detail? Thanks.

raladev:
- first issue fixed fully;
- [ ] second fixed partially, There is still issue with convertation when use closechaanel without params
Screen
- Description and logs for third.
[Screen 1](https://user-images.githubusercontent.com/29906866/83420105-cf348d00-a42e-11ea-89bb-a4dbb6fdfdfa.png)
[Screen 2](https://user-images.githubusercontent.com/29906866/83420301-1e7abd80-a42f-11ea-835b-f0e83e4f1a95.png)

Total balance before first withdrawal = 29.99979
Total balance after first withdrawal = 29.99978999
Total balance after all next withdrawals = 29.99978999

Question: who took my coins after first withdrawal?)

kilrau:
As agreed @erkarl to skim over the code changes, other than that this PR is well tested by @raladev

Issue #1574, opened 2020-05-22
Got free DAI because of failed swap from connext side


raladev:
### Your environment

Taker's env - master of simnet docker setup
Maker's env - lndltc and connext are from master of simnet docker setup, but xud was installed natively and tradehistory branch was used.

### Steps to reproduce

1. ./xucli -p 28886 buy 0.2 LTC/DAI 200 1 as maker
2. sell 0.2 LTC/DAI 200 1 as taker

### Actual behaviour

1. msg about successful swap for taker;
2. 40 DAI is received by taker;
4. xud node died because of the assertion;
3. 0.2 LTC was in pending_htlcs for ~30 sec and after that it was returned to taker's channel (he did not pay for DAI).

### Screens
[balances_before_the_swap](https://user-images.githubusercontent.com/29906866/82675463-82420100-9c4d-11ea-8e53-6875a90d1425.png)
[swap_proccess](https://user-images.githubusercontent.com/29906866/82675461-81a96a80-9c4d-11ea-833c-aadc1aa25986.png)
[LTC_Returned_back_to_taker](https://user-images.githubusercontent.com/29906866/82675459-80783d80-9c4d-11ea-884e-1dae5edc15ee.png)

### Logs

Note: Some of logs can be bigger then others because of re-installation cycles, dont try to match startup phases of them.

22/05/2020 13:25 - approx. time of the swap for taker
22/05/2020 06:25 - approx. time of the swap for maker

[takers_log.zip](https://github.com/ExchangeUnion/xud/files/4668501/takers_log.zip)
[makers_log.zip](https://github.com/ExchangeUnion/xud/files/4668514/makers_log.zip)

UPD: Add missed connext log for maker.
[freedai_connext_maker.log](https://github.com/ExchangeUnion/xud/files/4676865/freedai_connext_maker.log)

erkarl:
Looking at the logs indicates that the app install for `9012739211802103871b05e223b7afd56c41b9be6aba607c13687253cc9da7df` hash failed.

Node logs:
```
2020-05-22T10:52:43.116Z [CF-ProposeProtocol] [479265b7-d315-44fb-b1f1-aa2e62dadee8] Persisted app instance 0xd6a09199792b6b6687261c3e3f693c9ccce01c1a71b521631c87eb5f3c0a4db0 in 38 ms
[...]

PR #1583, opened 2020-05-26
Fix connext swap recovery

ghost:
This PR modifies `Connext.lookupPayment` to include the preimage in the case of a completed payment so that the swap recovery flow can use it to claim the payment.

In addition we're lowering the swap timeout to 35 seconds. It was previously 75 seconds for debugging purposes.

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -59,7 +59,7 @@ interface ConnextClient {
 const waitForPreimageByHash = (
   client: SwapClient,
   expectedHash: string,
-  errorTimeout = 75000,
+  errorTimeout = 34000,
 ): Promise<string> => {
   // create an observable that emits values when a preimage
   // event is triggered on the client
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -11,6 +11,7 @@ import SwapClient, {
   WalletBalance,
   TradingLimits,
   SwapClientInfo,
+  PaymentStatus,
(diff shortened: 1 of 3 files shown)


lib/connextclient/ConnextClient.ts
```diff
@@ -58,7 +59,7 @@ interface ConnextClient {
 const waitForPreimageByHash = (
   client: SwapClient,
   expectedHash: string,
-  errorTimeout = 75000,
+  errorTimeout = 34000,
```
sangaman, line 62: Is this supposed to be 35000? Or did you intentionally make it slightly lower than the swap timeout?

sangaman (approved):
Haven't tested this but code looks fine to me.

kilrau:
Definitely should test this going through the swap recovery flow @raladev

lib/connextclient/ConnextClient.ts
erkarl: Yep, this is the timeout for `ConnextClient.sendPayment` which should error before the global swap timeout. I think this is necessary because otherwise the subscription of observable is never cleaned up.

erkarl:
Once https://github.com/ExchangeUnion/xud-docker/pull/478 is merged this can be tested using the simnet docker setup on taker side. The maker side should use an external xud using this branch. The setup is analogical to the https://github.com/ExchangeUnion/xud/issues/1574 where the issue was first reported.

The above setup should prevent connext from sending the preimage to the maker's xud, triggering swap recovery flow.

raladev (approved):

PR #1591, opened 2020-06-01
temp: disable makerClient.removeInvoice

erkarl:


erkarl:
> I don't think we need to touch any of the code related to sanity swaps, mainly to limit code churn. The only two lines that really matter are at the bottom, line 960/972

Restored the sanity swaps' removeInvoice calls.

LePremierHomme:
> Also, as far as the tests are concerned can we just remove any checks as to whether the payment/invoice from maker to taker is canceled or not? That's not really a security a concern, since the taker is in control of the preimage.

The problem is that the test is required to leave with all channels being closed.. so that's the reason I added the HTLC expiration flow.

sangaman:
Thanks for the explanation. I suggest we go with #1592, which uses the same test fix commit, and also the same change to remove the lines that cancel the invoice when the maker's payment fails, but then takes the additional step of putting the swap into `SwapRecovery` where it will automatically claim payment if the outgoing payment is settled by the taker, or cancel the incoming invoice if the outgoing payment has failed for good.

sangaman:
Closing after merging #1592.

PR #1605, opened 2020-06-04
Grpc js

sangaman:
This migrates the `grpc` package to the pure javascript `grpc-js` implementation. `grpc-js` appears to be the more actively maintained package and the preferred implementation by Google projects, and it reduces the dependency on and overhead from compilers like `node-pre-gyp`.

This also updates `grpc-tools` and tweaks the way grpc code is generated from `.proto` files to be compatible with `grpc-js`.

Lastly this removes the grpc web proxy from xud. It did not appear to be compatible with `grpc-js` and it is currently not used for any purpose including testing or development. As the old approach relied on a custom fork of an old, unmaintained package to create the proxy, should the need arise for a web proxy in the future we can implement a more robust and updated solution.

BREAKING CHANGE: Removes grpc web proxy from xud.

With this change we can better integrate xud with projects that use its API like Arby. A follow-up step would be to export the generated grpc protbuf files in a separate package that can be cleanly imported without manually copying files or importing the entire xud project.

Edit: I believe this reintroduces the issue I described in https://github.com/grpc/grpc-node/issues/789, as far as I know it's still present in newer versions of grpc-tools/protoc, but it no longer impacts us as we aren't using channel ids or any other 64 bit integer field via grpc that exceeds javascript's 55 bit number precision.


erkarl:
Nit: we might want to have a non-chore commit message so this breaking change will automatically be reflected in the release notes.

I really like that as a consumer (arby) I'll only have to include `@grpc/grpc-js` dependency once this gets merged.

sangaman:
> Nit: we might want to have a non-chore commit message so this breaking change will automatically be reflected in the release notes.

As I understand any commit with breaking changes will be reflected in the changelog.

> I really like that as a consumer (arby) I'll only have to include @grpc/grpc-js dependency once this gets merged.

Yup that's the goal!

Something is causing this to fail simulation tests so I'm planning to do some manual testing to see what's going.

sangaman:
@erkarl Is this branch working for you currently? I'm able to run it and make xucli calls without issue, but the simulation tests get stuck after creating the rpc cert and I'm not sure why.

kilrau:
Give this another look as per call @sangaman

sangaman:
What's left to do here @LePremierHomme :

- Update newly added dependencies (particularly grpc-js) to latest versions.
- Rebase on master, this will also involve going through some newly added code that's using old references to the deprecated grpc dependency and updating them to grpc-js
- Investigate why simulation tests were failing.
- And perhaps most important, come up with an alternative to our current web proxy solution, which isn't compatible with grpc-js in it's current form. In the current state of the PR it simply removes the web proxy, since we weren't using it at the time, but now it's a dependency for the xud-ui project and is important to preserve. I know we had looked at and discussed several options in the past, it's been a while since I looked at it so I don't remember much off the top of my head but @erkarl or @kilrau might have some thoughts. There's been some discussion in #1661 and #1897. I can try to refresh my memory and take a look at the options as well, but at this point I'd be OK with any option that works. As part of this it would be nice to restore the web proxy tests.

LePremierHomme:
Still stuck here - there were various unrelated issues, some are still unresolved. 

1. I'm getting the following pre-crash memory errors on stderr (not stdout), which are associated with deletion of invoice subscription. It's coming from the C++ stack (nodejs core), and It doesn't happen consistently:
```
free(): double free detected in tcache 2
```
```
node[8399]: ../src/node_http2.cc:448:void node::http2::Http2Session::CheckAllocatedSize(size_t) const: Assertion `(current_nghttp2_memory_) >= (previous_size)' failed.
 1: 0xa03530 node::Abort() [node]
 2: 0xa035ae  [node]
 3: 0xa3ba6b node::mem::NgLibMemoryManager<node::http2::Http2Session, nghttp2_mem>::ReallocImpl(void*, unsigned long, void*) [node]
 4: 0xa3bc43 node::mem::NgLibMemoryManager<node::http2::Http2Session, nghttp2_mem>::FreeImpl(void*, void*) [node]
 5: 0x1865d81 nghttp2_session_close_stream [node]
 6: 0x186c4e4 nghttp2_session_mem_recv [node]
 7: 0xa2faa9 node::http2::Http2Session::ConsumeHTTP2Data() [node]
 8: 0xa2fe18 node::http2::Http2Session::OnStreamRead(long, uv_buf_t const&) [node]
 9: 0xb4af71 node::TLSWrap::ClearOut() [node]
10: 0xb4bb60 node::TLSWrap::OnStreamRead(long, uv_buf_t const&) [node]
11: 0xada258 node::LibuvStreamWrap::OnUvRead(long, uv_buf_t const*) [node]
12: 0x1385197  [node]
13: 0x1385b50  [node]
14: 0x138c545  [node]
15: 0x137a398 uv_run [node]
16: 0xa438d4 node::NodeMainInstance::Run() [node]17: 0x9d1735 node::Start(int, char**) [node]
18: 0x7f469681409b __libc_start_main [/lib/x86_64-linux-gnu/libc.so.6]
19: 0x96971c  [node]
```

2. Failing to use the LND's router service, as if it is not implemented or activated by LND's RPC server.
```
13/12/2020 19:06:14.963 [LND-LTC] trace: sending payment with request: {"dest":"AiKFk74pEEDCUglcueGvjT6M9c71wspSJmH+2m4SpOxt","amt":1000000,"amtMsat":0,"paymentHash":"5tr0dDs/dfBHDy/KkOIZX3anYuo3CbfgjoWfm718NiQ=","finalCltvDelta":1118,"paymentRequest":"","timeoutSeconds":90,"feeLimitSat":30000,"feeLimitMsat":0,"outgoingChanId":"0","outgoingChanIdsList":[],"lastHopPubkey":"","cltvLimit":0,"routeHintsList":[],"destCustomRecordsMap":[],"allowSelfPayment":false,"destFeaturesList":[],"maxParts":0,"noInflightUpdates":false}
13/12/2020 19:06:14.981 [LND-LTC] error: error event from sendPaymentV2: Error: 12 UNIMPLEMENTED: unknown service routerrpc.Router
    at Object.callErrorFromStatus (/xud-vol/node_modules/@grpc/grpc-js/build/src/call.js:31:26)
    at Object.onReceiveStatus (/xud-vol/node_modules/@grpc/grpc-js/build/src/client.js:327:49)
    at Object.onReceiveStatus (/xud-vol/node_modules/@grpc/grpc-js/build/src/client-interceptors.js:297:181)
    at /xud-vol/node_modules/@grpc/grpc-js/build/src/call-stream.js:129:78
    at processTicksAndRejections (internal/process/task_queues.js:79:11)
```

kilrau:
@sangaman

sangaman:
> I'm getting the following pre-crash memory errors on stderr (not stdout), which are associated with deletion of invoice subscription. It's coming from the C++ stack (nodejs core), and It doesn't happen consistently:

This is weird, what version of node is this on? Are you seeing this in manual tests?

> Failing to use the LND's router service, as if it is not implemented or activated by LND's RPC server.

For this, are you sure that the lnd you're running this against has the `routerrpc` service enabled? I'm noticing now that our `install-lnd.sh` script for simulation tests doesn't have `routerrpc` in its make tags, so that may be worth trying (I'm surprised it hasn't given us trouble before, actually).

LePremierHomme:
> This is weird, what version of node is this on? Are you seeing this in manual tests?

Looks like enforcing node version to `>=14.5.0` and *not* doing `invoiceSubscription.cancel()` solves these problems. Still looking into it.

> For this, are you sure that the lnd you're running this against has the routerrpc service enabled? I'm noticing now that our install-lnd.sh script for simulation tests doesn't have routerrpc in its make tags, so that may be worth trying (I'm surprised it hasn't given us trouble before, actually).

Good call, this indeed was missing.

sangaman:
> Looks like enforcing node version to `>=14.5.0` and _not_ doing `invoiceSubscription.cancel()` solves these problems. Still looking into it.

Yes, using `bigint` requires v14.5.0+ so that makes sense, but even with a sufficient version are you still getting problems wtih `invoiceSubscription.cancel()`? If so I can take a look to see if I have any ideas.

LePremierHomme:
> Yes, using `bigint` requires v14.5.0+ so that makes sense, but even with a sufficient version are you still getting problems wtih `invoiceSubscription.cancel()`? If so I can take a look to see if I have any ideas.

Yes, i've commented-out the 2 instances of `invoiceSubscription.cancel()` in `LndClient.ts`. Undoing any of them results in these memory errors and a crash.

sangaman:
> > Yes, using `bigint` requires v14.5.0+ so that makes sense, but even with a sufficient version are you still getting problems wtih `invoiceSubscription.cancel()`? If so I can take a look to see if I have any ideas.
> 
> Yes, i've commented-out the 2 instances of `invoiceSubscription.cancel()` in `LndClient.ts`. Undoing any of them results in these memory errors and a crash.

Looks like this is a bug with either node js or the grpc-js package, see these isuses:

https://github.com/grpc/grpc-node/issues/1464
https://github.com/grpc/grpc-node/issues/1652

As a workaround we may be able to try creating separate client instances, one for unary calls and one for subscriptions? It's a bit of a hassle but I don't have another idea right now, it doesn't look like fixing the root cause will be simple.

LePremierHomme:
I've tried creating a separate client solely for `subscribeSingleInvoice` calls, but it doesn't help - the process still crashes if the subscription is cancelled. 

I don't think it's sustainable to avoid the cancellation and to allow the memory leak. So I guess we should put this upgrade on hold while pushing for bug fix in nodejs (i'm not sure if it was reported). 

If we don't want to hold the upgrade, I guess we can utilize LND's REST server (web proxy), although I'm not even sure it supports stream endpoints. I can check that, but anyway this would be quite an ugly solution. 

Any thoughts?

kilrau:
> I don't think it's sustainable to avoid the cancellation and to allow the memory leak. So I guess we should put this upgrade on hold while pushing for bug fix in nodejs (i'm not sure if it was reported).

Could you check on that and open an issue if it was not reported yet?

> If we don't want to hold the upgrade, I guess we can utilize LND's REST server (web proxy), although I'm not even sure it supports stream endpoints. I can check that, but anyway this would be quite an ugly solution.

Before doing that I'd rather not upgrade tbh. Not sure what others think.

erkarl:
@LePremierHomme have you tried cancelling using `setTimout(() => cancel())` or `setImmediate`? I was experiencing something similar with arby and had to use that as a workaround.

LePremierHomme:
> @LePremierHomme have you tried cancelling using `setTimout(() => cancel())` or `setImmediate`? I was experiencing something similar with arby and had to use that as a workaround.

👑 @erkarl 👑 awesome, it worked!

> Could you check on that and open an issue if it was not reported yet?

Yes (i'll do that anyway)

---

The PR will be finalized soon, I'm just checking what we can do as a solution for the web proxy.

kilrau:
Nice work @LePremierHomme (and @erkarl !) - I assume this is ready for testing? Took the liberty and moved it out of draft state.

raladev:
- [ ]  arm build is failed - https://github.com/ExchangeUnion/xud-docker/runs/1600453433?check_suite_focus=true

raladev (changes requested):
- [ ] I noticed delay for response finishing after update, can it be fixed? Video - https://gofile.io/d/qqa515

kilrau:
Can't add you as reviewer since you opened the original PR, but could you give this another review? @sangaman

LePremierHomme:
@raladev checking.

LePremierHomme:
> - [ ] I noticed delay for response finishing after update, can it be fixed? Video - https://gofile.io/d/qqa515

After checking it, i’m not sure there’s a problem. I did notice a delay when compared to `master`, but it wasn’t consistent. I can do some benchmarks if we still find this annoying. 

> - [ ] arm build is failed - https://github.com/ExchangeUnion/xud-docker/runs/1600617588?check_suite_focus=true

`grpc-tools` doesn’t support ARM pre-compiled binaries. The problem existed before, so AFAIK [Dockerfile.aarch64](https://github.com/ExchangeUnion/xud-docker/blob/master/images/xud/Dockerfile.aarch64#L13) was removing the dependency, which isn't required for running xud. `grpc-tools` version changed, so the removal stopped working. I tried to fix it [here](https://github.com/ExchangeUnion/xud-docker/commit/d5552451acd98c281144b5ef64a7062d6999947e), but it still tries to install it for some reason. Checking it now.

---

Regarding the web proxy - our current https://github.com/ExchangeUnion/grpc-dynamic-gateway approach doesn’t work with `grpc-js`. I tried to quickly upgrade it but without success. I’m not sure whether this is the right direction, or if using `grpc-web` is an option for us, so i’ll wait until our daily call today for some discussion.

raladev:
> . I did notice a delay when compared to `master`, but it wasn’t consistent. I can do some benchmarks if we still find this annoying.

for me it reproduces consistently for each call and main problem is UI app affecting  because all proxy requests will work slow enough.

kilrau:
Not sure if a benchmark helps @LePremierHomme , maybe pair with @raladev if you can't produce reliably on your end.

LePremierHomme:
> for me it reproduces consistently for each call and main problem is affection on UI app because all proxy requests will work slow enough.

Do you see this problem on the UI app, and not only on the CLI? I thought that the delays are due to the overhead of every CLI call in constructing the client and connection from scratch (although this always was the case, so it doesn't explain a regression from `master`). If this happens also on a usable gRPC connection, then it's much more problematic. 

> I think you already spoke to @sangaman about this. As of now we don't use the webproxy afaik and this shouldn't be a block to merge. Wdyt? @sangaman @LePremierHomme @erkarl

Looks like that's the case, but i'll wait for a final discussion on Discord.

sangaman:
> > I think you already spoke to @sangaman about this. As of now we don't use the webproxy afaik and this shouldn't be a block to merge. Wdyt? @sangaman @LePremierHomme @erkarl
> 
> Looks like that's the case, but i'll wait for a final discussion on Discord.

I'm solidly on board to remove the webproxy for now and reintroduce it in a separate module/add-on via grpc-web running in a separate process.

sangaman:
I squashed all the commits into one, and made a few minor changes.

- Removed the changes from the `_debug_` commits `4fda631f` and `4fda631f` since they seemed temporary and unnecessary for the PR. 
- Added comments to the usage of `setImmediate` to explain why it's necessary.
- Removed the leftover `grpc` dependency and updated the `grpc-tools` and `grpc_tools_node_protoc_ts` dependencies to their latest version.
- Removed the WebProxy.spec.ts file and its mention in the jest config.
- Removed the web proxy xud config options.

raladev:
- [ ] PR contain failed tests @sangaman 
 - [ ] arm build still does not work @LePremierHomme https://github.com/ExchangeUnion/xud-docker/actions/runs/452723915
 - [ ] about xud-ui affection by delay: it does not affect regular rest calls, but we have integrated console and this console contains delays
Screenshot from 2020-12-30 16-54-48

sangaman:
>     * [ ]  PR contains failed tests @sangaman

Just flakiness, I'm restarting the failed test and it should work.

>     * [ ]  arm build still does not work @LePremierHomme https://github.com/ExchangeUnion/xud-docker/actions/runs/452723915

This depends on https://github.com/ExchangeUnion/xud-docker/pull/840 to remove the pre-compiled binary from grpc-tools.

>     * [ ]  about xud-ui affection by delay: it does not affect regular rest calls, but we have integrated console and this console contains delays
>       Screenshot from 2020-12-30 16-54-48

This one I'm not so sure about. It's just the calls within the console that are affected, not within docker or other xud-ui calls? How bad are the delays?

raladev:
> This depends on [ExchangeUnion/xud-docker#840](https://github.com/ExchangeUnion/xud-docker/pull/840) to remove the pre-compiled binary from grpc-tools.

i ran build for this branch and it was failed, so the fix does not help

sangaman:
> i ran build for this branch and it was failed, so the fix does not help

Sorry I left out an important detail that since I changed the grpc-tools version here, that PR needs to be updated too. I proposed a long term solution here https://github.com/ExchangeUnion/xud-docker/pull/840#pullrequestreview-560079126.

raladev (approved):

LePremierHomme:
I pushed a very minor cleanup commit. It can be squashed and merged. @sangaman

LePremierHomme (dismissed):

PR #1617, opened 2020-06-09
fix(swaps): go to recovery for all failures

### LINKED ISSUE #1606: SwapRecovery v2
https://github.com/ExchangeUnion/xud/tree/fix/swap-recovery still has an issue where an unclaimed payment (taker) is not entering the swap recovery path.

sangaman:
This fixes a bug where failed swaps in the `SendingPayment` phase as maker would only be tracked by the `SwapRecovery` module if they failed due to a timeout error.

Closes #1606.

This bug was due to an oversight in the changes from #1600.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -595,7 +595,7 @@ class LndClient extends SwapClient {
    * Sends a payment through the Lightning Network.
    */
   private sendPaymentSync = (request: lndrpc.SendRequest): Promise<lndrpc.SendResponse> => {
-    this.logger.trace(`sending payment of ${request.getAmt()} for ${request.getPaymentHashString()}`);
+    this.logger.trace(`sending payment with request: ${JSON.stringify(request.toObject())}`);
     return this.unaryCall<lndrpc.SendRequest, lndrpc.SendResponse>('sendPaymentSync', request);
   }
 
@@ -631,8 +631,7 @@ class LndClient extends SwapClient {
     if (!this.isConnected()) {
       throw swapErrors.FINAL_PAYMENT_ERROR(errors.UNAVAILABLE(this.currency, this.status).message);
     }
-
-    this.logger.trace(`sending payment with ${JSON.stringify(request.toObject())}`);
+    this.logger.debug(`sending payment of ${request.getAmt()} with hash ${request.getPaymentHashString()} to ${request.getDestString()}`);
     let sendPaymentResponse: lndrpc.SendResponse;
     try {
       sendPaymentResponse = await this.sendPaymentSync(request);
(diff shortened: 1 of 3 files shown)


kilrau:
Looks like travis failed due to an alias collission!? Restarted.

`2020-06-09 07:04:41.4141 [P2P][1] error: Unexpected error connecting to known peers on startup: 2.26 - alias MeadowOven refers to more than one node`

lib/swaps/Swaps.ts
```diff
@@ -1147,11 +1139,19 @@ class Swaps extends EventEmitter {
...
-      // if we are the maker and we have accepted a swap deal but we haven't yet started paying the taker
-      // then we should cancel
-      const swapClient = this.swapClientManager.get(deal.role === SwapRole.Maker ? deal.makerCurrency : deal.takerCurrency)!;
-      swapClient.removeInvoice(deal.rHash).catch(this.logger.error); // we don't need to await the remove invoice call
+    if (deal.role === SwapRole.Maker) {
+      if (deal.phase === SwapPhase.SwapAccepted) {
+        // if we are the maker and we have accepted a swap deal but we haven't yet started paying the taker
+        // then we should cancel the invoice for our incoming payment if one exists
+        const swapClient = this.swapClientManager.get(deal.makerCurrency)!;
+        swapClient.removeInvoice(deal.rHash).catch(this.logger.error); // we don't need to await the remove invoice call
```
erkarl, line 1147: Why not do the `removeInvoice` in `SwapRecovery`?

sangaman:
> Looks like travis failed due to an alias collission!? Restarted.
> 
> `2020-06-09 07:04:41.4141 [P2P][1] error: Unexpected error connecting to known peers on startup: 2.26 - alias MeadowOven refers to more than one node`

Wow... that should be extremely unlikely for a tiny network of nodes like the one the sim tests spin up.

lib/lndclient/LndClient.ts
```diff
@@ -631,8 +631,7 @@ class LndClient extends SwapClient {
     if (!this.isConnected()) {
       throw swapErrors.FINAL_PAYMENT_ERROR(errors.UNAVAILABLE(this.currency, this.status).message);
     }
-
-    this.logger.trace(`sending payment with ${JSON.stringify(request.toObject())}`);
+    this.logger.debug(`sending payment of ${request.getAmt()} with hash ${request.getPaymentHashString} to ${request.getDestString()}`);
```
erkarl, line 634: Should `getPaymentHashString` be a function call here?

lib/swaps/Swaps.ts
sangaman: Hm we could, my thinking though is that if we haven't started sending payment yet and are just in the acceptance phase, there are no funds that can possibly be lost and no uncertainty as to whether it's safe to cancel invoices.

lib/lndclient/LndClient.ts
sangaman: Yes absolutely, thanks for catching this.

PR #1621, opened 2020-06-09
fix(p2p): change connext_address to connext_identifier

### LINKED ISSUE #1611: connext_address -> connext_identifier
Implement https://github.com/opendexnetwork/opendex/pull/29

ghost:
BREAKING CHANGE: `connext_address` field in `NodeState` has been renamed
to `connext_identifier`.

Implements: https://github.com/opendexnetwork/opendex/issues/28
Closes: https://github.com/ExchangeUnion/xud/issues/1611


docs/api.md
```diff
@@ -91,9 +91,11 @@
     - [OrderSide](#xudrpc.OrderSide)
     - [Role](#xudrpc.Role)
   
+
```
kilrau, line 94: Why?

kilrau (approved):
CTRL+F `connextAddress` brought up things in xud/lib/utils/simnet-connext-channels.ts unchanged, but I guess that's on purpose.

docs/api.md
erkarl: Auto generated by `npm run proto`. Not sure if it's worth manually removing.

erkarl:
> CTRL+F `connextAddress` brought up things in xud/lib/utils/simnet-connext-channels.ts unchanged, but I guess that's on purpose.

That's intended, yes. In that context it really is `connextAddress` from `getinfo` gRPC call that we're dealing with.

kilrau:
> That's intended, yes. In that context it really is connextAddress from getinfo gRPC call that we're dealing with.

Gotcha, thanks!

proto/xudp2p.proto
```diff
@@ -31,7 +31,7 @@ message NodeState {
     map<string, string> lnd_pub_keys = 6;
     map<string, string> token_identifiers = 7;
     map<string, LndUris> lnd_uris = 8;
-    string connext_address = 9;
+    string connext_identifier = 9;
```
sangaman, line 34: Just a thought, but going forward I think it would make sense to keep this name even *if* for some reason connext renames this field again. "Identifier" is generic and we can use that for our p2p layer regardless of what connext calls it.

sangaman (approved):
Looks good. One other thought is that in simnet-connext-channels.ts there are still some references to `connextAddress`. Would it make sense to rename those variables as well?

kilrau:
> Looks good. One other thought is that in simnet-connext-channels.ts there are still some references to `connextAddress`. Would it make sense to rename those variables as well?

Had the same question: https://github.com/ExchangeUnion/xud/pull/1621#pullrequestreview-426977342 @sangaman

Issue #1625, opened 2020-06-10
DealTimedOut error instead of NoRouteFound error during swap


raladev:
### Background

1.There are 2 xud nodes with multi-hop connection through xud1.simnet for lnd. first one has fully synced btc node with active channel, but second has synced node with inactive btc channel.
2. Orderbook contains orders from second node + orders from xud1.simnet. Orders of second node are on first place and orders of xud1.simnet are on second place
[Screenshot from 2020-06-10 12-45-36](https://user-images.githubusercontent.com/29906866/84255170-ec4a1980-ab1a-11ea-9296-5bae44036877.png)

### Steps to reproduce

1. Send buy 0.2 ETH/BTC mkt order from first node

### Actual result:

1. 6 DealTimedOut errors instead of NoRouteErrors
2. No matching attempt with LaundryBrush after failed attepts with second node 

### Expected result:

1. NoRouteErrors
2. Matching attempt with LaundryBrush after failed attepts with second node 

### Logs:

First node(taker):
[deal_lndbtc.log](https://github.com/ExchangeUnion/xud/files/4757733/deal_lndbtc.log)
[deal_connext.log](https://github.com/ExchangeUnion/xud/files/4757734/deal_connext.log)
[deal_xud.log](https://github.com/ExchangeUnion/xud/files/4757735/deal_xud.log)

sangaman:
Examining the logs and the code, I'm pretty sure this is not due to no route found. Maker in this case is sending ETH, and Connext doesn't use routes so the internal `getRoute` call always succeeds. On the flip side, taker actually *does* find a route for BTC even tho there's an inactive channel on the second hop. An lnd node doesn't know which channels are active or not (besides its own) and for route-finding purposes will treat all channels as active.

The swap does appear to fail because the deal is not accepted in time, although I'm not sure why. There's nothing I can see in the maker's routine to decide whether to accept a swap that would take a long time for Connext. The logs from the maker would be helpful, if you can't get them I can try to reproduce. I've seen this `DealTimedOut` failure reason several times now when trying to swap with Connext, so we ought to get to the bottom of it. @erkarl If you have any ideas let me know.

sangaman:
This log may be relevant to the DealTimedOut error from @kilrau .

https://paste.ubuntu.com/p/rmpThFw7NB/

```
[...]

PR #1632, opened 2020-06-11
fix(backup): use async fs calls

sangaman:
This PR attempts to address issues and timeouts related to the Backup module writing new backups to disk synchronously. This can block the thread and prevent xud from proceeding with and responding to other matters requiring its attention. It appears to cause the swap acceptance logic to occasionally exceed the 10 second limit because accepting a swap involves several db writes which trigger the backup writes.

It also refactors the final steps of the routine to accept a swap deal to make it so that we update the deal phase (and in turn, update the db) in parallel with sending the swap accepted packet to our peer. This can reduce the total time needed to accept a deal request and begin the swap.

Likely related issue #1625.

### DIFF
--- a/lib/backup/Backup.ts
+++ b/lib/backup/Backup.ts
@@ -1,6 +1,6 @@
 import { createHash } from 'crypto';
 import { EventEmitter } from 'events';
-import fs from 'fs';
+import { FSWatcher, promises as fs, watch } from 'fs';
 import path from 'path';
 import Config from '../Config';
 import LndClient from '../lndclient/LndClient';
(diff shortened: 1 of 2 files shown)


kilrau:
Restarted, but looks like tests are failing.

sangaman:
After some more thought, I think I had overlooked the aspect of the xud backup running on a separate process. That shouldn't block the main xud thread, but it does have me thinking that maybe it's keeping the disk occupied, as it writes the entire xud database on each and every write.

A better approach (and I think we discussed this as a future enhancement when we first put in the backups) would be to intelligently replicate the database query by query rather than reading & writing the entire db.

So I'm closing this PR (and reopening another one just for the second refactoring commit that's unrelated to backups) and will see about opening another one for better xud db replication that's far less I/O intensive.

PR #1642, opened 2020-06-12
fix(swaps): cancel timer immediately on complete

### LINKED ISSUE #1634: Maker crashes on successful swap
The maker xud in a swap that succeeds can crash due to `Can not fail a completed deal.` This issue tracks changes to ensure this never happens.

Related logs: https://paste.ubuntu.com/p/5zYtfgZkXB/

sangaman:
This changes the order of events when a swap enters the `PaymentReceived` phase and is considered complete. Currently, we mark the swap as complete right away but we don't cancel the swap timer until after we've updated the database to persist the swap. When the database write is delayed at all, it presents the edge case risk where the swap timer hits its limit while are writing to the database and after we've already marked the swap as complete. This results in an assertion error as we try to fail a swap due to a timeout when that swap is already complete.

Instead, we cancel the swap timer immediately after marking the swap complete and only then persist the update to the database.

Fixes #1634.

### DIFF
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -1207,6 +1207,28 @@ class Swaps extends EventEmitter {
         assert(deal.phase === SwapPhase.SendingPayment, 'PaymentReceived can be only be set after SendingPayment');
         deal.completeTime = Date.now();
         deal.state = SwapState.Completed;
+
+        clearTimeout(this.timeouts.get(deal.rHash));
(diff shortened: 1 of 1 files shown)


kilrau:
Pushed this to xud-docker. You can run this branch with `-b fix/swapcomplete-timeout` once https://travis-ci.org/github/ExchangeUnion/xud-docker/builds/697578932 passed.

Issue #1111, opened 2019-07-19
`addpair`/`addcurrency` requires xud restart


erkarl:
I just noticed that in our `xud-docker` setup we're adding the trading pairs **after** connecting to the test nodes. This results in a state where we have the trading pairs enabled, but have 0 orders for the new pairs. The current solution is to restart xud.

I'd love this to work without the need for xud restart.

```
xud_1     | 19/07/2019 08:12:05.003 [P2P] info: Connecting to simnet XU network
xud_1     | 19/07/2019 08:12:05.014 [P2P] info: p2p server listening on 0.0.0.0:8885
xud_1     | 19/07/2019 08:12:05.024 [P2P] debug: Verifying reachability of advertised address: 0.0.0.0:8885
xud_1     | 19/07/2019 08:12:05.039 [HTTP] info: http server listening on 0.0.0.0:8887
xud_1     | 19/07/2019 08:12:05.053 [RPC] debug: Could not load gRPC TLS certificate. Generating new one
xud_1     | 19/07/2019 08:12:08.910 [P2P] debug: Connected pre-handshake to peer 127.0.0.1:47574
xud_1     | 19/07/2019 08:12:09.434 [P2P] debug: Peer (036b27d570ea4b0be00b794d7f75dfd4bdc30fe8ec5c3002d3bedcc9061c868690): closing socket. reason: ConnectedToSelf
xud_1     | 19/07/2019 08:12:09.441 [P2P] warn: could not open connection to peer (036b27d570ea4b0be00b794d7f75dfd4bdc30fe8ec5c3002d3bedcc9061c868690): cannot attempt connection to self
xud_1     | 19/07/2019 08:12:09.470 [P2P] info: Peer 036b27d570ea4b0be00b794d7f75dfd4bdc30fe8ec5c3002d3bedcc9061c868690 socket closed
xud_1     | 19/07/2019 08:12:09.481 [RPC] debug: gRPC TLS certificate created
xud_1     | 19/07/2019 08:12:09.487 [RPC] info: gRPC server listening on 0.0.0.0:8886
xud_1     | 19/07/2019 08:12:09.503 [P2P] verbose: Verified reachability of advertised address: 0.0.0.0:8885
xud_1     | 19/07/2019 08:12:09.512 [P2P] info: Connecting to known / previously connected peers
xud_1     | 19/07/2019 08:12:09.625 [P2P] debug: Connected pre-handshake to peer xud1.test.exchangeunion.com:8885
[...]

PR #1657, opened 2020-06-14
test(sim): connext integration

LePremierHomme:
Integrating connext to the simulation tests, using indra's docker setup which provides the ethprovider (at localhost:8545) and the connext node (at localhost:8080), among other necessary services. 

The simulation framework (running within the test docker container) manages a network of connext-client instances (https://github.com/connext/rest-api-client), one per xud node. It communicates with indra's docker setup containers via the host machine.

One test case was added: `testOrderMatchingAndSwapConnext`.
Once https://github.com/ExchangeUnion/xud/issues/1111 will be solved, the test should be refactored, to comply with other test cases conventions (currency/pair init, channel openings, state assertions, etc.).

Open task (@sangaman @erkarl @kilrau ): verify the expected ETH balance assertions in the test.


LePremierHomme:
@erkarl thanks. It can already be reviewed - I just had some issues with Travis, and currently also with GitHub Actions, but it's unrelated to the main work. 

I'll continue to work on the CI (Travis / GitHub Actions), to see what's the best way to run this, since it's pretty heavy and unstable.

erkarl:
Gave it another scroll-through review. Looking good - no major comments apart from cleaning up the TODOs in the code.

LePremierHomme:
@erkarl can you give a closer look at the amount values that are used in the new test case? I wanted someone else to verify that these are actually the expected values. 
If it's not clear enough, tell me and i'll describe it here in words.

test/simulation/docker-build.sh
```diff
@@ -1,5 +1,14 @@
...
 
-docker-compose build $@
-
+if [[ $@ == "connext" || $# == 0 ]]
+then
+    mkdir -p temp
+    pushd temp
+    git clone https://github.com/ConnextProject/indra.git
+    cd indra
+    git checkout ab4daf9b63726ab5ca3d1903d910a2682b8855ef
```
erkarl, line 9: think we can upgrade to `7.0.0.alpha6` here since our simnet is now also running that

test/simulation/docker-connext/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:12-alpine3.11
+
+RUN apk add --no-cache git bash python3 make g++ python
+
+WORKDIR /app
+
+RUN git clone https://github.com/connext/rest-api-client.git
+RUN cd rest-api-client && git pull
+RUN cd rest-api-client && git checkout b58fa301f041af579fc374e25bc51dc726e3dda8
```
erkarl, line 9: and here too - upgrade the client to `64dd1eceba441297f4c31f3389f3910a453d550c`

erkarl:
> @erkarl can you give a closer look at the amount values that are used in the new test case? I wanted someone else to verify that these are actually the expected values.
> If it's not clear enough, tell me and i'll describe it here in words.

Yes, some comments for the expected values would be helpful :). Thanks.

LePremierHomme:
@erkarl 
1) After funding Alice with 2000000000000000000 of ETH, its xud's ETH wallet balance is 200000000. Is that expected?
2) After Alice open an ETH channel with Bob, its total balance (channel+wallet) goes drops immediately from 200000000 to 199979000. Is that expected?

LePremierHomme:
I'm looking at the tests failure reason (also happens in @sangaman's PR, so it might not be related to newly added code), but regarding travis - wasn't it already removed from the repo settings?

erkarl:
> @erkarl
> 
>     1. After funding Alice with 2000000000000000000 of ETH, its xud's ETH wallet balance is 200000000. Is that expected?

XUD outputs amounts in "satoshis per coin", 10 to the power of 8.


```
aliceEthFundingInWei = 2000000000000000000
2000000000000000000
expectedXudBalance = 200000000 * 10 ** 8
20000000000000000
```

>     2. After Alice open an ETH channel with Bob, its total balance (channel+wallet) goes drops immediately from 200000000 to 199979000. Is that expected?

Who is Bob in this context? Connext currently does not support multi-hop. Technically all channels are to the node. The node's balance is expected to drop a little because the node is paying for the funding transaction fees.

LePremierHomme:
> XUD outputs amounts in "satoshis per coin", 10 to the power of 8.

So if 2000000000000000000 wei are 2 ether (10 ** 18), xud balance should be 2 * 10 ** 8 and that’s 200000000, which is indeed the asserted value. 
Can you explain how did you calculated `expectedXudBalance` in your comment (which doesn't match the asserted value)?

> Who is Bob in this context? Connext currently does not support multi-hop. Technically all channels are to the node. The node's balance is expected to drop a little because the node is paying for the funding transaction fees.

My bad, channel is indeed opened to the node. Regarding the balance - just wanted check if the dropped amount make sense.

erkarl:
> So if 2000000000000000000 wei are 2 ether (10 ** 18), xud balance should be 2 * 10 ** 8 and that’s 200000000, which is indeed the asserted value.
Can you explain how did you calculated expectedXudBalance in your comment (which doesn't match the asserted value)?

Oops, sorry you're right. All quantities received from xud are satoshis per coin (10 ** 8) so converting that to wei would be 10 ** 10.

```
const aliceEthFundingInWei = 2000000000000000000
const expectedXudBalance = 200000000 * 10 ** 10
aliceEthFundingInWei === expectedXudBalance
true
```

> My bad, channel is indeed opened to the node. Regarding the balance - just wanted check if the dropped amount make sense.

Yes, this is expected because at the moment the node is paying for the funding transaction fee.

LePremierHomme:
@erkarl looks like only your approval is pending before merging, unless @sangaman wants to have a look as well.

erkarl:
@LePremierHomme 
Looks like merging https://github.com/ExchangeUnion/xud/pull/1680 caused some merge conflicts.

Could you please give it a rebase before I give it final review.

LePremierHomme:
@erkarl it's good now.

LePremierHomme:
@erkarl @sangaman can you please try to run it on your machine? I want to verify that we don't have issues with the `172.17.0.1` gateway (the static IP equivalent of `host.docker.internal` for accessing  the `localhost` of the host machine).

erkarl:
> @erkarl @sangaman can you please try to run it on your machine? I want to verify that we don't have issues with the `172.17.0.1` gateway (the static IP equivalent of `host.docker.internal` for accessing the `localhost` of the host machine).

The tests passed for me.

```
PASS
ok      github.com/ExchangeUnion/xud-simulation 578.635s
```

PR #1641, opened 2020-06-12
fix(rpc): close listeners for streaming calls

sangaman:
This removes internal listeners that are created to handle streaming rpc calls. When the rpc call is canceled, we remove all listeners related to that call.

Fixes #1640.

PR #1665, opened 2020-06-16
refactor(rpc): use rxjs for streaming calls

sangaman:
This is a refactoring followup to #1641 per the review here https://github.com/ExchangeUnion/xud/pull/1641#pullrequestreview-429758175.

### DIFF
--- a/lib/grpc/GrpcService.ts
+++ b/lib/grpc/GrpcService.ts
@@ -1,5 +1,7 @@
 /* tslint:disable no-floating-promises no-null-keyword */
-import grpc, { status, ServerWritableStream } from 'grpc';
+import grpc, { ServerWritableStream, status } from 'grpc';
+import * as rxjs from 'rxjs';
+import { take } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


lib/grpc/GrpcService.ts
```diff
@@ -1,5 +1,7 @@
 /* tslint:disable no-floating-promises no-null-keyword */
-import grpc, { status, ServerWritableStream } from 'grpc';
+import grpc, { ServerWritableStream, status } from 'grpc';
+import * as rxjs from 'rxjs';
```
erkarl, line 3: Minor nit: think we can `import { fromEvent } from 'rxjs'` here since it's the only one being used in this file.

erkarl:
@sangaman I've pushed a fix for the simulation tests in this branch. The issue was that the `swapFailed$` was listening for the `swap.paid` event.

https://github.com/ExchangeUnion/xud/pull/1665/commits/51bf219b53f12fb4976b469d3ab34b7f62b5a9a8#diff-656a293cf843469fb25803adc7fee5bfR750

PR #1666, opened 2020-06-16
feat(config): enable xud encryption by default

sangaman:


sangaman:
I modified the tests so that they continue to not use encryption, which keeps things simple from a testing perspective.

raladev:
- [ ] incorrect path to seedutil for native setup
Screenshot from 2020-06-18 15-53-22

sangaman:
>     * [ ]  incorrect path to seedutil for native setup

I suspect this is because seedutil hasn't been compiled. Can you try running `npm run compile:seedutil` and try again? We perhaps should make this part of the regular installation procedure, although it introduces a dependency on Go it seems appropriate since it's required for the default procedure to create an xud noe.

raladev:
> > ```
> > * [ ]  incorrect path to seedutil for native setup
> > ```
> 
> I suspect this is because seedutil hasn't been compiled. Can you try running `npm run compile:seedutil` and try again? We perhaps should make this part of the regular installation procedure, although it introduces a dependency on Go it seems appropriate since it's required for the default procedure to create an xud noe.

I did it, it is a part of installation.
https://docs.exchangeunion.com/development/native-installation
and it was compiled, I used it manually from xud/seedutil folder.

I thought a little bit about that:
Its because I ran xud from xud/bin folder instead of xud/ folder.
So it cant find path from xud/bin because path of ./seedutil/seedutil.

sangaman:
@raladev I've identified and fixed the issue you describe, but since it's not related to xud encryption I'll handle it in another PR. Can you approve this PR so we can merge it?

PR #1671, opened 2020-06-17
chore: add GitHub actions

michael1011:
This PR adds build configurations GitHub actions. I split the GitHub action builds in three pieces:

- CI: compiling, linting and testing XUD
- Seedutil: compiling and testing the seedutil
- Simulation tests: building and running the simulation tests

The minor change in `parseGitCommit.js` was required because of the way GitHub actions clone the repository but it shouldn't affect any other builds at all.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,33 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        platform: [ubuntu-latest]
+        node-version: [12.x]
+
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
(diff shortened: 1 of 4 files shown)


michael1011:
> The only question I'd have is whether it makes sense to split the linting check into its own workflow? That's more in line with how travis is currently set up.

I don't think it makes sense tbh. Right now all the XUD related things (compile, lint and test) are in one workflow and I'd prefer having it that way (one workflow per application (although the simulation tests aren't technically a separate application but you get what I mean)).

sangaman:
> I don't think it makes sense tbh. Right now all the XUD related things (compile, lint and test) are in one workflow and I'd prefer having it that way (one workflow per application (although the simulation tests aren't technically a separate application but you get what I mean)).

That's fine with me. The "pending" travis test has actually passed so I'm going to go ahead and merge this.

kilrau:
Travis tests passed for a PR making itself obsolete. How ironic.

PR #1676, opened 2020-06-18
Add prettier

ghost:
This PR adds `prettier` development dependency to the project in order to enforce consistent style across the project.

2 new commands are added:
- `npm run prettier:check` - used by `npm run test` to make the tests fail if new code is not formatted according to prettier rules
- `npm run prettier:fix` - used by developers to automatically format the code according to prettier rules


erkarl:
Looks like the build is failing right now because the simulation tests are trying to patch xud code, but can't find the correct place because of the style changes.

package.json
```diff
@@ -29,7 +29,7 @@
     "slate": "proto2slate proto/xudrpc.proto",
     "start": "node dist/Xud.js",
     "stop": "cross-os stop",
-    "test": "npm run test:unit && npm run test:crypto && npm run test:int && npm run test:p2p && npm run test:jest",
+    "test": "npm run prettier:check && npm run test:unit && npm run test:crypto && npm run test:int && npm run test:p2p && npm run test:jest",
```
michael1011, line 32: Shouldn't the prettier check be parth of the `lint` script?

sangaman (commented):
My main feedback would be to make the rules less strict (at least for now) to minimize the amount of existing code that changes. 10k lines is a lot, and although I haven't looked through all of it it suggests to me that this is enforcing a lot of new style rather than existing style. I can make some suggestions on what to tweak if you'd like - a lot of the changes seem to be around splitting single line statements into multiple lines. It also looks like this is adding a lot of semi-colons to the end of member methods which we didn't do at all before from what I can tell.

Some changes to existing code is fine of course but I'd want to try to enforce mainly styles that are already very common. Fewer changes to existing code might also help with the custom-xud patch.

I also agree that prettier should be part of the `lint` script rather than `test` - in fact I think it might make sense to just incorporate prettier into the lint scripts rather than have separate scripts for prettier only. They both relate to code style.

erkarl:
> My main feedback would be to make the rules less strict (at least for now) to minimize the amount of existing code that changes. 10k lines is a lot, and although I haven't looked through all of it it suggests to me that this is enforcing a lot of new style rather than existing style. I can make some suggestions on what to tweak if you'd like - a lot of the changes seem to be around splitting single line statements into multiple lines. It also looks like this is adding a lot of semi-colons to the end of member methods which we didn't do at all before from what I can tell.
> 
> Some changes to existing code is fine of course but I'd want to try to enforce mainly styles that are already very common. Fewer changes to existing code might also help with the custom-xud patch.

I'm open to suggestions - https://prettier.io/docs/en/options.html

At the moment it's using almost default configuration. Once we agree upon the rules I'll modify the `custom-xud.patch` to make the simulation tests pass.

> I also agree that prettier should be part of the `lint` script rather than `test` - in fact I think it might make sense to just incorporate prettier into the lint scripts rather than have separate scripts for prettier only. They both relate to code style.

Agreed. Deleted the prettier scripts and moved them to `lint` and `lintNoFix`.

erkarl:
It also looks like that `tslint --fix` is overwriting prettier style changes. I'll also upgrade to `eslint` in this PR and configure it to not apply style related changes.

michael1011:
Isn't tslint going to be deprecated in favor of eslint with a TypeScript extension?

erkarl:
> Isn't tslint going to be deprecated in favor of eslint with a TypeScript extension?

It's deprecated, yes. That's why I mentioned above that I'll upgrade to eslint.

sangaman:
> I'll also upgrade to `eslint` in this PR and configure it to not apply style related changes.

Big +1 for this, something I've wanted to do for a while and started tinkering with at one point but didn't complete. FWIW I think it's OK if the linter makes style related changes as well, just so long as it doesn't conflict with prettier of course. But I think that's a good example where we should try to keep the style rules currently enforced by the linter (whether or not they are enforced by prettier or linting going forward).

I'll suggest some rules to tweak later tonight, thanks.

erkarl:
Putting this PR on hold until migration to eslint is complete: https://github.com/ExchangeUnion/xud/pull/1677

Issue #864, opened 2019-04-02
[Concept] Channel unbalanced alerts


kilrau:
Continuation of #680:
This issue is about using #680's cached values to send alert events via the gRPC to inform the exchange that a channel is unbalanced. Details TBD.

kilrau:
Since you are also on #680, I think it makes sense to also take this one @erkarl . I think we need a new `SubscribeAlerts` or more specific `SubscribeChannelAlerts`.

sangaman:
We ought to discuss this on the call this week to settle on a good approach.

kilrau:
Recap from the call: check other projects for how they transport non-call specific alerts. Apart from that no show-stoppers so far @sangaman @erkarl

erkarl:
Conclusion from the call: 1 call for generic events. Just type and description.

engwarrior:
Pls provide estimate for this task.

kilrau:
Still interested? @engwarrior

kilrau:
Would you be interested in this one? @rsercano

rsercano:
Yes of course, let me check :)

kilrau:
https://github.com/ExchangeUnion/xud-ui/issues/34 depends on this

PR #1680, opened 2020-06-18
 feat(swaps): new PreimageResolved swap phase 

### LINKED ISSUE #1654: eth/btc swap shows successful for taker, but failed for maker (was successful, no crash, slow disk involved)
### Background

Note: This swap being on the brink of timing out is legit since the taker disk has been busy/slow.

Orderbook pre-swap:
```
testnet > listorders

Trading pair: ETH/BTC
│ 0.001         │ 0.008       │ VersionUphold    │ 0.001         │ 0.01        │ VersionUphold    │
│ 0.001         │ 0.008       │ VersionUphold    │               │             │                  │
```
Taker: long, but successful swap
```
testnet > sell 0.001 eth/btc 0.008
matched 0.001 ETH @ 0.008 with peer VersionUphold order 90abe1a0-acd0-11ea-9d8c-d70762ca79c7, attempting swap...
successfully swapped 0.001 ETH with peer order 90abe1a0-acd0-11ea-9d8c-d70762ca79c7
testnet > listorders

Trading pair: ETH/BTC
│ 0.001         │ 0.008       │ VersionUphold    │ 0.001         │ 0.01        │ VersionUphold    │
```
```
testnet > tradehistory
Trades:
┌─────────────────┬───────────┬───────┬─────────────┬─────────────────────────────┬───────────┬───────────────────────┐
│ Execution       │ Price     │ Role  │ Order Id    │ Order Id    (Counterparty)  │ Swap Hash │ Executed At           │
├─────────────────┼───────────┼───────┼─────────────┼─────────────────────────────┼───────────┼───────────────────────┤
│ Sell 0.001 ETH  │ 0.008 BTC │ Taker │ 1e4772d0... │ 90abe1a0... (VersionUphold) │ 4fa363... │ 6/12/2020, 5:28:58 PM │
├─────────────────┼───────────┼───────┼─────────────┼─────────────────────────────┼───────────┼───────────────────────┤
```
Maker: nothing (that's an old swap you see there)
```
testnet > tradehistory
Trades:
┌─────────────────┬───────────┬───────┬─────────────┬────────────────────────────┬───────────┬───────────────────────┐
│ Execution       │ Price     │ Role  │ Order Id    │ Order Id    (Counterparty) │ Swap Hash │ Executed At           │
├─────────────────┼───────────┼───────┼─────────────┼────────────────────────────┼───────────┼───────────────────────┤
│ Buy 0.005 LTC   │ 0.09 BTC  │ Maker │ 071b92c0... │ N/A (ShadowCore)           │ f82c1f... │ 6/12/2020, 3:12:51 PM │
├─────────────────┼───────────┼───────┼─────────────┼────────────────────────────┼───────────┼───────────────────────┤
```

[Taker xud logs](https://paste.ubuntu.com/p/sPMXJJF3s7/) (pay attention to the `received unexpected swap failed packet` entry)
[Taker connext logs](https://paste.ubuntu.com/p/SrqRwP6kVg/) (shows successfully sending (installing) and later unlock event of eth payment htlc)
Taker lndbtc shows successfully receiving and settling lndbtc htlc payment:
```
2020-06-12 17:28:31.654 [INF] INVC: New single invoice subscription client: id=5, hash=4fa363f933ceed754201a2a979d82b06281c95302a1795395ecaf594c0ba5987
2020-06-12 17:28:57.208 [INF] CRTR: Processed channels=0 updates=3 nodes=0 in last 59.99990519s
2020-06-12 17:28:57.570 [INF] HSWC: ChannelLink(1769796:8:1): settling htlc 4fa363f933ceed754201a2a979d82b06281c95302a1795395ecaf594c0ba5987 as exit hop
```
[Maker xud logs](https://paste.ubuntu.com/p/VPPhBsKNQM/) (also pay attention to the `**this must be resolved manually**` entry)
[Maker connext logs](https://paste.ubuntu.com/p/vGkxDWdGsF/)(shows successfully receiving and unlocking eth payment htlc)
Maker lndbtc shows successfully sending lndbtc payment:
```
2020-06-12 17:28:57.878 [INF] SRVR: Adding preimage=6c180d37c22a37aba394e9a1704d7be64cad51ff126b3c5edc95ebfb8babcb75 to witness cache
```
There is nothing in the maker logs thereafter of swap recovery picking this up (probably because xud is convinced this swap failed indeed).

TODO: figure out how the taker xud came to the conclusion that the swap failed whereas it obviously didn't (and fix it up). Since we are very close to the 30s, my best guess is overlapping timeouts once more.

### LINKED ISSUE #1659: xud triggers SwapTimeout error for successful swap.
### Background

Reproduction: Rare

One of the nodes was run with arby, another one was run without arby and performed manual swaps. One of the swaps was marked as successful in the cli, but got SwapTimeOut in the logs.

[Screen](https://user-images.githubusercontent.com/29906866/84661138-93b3bb80-af22-11ea-88b3-26fb023ea713.png)

### Logs
Time of swap = 11:02:38
Swap hash = 222dc469651b4eb912df2fe5a9f627a3816e66fa2f41639faf4c1326f0c8d0e2

[maker_f2.zip](https://github.com/ExchangeUnion/xud/files/4780410/maker_f2.zip)
[taker_f2.zip](https://github.com/ExchangeUnion/xud/files/4780477/taker_f2.zip)

sangaman:
This creates a new phase for SwapDeals called `PreimageResolved` that represents the part after the maker has completed its payment to taker and resolved the preimage, but before it has settled its incoming payment using the preimage.

Swaps that have reached this phase are considered atomic - our outgoing payment has been settle and we've acquired the ability to settle our incoming payment - even if they're not complete. Therefore, the maker ends the swap timeout and gives the call to settle the incoming payment as long as it needs to complete while keeping the hold on the order to prevent it from being filled a second time.

Should a swap fail the `settleInvoice` call, it enters swap SwapRecovery where settling the payment will be attempted one more time but only after ensuring we are connected to the SwapClient responsible for that payment. Previously, deals that hit an error upon the `settleInvoice` would not fail the deal and send it to recovery until the timeout was reached.

We persist the preimage to the database immediately upon reaching the `PreimageResolved` phase and then use that preimage in SwapRecovery if it is available, rather than querying the swap client for the outgoing payment to lookup the preimage again.

It also introduces a new level of alert logging, with greater severity than the error level, to be used only for warning a user when loss of funds is possible and cannot be handled automatically by xud due to unexpected errors or conditions.

Resolves #1654. Resolves #1659.

### DIFF
--- a/lib/constants/enums.ts
+++ b/lib/constants/enums.ts
@@ -110,10 +110,21 @@ export enum ReputationEvent {
   SwapSuccess = 5,
   /** When a swap is accepted and is attempted to be executed but fails. */
   SwapFailure = 6,
-  /** When a swap is accepted and then fails due to exceeding time limits. */
+  /** When a swap fails due to exceeding time limits. */
   SwapTimeout = 7,
   /** When a swap fails due to unexpected or possibly malicious behavior. */
   SwapMisbehavior = 8,
+  /**
+   * The peer has behaved dishonestly during a swap in a way that strongly
+   * suggests it is running modified malicious code.
+   */
+  SwapAbuse = 9,
+  /**
+   * The peer completed a swap that was delayed beyond the expected deadline
+   * to complete or fail the swap deal, but the delay was not so long as to
+   * preclude the possibility of latency or lag.
+   */
+  SwapDelay = 10,
 }
 
 export enum SwapFailureReason {
--- a/lib/constants/enums.ts
+++ b/lib/constants/enums.ts
@@ -64,21 +64,26 @@ export enum SwapRole {
 }
 
 export enum SwapPhase {
-  /** The swap deal has been created locally. */
+  /** 0/5 The swap deal has been created locally. */
   SwapCreated = 0,
-  /** We've made a request to a peer to accept this swap. */
+  /** 1/5 We've made a request to a peer to accept this swap. */
   SwapRequested = 1,
-  /** The terms of the swap have been agreed to, and we will attempt to execute it. */
+  /** 2/5 The terms of the swap have been agreed to, and we will attempt to execute it. */
   SwapAccepted = 2,
   /**
-   * We have made a request to the swap client to send payment according to the agreed terms.
+   * 3/5 We have made a request to the swap client to send payment according to the agreed terms.
    * The payment (and swap) could still fail due to no route with sufficient capacity, lack of
    * cooperation from the receiver or any intermediary node along the route, or an unexpected
    * error from the swap client.
    */
   SendingPayment = 3,
   /**
-   * We have received the agreed amount of the swap and released the preimage to the
+   * 4/5 We have completed our outgoing payment and retrieved the preimage which can be used to
+   * settle the incoming payment locked by the same hash.
+   */
+  PreimageResolved = 5,
+  /**
+   * 5/5 We have received the agreed amount of the swap and released the preimage to the
    * receiving swap client so it can accept payment.
    */
   PaymentReceived = 4,
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -82,7 +82,7 @@ class Swaps extends EventEmitter {
   ) {
     super();
 
-    this.swapRecovery = new SwapRecovery(swapClientManager, logger);
+    this.swapRecovery = new SwapRecovery(swapClientManager, logger.createSubLogger('RECOVERY'));
     this.repository = new SwapRepository(this.models);
     this.bind();
   }
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -25,8 +25,10 @@ export type OrderToAccept = Pick<SwapDeal, 'quantity' | 'price' | 'localId' | 'i
 interface Swaps {
   on(event: 'swap.paid', listener: (swapSuccess: SwapSuccess) => void): this;
   on(event: 'swap.failed', listener: (deal: SwapDeal) => void): this;
+  on(event: 'swap.recovered', listener: (recoveredSwap: SwapDealInstance) => void): this;
   emit(event: 'swap.paid', swapSuccess: SwapSuccess): boolean;
   emit(event: 'swap.failed', deal: SwapDeal): boolean;
+  emit(event: 'swap.recovered', recoveredSwap: SwapDealInstance): boolean;
 }
 
 class Swaps extends EventEmitter {
@@ -221,6 +223,10 @@ class Swaps extends EventEmitter {
     this.swapClientManager.on('lndUpdate', this.pool.updateLndState);
     this.swapClientManager.on('raidenUpdate', this.pool.updateRaidenState);
     this.swapClientManager.on('connextUpdate', this.pool.updateConnextState);
+
+    this.swapRecovery.on('recovered', (recoveredSwap) => {
+      this.emit('swap.recovered', recoveredSwap);
+    });
   }
 
   /**
@@ -1411,6 +1417,8 @@ class Swaps extends EventEmitter {
   }
 
   public close = () => {
+    this.swapClientManager.removeAllListeners();
+    this.swapRecovery.removeAllListeners();
     this.swapRecovery.stopTimer();
   }
 }
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -12,6 +12,7 @@ import Peer from '../p2p/Peer';
 import Pool from '../p2p/Pool';
 import { generatePreimageAndHash, setTimeoutPromise } from '../utils/utils';
 import errors, { errorCodes } from './errors';
+import SwapClient, { PaymentState } from './SwapClient';
 import SwapClientManager from './SwapClientManager';
 import SwapRecovery from './SwapRecovery';
 import SwapRepository from './SwapRepository';
@@ -53,6 +54,20 @@ class Swaps extends EventEmitter {
   private static readonly SWAP_ACCEPT_TIMEOUT = 10000;
   /** The maximum time in milliseconds we will wait for a swap to be completed before failing it. */
   private static readonly SWAP_COMPLETE_TIMEOUT = 35000;
+  /**
+   * Additional time that the maker will wait for a swap to be completed before considering it timed
+   * out. This exists because the maker starts timing sooner and ends timing later than the taker.
+   * The maker starts timing as soon as it sends its SwapAccepted packet, but taker starts upon
+   * receiving that packet some short time later. Furthermore, the taker stops the timer as soon as
+   * it reveals the preimage and settles its incoming payment, whereas the maker doesn't stop until
+   * it receives the preimage.
+   */
+  private static readonly SWAP_COMPLETE_MAKER_BUFFER = 5000;
+  /**
+   * The time threshold in milliseconds after which we consider a counterparty abusive if they
+   * settle payment for a timed out swap.
+   */
+  private static readonly SWAP_ABUSE_TIME_LIMIT = 60000;
   /** The maximum time in milliseconds we will wait to receive an expected sanity swap init packet. */
   private static readonly SANITY_SWAP_INIT_TIMEOUT = 3000;
   /** The maximum time in milliseconds we will wait for a swap to be completed before failing it. */
@@ -202,19 +217,7 @@ class Swaps extends EventEmitter {
     this.pool.on('packet.swapAccepted', this.handleSwapAccepted);
     this.pool.on('packet.swapFailed', this.handleSwapFailed);
 
-    this.swapClientManager.on('htlcAccepted', async (swapClient, rHash, amount, currency) => {
-      try {
-        const rPreimage = await this.resolveHash(rHash, amount, currency);
-        await swapClient.settleInvoice(rHash, rPreimage, currency);
-
-        const deal = this.getDeal(rHash);
-        if (deal) {
-          await this.setDealPhase(deal, SwapPhase.PaymentReceived);
-        }
-      } catch (err) {
-        this.logger.error('could not settle invoice', err);
-      }
-    });
+    this.swapClientManager.on('htlcAccepted', this.handleHtlcAccepted);
     this.swapClientManager.on('lndUpdate', this.pool.updateLndState);
     this.swapClientManager.on('raidenUpdate', this.pool.updateRaidenState);
     this.swapClientManager.on('connextUpdate', this.pool.updateConnextState);
@@ -455,8 +458,6 @@ class Swaps extends EventEmitter {
       createTime: Date.now(),
     };
 
-    this.timeouts.set(rHash, setTimeout(this.handleSwapTimeout, Swaps.SWAP_ACCEPT_TIMEOUT, rHash, SwapFailureReason.DealTimedOut));
-
     this.addDeal(deal);
 
     // Make sure we are connected to both swap clients
@@ -560,8 +561,6 @@ class Swaps extends EventEmitter {
       role: SwapRole.Maker,
       createTime: Date.now(),
     };
-
-    this.timeouts.set(rHash, setTimeout(this.handleSwapTimeout, Swaps.SWAP_COMPLETE_TIMEOUT, rHash, SwapFailureReason.SwapTimedOut));
 
     // add the deal. Going forward we can "record" errors related to this deal.
     this.addDeal(deal);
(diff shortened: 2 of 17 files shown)


sangaman:
I updated this PR based on our discussions from last week, see the commit message for details (quoted below). I tried to thoroughly comment the code as well to make review and comprehension easier. The simulation tests still need some work to adjust to these changes (including updating the custom-xud patch) so I will finish that tomorrow, but the code can be tested and reviewed now.

> This draft commit modifies the timeout behavior for the maker in the swap. The maker will no longer "fail" a deal once it has begun sending its payment to the taker due to a swap timing out. Instead it will only notify the taker that its time limit has been reached, and a cooperative taker will cancel its invoice and call off the swap.
> 
> The maker also tracks the elapsed time for a swap, and if its payment to taker succeeds after the time out period it penalizes the taker for accepting payment and settling its invoice too late. Takers that delay before they settle payment can abuse the free options problem.
> 
> On the swap recovery side, we no longer fail a recovered deal if a settle invoice call fails. Instead, we continuosly attempt settle invoice calls until one succeeds. Longer term we can permanently stop settle invoice attempts if we can recognize the sorts of errors that indicate permanent failure such as htlc expiration or an unrecognized invoice hash.

erkarl:
Opened xud-docker PR so that this can be manually tested by @kilrau and @raladev 
https://github.com/ExchangeUnion/xud-docker/pull/568

lib/constants/enums.ts
```diff
@@ -77,6 +77,11 @@ export enum SwapPhase {
    * error from the swap client.
    */
   SendingPayment = 3,
+  /**
+   * We have completed our outgoing payment and retrieved the preimage which can be used to
+   * settle the incoming payment locked by the same hash.
+   */
+  PreimageResolved = 5,
```
kilrau, line 84: Since the enmus are not in line with the swap phases now, maybe add swap phase `1/5` to `5/5` to the comments .

kilrau:
As just discussed with @sangaman this an attempt of an exhaustive **overview** of what we discussed in the course of this PR and **what is/has to be covered here and what will be covered in follow-up PRs**:

- [x] add new phase on maker side `PreimageResolved`, when preimage was successfully written into xud db (**this PR**)
- [x] mark swap as successful, show in `tradingHistory` in this stage (**this PR**)
- [x]  adjust swap timeout to end at `PreimageResolved` (**this PR**)
- balance shown as pending since it's not claimed (https://github.com/ExchangeUnion/xud/issues/1684)
- [x] maker no longer fails a deal due to timeout(!). Instead punishes with neg reputation event after 40s (this is the new soft timeout where maker asks taker to gracefully cancel swap) (**this PR**)
- [x] after another 60s (new hard timeout after 100s in total) xud simply continues trying to claim the htlc. If taker doesn't cooperate, we can assume the swap client will automatically go on-chain before htlc timeout to claim funds (**this PR**)
- [ ] if taker releases preimage after 100s hard timeout, immediately BAN the peer because we consider this a free option exploit (**this PR**)
- [ ] Record swaps which were in swap recovery but then unexpectedly succeeded (taker did release the preimage, exploited free option) in db. Should show in `tradehistory`. Maybe we can mark it with `FOP Alert!` or something getting the users attention (**this PR**)
- additionally we would want to send send out an alert [via the new alert streaming call](https://github.com/ExchangeUnion/xud/issues/864) if xud does not successfully claim the htlc for e.g. 1h. The swap client might have crashed. (https://github.com/ExchangeUnion/xud/issues/1686)
- negotiate timeouts in swap request phase (https://github.com/ExchangeUnion/xud/issues/1685)
- [x] increase maker soft timeout to 40s, leave taker at 35s
- [x] retry indefinitely to claim htlc
- [ ] above mentioned unit/integration tests and simulation test fixes

lib/constants/enums.ts
sangaman: Done.

lib/swaps/Swaps.ts
```diff
@@ -677,6 +676,53 @@ class Swaps extends EventEmitter {
     return true;
   }
 
+  private handleHtlcAccepted = async (swapClient: SwapClient, rHash: string, amount: number, currency: string) => {
```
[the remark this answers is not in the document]
sangaman: I will make a follow-up issue for this one since the PR here is only breaking it into its own named function and I don't want this to get too much larger.

lib/swaps/Swaps.ts
```diff
@@ -202,19 +217,7 @@ class Swaps extends EventEmitter {
...
-
-        const deal = this.getDeal(rHash);
-        if (deal) {
-          await this.setDealPhase(deal, SwapPhase.PaymentReceived);
-        }
-      } catch (err) {
-        this.logger.error('could not settle invoice', err);
-      }
-    });
+    this.swapClientManager.on('htlcAccepted', this.handleHtlcAccepted);
```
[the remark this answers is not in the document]
sangaman: Done.

sangaman:
I added the suggested checks and new test case for the `PreimageResolved` status to our SwapRecovery test suite, that is all passing.

However I am stuck on the `maker_shutdown_after_1st_htlc` security simulation test, one that I know I've had issues with in the past. The xud logs & behavior appear to be correct, the maker shuts down and the taker times out and fails the deal. There is a lingering HTLC from taker to maker that maker can never claim because he does not know the preimage, we expect lnd to force close the channel after that HTLC expires but that's not happening. I'm not sure why as this PR should not change this scenario - it is not one where there is any risk of losing funds given that taker has control of the preimage and fails the deal. I will ping @LePremierHomme since he's worked on this test case in the past, but if it continues to give us problems I'd consider simplifying the test case to only ensure that taker properly times out and fails the deal, without putting in all these checks as to how lnd deals with a stuck/expired incoming HTLC.

Below is the error.

```
--- FAIL: TestSecurityUnsettledChannels (90.70s)
    --- PASS: TestSecurityUnsettledChannels/network_initialization (0.59s)
    --- FAIL: TestSecurityUnsettledChannels/maker_shutdown_after_1st_htlc (69.77s)
        tests-security.go:296: 
            	Error Trace:	tests-security.go:296
            	            				harnessTest.go:95
            	            				xud_test.go:267
            	Error:      	Not equal: 
            	            	expected: 1
            	            	actual  : 0
            	Test:       	TestSecurityUnsettledChannels/maker_shutdown_after_1st_htlc
```

lib/swaps/Swaps.ts
```diff
@@ -1192,45 +1301,70 @@ class Swaps extends EventEmitter {
...
+        // we treat the swap as having succeeded once we have resolved the preimage as the maker
+        succeedSwap(true);
+
+        /** The number of milliseconds elapsed since we started execting this swap. */
+        const elapsedMilliseconds = Date.now() - deal.executeTime!;
+
+        // we check that the taker did not release the preimage after the swap has timed out
+        // if so, they are misbehaving by completing a swap too late and possibly exploiting
+        // us with the free option problem
+        if (elapsedMilliseconds > Swaps.SWAP_COMPLETE_TIMEOUT + Swaps.SWAP_ABUSE_TIME_LIMIT) {
```
erkarl, line 1348: I think this code (alert + reputation event) below is never reached. I wasn't able to reproduce `ReputationEvent.SwapDelay` or `ReputationEvent.SwapAbuse` being added during my manual tests where I:
- modified the taker's xud code to never release the preimage
- manually claimed the payment after `Swaps.SWAP_COMPLETE_TIMEOUT + Swaps.SWAP_ABUSE_TIME_LIMIT`
- manually claimed the payment after `Swaps.SWAP_COMPLETE_TIMEOUT + Swaps.SWAP_COMPLETE_MAKER_BUFFER`

What happened in my test environment was that the maker just recovered the payment, but did not ban the peer for abuse:
```
26/06/2020 09:06:10.275 [SWAPS] debug: checking outgoing payment status for swap 7e745d10dc4ba8b15183bc5cfe964aae2c1b4e41c6e5c6ca05df8348111b7f81
getting token address for currency ETH
26/06/2020 09:06:10.395 [LND-BTC] debug: settling invoice for 7e745d10dc4ba8b15183bc5cfe964aae2c1b4e41c6e5c6ca05df8348111b7f81 with preimage c0457d76c9434d341ff07b1bf12ad0f19f77002491f395698de6890f7c2d68e8
26/06/2020 09:06:10.398 [SWAPS] info: recovered BTC swap payment of 100000 using preimage c0457d76c9434d341ff07b1bf12ad0f19f77002491f395698de6890f7c2d68e8
26/06/2020 09:06:10.400 [LND-BTC] debug: deleted invoice subscription for 7e745d10dc4ba8b15183bc5cfe964aae2c1b4e41c6e5c6ca05df8348111b7f81
```

lib/swaps/Swaps.ts
```diff
@@ -961,42 +1013,60 @@ class Swaps extends EventEmitter {
...
+              break;
+          }
+        };
 
-        throw err;
+        // the payment failed but we first double check its final status, so we
+        // only fail the deal when we know our payment won't go through. otherwise
+        // we extract the preimage if the payment went through in spite of the error
+        // or we fail the deal and go to SwapRecovery if it's still pending
+        const paymentStatus = await swapClient.lookupPayment(rHash, deal.makerCurrency, deal.destination);
```
erkarl, line 1056: Think this should be `deal.takerCurrency` as maker's payment is in taker's currency.

lib/swaps/Swaps.ts
sangaman: Thanks for the thorough test, I will investigate.

sangaman: I'm adding a test case for this to ensure the logic is correct.

lib/swaps/Swaps.ts
sangaman: Good catch, you are right. It's not a parameter that's used for LND but it wouldn't work for connext.

erkarl:
> I added the suggested checks and new test case for the `PreimageResolved` status to our SwapRecovery test suite, that is all passing.
> 
> However I am stuck on the `maker_shutdown_after_1st_htlc` security simulation test, one that I know I've had issues with in the past. The xud logs & behavior appear to be correct, the maker shuts down and the taker times out and fails the deal. There is a lingering HTLC from taker to maker that maker can never claim because he does not know the preimage, we expect lnd to force close the channel after that HTLC expires but that's not happening. I'm not sure why as this PR should not change this scenario - it is not one where there is any risk of losing funds given that taker has control of the preimage and fails the deal. I will ping @LePremierHomme since he's worked on this test case in the past, but if it continues to give us problems I'd consider simplifying the test case to only ensure that taker properly times out and fails the deal, without putting in all these checks as to how lnd deals with a stuck/expired incoming HTLC.
> 
> Below is the error.
> 
> ```
> --- FAIL: TestSecurityUnsettledChannels (90.70s)
>     --- PASS: TestSecurityUnsettledChannels/network_initialization (0.59s)
>     --- FAIL: TestSecurityUnsettledChannels/maker_shutdown_after_1st_htlc (69.77s)
>         tests-security.go:296: 
>             	Error Trace:	tests-security.go:296
>             	            				harnessTest.go:95
>             	            				xud_test.go:267
>             	Error:      	Not equal: 
>             	            	expected: 1
>             	            	actual  : 0
>             	Test:       	TestSecurityUnsettledChannels/maker_shutdown_after_1st_htlc
> ```

This test failure also happens in @moshababo's PR . https://github.com/ExchangeUnion/xud/pull/1657#issuecomment-650571724

lib/swaps/Swaps.ts
sangaman: The new test cases are passing, and what I realized here only afterwards is that the logs/code above is executing from the SwapRecovery portion of the code. The logic to ban/punish peers isn't in there - only as part of the regular swap flow.

I believe what's happening here is that the outgoing Connext payment attempt fails when it isn't claimed by the taker after some time (presumably not that long), thereby throwing the swapdeal into recovery. In there we eventually detect that the payment went through, fetch the preimage, and use it to claim our BTC payment.

The reason I didn't put the penalizing/banning logic there is that we wouldn't want to punish peers in cases where xud stopped/crashed during swaps on recovered them upon restarting. In such cases it's possible that the taker released the preimage in a timely way, but our node went down. Several minutes later we'd come back on, and we wouldn't actually know when the taker accepted our payment and released the preimage.

I can probably add some logic to distinguish between recoveries where xud crashed and recoveries where sending a payment failed but the payment still reports as "pending" - or maybe modify the logic to simply not go into swap recovery and monitor pending outgoing payments within Swaps.ts - thereby leaving SwapRecovery only for cases where our xud crashed. I think I'd lean towards the latter but either way it might make sense to do that in a followup PR since this is already pretty big.

Ideally the outgoing connext payment won't return an error simply due to exceeding a time limit (that puts us back into the sort of situation where the maker is failing the deal due to a timeout, which is what we're trying to move away from in this PR).

erkarl: > I can probably add some logic to distinguish between recoveries where xud crashed and recoveries where sending a payment failed but the payment still reports as "pending" - or maybe modify the logic to simply not go into swap recovery and monitor pending outgoing payments within Swaps.ts - thereby leaving SwapRecovery only for cases where our xud crashed.

Yeah, I think this makes a lot of sense to monitor them in `Swaps`. This way we can ensure that xud did not crash and taker delayed releasing the preimage intentionally.

>  I think I'd lean towards the latter but either way it might make sense to do that in a followup PR since this is already pretty big.

Sounds good.

erkarl:
Can we also disable the failing simulation tests? I believe @moshababo is working on a fix for that in another PR.

erkarl:
This PR looks largely good to go for me once the tests are passing.

sangaman:
@LePremierHomme  Does this PR require that commit you just pushed? I see that #1657 just passed sim tests, should I wait for that to be merged or should I disable the failing test case here and merge, then #1657 can rebase and reenable the test?

LePremierHomme:
Yes, just wanted to save you the troubles of messing with that. I removed my commit.

sangaman:
I've finally gotten all the test cases to pass. I had to modify the test where the taker shuts down - since the maker no longer times out a swap, it won't fail the swap until the HTLC has expired in cases where the taker becomes unresponsive. The test case now waits for the swap to fail only after ensuring the HTLC has expired.

Oddly, locally the `testTakerStallingAfter2ndHTLC` case only passes for me if I make it the first test case to run of the three, but since it's working in GitHub Actions as-is I'll leave it be and we can change or investigate later if necessary.

I also added a commit that should make the security tests much less flaky in regards to htlc expiration. It's now using a retry loop, waiting a few seconds and checking if the channel with the expired htlc has been force closed, rather than a fixed wait. This should make the test cases both faster and more reliable.

We can merge this once it gets an approval.

erkarl:
Created https://github.com/ExchangeUnion/xud/issues/1700 as a follow-up.

PR #1692, opened 2020-06-26
feat: generate 2048-bit tls certs

reliveyy:
This PR generates 2048-bit TLS certificate instead of 1024-bit. Related to https://github.com/ExchangeUnion/xud-docker/pull/533#issuecomment-650117698

### DIFF
--- a/lib/grpc/GrpcServer.ts
+++ b/lib/grpc/GrpcServer.ts
@@ -104,11 +104,11 @@ class GrpcServer {
    * @returns the cerificate and its private key
    */
   private generateCertificate = async (tlsCertPath: string, tlsKeyPath: string): Promise<{ tlsCert: string, tlsKey: string }> => {
-    const keys = pki.rsa.generateKeyPair(1024);
+    const keys = pki.rsa.generateKeyPair(2048);
     const cert = pki.createCertificate();
 
     cert.publicKey = keys.publicKey;
-    cert.serialNumber = String(Math.floor(Math.random() * 1024) + 1);
+    cert.serialNumber = `2048-${String(Math.floor(Math.random() * 100000000))}`;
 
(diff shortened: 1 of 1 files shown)


kilrau:
As discussed on the call, @sangaman can take over this PR to see if there is an easy way to make changes like this backwards compatible

lib/grpc/GrpcServer.ts
```diff
@@ -104,11 +104,11 @@ class GrpcServer {
    * @returns the cerificate and its private key
    */
   private generateCertificate = async (tlsCertPath: string, tlsKeyPath: string): Promise<{ tlsCert: string, tlsKey: string }> => {
-    const keys = pki.rsa.generateKeyPair(1024);
+    const keys = pki.rsa.generateKeyPair(2048);
     const cert = pki.createCertificate();
 
     cert.publicKey = keys.publicKey;
-    cert.serialNumber = String(Math.floor(Math.random() * 1024) + 1);
+    cert.serialNumber = String(Math.floor(Math.random() * 2048) + 1);
```
sangaman, line 111: I don't think this actually needs to change to use 2048 bit keys. Maybe we can make the serial number indicate that the cert is 2048 bit? Then we'd have an easy to check if existing certs are 2048 or 1024.

So something like this

```javascript
cert.serialNumber = `2048-${String(Math.floor(Math.random() * 100000000))}`;
```

That would give us serial numbers that look like "2048-" followed by a random number up to 8 digits. And I could check if a cert is 2048 bit just by seeing if the serial number starts with "2048".

kilrau:
Would you still want to merge this one? @sangaman

kilrau:
Could you squash and change the commit to the `feat: ` prefix? @reliveyy

erkarl (approved):

PR #1696, opened 2020-06-27
aliases now have a third part for collision avoidance

### LINKED ISSUE #1688: Fix alias collisions
### Background

We are seeing alias collisions frequently https://github.com/ExchangeUnion/xud/issues/1687 , especially on travis. Since statistically this is extremely unlikely to happen, this issue is about finding and fixing the root cause of such.

hatmer:
closes #1688

### DIFF
--- a/lib/utils/aliasUtils.ts
+++ b/lib/utils/aliasUtils.ts
@@ -7,19 +7,23 @@ import errors from '../service/errors';
  * @param pubkey a public key in hex format to convert to an alias
  */
 export const pubKeyToAlias = (pubkey: string): string => {
-  const getWord = (substring: string): string | undefined => {
+  const getWord = (substring: string, getIndex: boolean): string | undefined => {
     const asNumber = parseInt(substring, 16);
     if (!isNaN(asNumber)) {
       const index = asNumber % wordlist.length;
+      if (getIndex) {
+        return index.toString();
+      }
       return wordlist[index];
     }
     return undefined;
   };
   if (pubkey.length === 66) {
-    const a = getWord(pubkey.slice(0, 33));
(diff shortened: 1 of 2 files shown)


raladev:
Tested manually, works.
@kilrau @sangaman  What do you think about Number solution?
Screen

hatmer:
If the extra number is bad then an alternative solution is to simply modify the wordlist file to be a set of 2048^2 aliases so that pubkeys can map directly to >4 million aliases. I probably should have tried this first, so I'll do this now.

erkarl:
> If the extra number is bad then an alternative solution is to simply modify the wordlist file to be a set of 2048^2 aliases so that pubkeys can map directly to >4 million aliases. I probably should have tried this first, so I'll do this now.

Yeah, I think that's a better approach. Thoughts @kilrau @sangaman?

sangaman:
I do think that adding a number to the alias makes it less memorable than just a couple of words. I'd rather we do a third word if necessary, or increase our word list size.

My bigger concern is whether this is necessary - have we actually seen collisions in the wild? I suspect the issue with travis alias collisions is not due to the birthday problem but rather due to some issue with pseudorandomness with tests that are running back to back or near simultaneously (as @raladev  said). The odds that we'd get collisions with just 1 or 2 peers - and that's all our tests are using I believe - is extremely low.

> an alternative solution is to simply modify the wordlist file to be a set of 2048^2 aliases

Do you mean a wordlist of 2048^2 words, over 4 million? That sounds impossible unless I'm missing something.

hatmer:
new PR at #1703 that uses math to index into a 4million+ wordlist. Using a larger hardcoded wordlist was indeed impossible.

PR #1704, opened 2020-07-01
fix(swaps): cancel taker invoice on swap fail

### LINKED ISSUE #1695: xud does not cancel maker BTC invoice when payment expired
maker xud:
- received BTC
- sent ETH

maker's payment was not claimed => timelock expired

ERROR: xud did not cancel the BTC invoice

sangaman:
This commit ensures that the taker cancels its incoming invoice when it fails while attempting to send payment.

Fixes #1695


sangaman:
The tests were failing because the `testTakerStallingAfter2ndHTLC` sim test case was expecting the taker to stall on preimage release as well as stall on canceling the invoice/failing any incoming payments. So I modified the custom-xud patch to keep the original logic.

PR #1711, opened 2020-07-03
feat(p2p): improved ban handling

sangaman:
This makes several improvements to the way xud behaves when a node is banned.

1. Only emit the `node.ban` event a single time when a node is banned, previously it was possible for this to fire multiple times when multiple negative reputation events occur at roughly the same time, polluting the logs and sending multiple disconnecting packets to the peer.

2. Log the name of the reputation event that triggered the ban.

3. Only send the recent reputation events to a peer when we ban them, rather than JSON stringifying the entire ReputationEventInstance db object which contains a lot of internal metadata that's irrelevent to
peers.

### DIFF
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -1,11 +1,11 @@
 import { EventEmitter } from 'events';
 import { ReputationEvent } from '../constants/enums';
-import { NodeFactory, NodeInstance, ReputationEventInstance } from '../db/types';
+import { NodeFactory, NodeInstance } from '../db/types';
 import addressUtils from '../utils/addressUtils';
 import { pubKeyToAlias } from '../utils/aliasUtils';
+import errors from './errors';
 import P2PRepository from './P2PRepository';
 import { Address } from './types';
-import errors from './errors';
 
 export const reputationEventWeight = {
   [ReputationEvent.ManualBan]: Number.NEGATIVE_INFINITY,
@@ -25,8 +25,8 @@ export const reputationEventWeight = {
 // TODO: remove reputation events after certain amount of time
 
 interface NodeList {
-  on(event: 'node.ban', listener: (nodePubKey: string, events: ReputationEventInstance[]) => void): this;
-  emit(event: 'node.ban', nodePubKey: string, events: ReputationEventInstance[]): boolean;
+  on(event: 'node.ban', listener: (nodePubKey: string, events: ReputationEvent[]) => void): this;
+  emit(event: 'node.ban', nodePubKey: string, events: ReputationEvent[]): boolean;
 }
 
 /** Represents a list of nodes for managing network peers activity */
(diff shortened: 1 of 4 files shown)


lib/p2p/NodeList.ts
```diff
@@ -170,23 +170,26 @@ class NodeList extends EventEmitter {
...
 
-      if (node.reputationScore < NodeList.BAN_THRESHOLD) {
+      if (node.reputationScore < NodeList.BAN_THRESHOLD && !node.banned) {
         promises.push(this.setBanStatus(node, true));
 
-        const events = await this.repository.getReputationEvents(node);
-        this.emit('node.ban', nodePubKey, events);
-      } else if (node.banned) {
+        const negativeReputationEvents = (await this.repository.getReputationEvents(node))
+          .filter(e => reputationEventWeight[e.event] < 0).slice(0, 9)
```
erkarl, line 181: Just wondering why do we need to send up to last 10 reputation events to the peer rather than just the final one that triggered the ban?

lib/p2p/NodeList.ts
```diff
@@ -170,23 +170,26 @@ class NodeList extends EventEmitter {
...
       this.updateReputationScore(node, event);
 
-      if (node.reputationScore < NodeList.BAN_THRESHOLD) {
+      if (node.reputationScore < NodeList.BAN_THRESHOLD && !node.banned) {
         promises.push(this.setBanStatus(node, true));
 
-        const events = await this.repository.getReputationEvents(node);
-        this.emit('node.ban', nodePubKey, events);
-      } else if (node.banned) {
+        const negativeReputationEvents = (await this.repository.getReputationEvents(node))
```
erkarl, line 180: Maybe we can extract `getNegativeReputationEvents(events: ReputationEvents[], event: ReputationEvent) => ReputationEvents[]` from this function. This would be a pure function and easily unit testable that it:
- returns up to 10 reputation events
- outputs only negative events
- prepends current to the filtered list

lib/p2p/NodeList.ts
```diff
@@ -170,23 +170,26 @@ class NodeList extends EventEmitter {
     const node = this.nodes.get(nodePubKey);
```
erkarl, line 170: I noticed `class NodeList` doesn't have any tests. Perhaps this is a good spot to add an integration test to `public addReputationEvents()`?

lib/p2p/NodeList.ts
sangaman: That's just the way it was set up originally, it might be helpful in the case of repeated events triggering the failure so the peer has a better idea of why it was banned. I am fine with only sending the most recent event, however.

lib/p2p/NodeList.ts
sangaman: I'll handle this in a separate PR.

lib/p2p/NodeList.ts
erkarl: That makes sense. Was just curious. I'm OK with leaving it with up to 10 last negative events.

sangaman:
@erkarl I pushed a commit with the new test suite and test cases for getting negative reputation events.

Issue #1714, opened 2020-07-06
Simulation test failure


erkarl:
```
> xud@1.0.0-beta.4 test:sim:build /home/ar/xud
> (npm run test:sim:compile:xud && npm run test:sim:compile:custom-xud && cd test/simulation && ./docker-build.sh)


> xud@1.0.0-beta.4 test:sim:compile:xud /home/ar/xud
> (npm run compile && mkdir -p test/simulation/docker-xud/xud && cp -a dist test/simulation/docker-xud/xud && cp -a bin test/simulation/docker-xud/xud && cp package.json test/simulation/docker-xud/xud)


> xud@1.0.0-beta.4 precompile /home/ar/xud
> node parseGitCommit.js


> xud@1.0.0-beta.4 compile /home/ar/xud
> npm run precompile && tsc && cross-os postcompile


> xud@1.0.0-beta.4 precompile /home/ar/xud
> node parseGitCommit.js


> xud@1.0.0-beta.4 test:sim:compile:custom-xud /home/ar/xud
> (npm run test:sim:custom-xud:apply && npm run compile && mkdir -p test/simulation/docker-xud/custom-xud/dist && npm run test:sim:custom-xud:reverse && cp -a dist test/simulation/docker-xud/custom-xud && cp -a bin test/simulation/docker-xud/custom-xud && cp package.json test/simulation/docker-xud/custom-xud)


> xud@1.0.0-beta.4 test:sim:custom-xud:apply /home/ar/xud
> git apply --reject test/simulation/custom-xud.patch

Checking patch lib/Xud.ts...
Checking patch lib/swaps/SwapRecovery.ts...
Checking patch lib/swaps/Swaps.ts...
Applied patch lib/Xud.ts cleanly.
Applied patch lib/swaps/SwapRecovery.ts cleanly.
Applied patch lib/swaps/Swaps.ts cleanly.

> xud@1.0.0-beta.4 precompile /home/ar/xud
> node parseGitCommit.js


> xud@1.0.0-beta.4 compile /home/ar/xud
> npm run precompile && tsc && cross-os postcompile


> xud@1.0.0-beta.4 precompile /home/ar/xud
> node parseGitCommit.js


> xud@1.0.0-beta.4 test:sim:custom-xud:reverse /home/ar/xud
> git apply -R test/simulation/custom-xud.patch

~/xud/test/simulation/temp ~/xud/test/simulation
fatal: destination path 'indra' already exists and is not an empty directory.
[...]

PR #1720, opened 2020-07-08
fix(simtest): nil pointer deref bug

LePremierHomme:
Closing https://github.com/ExchangeUnion/xud/issues/1714.

The problem was with having a nil pointer dereference in a scenario where xud shutdown shortly after launch. This was fixed, but not the shutdown itself, which happens when trying to use an already-used port. I didn't manage to figure out what's causing this known instability, but I might try to use a different free-port generator technique if this happens too frequently. 

Also:
* Added log printing in case of failure, since we lost that after migrating to GitHub Actions. 
* Extended test timeout to 30m, so it will be longer than the test context timeout (for better error messages)

I managed to reproduce the problem after a few attempts, and this is how it looks now:

test
```
2020/07/08 06:12:46 xud: launching network...
2020/07/08 06:13:08 cannot start xud network: timeout waiting for xud to be ready
exit status 1
FAIL	github.com/ExchangeUnion/xud-simulation	164.413s
```

xud
```
08/07/2020 06:12:52.167 [GLOBAL] error: Unexpected error during initialization, shutting down...: Error: listen EADDRINUSE: address already in use 127.0.0.1:34365
    at Server.setupListenHandle [as _listen2] (net.js:1313:16)
    at listenInCluster (net.js:1361:12)
    at GetAddrInfoReqWrap.doListen [as callback] (net.js:1498:7)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:68:8)
```

### DIFF
--- a/test/simulation/docker-run.sh
+++ b/test/simulation/docker-run.sh
@@ -9,11 +9,16 @@ popd
 
 export DOCKER_CLIENT_TIMEOUT=120
 export COMPOSE_HTTP_TIMEOUT=120
-docker-compose run -v $PWD/temp/logs:/app/temp/logs test go test -v -timeout 20m -run=$@
-res=$?
+docker-compose run -v $PWD/temp/logs:/app/temp/logs test go test -v -timeout 30m -run=$@
+retVal=$?
 
 pushd temp/indra
 make reset
 popd
 
-exit $res
+if [[ retVal != 0 ]]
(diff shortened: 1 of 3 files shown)


LePremierHomme:
I think it should be closed, but i'll open a new one regarding the used ports failure.

Issue #1721, opened 2020-07-08
Simulation test failure


LePremierHomme:
### Background

CI simulation tests sometimes fail due to different nodes (lnd, xud, etc.) using an already-used port. We use a custom free-port generator, which apparently isn't stable enough. This error results in the node shutdown and the failure of the test. 

### Example

test
```
2020/07/08 06:12:46 xud: launching network...
2020/07/08 06:13:08 cannot start xud network: timeout waiting for xud to be ready
exit status 1
FAIL	github.com/ExchangeUnion/xud-simulation	164.413s
```

xud
```
08/07/2020 06:12:52.167 [GLOBAL] error: Unexpected error during initialization, shutting down...: Error: listen EADDRINUSE: address already in use 127.0.0.1:34365
    at Server.setupListenHandle [as _listen2] (net.js:1313:16)
    at listenInCluster (net.js:1361:12)
    at GetAddrInfoReqWrap.doListen [as callback] (net.js:1498:7)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:68:8)
```

kilrau:
Is this still relevant? @LePremierHomme

LePremierHomme:
Closed by https://github.com/ExchangeUnion/xud/pull/1803

PR #1723, opened 2020-07-10
feat(rpc): subscribeSwapsAccepted

### LINKED ISSUE #1662: SubscribeSwapsAccepted
A streaming call that would notify every time a swap is accepted with some basic info like order id, pair, amount...

sangaman:
This introduces a new streaming call that notifies listeners when a swap is accepted. This allows market makers to anticipate when trades will be completed earlier, as there may be a delay of 15-30 seconds between when a swap is accepted and when it is completed.

Closes #1662.

I still want to do some more manual testing before this is merged but it should be ready for review. It can be tested via cli with the `streamorders` call.


erkarl:
We already have an issue for the simtests port conflict: https://github.com/ExchangeUnion/xud/issues/1721

raladev:
- [ ] no SwapDealAccepted msg in streamorders call for taker (works for maker).
[Screenshot from 2020-07-13 20-36-53](https://user-images.githubusercontent.com/29906866/87335557-2f980d80-c549-11ea-817f-a3c000f5c2da.png)

- [ ] streamorders stream crashed after some time (1 hour), but I am not sure that this problem is connected with  SwapDealAccepted msg
[Screenshot from 2020-07-13 20-28-00](https://user-images.githubusercontent.com/29906866/87335597-3f175680-c549-11ea-9103-e2b5e9c1efd9.png)

erkarl:
> no SwapDealAccepted msg in streamorders call for taker (works for maker).

I believe we can only accept a swap as a maker. The same thing for the taker would be sending the swap request (trying to take the order).

raladev:
> > no SwapDealAccepted msg in streamorders call for taker (works for maker).
> 
> I believe we can only accept a swap as a maker. The same thing for the taker would be sending the swap request (trying to take the order).

https://opendex.network/read/04-swap-protocol

If i understood correctly, taker should receive SwapAccepted packet and this action should be displayed in streamorders call.

Sentence form misbehaviour table:
"Taker doesn't start the swap after receiving the SwapAccepted message"

sangaman:
Sorry that I wasn't clear, but at least the way it's designed now it only triggers the swap accepted event on the maker side upon accepting a swap request.

An event for when a swap that we requested (as taker) is accepted might be a nice feature as part of the `PlaceOrder` flow, right now we have events for when we have an order match, when a swap fails, and when a swap is rejected, but not when a swap request is accepted.

raladev:
> Sorry that I wasn't clear, but at least the way it's designed now it only triggers the swap accepted event on the maker side upon accepting a swap request.
> 
> An event for when a swap that we requested (as taker) is accepted might be a nice feature as part of the `PlaceOrder` flow, right now we have events for when we have an order match, when a swap fails, and when a swap is rejected, but not when a swap request is accepted.

Gotcha, then all is fine and i will create separate issue for streamorders call crash.

Issue #1727, opened 2020-07-13
(streamorders stream)/xud crashed after 1h of active arby using


raladev:
### Actual behaviour

![Screenshot from 2020-07-13 20-28-00](https://user-images.githubusercontent.com/29906866/87335597-3f175680-c549-11ea-9103-e2b5e9c1efd9.png)

### Expected behaviour

No crashes

kilrau:
Any idea for this one? Was this an existing issue or caused be the `SubscribeSwapAccepted` PR we just merged (which would be bad)? @sangaman

raladev:
> Any idea for this one? Was this an existing issue or caused be the `SubscribeSwapAccepted` PR we just merged (which would be bad)? @sangaman

It was only taker's node, so if I understood correctly there is no SwapAccepted logic for this side.

raladev:
got this crash today but with xud instead of streamorders call
Bump to P1
[xud_fall.log](https://github.com/ExchangeUnion/xud/files/4926229/xud_fall.log)

erkarl:
The error looks pretty deep into NodeJS core. I wonder if others are able to reproduce it...

sangaman:
This is indeed something very low level and I'm guessing system specific that isn't playing nice with grpc. I suspect it's related to node v14, which was released fairly recently. @raladev Can you confirm the node version you're using? 

This is also a good reminder for me to revisit #1605, as the grpc dependency we're currently using is slated to be deprecated after April 2021.

raladev:
> This is also a good reminder for me to revisit #1605, as the grpc dependency we're currently using is slated to be deprecated after April 2021.

```
native@ubuntu:~$ node
Welcome to Node.js v14.0.0.
Type ".help" for more information.
```

PR #1732, opened 2020-07-14
feat(orderbook): don't match same orders twice

### LINKED ISSUE #1595: Matching routine repedeately matching same order
### Precondition
You got No Route Error after installation or you can reliably reproduce it

### Steps to reproduce

1. place buy 1 LTC/BTC 0.001 order as a maker
2. sell 0.1 LTC/BTC mkt 1 as a taker

### Actual behaviour

10 No Route swap attempts with order of maker after  
[Screen](https://user-images.githubusercontent.com/29906866/83516041-095d6780-a4df-11ea-8cf5-dd42d9960b84.png)

After 10 attempts order disappeared from orderbook.

### Expected behaviour

only 1 No Route error per order, after that order should be removed from orderbook and new order should be used for match.

[logs.zip](https://github.com/ExchangeUnion/xud/files/4716883/logs.zip)

sangaman:
This temporarily removes peer orders from the order book when they fail swaps during the matching routine for a locally placed order. Placed orders that fail any swaps will retry matching for the quantity that failed, and previously this meant they may match again with the same order resulting in a swap that will likely fail yet again.

This ensures that the two orders won't be matched again and that any other orders in the order book with a matching price would be tried instead. Once matching is complete, the order is added back to the
order book so it may be swapped again in the future by which time the  reason for the swap failure may be resolved.

This replaces the `debug.testing` behavior for order matching.

Closes #1595.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -13,7 +13,7 @@ import Swaps from '../swaps/Swaps';
 import { SwapDeal, SwapFailure, SwapSuccess } from '../swaps/types';
 import { pubKeyToAlias } from '../utils/aliasUtils';
 import { derivePairId, ms, setTimeoutPromise } from '../utils/utils';
-import errors from './errors';
+import errors, { errorCodes } from './errors';
 import OrderBookRepository from './OrderBookRepository';
 import TradingPair from './TradingPair';
 import {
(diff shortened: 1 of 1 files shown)


raladev:
- [ ] it works with static env (when i place orders and test only with them), but does not with dynamic env (with enabled arby).
I just got 
`5 NOT_FOUND: order with id 3323fda0-c6ad-11ea-939b-efbe83bcf33b could not be found`
without retry attempts.
[Screenshot from 2020-07-15 17-03-27](https://user-images.githubusercontent.com/29906866/87563632-21bcc680-c6c8-11ea-8560-572c8c8e3ffc.png)
[Screenshot from 2020-07-15 17-14-38](https://user-images.githubusercontent.com/29906866/87563628-208b9980-c6c8-11ea-880f-ab1fe152a42f.png)
[Screenshot from 2020-07-15 17-15-14](https://user-images.githubusercontent.com/29906866/87563620-1f5a6c80-c6c8-11ea-95dc-f7f53423fa4d.png)
[Screenshot from 2020-07-15 18-21-22](https://user-images.githubusercontent.com/29906866/87563617-1e293f80-c6c8-11ea-93b4-6d3ef8f9440a.png)
- [ ] xud was crushed after the meeting with same error from this task https://github.com/ExchangeUnion/xud/issues/1727 (but in this task only streamorder call was crashed, xud was fine). Im not sure that it is connected with this pr. its just a mention.
[xud_fall.log](https://github.com/ExchangeUnion/xud/files/4926212/xud_fall.log)
- There is can be a problem with dynamic env. After SwapTimeoutErr with one order i will get it with new arby order (because oldest was removed by arby). So, as a result i will be banned. Seems for some errors we should remove all order of bad peer.

sangaman:
I believe I'm now handling the issue with orders not being found - sometimes the order that failed a swap will no longer be there by the time we go to remove it. Can you give it another test run @raladev ?

PR #1736, opened 2020-07-17
fix(connext): async getbalance & error handling

### LINKED ISSUE #1572: Connext client blocks getbalance request when container is down or connext node returns error
[Screenshot from 2020-05-22 14-51-05](https://user-images.githubusercontent.com/29906866/82665286-24f18400-9c3c-11ea-92a6-69ac8f08a98b.png)
[Screenshot from 2020-05-22 14-50-50](https://user-images.githubusercontent.com/29906866/82665292-258a1a80-9c3c-11ea-8ad0-937a386698f5.png)

### Actual behaviour

1. getbalance command returns connext error if connext got 500 error
2. getbalance command returns 'UNKNOWN: getaddrinfo ENOTFOUND connext' if connext container was stopped

### Expected behaviour

1. getbalance command returns all balances except of connext balance
2. getbalance command returns all balances except of connext balance

sangaman:
This improves the parallelization and error handling of the `GetBalance` call particularly as it pertains to Connext. Behind the scenes, Connext has a single call that returns both on chain and off chain balance, unlike lnd where the two calls are separate. xud was calling this twice, which puts unnecessary load on Connext.

Instead, Connext now tracks when there is already an ongoing balance request that it is waiting for a response for. Any `getBalance` calls that happen while there is this wait will listen for the same response rather than making separate calls.

Also, the error handling is changed for the `GetBalance` call such that if we get errors for only some of the currencies for which we are checking balances, we still return the balances for currencies which did not give us any errors. This means that a connext error will not cause the entire `GetBalance` call to fail when there are active lnd currencies.

Fixes #1572.

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -106,6 +106,8 @@ class ConnextClient extends SwapClient {
   private webhookhost: string;
   private unitConverter: UnitConverter;
   private seed: string | undefined;
+  /** A map of currencies to promises representing balance requests. */
+  private getBalancePromises = new Map<string, Promise<ConnextBalanceResponse>>();
   private _totalOutboundAmount = new Map<string, number>();
 
   /**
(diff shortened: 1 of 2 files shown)


lib/connextclient/ConnextClient.ts
```diff
@@ -556,11 +558,22 @@ class ConnextClient extends SwapClient {
...
+  private getBalance = (currency: string): Promise<ConnextBalanceResponse> => {
+    // check if we already have a balance request that we are waiting a response for
+    // it's not helpful to have simultaneous requests for the current balance, as they
+    // should return the same info.
+    let getBalancePromise = this.getBalancePromises.get(currency);
+    if (!getBalancePromise) {
+      // if not make a new balance request and store the promise that's waiting for a response
+      const tokenAddress = this.getTokenAddress(currency);
+      getBalancePromise = this.sendRequest(`/balance/${tokenAddress}`, 'GET').then((res) => {
+        this.getBalancePromises.delete(currency); // clear the stored promise
```
erkarl, line 570: At the moment the pending promise is only cleaned up in case of a happy flow. This cleanup should probably go into `finally` block.

PR #1739, opened 2020-07-20
test(sim): PreimageResolved phase instability tests

LePremierHomme:
Closing https://github.com/ExchangeUnion/xud/issues/1712.

The previously existing `makerCrashedAfterSend` test wasn’t working well, as it was simulating a crash after swap already completed, due to an issue in the test code.

Instead, 4 tests were added:

* maker crash after send payment, before `PreimageResolved` phase; lnd
* maker crash after send payment, before `PreimageResolved` phase; connext
* maker crash after send payment, after `PreimageResolved` phase; lnd
* maker crash after send payment, after `PreimageResolved` phase; connext

As before, code structure is still messy, waiting for the upcoming refactoring.

PR #1745, opened 2020-07-23
test: remove temp/indra folder on clean

sangaman:
The `temp/indra` folder for simulation tests was not being removed when running `npm run:sim:clean`. A stale indra folder could cause tests to fail and would need to be manually deleted. This PR deletes it automatically with the clean script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "test:sim:custom-xud:apply": "git apply --reject test/simulation/custom-xud.patch",
     "test:sim:custom-xud:reverse": "git apply -R test/simulation/custom-xud.patch",
     "test:sim:clean": "(cd test/simulation && ./docker-clean.sh)",
+    "test:sim:clean:indra": "rm -rf test/simulation/temp/indra",
     "test:sim:clean:xud": "(cd test/simulation && ./docker-clean.sh xud)",
     "test:sim:clean:lnd": "(cd test/simulation && ./docker-clean.sh lnd)",
     "test:sim:clean:connext": "(cd test/simulation && ./docker-clean.sh connext)",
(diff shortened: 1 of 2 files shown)


sangaman:
> @sangaman thanks for checking. I originally thought to avoid doing that on regular cleanup, because re-downloading/installing indra is very resource-intensive, but looks like the build script wasn't compatible with that, so i'll look into it.

_Originally posted by @LePremierHomme in https://github.com/ExchangeUnion/xud/pull/1739#issuecomment-662986031_

I can make a separate `test:sim:clean:indra` script that cleans the indra folder so the default clean script doesn't touch it. It was definitely resource intensive for me to run the tests after deleting this folder.

erkarl:
> > @sangaman thanks for checking. I originally thought to avoid doing that on regular cleanup, because re-downloading/installing indra is very resource-intensive, but looks like the build script wasn't compatible with that, so i'll look into it.
> 
> _Originally posted by @LePremierHomme in [#1739 (comment)](https://github.com/ExchangeUnion/xud/pull/1739#issuecomment-662986031)_
> 
> I can make a separate `test:sim:clean:indra` script that cleans the indra folder so the default clean script doesn't touch it. It was definitely resource intensive for me to run the tests after deleting this folder.

I'd be OK with a separate clean script. The only concern I have is that we'll then need to run 2 commands for a completely "clean" environment. 

```
test:sim:clean
test:sim:clean:indra
test:sim:build
```

LePremierHomme:
I'm also in favor of having to explicitly clean indra (via separate script). Otherwise the wait time after every general clean will be too long, while in most of them upgrading indra isn't necessary.

sangaman:
@erkarl @LePremierHomme I updated the PR with a separate script as discussed.

PR #1759, opened 2020-07-27
feat(config): rename debug.testing to strict

### LINKED ISSUE #1757: Make `debug.testing` default (& rm the option) add new `production.strict` option
As [discussed on discord](https://discordapp.com/channels/547402601885466658/709337789304668160/735964844712525897), I'd like to turn things around: instead of being very strict and gradually adding more things to `debug.testing` because they hinder real-world usage, I'd open it up to the best degree possible and slowly start getting stricter once things mature.

I propose to remove the current  `debug.testing` option and make the behavior behind this flag the the default behavior.

Some of the features this is disabling will become useful though in future, that's why I propose to add a new `production.strict` option that enables the following:
- peers mutually banning each other because of severe L2 payment errors like "NO ROUTE FOUND".
- disabling currencies for a certain peer because of swap failure
- removing all orders of a peer because of swap failure

sangaman:
This replaces the `debug.testing` option with a `strict` option and makes the default behavior non-strict, equivalent to running xud with `debug.testing` before.

I went with `strict` as opposed to `production.strict` because it's less to type out that way and because I can't envision any other options that would go under a "production" category of config options.

It also allows severe reputation events to be registered even when not running in strict mode. Previously only manual reputation events were allowed when not in strict mode.

Closes #1757.

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -928,8 +928,15 @@ class Pool extends EventEmitter {
     peer.on('reputation', async (event) => {
       this.logger.debug(`Peer (${peer.label}): reputation event: ${ReputationEvent[event]}`);
       if (peer.nodePubKey) {
-        // we only add manual reputation events when not in strict mode to prevent unintentional bans
-        if (this.strict || event === ReputationEvent.ManualBan || event === ReputationEvent.ManualUnban) {
+        // when in strict mode, we add all reputation events
+        // otherwise, we only add manual or severe reputation events to prevent unintentional bans
+        if (this.strict
+          || event === ReputationEvent.ManualBan
+          || event === ReputationEvent.ManualUnban
+          || event === ReputationEvent.SwapAbuse
+          || event === ReputationEvent.SwapMisbehavior
+          || event === ReputationEvent.WireProtocolErr
+          || event === ReputationEvent.InvalidAuth) {
           await this.addReputationEvent(peer.nodePubKey, event);
         }
       }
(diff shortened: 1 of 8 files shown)


lib/p2p/Pool.ts
```diff
@@ -928,11 +928,16 @@ class Pool extends EventEmitter {
     peer.on('reputation', async (event) => {
       this.logger.debug(`Peer (${peer.label}): reputation event: ${ReputationEvent[event]}`);
       if (peer.nodePubKey) {
-        if (this.testing && event !== ReputationEvent.ManualBan && event !== ReputationEvent.ManualUnban) {
-          // we don't add non-manual reputation events when in debug/testing mode to prevent unintentional bans
-          return;
+        // we only add manual reputation events when not in strict mode to prevent unintentional bans
+        if (this.strict
```
erkarl, line 932: Should this be `!this.strict` so it's in line with the comment above?

sangaman: The `if` statement is correct, but the comment isn't very clear, so I'll reword the comment. We want to add reputation events when in strict mode OR if the reputation event is manual or severe (which we add in both strict and non strict modes).

kilrau:
Could you quickly outline which features will be behind `strict` now? It's not really obvious to me from the code changes.

sangaman:
> Could you quickly outline which features will be behind `strict` now? It's not really obvious to me from the code changes.

Yep sorry I had meant to summarize:

- No negative reputation events besides manual banning, swap abuse (free option problem), or p2p protocol violations. This means no banning peers for regular swap failures, disconnections, packet delays, etc...
- Activates all trading pairs for peers, even if we can't find a route to them at the time we connect.
- Never deactivates trading pairs due to swap failures.

PR #1760, opened 2020-07-27
feat(connext): requestColleteral for /deposit

### LINKED ISSUE #1756: Add `requestCollateral` to `openchannel eth` flow
Problem: when swapping for the first time, indra will need to collateralize the channel which often results in a timeout (on-chain confirmation taking too long), but at least in a very slow swap. Bad first time UX.

Solution: In addition to [increasing swap timeout to 90s](https://github.com/ExchangeUnion/xud/issues/1752) to avoid most swap timeouts when a live collateralization is involved, we want to request collateral from the indra node right away in the `openchannel` flow via `client.requestCollateral`. The indra node decides based on collateralization profile if and how much to collateralize. E.g. we are targeting to collateralize the same as the deposit amount to have a 50:50 balanced channel.

cc @ArjunBhuptani please point us to how such a collateralization profile looks like.

sangaman:
This automatically requests that Connext collateralize our channel each time we deposit funds to the channel.

Closes #1756.

Depends on: https://github.com/connext/rest-api-client/pull/61

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -590,6 +590,7 @@ class ConnextClient extends SwapClient {
     await this.sendRequest('/deposit', 'POST', {
       assetId,
       amount: BigInt(units).toString(),
+      requestCollateral: true,
     });
   }
 


sangaman:
This uses the originally proposed approach to requesting collateral for new channels, but we went a different direction and this PR and `requestCollateral` flag no longer applies.

PR #1765, opened 2020-07-29
feat(lnd): handle delayed macaroon creation

sangaman:
This improves the initialization flow for LndClient when creating a new wallet in lnd. Sometimes `admin.macaroon` is not created immediately after calling `InitWallet` on lnd. Previously, this could lead to us being stuck in a loop on `awaitWalletInit` immediately after creating or restoring a wallet in lnd. If there are no macaroons by the first time we attempt to verify the connection, we treated it as if lnd had never received an `InitWallet` command and wait for that command to happen, however that call never comes since it had already happened.

Instead, this adds new logic to detect the above scenario and skip ahead to simply wait for `admin.macaroon` to be written to disk and made available for lnd calls by xud.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -164,7 +164,7 @@ class LndClient extends SwapClient {
       try {
         await this.loadMacaroon();
       } catch (err) {
-        this.logger.warn(`could not load macaroon at ${macaroonpath}`);
+        this.logger.info(`could not load macaroon at ${macaroonpath}, this is normal before creating a new wallet`);
       }
     } else {
       this.logger.info('macaroons are disabled');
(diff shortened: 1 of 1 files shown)


kilrau:
This one could potentially fix the xud restart we are currently enforcing in the xud-docker create flow. Let's test without restart with this branch @raladev

lib/lndclient/LndClient.ts
```diff
@@ -384,16 +384,27 @@ class LndClient extends SwapClient {
    * to verify the connection to lnd.
    */
   private awaitWalletInit = async () => {
-    // we are waiting for lnd to be initialized by xud and for the lnd macaroons to be created
-    this.logger.info('waiting for wallet to be initialized...');
-
     /**
-     * A promise that resolves to `true` when the lnd wallet is created via an InitWallet call,
-     * resolves to `false` if we close the client before the lnd wallet is created.
+     * Whether he lnd wallet has been created via an InitWallet call,
```
erkarl, line 388:
```suggestion
     * Whether the lnd wallet has been created via an InitWallet call,
```

sangaman:
I'm thinking here it might make sense to explicitly wait for admin.macaroon to be created after xud performs the InitWallet call to lnd. That way we can wait to respond to the xud `InitNode` call until all lnds have finished creating admin.macaroon and are ready to receive calls. I'll add that in a separate commit.

sangaman:
Since this appears to fix our restore issues and is working well in manual tests, I'm going to merge this PR as is (with karl's typo fix) and then open a separate PR for my proposal above.

kilrau:
Sounds good. Go ahead, merging is not blocked anymore.

PR #1768, opened 2020-07-29
fix(lnd): totalOutboundAmount

### LINKED ISSUE #1636: LND balancecheck on placeorder to not include inactive/closed channelbalance
This issue is about fixing the current lnd balancecheck on placeorder which allows me to place order with inactive/closed channels channels.

As @sangaman mentioned in https://github.com/ExchangeUnion/xud/issues/1609#issuecomment-642434598 :
This is a little more involved, we'd need to stop relying on the `ChannelBalance` lnd call since it always includes inactive channel balance. Would we want to switch to the logic for determining trading limits as the new lnd heartbeat call? It will continuously update our trading capacities/limits more precisely than `ChannelBalance` can.

Which is probably what we would need to do. Low prio, no milestone/project for now.

ghost:
Fixes https://github.com/ExchangeUnion/xud/issues/1636

Updating the `totalOutboundAmount` was missing from the original PR. In this context I think it makes sense to set it to `maximumOutboundAmount`. I originally thought about removing `totalOutboundAmount` field from the swap client, but connext swap client relies on it for multi currency support.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -688,10 +688,12 @@ class LndClient extends SwapClient {
     let maxInbound = 0;
     let balance = 0;
     let inactiveBalance = 0;
+    let totalOutboundAmount = 0;
     channels.toObject().channelsList.forEach((channel) => {
       if (channel.active) {
         balance += channel.localBalance;
         const outbound = channel.localBalance - channel.localChanReserveSat;
+        totalOutboundAmount += outbound;
         if (maxOutbound < outbound) {
           maxOutbound = outbound;
         }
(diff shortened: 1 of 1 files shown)


sangaman (dismissed):
Thanks. I'm only wondering if it makes sense to set this to the sum of all outbound amounts, reflecting the possibility of AMP payments and/or orders that can be filled piece by piece.

kilrau:
I think that makes sense. Sum of all outbound amounts=max sell and sum of all inbound amounts=max buy.

erkarl:
> I think that makes sense. Sum of all outbound amounts=max sell and sum of all inbound amounts=max buy.

Oh yeah, since we have AMP now we can possibly fill those orders.

erkarl:
> Thanks. I'm only wondering if it makes sense to set this to the sum of all outbound amounts, reflecting the possibility of AMP payments and/or orders that can be filled piece by piece.

Updated the PR accordingly.

raladev (approved):

sangaman (approved):

PR #1769, opened 2020-07-29
feat(lnd): wait for macaroon after InitWallet

sangaman:
This waits for the admin.macaroon to be created after calling InitWallet.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -10,8 +10,8 @@ import { LightningClient, WalletUnlockerClient } from '../proto/lndrpc_grpc_pb';
 import * as lndrpc from '../proto/lndrpc_pb';
 import swapErrors from '../swaps/errors';
 import SwapClient, { ChannelBalance, ClientStatus, PaymentState, SwapClientInfo, TradingLimits } from '../swaps/SwapClient';
-import { SwapDeal, CloseChannelParams, OpenChannelParams } from '../swaps/types';
-import { base64ToHex, hexToUint8Array } from '../utils/utils';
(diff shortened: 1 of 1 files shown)


kilrau:
Follow-up to #1765 ? A better description why this is needed would be helpful I think.

sangaman:
> Follow-up to #1765 ? A better description why this is needed would be helpful I think.

Yes, I had meant to reference this comment: https://github.com/ExchangeUnion/xud/pull/1765#issuecomment-665642564

> I'm thinking here it might make sense to explicitly wait for admin.macaroon to be created after xud performs the InitWallet call to lnd. That way we can wait to respond to the xud InitNode call until all lnds have finished creating admin.macaroon and are ready to receive calls. I'll add that in a separate commit.

This makes loading/waiting for the admin.macaroon file part of the `InitNode` flow, which seems sensible to me.

raladev:
it does not work well. I got timeout err reliably when create new wallet for both lnds.
Screenshot from 2020-07-30 18-26-40
Screenshot from 2020-07-30 18-26-16

kilrau:
Do we still want this one since there is a conflict with how xud-docker handles the `create` flow? @sangaman

erkarl:
Do we still want this?

sangaman:
I think it probably makes sense to close - as it stands it doesn't seem there isn't a need for it and clearly there's some issue with my implementation as-is and it's not clear to me what it is. The general idea still makes sense to me, though (that we should wait for the macaroon to be created by lnd before we move along with making calls to lnd). In case we face any issues regarding lnd macaroon creation I figure we can revisit this, but it's probably not worth the attention right now.

PR #1772, opened 2020-07-30
fix(swaps): don't log non-existent route

sangaman:
This fixes a log statement to not attempt to log a route when none exists, which would cause `undefined` to be logged.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -870,7 +870,7 @@ class LndClient extends SwapClient {
     if (route) {
       this.logger.debug(`found a route to ${destination} for ${units} units with finalCltvDelta ${finalLock}: ${route}`);
     } else {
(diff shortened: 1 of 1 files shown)


erkarl:
Tests are failing with:
```
   1) P2P Pool Tests
       should update a node on new handshake:

      AssertionError: expected '123.123.123.123' to equal '86.75.30.9'
      + expected - actual

      -123.123.123.123
      +86.75.30.9
      
      at /home/runner/work/xud/xud/test/integration/Pool.spec.ts:103:49
      at fulfilled (test/integration/Pool.spec.ts:24:58)

  2) P2P Pool Tests
       reconnect logic
         "before each" hook for "should not reconnect upon shutdown inbound":
     node 022a49046f252821dfcb3a3f7c8e203af97fb2b8223d8b9204380ccdce6910755d at 321.321.321.321:1337 already connected
```

Think this integration test is a bit flaky - have seen this error in other PRs.

PR #1783, opened 2020-08-04
feat(connext): request collateral after deposit

### LINKED ISSUE #1756: Add `requestCollateral` to `openchannel eth` flow

sangaman:
This calls the new `/request-collateral` endpoint after depositing funds to a connext channel. We do not await a response from `/request-collateral` but rather allow it to complete asynchronously after the `/deposit` call finishes.

Closes #1756.

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -591,6 +591,13 @@ class ConnextClient extends SwapClient {
       assetId,
       amount: BigInt(units).toString(),
     });
+    this.sendRequest('/request-collateral', 'POST', {
+      assetId,
+    }).then(() => {
+      this.logger.verbose(`collateralized channel for ${currency}`);
+    }).catch((err) => {
+      this.logger.error(`failed requesting collateral for ${currency}`, err);
+    });
   }
 
   public closeChannel = async ({ units, currency, destination }: CloseChannelParams): Promise<void> => {


lib/connextclient/ConnextClient.ts
```diff
@@ -591,6 +591,11 @@ class ConnextClient extends SwapClient {
       assetId,
       amount: BigInt(units).toString(),
     });
+    this.sendRequest('/request-collateral', 'POST', {
```
erkarl, line 594: Maybe also add a info logging here to notify when the `/request-collateral` request completes?

sangaman: That would be quite helpful indeed.

raladev:
results In short - I cant find reliable legitimate way to get cancel from counterparty.

- tried connext shutdown - fund was returned back from the node to sender because of 60 timeout
- tried xud shutdown + connext shutodwn - it does not work because i lost paymentId ( got cant find a incoming tranfer msg) and funds has not been returned to sender
- tried lndbtc shutdown during swap and upping it back -  got cant find a incoming tranfer msg and funds has not been returned to sender

IMO cancel should work in third case, so:

Steps:
1. Place 8 sell BTC/ETH orders as a maker
2. Send 1 buy order as a maker that will be matched with all 8 orders ( [Screenshot from 2020-08-05 22-33-15](https://user-images.githubusercontent.com/29906866/89463132-24668500-d777-11ea-876a-7ed81b43e692.png)
)
3. Wait first successful match in xud log
4. docker rm -f simnet_lndbtc_1   (after this channel of taker that was connected to simnet node was closed with no reason)
5. Check taker ETH balance -> Full amount of 8 orders was blocked  ( [Screenshot from 2020-08-05 22-33-15](https://user-images.githubusercontent.com/29906866/89463132-24668500-d777-11ea-876a-7ed81b43e692.png)(0.0008*8+0.0007*7+0.0006*6+0.0005*5+0.0004*4+0.0003*3+0.0002*2+0.0001*1=0.0204 ETH)
5. Wait first recovery attempt from maker -> all recovery attempts were failed because of disabled BTC node
6. Return lndbtc node back
7. Wait second recovery attempt

Actual result:

1. 1 successful recovery for released ETH payment (because 2 swaps were processed before disablin lndbtc, but only one was processed by xud. So, recovery solved this problem). As a result we have 2 successful swaps of 8 - for 0.0001(0.0001*1) ETH and for 0.0025(0.0005*5).

2. 
```
swap for a7bbc94ac3bac589488dba7ceeed3cada3ba763a613a100776eee65624d0af13 still has pending payments and will be monitored
swap for 2b48974858ec68b34d174c1580986debf5e4d654948bfd84c9b661cf7d42a73d still has pending payments and will be monitored
```
a7bbc94ac3bac589488dba7ceeed3cada3ba763a613a100776eee65624d0af13 swap is for 0.0003*3 ETH
2b48974858ec68b34d174c1580986debf5e4d654948bfd84c9b661cf7d42a73d swap is for 0.0006*6 ETH

3.
```
could not find expected incoming transfer with hash eaaef273e2bdd497962829e3e36bf84a169bf07cfa485fbd9acf3ac3943dd426
could not find expected incoming transfer with hash 1dc7108f3461f4844bfccb4cc9377447b0a437cf10d7ef030122bf15442bae51
could not find expected incoming transfer with hash 63b53468480f0eed9ca37789fc931d0ec4eb4a8e7eff3760371f59bfd50abe4a
could not find expected incoming transfer with hash 5398e3d05d3a8aa72ff1514dc36bab0e25e9f2c0ccca8142753edee4ea93d0b9
```
eaaef273e2bdd497962829e3e36bf84a169bf07cfa485fbd9acf3ac3943dd426 swap is for 0.0007*7 ETH
1dc7108f3461f4844bfccb4cc9377447b0a437cf10d7ef030122bf15442bae51 swap is for 0.0008*8 ETH
63b53468480f0eed9ca37789fc931d0ec4eb4a8e7eff3760371f59bfd50abe4a swap is for 0.0004*4 ETH 
5398e3d05d3a8aa72ff1514dc36bab0e25e9f2c0ccca8142753edee4ea93d0b9 swap is for 0.0002*2 ETH

4. taker still has 0.0204 of blocked amount, so nothing was returned back

[connext_taker_nocancel.log](https://github.com/ExchangeUnion/xud/files/5031033/connext_taker_nocancel.log)
[xud_taker_nocancel.log](https://github.com/ExchangeUnion/xud/files/5031034/xud_taker_nocancel.log)
[connext_maker_nocancel.log](https://github.com/ExchangeUnion/xud/files/5031035/connext_maker_nocancel.log)
[xud_maker_nocancel.log](https://github.com/ExchangeUnion/xud/files/5031036/xud_maker_nocancel.log)


5. Wait third attempt of recovery ->
Got only this and nothing about other 6 swaps. Also i checked balance of ETH for the taker. Nothing was returned back, he still has 0.0204 blocked ETH.
```
swap for a7bbc94ac3bac589488dba7ceeed3cada3ba763a613a100776eee65624d0af13 still has pending payments and will be monitored
swap for 2b48974858ec68b34d174c1580986debf5e4d654948bfd84c9b661cf7d42a73d still has pending payments and will be monitored
```

6. VERY IMPORTANT NOTE:
maker's lndbtc has only one pending htlc (for 30001 satoshi) - one of the pending swaps
[Screenshot from 2020-08-05 23-58-37](https://user-images.githubusercontent.com/29906866/89463462-a656ae00-d777-11ea-96c4-4f183a3c7d7a.png)



QUESTIONS:
1. How much is there real cases (without modified xud) when cancellation from counterparty can be reached?
2. Should cancellation work in third case for four transfers from third point (with incoming error)?
3. Why i got only 1 pending htlc but 2 monitored swaps?

PR #1785, opened 2020-08-05
feat(orderbook): prevent sub-satoshi order & match

### LINKED ISSUE #1594: Minimum order size check: swap amount smaller 1 sat fails with "no route found"
### Steps to reproduce

1. place order sell 0.0000001 ETH/BTC 0.02  as a maker
2. place order buy 0.0000001 ETH/BTC mkt as a taker

### Actual behaviour

No route error because btc amount is 0.0000001*0.02=0.**00000000**2

### Expected behaviour

User cant place order with swap amount < 8 decimals
[Screenshot from 2020-06-02 12-37-18](https://user-images.githubusercontent.com/29906866/83513947-5fc8a700-a4db-11ea-9310-a10df44e48be.png)

[logs.zip](https://github.com/ExchangeUnion/xud/files/4716771/logs.zip)

sangaman:
This prevents xud from accepting and matching orders that would result in less than a full satoshi from being exchanged for both currencies involved in the swap. It does this three ways:

1. Preventing limit orders from being placed locally that do not meet the satoshi requirement.

2. Preventing peer orders from being added to the order book if they do not meet the satoshi requirement.

3. Preventing matches that do not meet the satoshi requirement, this is needed for handling locally placed market orders since they do not have a price assigned to them until they match with an order.

It is still possible for "dust" orders that cannot be traded to get stuck in the orderbook if they are partially removed or matched leaving a tiny fraction behind. A follow-up PR can automatically remove these.

Closes #1594.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -124,7 +124,7 @@ class OrderBook extends EventEmitter {
 
   private static createOutgoingOrder = (order: OwnOrder): OutgoingOrder => {
     const { createdAt, localId, initialQuantity, hold, ...outgoingOrder } = order;
-    return outgoingOrder ;
+    return outgoingOrder;
   }
 
   private checkThresholdCompliance = (order: OwnOrder | IncomingOrder) => {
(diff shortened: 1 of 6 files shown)


raladev:
SIm test should be fixed:
 ```
            	            	rpc error: code = InvalidArgument desc = order does not meet the minimum BTC quantity of 1 satoshis
```

sangaman:
> SIm test should be fixed:
> 
> ```
>        	Error:      	Received unexpected error:
>            	            	rpc error: code = InvalidArgument desc = order does not meet the minimum BTC quantity of 1 satoshis
> ```

In investigating this, I realized my original approach was mistakenly strict. I don't need to be concerned about the base currency, we already have a minimum of 1 satoshi on that as it's the smallest currency allowed. But when the price is less than 1, we may have less than 1 satoshi on the quote currency side. So I threw out the checks of `(quantity / price) < 1` and am only checking if `(quantity * price) < 1`.

raladev:
works good, but we need to solve problem of multihop swap or increase limit to 100 sat instead on 1.

For now multihop swaps with qty less then 33 sat are failing because of noRouteErr.

PR #1790, opened 2020-08-05
test: fix openPeer await

sangaman:
This fixes test flakiness with the mocha integration tests by ensuring that we wait for all follow-up actions to an opened peer to occur before continuing with test cases that may rely on those follow-up actions, such as database writes.

I'm going to restart the github actions tests several times to ensure we don't see the test failures this aims to fix.


sangaman:
This has passed the CI tests about a dozen times in GitHub actions, I'm going to consider the related issues fixed.

PR #1796, opened 2020-08-06
fix(unitconverter): USDT unit conversion

sangaman:
This fixes the USDT unit conversion to reflect the mainnet USDT contract which only uses 6 decimal places instead of 18.


sangaman:
Closing in favor of https://github.com/ExchangeUnion/xud/pull/1795

PR #1803, opened 2020-08-08
test(sim): auto retry, safer free ports

LePremierHomme:
Introducing changes for improving simtests stability.

* Retry tests after failure (up to 3 times).
We discussed that this should be done only after unique failures, but after spending some time on that and trying different ways to capture stdout/stderr for comparison, my conclusion is that it won't be stable enough, especially if we'll do it from the bash script, and dealing with `docker-compose` output (which isn't very consistent).
Doing that from within the test code is better, but involves with some issues: 1) logging the output in real-time and capturing it at the same time isn't very straightforward, and 2) we still sometimes log data with some uniqueness, and trying to enforce the test code to not do that isn't worthwhile IMO.
So I ended up doing 3 re-runs after every failure, and the downside of that is that if we have a genuine/consistent failure, the last 2 retries could have been avoided.
* Wrapping `phayes/freeport` util with some additional thread safety code and checking of ports re-use.

In addition, returning the tests to run on `master` branch CI (on `push` event). We previously removed `push` so that it won't need to run twice on every PR, but it caused the tests to be left out of `master` CI.

### DIFF
--- a/.github/workflows/simtest.yml
+++ b/.github/workflows/simtest.yml
@@ -1,6 +1,10 @@
 name: Simulation tests
 
-on: [pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - master
 
 jobs:
   build:
--- a/test/simulation/connexttest/client.go
+++ b/test/simulation/connexttest/client.go
@@ -3,6 +3,7 @@ package connexttest
 import (
 	"bytes"
 	"fmt"
+	"github.com/ExchangeUnion/xud-simulation/shared"
 	"io"
 	"os"
 	"os/exec"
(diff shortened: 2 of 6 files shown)


sangaman:
I'm going to restart the sim tests a few times to ensure we don't get any more failures due to ports in use, and also see how it behaves when it hits a failure and auto-retries, but this looks pretty good to me otherwise.

erkarl:
Running this branch locally passed for me 3 times in a row.

kilrau:
Great! Good job @LePremierHomme ! Could you force push this branch with all signed commits? This is a new merge requirement I put in place the other day :)

PR #1808, opened 2020-08-11
fix: restrict reputation events in non-strict mode

### LINKED ISSUE #1802: Negative reputation events (e.g. no route error), even though strict mode off
Steps:
1. Reproduce no route err with strict mode = true (can be reproduced on simnet with multihop configuration and without outbound balance)
2. Reproduce no route err with strict mode = false

Actual result:
[Screenshot strict=true](https://user-images.githubusercontent.com/29906866/89675177-2bfe6900-d8f2-11ea-9532-79839d188f65.png)
[Screenshot strict=false](https://user-images.githubusercontent.com/29906866/89675172-2acd3c00-d8f2-11ea-883e-1f6f21a27211.png)

1 There is no difference in this cases, they all works like strict mode = true, i got minus 10 reputation because of swap.failure 
2 there is no logging of reputation events
3 Currency has been deactivated in both cases

Note:
Looks like there is another call of addReputationEvent in the code that is used bypassing the strict condition.
OR
strict is always true, but then idk why there is no reputation event logging

sangaman:
This ensures we only add severe and manual reputation events to peers when not running in strict mode by filtering both events that arise from the peer itself as well as those originating from other sources, such as failed swaps. Previously we only checked on events arising from the peer, which, for example, would include packet delays but not a failure to find routes.

Fixes #1802

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -675,8 +675,19 @@ class Pool extends EventEmitter {
   }
 
   // A wrapper for [[NodeList.addReputationEvent]].
-  public addReputationEvent = (nodePubKey: string, event: ReputationEvent) => {
-    return this.nodes.addReputationEvent(nodePubKey, event);
+  public addReputationEvent = async (nodePubKey: string, event: ReputationEvent) => {
+    // when in strict mode, we add all reputation events
+    // otherwise, we only add manual or severe reputation events to prevent unintentional bans
+    if (this.strict
+      || event === ReputationEvent.ManualBan
+      || event === ReputationEvent.ManualUnban
+      || event === ReputationEvent.SwapAbuse
+      || event === ReputationEvent.SwapMisbehavior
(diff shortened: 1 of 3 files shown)


raladev:
For now:
1. not-counting reputation for swap failures works (tested with noRouteErr err and sendPaymentErr)
But:
2. currency deactivation still exists for both sides (maker/taker) when strict=false

ETH deactivation in taker logs (left side)
Screenshot from 2020-08-11 18-47-47
Screenshot from 2020-08-11 18-47-41

BTC deactivation in maker logs(right side):
Screenshot from 2020-08-11 18-44-44
Screenshot from 2020-08-11 18-44-40

kilrau:
Swap failures (apat from free option delay) should not add negative reputation events neither.

sangaman:
@raladev I pushed a fix for the currency deactivating, a simple logic switcheroo that took me several looks at to catch. Thanks.

kilrau:
Restarted simulation tests once more

PR #1812, opened 2020-08-12
feat(p2p): replace order in single packet

### LINKED ISSUE #1805: Replace order call should be atomic
Steps:
1. buy 2 ltc/usdt 0.01 1
2. buy 3 ltc/usdt 0.01 2
3. buy 2 ltc/usdt 0.01 1 -r 2 - > `6 ALREADY_EXISTS: order with local id 1 already exists`
4. Check order book

Actual result:
orderbook contains only one order (first) because second one was removed, but new was not placed because id error

Expected result:
orderbook contains 2 orders, order replacement is failed.

### LINKED ISSUE #1806: Replace order call should issue new order with *same id* by default
Currently e.g. `buy 2 ltc/btc 0.01 -r 1` places the new transaction with a new (non-1) local id. Workaround is explicitely stating the new order's id, if we want it to be `1` too: `buy 2 ltc/btc 0.01 1 -r 1`. This shouldn't be required and xud should issue the re-place order with the id as the previous one by default.

sangaman:
This modifies the procedure for replacing an order in the order book. Previously, it would go like this:

1. Remove old order & send order invalidation to peers.
2. Place new order and wait for matching to complete (which may take
some time if it involves swap attempts).
3. Send new order to peers.

Now it goes:

1. Put the old order on hold, preventing further swaps/matching.
2. Place the new order and wait for matching to complete.
3. Remove the old order from the order book.
4. Send a single packet to peers with new order info and old order id.
5. Peers that receive this packet will remove the old order and add the new one sequentially.

By default, replaced orders will have the same local order id as the order they are replacing unless a different id is specified.

Closes #1805. Closes #1806.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -17,7 +17,8 @@ import errors, { errorCodes } from './errors';
 import OrderBookRepository from './OrderBookRepository';
 import TradingPair from './TradingPair';
 import {
-  IncomingOrder, isOwnOrder, Order, OrderBookThresholds, OrderIdentifier, OrderPortion, OutgoingOrder, OwnLimitOrder, OwnMarketOrder,
+  IncomingOrder, isOwnOrder, Order, OrderBookThresholds, OrderIdentifier,
+  OrderInvalidation, OrderPortion, OutgoingOrder, OwnLimitOrder, OwnMarketOrder,
   OwnOrder, Pair, PeerOrder, PlaceOrderEvent, PlaceOrderEventType, PlaceOrderResult,
 } from './types';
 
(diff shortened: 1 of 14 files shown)


raladev:
Commit: a3a70ac3c5ccd056865654d5a021668dcd7384f0
- [ ] it seems replace logic does not work for now, just new order is placed to orderbook without removing old one.
[Screenshot from 2020-08-13 13-55-51](https://user-images.githubusercontent.com/29906866/90127036-38a91400-dd6d-11ea-916c-b85929471616.png)
[Screenshot from 2020-08-13 13-55-14](https://user-images.githubusercontent.com/29906866/90127037-39da4100-dd6d-11ea-996b-03a89491f7b4.png)

sangaman:
I made several fixes and new tests, including some manual testing. I mainly just want to make sure the simulation tests here pass, since I was having some trouble running them locally. I'll check back tomorrow and make sure everything is good with the tests, but in the meantime this is ready for testing/review.

raladev:
- [ ] utils matching output is still broken
it prints nothing except of no matches found when match was founded and swap was successful
[Screen utils](https://user-images.githubusercontent.com/29906866/90247280-ed5f3600-de3e-11ea-839b-91a446ca1051.png)
[Screen log](https://user-images.githubusercontent.com/29906866/90247284-ee906300-de3e-11ea-872b-3a515965065f.png)
- [ ] replace with different ids works (~0.2 sec is required to deliver packet to remote node), but if match was founded, replace will be performed only after 20-30 sec (time of swap). So, real testing with arby is required, it can be a bottleneck.

sangaman:
> Would it be possible to use the same order id when replacing the order?

Yes I can try and handle #1806 here as well.

sangaman:
Replaced orders now use the id of the order they are replacing by default, all tests are passing, and the issue with the cli not displaying matches has been fixed. Thanks!

PR #1821, opened 2020-08-18
feat(orderbook): automatically remove dust orders

### LINKED ISSUE #1798: Set smallest order amount to actual
Follow-up of https://github.com/ExchangeUnion/xud/pull/1785:

> It is still possible for "dust" orders that cannot be traded to get stuck in the orderbook if they are partially removed or matched leaving a tiny fraction behind. A follow-up PR can automatically remove these.

Further, the actual dust limit is a per-channel policy in lightning (3*relay fee) and is usually much higher than 1 satoshi (in our example 33 sats in a 2 hop channel with boltz). Calculating it is hard, since it's per channel and varies, so setting a sufficiently high fixed default of e.g. 100 satoshis for bitcoin is probably the more realistic approach. We don't know yet if something like dust applies to connext yet.

sangaman:
This removes dust orders from the order book as they arise due to orders being partially removed or matched. Dust orders are defined as orders where either side of the trade has less than 100 satoshis of volume.

This logic exists in three places. During matching, we check whether the remainder of any maker order is beneath the dust limit. During partial order removal, we check whether the remainder is beneath the dust limit and, if so, remove the entire order instead. Lastly, any remaining order after matching is checked against the dust limit as well.

Closes #1798.

Follow-up to #1785.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -211,7 +211,7 @@ class OrderBook extends EventEmitter {
     currencies.forEach(currency => this.currencyInstances.set(currency.id, currency));
     pairs.forEach((pair) => {
       this.pairInstances.set(pair.id, pair);
-      this.tradingPairs.set(pair.id, new TradingPair(this.logger, pair.id, this.nomatching));
+      this.addTradingPair(pair.id);
     });
 
     this.pool.updatePairs(this.pairIds);
@@ -286,10 +286,25 @@ class OrderBook extends EventEmitter {
 
     const pairInstance = await this.repository.addPair(pair);
     this.pairInstances.set(pairInstance.id, pairInstance);
-    this.tradingPairs.set(pairInstance.id, new TradingPair(this.logger, pairInstance.id, this.nomatching));
+    this.addTradingPair(pairInstance.id);
 
     this.pool.updatePairs(this.pairIds);
     return pairInstance;
+  }
+
+  private addTradingPair = (pairId: string) => {
+    const tp = new TradingPair(this.logger, pairId, this.nomatching);
+    this.tradingPairs.set(pairId, tp);
+    tp.on('peerOrder.dust', (order) => {
+      this.removePeerOrder(order.id, order.pairId);
+    });
+    tp.on('ownOrder.dust', (order) => {
+      this.removeOwnOrder({
+        orderId: order.id,
+        pairId: order.pairId,
+        quantityToRemove: order.quantity,
+      });
+    });
   }
 
   public addCurrency = async (currency: CurrencyCreationAttributes) => {
@@ -342,9 +357,13 @@ class OrderBook extends EventEmitter {
     }): Promise<PlaceOrderResult> => {
     const stampedOrder = this.stampOwnOrder(order, replaceOrderId);
 
-    if (order.quantity * order.price < 1) {
+    if (order.quantity < TradingPair.QUANTITY_DUST_LIMIT) {
+      const baseCurrency = order.pairId.split('/')[0];
+      throw errors.MIN_QUANTITY_VIOLATED(TradingPair.QUANTITY_DUST_LIMIT, baseCurrency);
+    }
+    if (order.quantity * order.price < TradingPair.QUANTITY_DUST_LIMIT) {
       const quoteCurrency = order.pairId.split('/')[1];
-      throw errors.MIN_QUANTITY_VIOLATED(1, quoteCurrency);
+      throw errors.MIN_QUANTITY_VIOLATED(TradingPair.QUANTITY_DUST_LIMIT, quoteCurrency);
     }
 
     if (this.nomatching) {
@@ -374,6 +393,11 @@ class OrderBook extends EventEmitter {
   }): Promise<PlaceOrderResult> => {
     if (this.nomatching) {
       throw errors.MARKET_ORDERS_NOT_ALLOWED();
+    }
+
+    if (order.quantity < TradingPair.QUANTITY_DUST_LIMIT) {
+      const baseCurrency = order.pairId.split('/')[0];
+      throw errors.MIN_QUANTITY_VIOLATED(TradingPair.QUANTITY_DUST_LIMIT, baseCurrency);
     }
 
     const stampedOrder = this.stampOwnOrder({ ...order, price: order.isBuy ? Number.POSITIVE_INFINITY : 0 });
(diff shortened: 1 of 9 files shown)


kilrau:
Can't really remember why we decided to have 1 satoshi as limit for `PlaceOrder` (https://github.com/ExchangeUnion/xud/pull/1785), which ends up in an untradable (because unroutable via lightning) dust order the same way a remainder of a matching process does. Shouldn't the 100 satoshi limit apply to what was addressed in https://github.com/ExchangeUnion/xud/pull/1785 too?

```
mainnet > sell 0.00000001 btc/usdt 11123
remaining 0.00000001 BTC entered the order book as 65016610-e182-11ea-b2d4-49514e958013
no matches found
mainnet > listorders

Trading pair: BTC/USDT
│               │             │                  │ 0.00000001    │ 11123       │ FoilTrip         │
```

sangaman:
> Can't really remember why we decided to have 1 satoshi as limit for `PlaceOrder` (#1785), which ends up in an untradable (because unroutable via lightning) dust order the same way a remainder of a matching process does. Shouldn't the 100 satoshi limit apply to what was addressed in #1785 too?

Thanks for catching this, that's one more place I needed to change the limit from 1 to 100. I replaced the numeric values with a constant so in the future we only need to update one line if we want to increase/decrease 100, so that there won't be oversights like that.

lib/orderbook/OrderBook.ts
```diff
@@ -286,12 +286,27 @@ class OrderBook extends EventEmitter {
...
+    this.addTradingPair(pairInstance.id);
 
     this.pool.updatePairs(this.pairIds);
     return pairInstance;
   }
 
+  private addTradingPair = (pairId: string) => {
+    const tp = new TradingPair(this.logger, pairId, this.nomatching);
+    this.tradingPairs.set(pairId, tp);
+    tp.on('peerOrder.dust', (order) => {
```
erkarl, line 298: Should we also add code to clean up these listeners before shutting down?

lib/orderbook/TradingPair.ts
```diff
@@ -365,11 +390,17 @@ class TradingPair {
...
 
       const matchingQuantity = getMatchingQuantity(remainingOrder, makerAvailableQuantityOrder);
-      if (matchingQuantity <= 0) {
-        // there's no match with the best available maker order, so end the matching routine
-        break;
-      } else if (makerOrder.quantity * makerOrder.price < 1) {
-        // there's a match but it doesn't meet the 1 satoshi minimum on both sides of the trade
+      if (matchingQuantity * makerOrder.price < TradingPair.QUANTITY_DUST_LIMIT) {
+        // there's no match with the best available maker order OR there's a match
+        // but it doesn't meet the 1 satoshi minimum on both sides of the trade
```
erkarl, line 395:
```suggestion
        // but it doesn't meet the minimum on both sides of the trade
```

lib/orderbook/OrderBook.ts
sangaman: I don't believe it's necessary as far as graceful xud shutdowns go, and the `OrderBook` class currently doesn't have a close/shutdown method. `OrderBook` already listens to `Pool` and `Swaps` events that don't get explicitly removed and doesn't have any issues, so I was going to leave that alone. I know in other spots we are explicitly removing listeners (even when not necessary) but I stuck to the pattern in this class. Let me know if that sounds good to you.

erkarl (commented):
Tested and this works for maker orders.

Should this also work for taker orders?
```
➜  xud git:(dust-order-removal) ✗ ./bin/xucli buy 0.01 ETH/BTC 1
matched 0.00999901 ETH @ 1 with peer ProsperModify order dbdf3cb0-e1fc-11ea-89a6-71ed0ac19342, attempting swap...
successfully swapped 0.00999901 ETH with peer order dbdf3cb0-e1fc-11ea-89a6-71ed0ac19342
remaining 0.00000099 ETH entered the order book as d9752fa0-e1fe-11ea-943f-535d84ff4505
➜  xud git:(dust-order-removal) ✗ ./bin/xucli orderbook ETH/BTC

Trading pair: ETH/BTC
│ Quantity          │ Price             │ Price             │ Quantity          │
│ 0.00000099        │ 1                 │                   │                   │
└───────────────────┴───────────────────┴───────────────────┴───────────────────┘
```

kilrau:
It should. This part should have been discarded:
`remaining 0.00000099 ETH entered the order book as d9752fa0-e1fe-11ea-943f-535d84ff4505`

Since I can't make it out from the git diff: what are we printing when discarding an order because sub-dust threshold? Should be something along the lines `Discarded remaining 0.00000099 ETH because quantity below dust level`

sangaman:
> Tested and this works for maker orders.
> 
> Should this also work for taker orders?
> 
> ```
> ➜  xud git:(dust-order-removal) ✗ ./bin/xucli buy 0.01 ETH/BTC 1
> matched 0.00999901 ETH @ 1 with peer ProsperModify order dbdf3cb0-e1fc-11ea-89a6-71ed0ac19342, attempting swap...
> successfully swapped 0.00999901 ETH with peer order dbdf3cb0-e1fc-11ea-89a6-71ed0ac19342
> remaining 0.00000099 ETH entered the order book as d9752fa0-e1fe-11ea-943f-535d84ff4505
> ➜  xud git:(dust-order-removal) ✗ ./bin/xucli orderbook ETH/BTC
> 
> Trading pair: ETH/BTC
> ┌───────────────────────────────────────┬───────────────────────────────────────┐
> │ Buy                                   │ Sell                                  │
> ├───────────────────┬───────────────────┼───────────────────┬───────────────────┤
> │ Quantity          │ Price             │ Price             │ Quantity          │
> ├───────────────────┼───────────────────┼───────────────────┼───────────────────┤
> │ 0.00000099        │ 1                 │                   │                   │
> └───────────────────┴───────────────────┴───────────────────┴───────────────────┘
> ```

Hmmm yes it definitely should, I'll look into this and also create a new simulation test for this exact test case.


> It should. This part should have been discarded:
> `remaining 0.00000099 ETH entered the order book as d9752fa0-e1fe-11ea-943f-535d84ff4505`
> 
> Since I can't make it out from the git diff: what are we printing when discarding an order because sub-dust threshold? Should be something along the lines `Discarded remaining 0.00000099 ETH because quantity below dust level`

From the cli perspective nothing has changed, what should print right now is just that there were no more matches and the remaining quantity was discarded. It wouldn't know *why* there were no matches, that would take some changes on the rpc level to add fields to the `PlaceOrder` response indicating the reason for no more matches.

kilrau:
Ok, as long as it says "discarded remaining quantity of X" its fine imo

erkarl:
Taker side also seems to now work :+1: 
```
➜  xud git:(dust-order-removal) ✗ ./bin/xucli buy 0.01 ETH/BTC 1
matched 0.00999901 ETH @ 1 with peer InjuryClump order 79c44170-e2be-11ea-84c5-7b6f8485b9c1, attempting swap...
successfully swapped 0.00999901 ETH with peer order 79c44170-e2be-11ea-84c5-7b6f8485b9c1
no more matches found, 0.00000099 qty will be discarded
```

erkarl (approved):

PR #1824, opened 2020-08-19
refactor: remove raiden support

### LINKED ISSUE #1516: Post-connext cleanup
- remove raiden warning from logs, e.g. `xud tries to find riden backup location, but i think it shuld be a part of task with 'raiden things' - : Error: ENOENT: no such file or directory, watch '/root/.raiden';`
- logs still contain
```
09/06/2020 13:53:33.422 [RAIDEN] info: trying to verify connection to raiden with uri: raiden:5001
09/06/2020 13:53:33.427 [RAIDEN] error: Error: getaddrinfo ENOTFOUND raiden
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:66:26)
```
- `listpeers -j` contains `"raidenAddress": "",`
- [debug]
raidenDirectChannelChecks = true

sangaman:
Closes #1516.

BREAKING CHANGE: removes raiden support & p2p packet fields


lib/constants/enums.ts
```diff
@@ -1,7 +1,6 @@
 /** An enumeration of payment channel network clients that support token swaps. */
 export enum SwapClientType {
   Lnd,
-  Raiden,
```
erkarl, line 4: I wonder if this breaks existing xud environments since in the database it's 0, 1, 2, and now 0, 1?

test/jest/__snapshots__/NodeKey.spec.ts.snap
```diff
@@ -1,3 +1,3 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`NodeKey it should derive a child seed from private key 1`] = `"crouch cause bicycle tomato choice city produce ocean canyon rare laugh sudden close angle cage item special jeans either heavy skill bracket lift butter"`;
+exports[`NodeKey it should derive a child seed from private key 1`] = `"chat toward myself element estate miracle merit pudding next net process idle mobile hawk transfer chimney border gate patient artefact square strike worth already"`;
```
erkarl, line 3: I'm curious why this snapshot update is necessary? Maybe I'm missing something, but there are no changes to `NodeKey.ts`

erkarl (commented):
:joy: :partying_face: :+1:

kilrau:
Let's leave this one open until @raladev is back for manual testing

test/jest/__snapshots__/NodeKey.spec.ts.snap
sangaman: We use the name of the swap client in the entropy used to generate the child seed, so it makes sense that this is changing because it's going from raiden to connext.

erkarl: Ah, right. It's because the enum changed. Thanks for clearing this up.

erkarl:
One minor nit. I'm not sure if `refactor: *` commit messages show up in the release notes? Maybe we change it since it'll be a breaking change on the p2p layer (we removed the `raidenAddress` field).

sangaman:
> One minor nit. I'm not sure if `refactor: *` commit messages show up in the release notes? Maybe we change it since it'll be a breaking change on the p2p layer (we removed the `raidenAddress` field).

Good point, it should get the BREAKING CHANGE label (which will show up in release notes).

erkarl:
@sangaman could you please rebase this so that it's testable via xud-docker setup (the contract address for USDT changed and it will cause infinite requests to the faucet if running the old version).

sangaman:
> @sangaman could you please rebase this so that it's testable via xud-docker setup (the contract address for USDT changed and it will cause infinite requests to the faucet if running the old version).

Done.

erkarl (approved):

kilrau (approved):
Was a bit complicated to test because of the breaking changes on p2p layer, but looking good. Tested new environment and upgrading an existing environment with this branch.

PR #1826, opened 2020-08-19
feat(connext): mark payment as failed when receiving 404 status

### LINKED ISSUE #1708: xud should cancel invoice when receiving 404 from connext
Depends on: https://github.com/connext/rest-api-client/issues/31

**Problem**
At the moment connext's `/hashlock-status` endpoint (for `swapClient.lookupPayment`) only returns 500 status codes which is quite bad since if the connext transfer fails for some reason so that the HTLC is never created it will go into recovery and start querying it.

**Solution**
It should get 404 status code back from connext if the HTLC does not exist. That would mean that it was never created and we can also cancel the invoice on the other chain.

ghost:
This PR adds logic to mark the payment state as failed when we receive a 404 status code from connext's `/hashlock-status` endpoint. At this point we can be sure that the payment was not created and it's safe to do so.

Closes https://github.com/ExchangeUnion/xud/issues/1708

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -433,8 +433,6 @@ class ConnextClient extends SwapClient {
     try {
       const assetId = this.getTokenAddress(currency);
       const transferStatusResponse = await this.getHashLockStatus(rHash, assetId);
-      // TODO: Edge case. Once https://github.com/connext/rest-api-client/issues/31 is
-      // implemented check for the response code 404 and set the payment as failed.
 
       switch (transferStatusResponse.status) {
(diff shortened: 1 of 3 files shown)


sangaman (dismissed):
It would be good to have a test case for this too, either in this PR or a future one right now we have a similar test (404 for an rHash not found) for raiden but that will be removed soon.

erkarl:
> It would be good to have a test case for this too, either in this PR or a future one right now we have a similar test (404 for an rHash not found) for raiden but that will be removed soon.

Added some test cases.

sangaman (approved):
Looks great, thank you.

PR #1831, opened 2020-08-24
fix: improve HttpServer error logging

ghost:
This improves the error message printed by HttpServer.

### DIFF
--- a/lib/http/HttpServer.ts
+++ b/lib/http/HttpServer.ts
@@ -41,7 +41,7 @@ class HttpServer {
         }
       } catch (err) {
         const msg = `processing request to ${req.url} failed`;
-        this.logger.error(`${msg}: ${err}`);
+        this.logger.error(msg, err);
         statusCode = 500;
         resJson = { message: msg };
       }


lib/http/HttpServer.ts
```diff
@@ -41,7 +41,7 @@ class HttpServer {
         }
       } catch (err) {
         const msg = `processing request to ${req.url} failed`;
-        this.logger.error(`${msg}: ${err}`);
```
sangaman, line 44: This could be

`this.logger.error(msg, err);`

Because the `error` log method already accepts a second parameter for the error and will automatically JSON.stringify it if it's an object.

sangaman (commented):
Good catch, I made a suggestion to just make this more in line with the way we log errors elsewhere, although it does the same main fix of printing the full object json.

Also, I think rather than `chore` this should be a `fix` or `feat` commit or maybe just `refactor` if you think it's too minor, but chore should just be build/project related stuff and not code changes as I understand it (even if minor).

lib/http/HttpServer.ts
erkarl: Thanks - good catch. Updated @sangaman.

sangaman (approved):

Issue #1771, opened 2020-07-30
Tests flakiness


sangaman:
This issue is for tracking intermittent test failures that cause our tests to be flaky and require manual restarts even when there is no underlying issue in the xud code itself. Since there seem to be quite a few, I figured it would be easier to track them all in one issue. These primarily effect simulation tests but also occasionally our mocha/jest tests.

sangaman:
Failure at the start of running simulation tests, a new one I haven't seen before.

```
 === RUN   TestIntegration
2020/07/30 13:23:37 Connext node reachability failure: invalid response at http://172.17.0.1:8080/config: [123 34 115 116 97 116 117 115 67 111 100 101 34 58 52 48 52 44 34 109 101 115 115 97 103 101 34 58 34 67 97 110 110 111 116 32 71 69 84 32 47 99 111 110 102 105 103 34 44 34 101 114 114 111 114 34 58 34 78 111 116 32 70 111 117 110 100 34 125]
exit status 1
FAIL	github.com/ExchangeUnion/xud-simulation	0.083s
~/work/xud/xud/test/simulation/temp/indra ~/work/xud/xud/test/simulation
```

sangaman:
This happens occasionally with the Mocha tests, I've seen it several times both locally and on GitHub Actions.

```
   1) P2P Pool Tests
       should update a node on new handshake:

      AssertionError: expected '123.123.123.123' to equal '86.75.30.9'
      + expected - actual

      -123.123.123.123
      +86.75.30.9
      
      at /home/runner/work/xud/xud/test/integration/Pool.spec.ts:103:49
      at Generator.next (<anonymous>)
      at fulfilled (test/integration/Pool.spec.ts:24:58)

  2) P2P Pool Tests
       reconnect logic
         "before each" hook for "should not reconnect upon shutdown inbound":
     node 022a49046f252821dfcb3a3f7c8e203af97fb2b8223d8b9204380ccdce6910755d at 321.321.321.321:1337 already connected
```

sangaman:
This one I've seen for a while, it happens intermittently and more often locally. I've tried looking into this a bit before but was stumped.

```
  1) Parser
       test TCP segmentation/concatenation support
         should parse encrypted Ping SessionInit Ping concatenated and split on byte 15 from each packet beginning:

[...]

PR #1840, opened 2020-08-26
test(sim): extend gRPC TLS certificate timeouts

sangaman:
This aims to decrease failed simulation tests due to the TLS certificate taking too long to be created by xud.

Related: https://github.com/ExchangeUnion/xud/issues/1771#issuecomment-681108205

I plan to restart the simulation tests for this several times to see if the error referenced above reappears or not.

### DIFF
--- a/test/simulation/xudtest/node.go
+++ b/test/simulation/xudtest/node.go
@@ -3,7 +3,6 @@ package xudtest
 import (
 	"bytes"
 	"fmt"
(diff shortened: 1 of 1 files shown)


sangaman:
The error is still happening in this PR and others, I think the only sensible step is to revert #1692. I'll close this if the test failures go away after reverting.

kilrau:
Looks like it worked https://github.com/ExchangeUnion/xud/pull/1841#event-3698317311. Closing.

Issue #1643, opened 2020-06-12
Don't throw assertion on xucli argument error


kilrau:
We should try catching all of these kind of xucli argement error so that it doesn't throw an assertion. Example:
```
testnet > openchannel 0.01 BTC BrassMarble 0.005
xucli openchannel <currency> <amount> [node_identifier] [push_amount]

open a payment channel with a peer

Positionals:
  currency  the ticker symbol for the currency               [string] [required]
  amount    the amount to be deposited into the channel      [number] [required]

Options:
  --help             Show help                                         [boolean]
  --version          Show version number                               [boolean]
  --rpcport, -p      RPC service port                                   [number]
  --rpchost, -h      RPC service hostname                               [string]
  --tlscertpath, -c  Path to the TLS certificate of xud                 [string]
  --json, -j         Display output in json format    [boolean] [default: false]
  --xudir, -x        Data directory for xud                             [string]
  --node_identifier  the node key or alias of the connected peer to open the
                     channel with                                       [string]
  --push_amount      the amount to be pushed to the remote side of the channel
                                                           [number] [default: 0]

Examples:
  xucli openchannel BTC 0.1                 open an 0.1 BTC channel by node key
  028599d05b18c0c3f8028915a17d603416f7276c
  822b6b2d20e71a3502bd0f9e0b
  xucli openchannel BTC 0.1 CheeseMonkey    open an 0.1 BTC channel by alias
  xucli openchannel BTC 0.1 CheeseMonkey    open an 0.1 BTC channel by alias and
  0.05                                      push 0.05 to remote side
  xucli openchannel ETH 0.5                 deposit 0.5 into an ETH Connext
                                            channel without specifying a remote
                                            node

Error [AssertionError]: Assertion failed
    at new goog.asserts.AssertionError (/app/node_modules/google-protobuf/google-protobuf.js:81:876)
    at Object.goog.asserts.doAssertFailure_ (/app/node_modules/google-protobuf/google-protobuf.js:82:257)
    at Object.goog.asserts.assert (/app/node_modules/google-protobuf/google-protobuf.js:83:83)
    at jspb.BinaryWriter.writeUint64 (/app/node_modules/google-protobuf/google-protobuf.js:483:419)
    at Function.proto.xudrpc.OpenChannelRequest.serializeBinaryToWriter (/app/dist/proto/xudrpc_pb.js:6591:12)
[...]

PR #1846, opened 2020-08-28
added param check to walletwithdraw/addpair commands

### LINKED ISSUE #1844: Syntax err in `walletwithdraw` response
### Actual behaviour

```
simnet > walletwithdraw
xucli walletwithdraw [amount] [currency] [destination] [fee]

withdraws on-chain funds from xud

Options:
  --help             Show help                                         [boolean]
  --version          Show version number                               [boolean]
  --rpcport, -p      RPC service port                                   [number]
  --rpchost, -h      RPC service hostname                               [string]
  --tlscertpath, -c  Path to the TLS certificate of xud                 [string]
  --json, -j         Display output in json format    [boolean] [default: false]
  --xudir, -x        Data directory for xud                             [string]
  --amount           the amount to withdraw                             [number]
  --currency         the ticker symbol of the currency to withdraw.     [string]
  --destination      the address to send withdrawn funds to             [string]
  --fee              the fee in satoshis (or equivalent) per byte       [number]
  --all              whether to withdraw all available funds           [boolean]

Examples:
  xucli withdraw 0.1 BTC                    withdraws 0.1 BTC
  1BitcoinEaterAddressDontSendf59kuE
  xucli withdraw 0.1 BTC                    withdraws 0.1 BTC using 20 sats/byte
  1BitcoinEaterAddressDontSendf59kuE 20
  xucli withdraw --all --currency BTC       withdraws all BTC
  --address
  1BitcoinEaterAddressDontSendf59kuE

TypeError: Cannot read property 'toUpperCase' of undefined
    at /app/dist/cli/commands/walletwithdraw.js:44:39
    at /app/dist/cli/commands/walletwithdraw.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/app/dist/cli/commands/walletwithdraw.js:4:12)
    at Object.exports.handler (/app/dist/cli/commands/walletwithdraw.js:42:29)
    at Object.runCommand (/app/node_modules/yargs/lib/command.js:237:40)
    at Object.parseArgs [as _parseArgs] (/app/node_modules/yargs/yargs.js:1090:30)
    at Object.get [as argv] (/app/node_modules/yargs/yargs.js:1024:21)
    at Object.<anonymous> (/app/bin/xucli:43:3)
```
Expected result:
No the syntax err

rsercano:
attempts to resolve #1844

But IMO, this's not enough, because there're loads of missing argument checks in commands, I guess we need something alike in Service.ts.

Do you want me to implement similar arg checks of Service.ts to commands?

### DIFF
--- a/lib/cli/commands/walletwithdraw.ts
+++ b/lib/cli/commands/walletwithdraw.ts
@@ -33,6 +33,10 @@ export const builder = (argv: Argv) => argv
   .example('$0 withdraw --all --currency BTC --address 1BitcoinEaterAddressDontSendf59kuE', 'withdraws all BTC');
 
 export const handler = async (argv: Arguments<any>) => {
+  if (!argv.currency) {
+    throw 'currency is not valid';
+  }
+
   const request = new WithdrawRequest();
   request.setCurrency(argv.currency.toUpperCase());
   if (argv.all) {
--- a/lib/cli/commands/walletwithdraw.ts
+++ b/lib/cli/commands/walletwithdraw.ts
@@ -1,16 +1,16 @@
 import { Arguments, Argv } from 'yargs';
 import { WithdrawRequest } from '../../proto/xudrpc_pb';
 import { callback, loadXudClient } from '../command';
-import { coinsToSats } from '../utils';
+import { argChecks, coinsToSats } from '../utils';
 
-export const command = 'walletwithdraw [amount] [currency] [destination] [fee]';
+export const command = 'walletwithdraw <amount> <currency> <destination> [fee]';
 
 export const describe = 'withdraws on-chain funds from xud';
 
 export const builder = (argv: Argv) => argv
   .option('amount', {
     description: 'the amount to withdraw',
-    type: 'number',
+    type: 'string',
   })
   .option('currency', {
     description: 'the ticker symbol of the currency to withdraw.',
@@ -24,23 +24,17 @@ export const builder = (argv: Argv) => argv
     description: 'the fee in satoshis (or equivalent) per byte',
     type: 'number',
   })
-  .option('all', {
-    description: 'whether to withdraw all available funds',
-    type: 'boolean',
-  })
   .example('$0 walletwithdraw 0.1 BTC 1BitcoinEaterAddressDontSendf59kuE', 'withdraws 0.1 BTC')
   .example('$0 walletwithdraw 0.1 BTC 1BitcoinEaterAddressDontSendf59kuE 20', 'withdraws 0.1 BTC using 20 sats/byte')
-  .example('$0 walletwithdraw --all --currency BTC --address 1BitcoinEaterAddressDontSendf59kuE', 'withdraws all BTC');
+  .example('$0 walletwithdraw all BTC 1BitcoinEaterAddressDontSendf59kuE', 'withdraws all BTC');
 
 export const handler = async (argv: Arguments<any>) => {
-  if (!argv.currency) {
-    throw 'currency is not valid';
-  }
+  argChecks.NUMBER_CHECK({ param: argv.amount, paramName: 'amount', allowedValues: ['all'] });
 
   const request = new WithdrawRequest();
   request.setCurrency(argv.currency.toUpperCase());
-  if (argv.all) {
-    request.setAll(argv.all);
+  if (argv.amount === 'all') {
+    request.setAll(true);
   } else {
     request.setAmount(coinsToSats(argv.amount));
   }
--- a/lib/cli/commands/walletwithdraw.ts
+++ b/lib/cli/commands/walletwithdraw.ts
@@ -3,16 +3,16 @@ import { WithdrawRequest } from '../../proto/xudrpc_pb';
 import { callback, loadXudClient } from '../command';
 import { argChecks, coinsToSats } from '../utils';
 
-export const command = 'walletwithdraw <amount> <currency> <destination> [fee]';
+export const command = 'walletwithdraw <amount> <currency> [destination] [fee]';
 
 export const describe = 'withdraws on-chain funds from xud';
 
 export const builder = (argv: Argv) => argv
-  .option('amount', {
-    description: 'the amount to withdraw',
+  .positional('amount', {
+    description: 'the amount to withdraw, `all` withdraws everything',
     type: 'string',
   })
-  .option('currency', {
+  .positional('currency', {
     description: 'the ticker symbol of the currency to withdraw.',
     type: 'string',
   })
@@ -36,7 +36,7 @@ export const handler = async (argv: Arguments<any>) => {
   if (argv.amount === 'all') {
     request.setAll(true);
   } else {
-    request.setAmount(coinsToSats(argv.amount));
+    request.setAmount(coinsToSats(parseFloat(argv.amount)));
   }
   request.setDestination(argv.destination);
   request.setFee(argv.fee);
--- a/lib/cli/commands/walletwithdraw.ts
+++ b/lib/cli/commands/walletwithdraw.ts
@@ -28,9 +28,9 @@ export const builder = (argv: Argv) => argv
     description: 'whether to withdraw all available funds',
     type: 'boolean',
   })
-  .example('$0 withdraw 0.1 BTC 1BitcoinEaterAddressDontSendf59kuE', 'withdraws 0.1 BTC')
-  .example('$0 withdraw 0.1 BTC 1BitcoinEaterAddressDontSendf59kuE 20', 'withdraws 0.1 BTC using 20 sats/byte')
-  .example('$0 withdraw --all --currency BTC --address 1BitcoinEaterAddressDontSendf59kuE', 'withdraws all BTC');
+  .example('$0 walletwithdraw 0.1 BTC 1BitcoinEaterAddressDontSendf59kuE', 'withdraws 0.1 BTC')
+  .example('$0 walletwithdraw 0.1 BTC 1BitcoinEaterAddressDontSendf59kuE 20', 'withdraws 0.1 BTC using 20 sats/byte')
+  .example('$0 walletwithdraw --all --currency BTC --address 1BitcoinEaterAddressDontSendf59kuE', 'withdraws all BTC');
 
 export const handler = async (argv: Arguments<any>) => {
   if (!argv.currency) {
--- a/lib/cli/commands/walletwithdraw.ts
+++ b/lib/cli/commands/walletwithdraw.ts
@@ -1,9 +1,9 @@
 import { Arguments, Argv } from 'yargs';
 import { WithdrawRequest } from '../../proto/xudrpc_pb';
 import { callback, loadXudClient } from '../command';
-import { argChecks, coinsToSats } from '../utils';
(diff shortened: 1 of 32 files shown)


kilrau:
> Do you want me to implement similar arg checks of Service.ts to commands?

I guess that makes sense. Wdyt? @sangaman

raladev:
1. Suggestion about walletwithdraw:

For now walletwithdraw validation is a little bit different from others because all parameters marked as optional (ant it is the reason of the bug).
```
simnet > walletwithdraw
xucli walletwithdraw [amount] [currency] [destination] [fee]

currency is not valid

simnet > walletdeposit
xucli walletdeposit <currency>

Not enough non-option arguments: got 0, need at least 1

simnet > 
``` 

But first 2 arguments should be non-option for walletwithdraw, and reason why it is not non-option is --all parameter (not sure).

So my suggestion is remove --all parameter and make first 2 arguments of withdrawal non-option how it should be,  because currency not valid means that currency is required param but it is marked as optional.

2. About other fileds without validation, we have issue for that - https://github.com/ExchangeUnion/xud/issues/1643

kilrau:
> So my suggestion is remove --all parameter and make first 2 arguments of withdrawal non-option how it should be, because currency not valid means that currency is required param but it is marked as optional.

Agreed, all of these should be mandatory: `[amount] [currency] [destination]`. Further, I suggest we can have one exception for `amount` and that is the string `all` (not case sensitive as always) and add it to the examples of the help out put to make it clear how withdrawing `all` funds works. So the only optional parameter is `[fee]`, no separate `[all]` parameter. Sounds good?

raladev (changes requested):
Above

kilrau:
Yeah https://github.com/ExchangeUnion/xud/pull/1846#issuecomment-682466538 is still to be done by @rsercano

rsercano:
Yes, I'll do asap, sorry for late response

rsercano:
Added

raladev:
-  [  ] all option does not work for eth. There was 'This option is not supported' msg instead of 500 err.
```
 walletwithdraw all eth 0x1E570c337937F6dF5D731E7bE43c1d354ec45d26
Error: 2 UNKNOWN: connext server error 500: Internal Server Error
```
```
[1599138446059] INFO  (70 on connext): received request
    url: "/onchain-transfer"
    id: 173
    reqId: 173
[1599138446061] INFO  (70 on connext): Getting wallet for index: 0
[1599138446072] INFO  (70 on connext): Getting wallet for publicIdentifier: indra8mhr8qoXJL5Qn8NRhzkRg1FPGyESMRhSzkEz5ArgXS3zvsNfDc
[1599138446129] ERROR (70 on connext): invalid BigNumber string (argument="value", value="", code=INVALID_ARGUMENT, version=bignumber/5.0.1)
    Error: invalid BigNumber string (argument="value", value="", code=INVALID_ARGUMENT, version=bignumber/5.0.1)
        at Logger.makeError (/app/node_modules/@ethersproject/logger/lib/index.js:179:21)
        at Logger.throwError (/app/node_modules/@ethersproject/logger/lib/index.js:188:20)
        at Logger.throwArgumentError (/app/node_modules/@ethersproject/logger/lib/index.js:191:21)
        at Function.BigNumber.from (/app/node_modules/@ethersproject/bignumber/lib/bignumber.js:167:27)
        at /app/build/src/helpers/ethereum.js:165:51
        at step (/app/build/src/helpers/ethereum.js:33:23)
        at Object.next (/app/build/src/helpers/ethereum.js:14:53)
        at /app/build/src/helpers/ethereum.js:8:71
        at new Promise (<anonymous>)
        at __awaiter (/app/build/src/helpers/ethereum.js:4:12)
[1599138446137] INFO  (70 on connext): request completed
    url: "/onchain-transfer"
    statusCode: 500
    reqId: 173
[1599138446638] INFO  (70 on connext): Getting client for publicIdentifier: indra8mhr8qoXJL5Qn8NRhzkRg1FPGyESMRhSzkEz5ArgXS3zvsNfDc
```

- [ ] all option does not work for btc too :( 
```
simnet > walletwithdraw all btc sb1qd3hns0kd5z6xl7l0zftv37nmj673gjhfjzhvp0
Error: 2 UNKNOWN: 2 UNKNOWN: transaction output is dust
```

rsercano:
Interesting, I didn't change how all option behaves I just changed how to obtain it, but will check.

raladev:
> Interesting, I didn't change how all option behaves I just changed how to obtain it, but will check, thanks. (It actually works how it was with --all)

About first one:
https://github.com/ExchangeUnion/xud/blob/5e3ad04ed8f2e35cb06479d2172d25bffa002768/lib/connextclient/ConnextClient.ts#L723
maybe value of all parameter for connext client was changed after your changes

About second one:
Not sure that it was fine before

rsercano:
Ah nice catch, completely my bad that I forgot to set all as a boolean since we changed how it comes, now fixed.

raladev (dismissed):

lib/cli/commands/walletwithdraw.ts
```diff
@@ -1,16 +1,16 @@
 import { Arguments, Argv } from 'yargs';
 import { WithdrawRequest } from '../../proto/xudrpc_pb';
 import { callback, loadXudClient } from '../command';
-import { coinsToSats } from '../utils';
+import { argChecks, coinsToSats } from '../utils';
 
-export const command = 'walletwithdraw [amount] [currency] [destination] [fee]';
+export const command = 'walletwithdraw <amount> <currency> <destination> [fee]';
```
sangaman, line 6: I think we want to leave destination as optional as it's not required for connext.

lib/cli/commands/walletwithdraw.ts
```diff
@@ -1,16 +1,16 @@
...
 
 export const describe = 'withdraws on-chain funds from xud';
 
 export const builder = (argv: Argv) => argv
   .option('amount', {
     description: 'the amount to withdraw',
```
sangaman, line 12: We should mention that typing in "all" withdraws everything.

lib/cli/commands/walletwithdraw.ts
```diff
@@ -1,16 +1,16 @@
...
-import { coinsToSats } from '../utils';
+import { argChecks, coinsToSats } from '../utils';
 
-export const command = 'walletwithdraw [amount] [currency] [destination] [fee]';
+export const command = 'walletwithdraw <amount> <currency> <destination> [fee]';
 
 export const describe = 'withdraws on-chain funds from xud';
 
 export const builder = (argv: Argv) => argv
   .option('amount', {
```
sangaman, line 11: This should be `positional` rather than `option`.

lib/cli/commands/walletwithdraw.ts
```diff
@@ -1,16 +1,16 @@
...
     description: 'the amount to withdraw',
-    type: 'number',
+    type: 'string',
   })
   .option('currency', {
```
sangaman, line 15: `positional` here as well.

lib/cli/utils.ts
```diff
@@ -80,3 +80,12 @@ export const waitForCert = (certPath: string) => {
     }
   });
 };
+
+export const argChecks = {
```
sangaman, line 84: I don't think we need the whole `argChecks` thing here, which I intended as a set of reusable server side argument checks for similar parameters across multiple calls.

I think instead we should check if `argv.amount === 'all'` as we already do, else check `!isNan()`. If it's not a number and not `all`, then `console.error` a message and `process.exit(1)` which is how we handle similar situations for other commands.

lib/cli/commands/walletwithdraw.ts
```diff
@@ -24,19 +24,17 @@ export const builder = (argv: Argv) => argv
...
+  argChecks.NUMBER_CHECK({ param: argv.amount, paramName: 'amount', allowedValues: ['all'] });
+
   const request = new WithdrawRequest();
   request.setCurrency(argv.currency.toUpperCase());
-  if (argv.all) {
-    request.setAll(argv.all);
+  if (argv.amount === 'all') {
+    request.setAll(true);
   } else {
     request.setAmount(coinsToSats(argv.amount));
```
sangaman, line 39: `coinsToSats` exepcts a `number` but we've defined `amount` as a `string` above so I'm concerned yargs would give us a string. I'm thinking we should do `parseFloat(argv.amount)` here.

sangaman (changes requested):

lib/cli/commands/walletwithdraw.ts
rsercano: Gotcha

lib/cli/utils.ts
rsercano: I put it separately just as a beginning for all command checks wouldn't it be better to move all arg checks here? As we already do in `Service.ts` ?

lib/cli/commands/walletwithdraw.ts
rsercano: True that!

raladev (changes requested):
https://github.com/ExchangeUnion/xud/pull/1846#discussion_r484456741
- [ ] It is not true, destination is always required) (maybe @sangaman confused it with closechannel ETH - for closechannel command with eth pair peer is not required)

sangaman:
>     * [ ]  It is not true, destination is always required) (maybe @sangaman confused it with closechannel ETH - for closechannel command with eth pair peer is not required)

Yes you're exactly correct, that's the mistake I made. So we *should* make `destination` required and a `positional` rather than `option` - my bad for the confusion.

lib/cli/utils.ts
sangaman: Do you mean move all argument checks from the server to the cli? That's a major refactor and also would mean that other consumers of the xud API would have to do their own arg checking. I think it makes sense to leave the majority of these checks on the server side, and have the cli tool just ensure that it is using the correct types for all the parameters (we'd get ugly grpc errors otherwise). This is sort of a special case where we're trying to have command line arg assign values to two separate request parameters (`all` and `amount`).

kilrau:
Ok. Clear now? @rsercano

lib/cli/utils.ts
rsercano: Alright gotcha!

rsercano:
Yes it's and fixed! @kilrau I'm just not sure if I can keep addpair arg check within this PR: https://github.com/ExchangeUnion/xud/pull/1846/files#diff-991abee658202c7a1152694a87d47368R30

raladev (approved):

raladev (dismissed):

kilrau:
> Yes it's and fixed! @kilrau I'm just not sure if I can keep addpair arg check within this PR: https://github.com/ExchangeUnion/xud/pull/1846/files#diff-991abee658202c7a1152694a87d47368R30

Not sure if that check makes sense. @sangaman ?

sangaman (approved):
Looks good! I think we just want to squash the commits.

Edit: Looks like they need to be signed as well. @rsercano see if you can get these squashed and signed if not I can take care of that tomorrow.

rsercano:
I'm going to create a new PR, sorry for breaking this one

rsercano:
New one: #1889

PR #1857, opened 2020-09-01
feat(connext): unlock expired transfer apps

### LINKED ISSUE #1849: xud should unlock expired connext transfer apps
1. maker sends connext payment
2. taker never claims it
3. transfer expires
->
when xud detects that the transfer has expired we should hit the `/hashlock-transfer` endpoint with the payment id in order to reclaim funds back into our channel balance (from the transfer app).

We don't need to store the payment id in the database because we can derive it from from the assetId and hash of the transfer.
```
const paymentId = soliditySha256(["address", "bytes32"], [params.assetId, lockHash]);
```

ghost:
This PR adds logic to attempt to unlock expired connext transfer apps when receiving `EXPIRED` response from `/hashlock-status`.

Closes https://github.com/ExchangeUnion/xud/issues/1849

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -34,8 +34,9 @@ import {
   ConnextDepositResponse,
 } from './types';
 import { parseResponseBody } from '../utils/utils';
-import { Observable, fromEvent, from, combineLatest, defer } from 'rxjs';
-import { take, pluck, timeout, filter, mergeMap } from 'rxjs/operators';
+import { Observable, fromEvent, from, combineLatest, defer, timer } from 'rxjs';
+import { take, pluck, timeout, filter, mergeMap, catchError, mergeMapTo } from 'rxjs/operators';
+import { sha256 } from '@ethersproject/solidity';
 
 const MAX_AMOUNT = Number.MAX_SAFE_INTEGER;
 
@@ -406,10 +407,11 @@ class ConnextClient extends SwapClient {
    */
   public settleInvoice = async (rHash: string, rPreimage: string, currency: string) => {
     this.logger.debug(`settling ${currency} invoice for ${rHash} with preimage ${rPreimage}`);
-    const assetId = this.tokenAddresses.get(currency);
+    const assetId = this.getTokenAddress(currency);
     await this.sendRequest('/hashlock-resolve', 'POST', {
       assetId,
       preImage: `0x${rPreimage}`,
+      paymentId: sha256(['address', 'bytes32'], [assetId, `0x${rHash}`]),
     });
     this.expectedIncomingTransfers.delete(rHash);
   }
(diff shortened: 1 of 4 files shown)


raladev:
1. it would be good to have simtest for this case (if we can control expiry blocks number on the connext side)
2. or at least integration test (mocking hashlock response with/without EXPIRED field)
3. Also, @erkarl  may u check tests status? it seems they are in infinite loop. I  restarted them once and got same result.

lib/connextclient/ConnextClient.ts
```diff
@@ -455,6 +457,20 @@ class ConnextClient extends SwapClient {
             preimage: transferStatusResponse.preImage?.slice(2),
           };
         case 'EXPIRED':
+          const expiredTransferUnlocked$ = defer(() => from(this.settleInvoice(rHash, '', currency))).pipe(
```
sangaman, line 460: I'm a bit confused as to what this is doing (some comments might be good). Why are we "settling" a payment where we are the sender? I know that "invoice" is not exactly a connext term/concept, maybe we should just use the connext call/terminology directly here rather than `settleInvoice` which is a generic SwapClient method.

erkarl: When the connext transfer (HTLC) expires the funds are not automatically returned to the channel balance. Expired transfer apps can be unlocked by both the receiver and sender by calling `/hashlock-resolve` with the `paymentId`.

In this PR the `settleInvoice` function has been modified to always include the `paymentId` in the body of the request in case the transfer has been expired.

It's worth noting that the node also periodically cleans up the expired transfer apps, but from UX perspective it's better if this happens from xud side. This ensures that the funds locked in the transfer app are available to the user sooner.

I've added a short comment above the `settleInvoice` call.

> maybe we should just use the connext call/terminology directly here rather than settleInvoice which is a generic SwapClient method

Are you suggesting to create a wrapper function here that internally calls `settleInvoice`?

erkarl:
>     2. or at least integration test (mocking hashlock response with/without EXPIRED field)

Integration test already exists in this PR.
 
>     3. Also, @erkarl  may u check tests status? it seems they are in infinite loop. I  restarted them once and got same result.

Tests should now pass.

erkarl:
This PR requires us to update the connext client to 1.3.3 and node to latest.

lib/connextclient/ConnextClient.ts
sangaman: OK, thanks for the explanation this makes sense. I think it would be great to put a brief summary of this in a line or two of comments right above this call.

I think calling `settleInvoice` here with a blank preimage is still a bit odd, I would just go ahead and call `/hashlock-resolve` directly like so:

```
await this.sendRequest('/hashlock-resolve', 'POST', {
  assetId,
  paymentId: sha256(['address', 'bytes32'], [assetId, `0x${rHash}`]),
});
```

And the settleInvoice call can be left as it currently is unless you think the payment id is useful to include even when we have the preimage?

erkarl: Updated.

> And the settleInvoice call can be left as it currently is unless you think the payment id is useful to include even when we have the preimage?

Yep, removed payment id from settleInvoice.

sangaman (approved):
Looks great, thank you.

PR #1859, opened 2020-09-01
chore(connext): add integration tests for withdraw

ghost:


sangaman (approved):
Looks good, I think the commit can be a `test` rather than `chore` though for the message.

erkarl:
> Looks good, I think the commit can be a test rather than chore though for the message.

Oops, good point. Was about to change this, but this got merged.

raladev:
> > Looks good, I think the commit can be a test rather than chore though for the message.
> 
> Oops, good point. Was about to change this, but this got merged.

I changed commit message before merging

erkarl:
> > > Looks good, I think the commit can be a test rather than chore though for the message.
> > 
> > 
> > Oops, good point. Was about to change this, but this got merged.
> 
> I changed commit message before merging

Double oops, thanks.

PR #1867, opened 2020-09-04
Open/close channel txid show instead of success

### LINKED ISSUE #1860: Show txid instead of `success`
### How it is

Typical flow of a connext channel deposit:
```
mainnet > openchannel usdt 100
success
```

Connext client submits the txid in the response, let's display it just like we already do for other on-chain operations:
```
{
tx id: "123"
}
```

rsercano:
attempts to resolve #1860 

image
image

- Unfortunately I couldn't test connext even I implemented open channel response for txhash, because my connext somehow returns 500 always :(

- Connext closechannel (withdraw) doesn't return a tx hash, so I returned an empty array and added a TODO to this part, let me know if that's not correct.

- Finally lnd swap client doesn't return tx id for me instead it returns https://github.com/lightningnetwork/lnd/blob/3ae46d81f4a2edad06ef778b2940d9b06386d93b/lnrpc/rpc.proto#L779 so I used it and converted it to object and then to string (https://github.com/ExchangeUnion/xud/compare/openclosechannel-txid-show?expand=1#diff-f1d7503e017b27e9c93329057cd73f05R784), not sure if that's the correct way, please let me know. But no worries if it returns tx id string I print it instead of converting bytes.


raladev:
- [ ] it would be good to have one naming for all commands (transactionId/transactionIdsList)

- [ ] @erkarl can we return txid for closechannel eth?

- [ ] txid from openchannel definitely does not look like valid txid, something wrong with convertion

```
openchannel
- LTC

{
  "transactionId": "ZcGmT7YkNLs/DkwjI9nl3KOtGNNnEjdi4WWzX55Lebg="
}

- BTC

{
  "transactionId": "4opi3rWGsgQXQBDvdKSymI0NCIdUasC099B6CMViRk0="
}


- ETH

{
  "transactionId": "0x985e0889d3abc577b312198d611894cd7986c8a5e1aa3abbbfcc0bd86b801408"
}

- USDT

simnet > openchannel USDT 5

{
  "transactionId": "0x62854e0db94ab99d9b2fb9710bc93156f479f7b70c25a618abf4966371ae9e7c"
}


withdraw
- LTC

{
  "transactionId": "821f3baa82572e6e1cd891f4c3f4430c64d111ca0d0f37345bc106af9d657b06"
}


- BTC

{
  "transactionId": "8c88bdf19b6e7cf9cb1b83e04f27eb5f8f8e8de20a9e99a49e79adfcfd1cec24"
}

- ETH

{
  "transactionId": "0x7cd0ebef1d5b2b93d1e75c62d4cb6fe5ec7be66d777174b8f8a0cd0fb386a986"
}


- USDT

{
  "transactionId": "0x11886ceef823e4a591f5c9214d19beb3ce3660c504286359608f40585743651e"
}



closechannel
- LTC 

{
  "transactionIdsList": [
    "0be2952abb5594fe65507f17015b91a66e702f3711e456cd94746dcde84a2b8d"
  ]
}


- BTC

{
  "transactionIdsList": [
    "912c16d107a51f2923fbc1ba21c282bb9a69e7df9dbde284aa4cb74dd7551003"
  ]
}


- ETH 

simnet > closechannel ETH

{
  "transactionIdsList": []
}


- USDT

{
  "transactionIdsList": []
}

```

raladev (changes requested):
Above

rsercano:
> it would be good to have one naming for all commands (transactionId/transactionIdsList)
Sure but, closechannel returns a list of transactions since it closes multiple channels.

>  txid from openchannel definitely does not look like valid txid, something wrong with convertion
Correct but I'm lost here, a little help would be awesome as I mentioned in the third element in my PR description. @raladev @sangaman

kilrau:
>  it would be good to have one naming for all commands (transactionId/transactionIdsList)

Imo fine as is.

erkarl:
> @erkarl can we return txid for closechannel eth?

Yes, we can and should.

rsercano:
Let me know when I can implement it please @erkarl, also @sangaman could you please help me with showing txid for lnd grpc openchannel call, I'm really confused there.

proto/xudrpc.proto
```diff
@@ -400,7 +400,10 @@ message CloseChannelRequest {
   uint64 amount = 5 [json_name = "amount"];
 }
 
-message CloseChannelResponse {}
+message CloseChannelResponse {
+  // The id of the transaction.
```
sangaman, line 404: This should be plural, since it's returning multiple txids. And maybe it should say that it's one transaction per channel close.

lib/swaps/SwapClientManager.ts
```diff
@@ -437,7 +438,7 @@ class SwapClientManager extends EventEmitter {
       amount: pushAmount,
     });
 
-    await swapClient.openChannel({
+    return await swapClient.openChannel({
```
sangaman, line 441: Same here for `return` and `await`.

lib/service/Service.ts
```diff
@@ -248,7 +248,7 @@ class Service {
       remoteIdentifier = peer.getIdentifier(swapClientType, currency);
     }
 
-    await this.swapClientManager.closeChannel({
+    return await this.swapClientManager.closeChannel({
```
sangaman, line 251: Minor nit but we don't need both `return` and `await` here.

test/jest/LndClient.spec.ts
```diff
@@ -5,6 +5,14 @@ import { getValidDeal } from '../utils';
 import { SwapRole } from '../../lib/constants/enums';
 import { ClientStatus } from '../../lib/swaps/SwapClient';
 
+const getOpenChannelSyncResponse = () => {
```
sangaman, line 8: Also minor but does this need to be a function? It can just be an object with `hasFundingTxidStr` and `getFundingTxidStr` I think, it confused me a bit as currently used.

proto/xudrpc.proto
```diff
@@ -579,7 +582,10 @@ message OpenChannelRequest {
   uint64 push_amount = 4 [json_name = "push_amount"];
 }
 
-message OpenChannelResponse {}
+message OpenChannelResponse {
+  // The id of the transaction.
```
sangaman, line 586: How about "the id of the transaction that opened the channel"

lib/lndclient/LndClient.ts
```diff
@@ -780,7 +780,8 @@ class LndClient extends SwapClient {
       await this.connectPeerAddresses(uris);
     }
 
-    await this.openChannelSync(remoteIdentifier, units, pushUnits);
+    const openResponse = await this.openChannelSync(remoteIdentifier, units, pushUnits);
+    return openResponse.hasFundingTxidStr() ? openResponse.getFundingTxidStr() : openResponse.toObject().fundingTxidBytes.toString();
```
sangaman, line 784: Try `return openResponse.getFundingTxidBytes_asB64();` here to return the txId in the correct format. I remember having some issues with gettting grpc bytes fields to give us a string in the desired format but I believe this is what we want here for txids.

lib/lndclient/LndClient.ts
```diff
@@ -1079,20 +1080,21 @@ class LndClient extends SwapClient {
...
     channels.forEach((channel) => {
       if (channel.getRemotePubkey() === remoteIdentifier) {
         const [fundingTxId, outputIndex] = channel.getChannelPoint().split(':');
         const closePromise = this.closeChannelSync(fundingTxId, Number(outputIndex), force);
         closePromises.push(closePromise);
       }
     });
-    await Promise.all(closePromises);
+
+    return await Promise.all(closePromises);
```
sangaman, line 1092: No need for `await` here as well.

sangaman (changes requested):

proto/xudrpc.proto
rsercano: Changed as `The id of the transaction per channel close.`

test/jest/LndClient.spec.ts
rsercano: I actually wanted to keep same structure as `getSendPaymentSyncResponse` but I rather to keep as object too :)

lib/lndclient/LndClient.ts
rsercano: Thank you! :) This returns txids like `YJpU9BwREyVWLrjB/BEvUTQm7foeKOjXgdhRdMVQKHk=`

rsercano:
Thank you @sangaman for your great help, but still I can't get the correct tx id I suppose, apart from that other reviews are closed, @raladev and @kilrau could you please check from your perspective?

kilrau:
> * return txid for closechannel eth
> * txid from openchannel definitely does not look like valid txid, something wrong with convertion

If these two are done, good from my side

rsercano:
Still having issue with:

- txid from openchannel definitely does not look like valid txid, something wrong with convertion

:(

raladev:
> Still having issue with:
> 
> * txid from openchannel definitely does not look like valid txid, something wrong with convertion
> 
> :(

https://github.com/ExchangeUnion/xud/pull/1867#discussion_r484712660

@rsercano  u need to convert this base64 string to hex format
https://api.lightning.community/?python#openchannelsync

base64 -> bytes -> hex in python:
```
    a = 'ZcGmT7YkNLs/DkwjI9nl3KOtGNNnEjdi4WWzX55Lebg='
    b = base64.b64decode(a)# b = unreadable bytes
    result = b.hex()  #result = 65c1a64fb62434bb3f0e4c2323d9e5dca3ad18d367123762e165b35f9e4b79b8
```

lib/lndclient/LndClient.ts
sangaman: Sorry I had a brain fart and forgot that txids are in hex not in base64. We already have a helper function for this. You should be able to do `base64ToHex(openResponse.getFundingTxidBytes_asB64())`.

rsercano:
Thank you both @raladev and @sangaman fixed.

raladev (dismissed):

raladev:
it seems that connext client is returning txid for withdraw(closechannel) call @rsercano 
https://github.com/connext/rest-api-client/blob/master/src/client.ts#L316

kilrau:
> it seems that connext client is returning txid for withdraw(closechannel) call @rsercano
> https://github.com/connext/rest-api-client/blob/master/src/client.ts#L316

Oh yes, that should work in this PR if not the case yet.

lib/lndclient/LndClient.ts
```diff
@@ -780,7 +780,8 @@ class LndClient extends SwapClient {
       await this.connectPeerAddresses(uris);
     }
 
-    await this.openChannelSync(remoteIdentifier, units, pushUnits);
+    const openResponse = await this.openChannelSync(remoteIdentifier, units, pushUnits);
+    return openResponse.hasFundingTxidStr() ? openResponse.getFundingTxidStr() : base64ToHex(openResponse.getFundingTxidBytes_asB64());
```
sangaman, line 784: Do we need this ternary statement here? I think just `return base64ToHex(openResponse.getFundingTxidBytes_asB64())` is enough to keep things simple. Might be worth tidying up but I don't think it's necessarily worth holding up the PR.

rsercano: According to the documentation this's the field that actually should have transaction id, so I thought it would worth to check it with this statement.

rsercano:
> > it seems that connext client is returning txid for withdraw(closechannel) call @rsercano
> > https://github.com/connext/rest-api-client/blob/master/src/client.ts#L316
> 
> Oh yes, that should work in this PR if not the case yet.

Yeap checking now @raladev @kilrau

rsercano:
Implemented close channel txhash too for connext @kilrau @raladev 

image

But would you please squash and merge my commits for this one @sangaman I literally suck at it, and your help would be really appreciated..

raladev (approved):

lib/lndclient/LndClient.ts
sangaman: Which documentation?

sangaman:
> But would you please squash and merge my commits for this one @sangaman 

Done. 

Btw in spots like this the trick I've learned is to do `git merge origin/master` and then `git reset --soft origin/master` so that you're left with the same commit history as master and just the uncommitted changes from the PR/branch. Then you can add that as a single commit.

Are you able to run one more quick check @raladev ? I can give it a quick manual test too later today.

raladev (approved):

sangaman (approved):

rsercano:
Thank you @sangaman ☺️

PR #1871, opened 2020-09-04
feat(p2p): don't log empty order packets

sangaman:
This skips the log statement that logs when a peer sends us zero orders for a given trading pair, as this is more or less a non-event and is currently logged fairly frequently. It also lowers the level of logging when we do get orders from `verbose` to `debug` as it is a common and not particularly important event.

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -798,14 +798,17 @@ class Pool extends EventEmitter {
       }
       case PacketType.Orders: {
         const receivedOrders = (packet as packets.OrdersPacket).body!;
-        this.logger.verbose(`received ${receivedOrders.length} orders from ${peer.label}`);
-        receivedOrders.forEach((order) => {
-          if (peer.isPairActive(order.pairId)) {
-            this.emit('packet.order', { ...order, peerPubKey: peer.nodePubKey! });
-          } else {
-            this.logger.debug(`received order ${order.id} for deactivated trading pair`);
-          }
-        });
+        if (receivedOrders.length > 0) {
+          this.logger.debug(`received ${receivedOrders.length} orders from ${peer.label}`);
+          receivedOrders.forEach((order) => {
+            if (peer.isPairActive(order.pairId)) {
+              this.emit('packet.order', { ...order, peerPubKey: peer.nodePubKey! });
+            } else {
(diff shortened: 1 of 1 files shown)


erkarl (approved):

raladev:
Not sure that we should remove receiving if sending is still logging.

raladev (commented):
Above

kilrau:
Agree, that we should ommit logs on the sending side in the case of "no orders" too.

> It would be better to remove sending for such case or just do nothing.

Not sure I understand. The sender still needs to send the empty order packet, otherwise the receiver doesn't know the sender doesn't have any orders for that pair.

raladev:
> Not sure I understand. The sender still needs to send the empty order packet, otherwise the receiver doesn't know the sender doesn't have any orders for that pair.

So, it looks pretty strange - "Lets not log 0 order packets because there is too much such strings  in logs, but we still will log non-empty order packets"

U said that sending empty order packet is required action, then it definitely SHOULD be logged, there is no difference between empty order packet and non-empty, both are required. And if we will log sending, we need to log receiving too. (Sending it is an action that should be logged and logging should not be based on action parameters (2 or 0 orders).)

Then maybe lets just move this logging to trace level and will not omit it?

sangaman:
I agree with not logging empty order packets when they're being sent, I wasn't aware that was being done but I'll go check and make sure again.

> > Not sure I understand. The sender still needs to send the empty order packet, otherwise the receiver doesn't know the sender doesn't have any orders for that pair.
> 
> So, it looks pretty strange - "Lets not log 0 order packets because there is too much such strings in logs, but we still will log non-empty order packets"
> 
> U said that sending empty order packet is required action, then it definitely SHOULD be logged, there is no difference between empty order packet and non-empty, both are required. And if we will log sending, we need to log receiving too. (Sending it is an action that should be logged and logging should not be based on action parameters (2 or 0 orders).)
> 
> Then maybe lets just move this logging to trace level and will not omit it?

So we do indeed need the empty orders packet on a p2p level, as kilian said. We already log every single packet on trace level (here https://github.com/ExchangeUnion/xud/blob/master/lib/p2p/Peer.ts#L797), so when we really want to track every single action we'll be able to see that. I was just thinking that we don't need additional logging for empty order packets, particularly at the `verbose` level, which is what I was trying to change here.

raladev (approved):

PR #1872, opened 2020-09-05
fix: getbalance defaults to 0

### LINKED ISSUE #1420: getbalance defaulting to 0
Restoring without backup files, with a mnemonic only (which has on-chain tBTC & tLTC funds) ends up with an empty wallet.

rsercano:
attempts to resolve #1420 

- I added a new indicator to ChannelBalance named noChannels that indicates there was no channels to calculate balance, so while printing out channel balance, xud checks this field and formats as below. 

image


kilrau:
I find `nochannels` confusing. What this PR should do: return `?` if `getbalance` couldn't fetch lndbtc/lndltc/connext balance within the timeout. Don't display 0. See https://github.com/ExchangeUnion/xud/issues/1420#issuecomment-594096317

kilrau:
Could you help @rsercano here @sangaman ?

raladev:
> `nels` confusing. What this PR should do: return `?` if `getbalance` couldn't fetch l

I suggest to change ? to 'timeout' or to something more obvious, because ? a little bit scary and confused :)

kilrau:
> I suggest to change ? to 'timeout' or to something more obvious, because ? a little bit scary and confused :)

Agree. Let's go with `timeout` @rsercano

rsercano:
Yes of course, but the case is if we don't have a channel what are we going to show? I mean timeout for timed out requests but what about no channel case?

kilrau:
> Yes of course, but the case is if we don't have a channel what are we going to show? I mean timeout for timed out requests but what about no channel case?

no channel = no balance = `0`

rsercano:
Sorry for misunderstanding I just got the point, fixing now.

rsercano:
image

This's an example when lndltc docker image stopped.

kilrau:
> image
> 
> This's an example when lndltc docker image stopped.

:+1: Suggestion: maybe one simple `timeout` without any repetitive `(timeout | timeout)` / `(timeout)` is better.

rsercano:
Yes of course :) 
image

raladev (changes requested):
- [ ] does not work for connext
[Screenshot from 2020-09-09 16-07-57](https://user-images.githubusercontent.com/29906866/92602537-eea54680-f2b6-11ea-929a-b0d41499e1a4.png)
- [ ] first call after lnd killing returns nothing for required currency and only for next calls i got timout output
[Screenshot from 2020-09-09 15-51-28](https://user-images.githubusercontent.com/29906866/92602581-fe248f80-f2b6-11ea-81eb-160c840fadc2.png)
- Why first call after lnd killing took some time (10-15 sec), but for next calls 1-2 sec is required?

kilrau:
> Why first call after lnd killing took some time (10-15 sec), but for next calls 1-2 sec is required?

Hmm, that might be legit. Test if your lnd `channelbalance` call also takes significantly longer first time after restarting lnd in this environment.

sangaman (changes requested):
Sorry for being a bit late to the party here. I find the "timeout" output to be a bit unnatural, and changing all the balance fields on the grpc response from unsigned to signed ints is technically a breaking change for v1.0.0 software and seems a bit extreme.

My idea here originally was to simply return nothing when we can't fetch the balance for a currency. After digging into this code again, I'm wondering whether the original issue actually still exists in master. The relevant code was modified in #1736 which happened well after the original issue was opened, and I believe that may have fixed the main issue here as well. @kilrau or @raladev can you please try to reproduce on master? I suspect you simply won't see any balance for the currency whose swap client is not online or not responsive, that's what I'm seeing with some quick tests locally when swap clients aren't online.

The one thing we could do is add an actual timeout to handle cases where a swap client is hanging or stuck - in other words each currency has 5 seconds or so to fetch a balance otherwise it returns nothing. We could definitely do that, but it's not actually being done here, there's no time limit on the `channelBalance()`/`walletBalance()` calls.

raladev:
> I believe that may have fixed the main issue here as well. @kilrau or @raladev can you please try to reproduce on master? I suspect you simply won't see any balance for the currency whose swap client is not online or not responsive, that's what I'm seeing with some quick tests locally when swap clients aren't online.

Same for me - line disappears if something wrong with l2 client.

rsercano:
> The one thing we could do is add an actual timeout to handle cases where a swap client is hanging or stuck - in other words each currency has 5 seconds or so to fetch a balance otherwise it returns nothing. We could definitely do that, but it's not actually being done here, there's no time limit on the channelBalance()/walletBalance() calls.

If that's what we want I can definitely change the implementation as you described, I thought we shouldn't disappear the currency line in grpc call and that was the issue, could you guys please confirm ?

sangaman:
> If that's what we want I can definitely change the implementation as you described, I thought we shouldn't hide the currency line in grpc call and that was the issue, could you guys please confirm ?

Maybe I am misunderstanding but I thought the main goal was just not to display 0 balance when the balance is unknown, which is already the current behavior.

rsercano:
> > If that's what we want I can definitely change the implementation as you described, I thought we shouldn't hide the currency line in grpc call and that was the issue, could you guys please confirm ?
> 
> Maybe I am misunderstanding but I thought the main goal was just not to display 0 balance when the balance is unknown, which is already the current behavior.

But balance's unknown status is too generic definition to me, there can be several cases related with 0 balance, can't be? On the other hand putting a timeout for balance calls on GRPC call of LND makes sense and I'm going to change this to this way at first

kilrau:
> Maybe I am misunderstanding but I thought the main goal was just not to display 0 balance when the balance is unknown, which is already the current behavior.

That's precisely the goal. Just that this was not the case for me a while back: https://github.com/ExchangeUnion/xud/issues/1420#issuecomment-594096317 . I understand that we are caching balances, but `getbalance` should trigger fetching current balances from lnds and connext. If this fails, my balance is defacto **unknown** and at least the *next* `getbalance` call should show `timeout` or `unknown` until the last lnd/connext balance call was successful again and the cached balance was updated. Could you clarify if this is already the case @sangaman

sangaman:
> That's precisely the goal. Just that this was not the case for me a while back: [#1420 (comment)](https://github.com/ExchangeUnion/xud/issues/1420#issuecomment-594096317) . I understand that we are caching balances, but `getbalance` should trigger fetching current balances from lnds and connext. If this fails, my balance is defacto **unknown** and at least the _next_ `getbalance` call should show `timeout` or `unknown` until the last lnd/connext balance call was successful again and the cached balance was updated. Could you clarify if this is already the case @sangaman

Yes, if we can't get the balance for a given swap client at the time of the xud `GetBalance` rpc call then the response won't include any balance for that swap client. This is fairly easy to test by just shutting down a swap client and then calling getbalance without any currency specified. Apologies for the confusion over #1420, it should have been closed.

rsercano:
> > That's precisely the goal. Just that this was not the case for me a while back: [#1420 (comment)](https://github.com/ExchangeUnion/xud/issues/1420#issuecomment-594096317) . I understand that we are caching balances, but `getbalance` should trigger fetching current balances from lnds and connext. If this fails, my balance is defacto **unknown** and at least the _next_ `getbalance` call should show `timeout` or `unknown` until the last lnd/connext balance call was successful again and the cached balance was updated. Could you clarify if this is already the case @sangaman
> 
> Yes, if we can't get the balance for a given swap client at the time of the xud `GetBalance` rpc call then the response won't include any balance for that swap client. This is fairly easy to test by just shutting down a swap client and then calling getbalance without any currency specified. Apologies for the confusion over #1420, it should have been closed.

Then this can be closed too

kilrau:
Yes.

Issue #1881, opened 2020-09-09
incorrect filter for limit flag of listorders command


raladev:
Actual result:
limit flag cuts part of orderbook from the top - orders that should be used for swap will be cutted. Orders form bottom of the order book will be displayed

![Screenshot from 2020-09-09 17-34-57](https://user-images.githubusercontent.com/29906866/92636418-fd542380-f2df-11ea-810d-b2c70e0befdc.png)

Expected result:
top of orderbook should be displayed

PR #1883, opened 2020-09-09
fix: listorders limit displays first orders instead of last

raladev:
Fix of #1881, but need ACK from @kilrau that it is a bug :)

Screenshot from 2020-09-09 22-08-37

### DIFF
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -471,8 +471,8 @@ class Service {
       sellArray = sortOrders(sellArray, false);
 
       if (limit > 0) {
-        buyArray = buyArray.slice(0, limit);
-        sellArray = sellArray.slice(0, limit);
+        buyArray = buyArray.slice(-limit);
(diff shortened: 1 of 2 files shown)


kilrau:
ACK, #1881 is a bug! `--limit` should list the X **best** orders, currently it lists the X **worst** orders.

sangaman (approved):
The root cause here as I see it is that we're sorting our orders backwards from what I would expect. The buy orders *increase* in price while the sell orders *decrease* in price. I'd expect the first element in each array (buy and sell) to be the "best" price. Therefore, taking the first n orders from the array indeed gives us the n worst orders instead of n best. Our cli tool is designed with this "backwards" order in mind, which is why they display properly.

This fix looks good, but I'd be inclined to revisit this issue and make the order sorting go from best to worst (and change xucli accordingly). I can see it causing more confusion down the road, although it's a breaking change so we'd need to make sure it doesn't cause any issues with arby.

PR #1864, opened 2020-09-03
fix: added check for peer's pairs on order invalidation

rsercano:
attempts to fix #1530 

Not sure if there's any other order event to check, also it would be great if you can tell me if I have to check pair's active status before emitting the event.

PR #1884, opened 2020-09-10
fix: don't activate unsupported pairs with peers

sangaman:
This ensures that, even when not running in strict mode, we only activate trading pairs with peers when we support both currencies of the pair locally. Previously we would activate every trading pair, like
BTC/LTC when we don't have a swap client for LTC, resulting in us requesting and receiving order updates for trading pairs that are irrelevent to us.

Related https://github.com/ExchangeUnion/xud/pull/1864#issuecomment-687140868:

> after adding new pair, the peer with new pair sends packet update to his peers. And after that these peers send GetOrders packet to this peer even if they have no this pair. It would be good to not ask orders if we have no this new pair.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -994,6 +994,11 @@ class OrderBook extends EventEmitter {
       if (peer.isPairActive(pairId)) {
         return false; // don't verify a pair that is already active
       }
+      if (!this.tradingPairs.has(pairId)) {
+        // if we don't support the trading pair locally, then we don't want to verify or activate it
+        return false;
+      }
+
       const [baseCurrency, quoteCurrency] = pairId.split('/');
       const peerCurrenciesEnabled = !peer.disabledCurrencies.has(baseCurrency)
         && !peer.disabledCurrencies.has(quoteCurrency);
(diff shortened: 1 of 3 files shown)


rsercano:
> This ensures that, even when not running in strict mode, we only activate trading pairs with peers when we support both currencies of the pair locally. Previously we would activate every trading pair, like
> BTC/LTC when we don't have a swap client for LTC, resulting in us requesting and receiving order updates for trading pairs that are irrelevent to us.
> 
> Related [#1864 (comment)](https://github.com/ExchangeUnion/xud/pull/1864#issuecomment-687140868):
> 
> > after adding new pair, the peer with new pair sends packet update to his peers. And after that these peers send GetOrders packet to this peer even if they have no this pair. It would be good to not ask orders if we have no this new pair.

That's a more intuitive way than the first try that I made, but why don't we return just a simple `bool` for `has` method?

sangaman:
> That's a more intuitive way than the first try that I made, but why don't we return just a simple bool for has method?

Good catch, this was just an oversight. I meant to change it to a bool (copied from the `get`) method but just missed it. I'll update the PR.

raladev:
- [ ] still see GetOrders packet
Screenshot from 2020-09-10 17-42-27
Screenshot from 2020-09-10 17-41-35

raladev (changes requested):
Above

sangaman:
>     * [ ]  still see GetOrders packet

Is the issue here that you're sending a GetOrders packet for USDT/BTC when locally you only support BTC/USDT? That actually looks like a slightly different issue from what I originally set out to fix, but it would make sense to fix that as well. Here we support both currencies of the trading pair but we don't actually support the trading pair itself, which does not seem to be a situation that's currently handled properly by the code. I'll push a fix in a separate commit.

sangaman:
I actually went ahead and amended the existing commit since it's so closely related. @raladev could you please try this branch again and see if you're sending the expected `GetOrders` packets? Thanks.

raladev (approved):

sangaman:
I fixed the tests (that were expecting even the currencies we don't locally support to get activated), could you review one more time real quick @raladev?

raladev (approved):

PR #1893, opened 2020-09-11
fix(connext): remove BigInt to avoid precision loss

### LINKED ISSUE #1887: [Connext]  walletdeposit returns 500 server error
### Background

1. Successful xud setup, using all-default configuration.
2. Successful `walletdeposit eth`.   0.8 ETH appear as expected in the Wallet Balance column of `getbalance`.
3. Attempts to `openchannel` fail with the same error:
```
mainnet > openchannel ETH 0.8
Error: 2 UNKNOWN: failed to open ETH channel for 80000000: connext server error 500: Internal Server Error
```

### Your environment

* version of `xud`: 1.0.1-d04e4bd5    
* which operating system (`uname -a` on *Nix): `Linux whi1 4.15.0-117-generic #118-Ubuntu SMP Fri Sep 4 20:02:41 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`
* version of `btcd`, `bitcoind`, `ltcd`, `litecoind`, `geth`, `parity`, `connext`, `lnd` or other chain backend: all xud defaults for light-node:
connext 1.3.3
LND-BTC 0.11.0-beta commit=v0.11.0-beta 
LND-LTC 0.11.0-beta.rc1 commit=v0.11.0-beta.rc1 

* any other relevant environment details:

### Steps to reproduce

1. deposit ETH
2. openchannel on ETH

### Expected behaviour

channel opens

### Actual behaviour

500 server error, shown above

ghost:
This removes BigInt conversion from `ConnextClient` since it was causing precision loss when sending requests to Connext API.

EDIT by @kilrau : fixes https://github.com/ExchangeUnion/xud/issues/1887

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -350,7 +350,7 @@ class ConnextClient extends SwapClient {
       let secret;
       if (deal.role === SwapRole.Maker) {
         // we are the maker paying the taker
-        amount = BigInt(deal.takerUnits).toString();
+        amount = deal.takerUnits.toString();
         tokenAddress = this.tokenAddresses.get(deal.takerCurrency)!;
         const executeTransfer = this.executeHashLockTransfer({
           amount,
@@ -368,7 +368,7 @@ class ConnextClient extends SwapClient {
         secret = preimage;
       } else {
         // we are the taker paying the maker
-        amount = BigInt(deal.makerUnits).toString();
+        amount = deal.makerUnits.toString();
         tokenAddress = this.tokenAddresses.get(deal.makerCurrency)!;
         lockTimeout = deal.makerCltvDelta!;
         secret = deal.rPreimage!;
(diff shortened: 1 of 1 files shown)


sangaman (approved):
Do we know why BigInt was causing loss of precision? I couldn't find anything about it and logically it seems like it should work as it was. However, since `units` is already a number type, it's not like we can gain back any precision by converting to BigInt either, so the new approach looks like an improvement. If we still have issues with this maybe we need to revisit the amount/unit conversion logic and consider using other types than the native JS number.

erkarl:
> Do we know why BigInt was causing loss of precision? 

I'm not 100% sure, but the steps conversion looked like Number -> BigInt -> String VS Number -> String in this PR. Guess it ties back to how floating point numbersin JavaScript work.

> If we still have issues with this maybe we need to revisit the amount/unit conversion logic and consider using other types than the native JS number.

I'm pretty convinced there are still edge cases that our current implementation does not cover. Ideally, I'd like to see all quantities in/out of xud gRPC layer in string format and removing the amount/unit (satoshis per coin) from internal xud logic.

raladev (approved):

PR #1900, opened 2020-09-15
fix: implemented showing all pairs instead of active ones for listpeers

### LINKED ISSUE #1801: `listpeers` should show all pairs of peer
I just noticed that listpeers shows only the pairs of a peer that are currently active on my local node too. In my example, lndltc is down and ETH/BTC is the only functioning pair I have. All peers in listpeers show ETH/BTC as their pairs only, even though they have many more pairs. I don't think that's what we want, I htink I want to see all supported pairs of a peer there, even if I do not have them.

rsercano:
attempts to resolve #1801 

This should be as simple as this change as per comments on the issue, could you please check @sangaman @raladev @kilrau 
image

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -179,7 +179,7 @@ class Peer extends EventEmitter {
       alias: this.alias,
       nodePubKey: this.nodePubKey,
       inbound: this.inbound,
-      pairs: Array.from(this.activePairs),
+      pairs: this.nodeState ? this.nodeState.pairs : [],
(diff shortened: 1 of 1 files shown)


raladev (approved):

sangaman (approved):
Yep this is a simple change but it should achieve the effect of showing all pairs a peer advertises and not just the ones that are active.

As an aside I wonder if it would make sense to show *whether* each pair is active. If we think that would be desirable we can open a followup issue for an enhancement.

Issue #1576, opened 2020-05-23
Seedutil: derive child seed for swap clients (allow creation any time)


kilrau:
Change lndbtc/lndltc to derive a child seed (or rather the entropy for such) from the seedutil master seed and create with these, just like connext already does.

This allows allows swap clients to be created any time, `xucli create` to be called any time, the new behavior of `xucli create` could be along the lines: `xud` detects connected swap clients which aren't created yet (based on files which only exist if client is created since client can be locked). `xucli create output` will tell user which clients will be created so the user can sanity check if all she wants are listed. If user confirms, xud will create these new swap clients without touching existing ones.

I propose to split this up into two:
1. Derive entropy from seedutil master seed, create lnd swap clients with this (connext already does)
2. UX flow allowing to create a swap client after initial create as a follow-up issue

kilrau:
Bumping prio since this is  a dependency for some other issues, see https://github.com/ExchangeUnion/xud/issues/1815#issuecomment-689555436

PR #1902, opened 2020-09-16
feat(rpc): show connext status on create/restore node

### LINKED ISSUE #1601: Connext: add creation result to `create` output
How it is:
`The following wallets were initialized: LTC, BTC`

How it should be:
`The following wallets were initialized: BTC, LTC,  ETH`

rsercano:
attempts to resolve #1601.

But there's a logical issue here, when we start xud, naturally everything is locked, and we're not able to receive node key. But connext requires node key to be initialized. So while creating or restoring a node **connext's state is not initialized.** 

For this one, I directly took all token addresses available from connext and extracted available currencies from it to show the message, but I think we should show it with a different message, since it's misleading to say it's initialized even we don't know yet. 

@kilrau @sangaman @raladev @erkarl


raladev:
not sure that it is worth to add such explanation into installation flow, because wallet will be initialized after unlocking if all is fine and first unlock is a part of installation.

raladev (dismissed):

erkarl:
Just to confirm. The current implementation displays:
```
The following wallets were initialized: BTC, LTC, ERC20(ETH, USDT, DAI)
```
but the corresponding issue requested:
```
The following wallets were initialized: BTC, LTC, ETH
```

Is this OK? WDYT @kilrau @sangaman

erkarl (approved):

kilrau:
Let's change to the more generic `BTC, LTC, ETH` because `BTC, LTC, ERC20(ETH, USDT, DAI)` is wrong the moment we add/remove a token.

raladev:
> Let's change to the more generic `BTC, LTC, ETH` because `BTC, LTC, ERC20(ETH, USDT, DAI)` is wrong the moment we add/remove a token.

I a little bit misunderstand why is it wrong? ETH,USDT,DAI is dynamically builded string, so if we add new predefined ERC token to xud it will be in the list.

kilrau:
> I a little bit misunderstand why is it wrong? ETH,USDT,DAI is dynamically builded string, so if we add new predefined ERC token to xud it will be in the list.

Yes, but:
- `The following wallets were initialized: BTC, LTC, ETH` will only be shown once on creation, tokens may change afterwards and the user shouldn't get the feeling only the shown tokens can be used
- `ERC20(ETH` is technically wrong since ETH is not an ERC20 token
- I would argue almost everyone knows that an ETH wallet can deal with all ETH ERC20 tokens - it's just simpler

rsercano:
Changed the output as @kilrau described, but didn't want to change GRPC call, it still shows all supported ERC20 tokens just in case somebody else use it and be on the same page with supported lnds field.

npm-shrinkwrap.json
```diff
@@ -5970,7 +5970,7 @@
       "dependencies": {
         "camelcase-keys": {
           "version": "2.1.0",
-          "resolved": "https://registry.npmjs.org/camelcase-keys/-/camelcase-keys-2.1.0.tgz",
+          "resolved": "http://registry.npmjs.org/camelcase-keys/-/camelcase-keys-2.1.0.tgz",
```
erkarl, line 5973: Why is there a diff here? Seems unnecessary.

proto/xudrpc.proto
```diff
@@ -422,6 +422,8 @@ message CreateNodeResponse {
   repeated string seed_mnemonic = 1;
   // The list of lnd clients that were initialized.
   repeated string initialized_lnds = 2;
+  // The list of connext wallets that were initialized.
+  repeated string initialized_connexts = 3;
```
erkarl, line 426: Think we can change this to a boolean `initialized_connext` since we're not actually using the values in this array for the CLI.

erkarl (changes requested):

kilrau (changes requested):
> but didn't want to change GRPC call, it still shows all supported ERC20 tokens just in case somebody else use it and be on the same page with supported lnds field.

The grpc response should be changed too

rsercano:
> > but didn't want to change GRPC call, it still shows all supported ERC20 tokens just in case somebody else use it and be on the same page with supported lnds field.
> 
> The grpc response should be changed too

Alright I'll fix as per @erkarl's review soon

npm-shrinkwrap.json
rsercano: Correct, my bad

rsercano:
I also added go folder in simulation tests into gitignore since it caused a lot of trouble to me and seems like shouldn't be in the version control.

kilrau:
R4R? If so, please re-request reviews

lib/cli/commands/create.ts
```diff
@@ -28,6 +28,10 @@ const formatOutput = (response: CreateNodeResponse.AsObject) => {
       walletInitializedMessage += response.initializedLndsList.join(', ');
     }
 
+    if (response.initializedConnext) {
+      walletInitializedMessage += ', ETH';
```
raladev, line 32: - [ ] here is may be the situation when lnd clients already initialized (for example, when native xud is used). 
it is not common but still the case. In this case BTC and LTC will be missed in the string and as a result we will get ', ETH'. it would be good to handle this case.

raladev (changes requested):
- [ ] here is may be the situation when lnd clients already initialized (for example, when native xud is used). 
it is not common but still the case. So if i inderstood correctly

lib/cli/commands/create.ts
rsercano: Ah very good catch, sorry for this I'll handle it!

.gitignore
```diff
@@ -78,6 +78,8 @@ seedutil/seedutil
 test/simulation/temp
 test/simulation/docker-xud/xud
 test/simulation/docker-xud/custom-xud
+test/simulation/go
+temp_file
```
sangaman, line 82: temp_file should be removed. Also, I don't have a `test/simulation/go` folder in my repo, I think you should be able to simply delete it and remove it from here as well.

lib/cli/commands/restore.ts
```diff
@@ -32,6 +32,10 @@ const formatOutput = (response: RestoreNodeResponse.AsObject) => {
     walletRestoredMessage += response.restoredLndsList.join(', ');
   }
 
+  if (response.restoredConnext) {
+    walletRestoredMessage += ', ETH';
```
sangaman, line 36: Same concern here with the leading comma re: raladev's comment above.

proto/xudrpc.proto
```diff
@@ -422,6 +422,8 @@ message CreateNodeResponse {
   repeated string seed_mnemonic = 1;
   // The list of lnd clients that were initialized.
   repeated string initialized_lnds = 2;
+  // The indicator that shows if connext wallets were initialized.
```
sangaman, line 425: "Wallets" sounds a bit off here since there is really only one connext wallet, I'd just say "Whether the connext wallet was initialized." and have a similar comment for the restored node response below.

sangaman (changes requested):
This is currently *always* including Connext on the list of initalized/restored wallets, even when Connext is offline or malfunctioning (as long as it's not disabled in the config0. Unfortunately the way we handle initializing Connext is different from lnd since it requires us to pass a seed that is derived from the unencrypted nodekey each time we start xud, and Connext is currently initialized separately after the routine for unlocking the xud node and its swap clients. That part would probably need some significant refactoring for this feature to really function as I would expect.

We could still merge this as a sort of placeholder, but I think a follow-up issue would make sense in that case.

rsercano:
Thanks for your time @sangaman @raladev for reviewing this, I fixed the points could you please take a look at again?

kilrau:
> We could still merge this as a sort of placeholder, but I think a follow-up issue would make sense in that case.

Since we want lnds to be initialized the same way as connext soon too (derive child seed from seedutil), I agree that modifying where/when we display which wallets were initialized to be a follow-up issue *after* https://github.com/ExchangeUnion/xud/issues/1576 is done.

raladev (approved):

sangaman (approved):

kilrau (approved):

Issue #1797, opened 2020-08-06
Move to BigNumber (get rid of 8 decimal limitation within xud)


kilrau:
By @erkarl :
> In general I think we should consider doing the refactor from the satoshis per coin unit to something like BigNumber soon. It'll reduce the complexity of having to deal with amounts and units in the codebase and would allow us to send the amount in string format over gRPC.

sangaman:
It's a major refactor. The idea then is to use "whole coins" everywhere internally for quantity?

erkarl:
> It's a major refactor. 

For sure. Definitely not something we'd want to do in the next few weeks.

> The idea then is to use "whole coins" everywhere internally for quantity?

Exactly.

PR #1904, opened 2020-09-17
feat: bigint units & decimalplaces from db

### LINKED ISSUE #1054: UNITS_PER_CURRENCY to be fetched from db

### LINKED ISSUE #1888: Unable to deposit newly-added ERC20 currency, without environment restart
### Background

1. Successful light client xud setup.
2. Successful walletdeposit ETH.
3. Successfully added DAI via addcurrency.
4. Successfully added ETH/DAI pair via addpair.
5. `walletdeposit DAI` failed:
```
mainnet > walletdeposit dai
Error: 9 FAILED_PRECONDITION: swapClient for currency DAI not found
```

Describe your issue here.

### Your environment

- version of xud: 1.0.1-d04e4bd5
- which operating system (uname -a on *Nix): Linux whi1 4.15.0-117-generic #118-Ubuntu SMP Fri Sep 4 20:02:41 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
- version of btcd, bitcoind, ltcd, litecoind, geth, parity, connext, lnd or other chain backend: all xud defaults for light-node:
connext 1.3.3
LND-BTC 0.11.0-beta commit=v0.11.0-beta
LND-LTC 0.11.0-beta.rc1 commit=v0.11.0-beta.rc1

any other relevant environment details:

### Steps to reproduce

see above.

### Expected behaviour

DAI deposit address generated.

### Actual behaviour

error shown above.

sangaman:
This PR makes two significant changes.

1. It uses the new `bigint` type - supported as of node LTS v14 - for internal representations of the base "units" of any currencies/tokens. ETH wei, having 10^18 units per ether, can exceed the capacity of javascript's `number` type and lead to unexpected rounding or scientific notation when converting to string that has required hacky workarounds. Using `bigint` gives us native integer support for base tokens for all tokens, regardless of how many decimal places they can be split into.

2. It reads the aforementioned decimal places from the database for every token. Previously, we hardcoded the units per tokens for a handful of known ERC20 tokens. Adding other currencies to be supported by xud would have required a code change. Now, new tokens can be supported with changes to the database via rpc calls, without requiring code updates.

Bigint requires ES2020. ES2020 is only supported by node v14+, which has some known incompatibilities and deprecations with existing xud dependencies. This updates any of these dependencies which had newer versions published without other breaking changes.

Closes #1054. Closes #1888. Related to #1797.


sangaman:
> Looking at this again I'm wondering why we'd want do this instead of #1797 right away? @sangaman

That is proposing a different approach and is a bigger refactor, changing the units we use for `quantity` from satoshis to whole coins. I think using bigint like I'm doing here accomplishes the same thing in terms of ensuring we never lose precision but still keeping the internal definition of `quantity` the same. v14 LTS starts in a week though so I'll revisit this soon.

kilrau:
Should this close https://github.com/ExchangeUnion/xud/issues/1972 too? If so, mark it please as such @sangaman

rsercano:
I'll try to fix tests as Daniel already made the implementation, not sure if it's done yet tho, but my first prio is to fix tests, so what I've done so far:

- Added a new utility function to serialize bigint while logging with `JSON.stringify` since it doesn't know how serialize bigint.
- Fixed `test/integration` tests.
- Moved `unitConverter` in Orderbook to class variable to make it changeable in test code. 

I'll continue with other tests.

rsercano:
I have a question here please:

- BigInt only accepts integer and not decimal numbers, so how it resolves the situations like we need to buy/send decimal (0.0011)?

image

kilrau:
Thanks for working on this, Sercan! Any code-related questions, you can ping @sangaman here and once you consider it ready for testing, don't forget to move it out of draft state and ping @raladev

rsercano:
I have a question regarding amount - unit conversion:

- How can we assign **amount** to bigint since it can be in decimals? (Currently this causes issues on tests) @sangaman @raladev @erkarl 

So let's say I try to withdraw 0.003 BTC, when we try to convert this to units it's still 0.003 BTC which is not an **integer**. But when I try to withdraw 1999999 ETH, it's actually (1999999 * 10 ** 10) = **19999990000000000** in smallest unit and it exceeds the number and have to be a `[bigint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)`.

rsercano:
> I have a question regarding amount - unit conversion:
> 
> * How can we assign **amount** to bigint since it can be in decimals? (Currently this causes issues on tests) @sangaman @erkarl
> 
> So let's say I try to withdraw 0.003 BTC, when we try to convert this to units it's still 0.003 BTC which is not an **integer**. But when I try to withdraw 1999999 ETH, it's actually (1999999 * 10 ** 10) = **19999990000000000** in smallest unit and it exceeds the number and have to be a [bigint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)

Taking this into consideration:

- I changes these lines: https://github.com/ExchangeUnion/xud/pull/1904/files#diff-fc46ed01d1107d5b3a46d86e01f75ee730923fa732fdb8adbafcf92f1edd1f5aR47 and https://github.com/ExchangeUnion/xud/pull/1904/files#diff-fc46ed01d1107d5b3a46d86e01f75ee730923fa732fdb8adbafcf92f1edd1f5aR49 to convert after multiply otherwise we can't convert `amount` to BigInt since amount can be in decimals.
But this naturally causes this: https://github.com/ExchangeUnion/xud/pull/1904/files#diff-8d89d0d687ecb017c9deab79ce3cd4d038a79ecec0af6f128af7285100c1763bR535 not to be `999999999999999992000000n` because of `BigInt(Number/Number)`  instead of `BigInt/BigInt`. Please have a look @sangaman 

- I fixed the entire tests except simulation.
- I moved unit tests for `Swaps['calculateMakerTakerAmounts']` into integration tests by using `Swaps` instance since it's no longer static.
- I commented out an obsolete integration test snapshot, not sure how it was working before: https://github.com/ExchangeUnion/xud/pull/1904/commits/927b27a33a784dfa58db31ad81ed956fa82bea38#diff-e2689a2ad2f943ae0980b40c1b8126532f221687d645f0cbbe1c962258d34f33R3

rsercano:
I guess I found the issue with simulation tests, with the current implementation we always fetch currencies from DB upfront, and when a new currency is added, it's being ignored. Simulation tests always add currencies after initialization, therefore all related tests are throwing currency decimal not found error, because `UnitConverter` has no idea of it.

sangaman:
> > I have a question regarding amount - unit conversion:
> > 
> > * How can we assign **amount** to bigint since it can be in decimals? (Currently this causes issues on tests) @sangaman @erkarl
> > 
> > So let's say I try to withdraw 0.003 BTC, when we try to convert this to units it's still 0.003 BTC which is not an **integer**. But when I try to withdraw 1999999 ETH, it's actually (1999999 * 10 ** 10) = **19999990000000000** in smallest unit and it exceeds the number and have to be a [bigint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)
> 
> Taking this into consideration:
> 
>     * I changed these lines: [#1904 (files)](https://github.com/ExchangeUnion/xud/pull/1904/files#diff-fc46ed01d1107d5b3a46d86e01f75ee730923fa732fdb8adbafcf92f1edd1f5aR47) and [#1904 (files)](https://github.com/ExchangeUnion/xud/pull/1904/files#diff-fc46ed01d1107d5b3a46d86e01f75ee730923fa732fdb8adbafcf92f1edd1f5aR49) to convert after multiply otherwise we can't convert `amount` to BigInt since amount can be in decimals.
>       But this naturally causes this: [#1904 (files)](https://github.com/ExchangeUnion/xud/pull/1904/files#diff-8d89d0d687ecb017c9deab79ce3cd4d038a79ecec0af6f128af7285100c1763bR535) not to be `999999999999999992000000n` because of `BigInt(Number/Number)`  instead of `BigInt/BigInt`. Please have a look @sangaman
> 
>     * I fixed the entire tests except simulation.
> 
>     * I moved unit tests for `Swaps['calculateMakerTakerAmounts']` into integration tests by using `Swaps` instance since it's no longer static.
> 
>     * I commented out an obsolete integration test snapshot, not sure how it was working before: [927b27a#diff-e2689a2ad2f943ae0980b40c1b8126532f221687d645f0cbbe1c962258d34f33R3](https://github.com/ExchangeUnion/xud/commit/927b27a33a784dfa58db31ad81ed956fa82bea38#diff-e2689a2ad2f943ae0980b40c1b8126532f221687d645f0cbbe1c962258d34f33R3)
> 
>     * Another consideration came to my mind is we're still using number while calling LND grpc calls for such as `openChannelSync`.
> 
> 
> **Therefore**
> 
>     * Need to check simulation tests if working.
> 
>     * Need to check if we can use bigint while calling LND grpc calls.

`amount` is always an integer representing satoshis when used internally and on the grpc layer, so we shouldn't need to do any sort of conversions as in your first bullet point. I'll review your changes when I have some more time.


> I guess I found the issue with simulation tests, with the current implementation we always fetch currencies from DB upfront, and when a new currency is added, it's being ignored. Simulation tests always add currencies after initialization, therefore all related tests are throwing currency decimal not found error, because `UnitConverter` has no idea of it.

Yes, we'll need to change this in the simulation tests. That was the main work that was left to do as of the last time I was working on this PR.

rsercano:
Thanks Daniel for checking @sangaman 

But there's even a test case where we try to withdraw `0.0000011` ETH and there's a code block which try to convert it to units in withdraw code of connext, so it was causing issues before the change I did:

`BigInt(Number/Number) instead of BigInt/BigInt.`

rsercano:
I'm really stuck at this point for sim tests, with this change: https://github.com/ExchangeUnion/xud/blob/bigint/lib/orderbook/OrderBook.ts#L346 I'm doing the same thing what sim tests are doing locally and it's working but I cant make it work on simulation test.

- Remove LTC from currencies and pairs from DB.
- Don't init DB (comment out it)
- Start xud
- Try to add order for LTC/BTC and get error that pair doesn't exist.
- Try to add pair and get LTC doesn't exist error.
- Try to add currency LTC for Lnd with 8 decimals -> success
- Try to add pair LTC/BTC -> success
- Try to add order for LTC/BTC -> success.

Same flow doesn't work on simulation test 😕 @moshababo can you help me ?

kilrau:
> For this one, don't we have to fix this in general? Because when somebody adds a currency after xud initialization it should work as soon as it's added without restart. Therefore I added this line: https://github.com/ExchangeUnion/xud/blob/bigint/lib/orderbook/OrderBook.ts#L346

Yes, I believe that should be "fixed in general" to not require a xud restart.

sangaman:
> But there's even a test case where we try to withdraw `0.0000011` ETH and there's a code block which try to convert it to units in withdraw code of connext ([https://github.com/ExchangeUnion/xud/blob/bigint/lib/connextclient/ConnextClient.ts#L850](https://github.com/ExchangeUnion/xud/blob/bigint/lib/connextclient/ConnextClient.ts?rgh-link-date=2020-11-28T16%3A34%3A08Z#L850)), so it was causing issues before the change I did: (It's not only for test-case but it should be possible to withdraw decimals I mean, maybe we should change conversion and write a new method for connext then, I don't really know and waiting feedback from you in this matter)
> `BigInt(Number/Number) instead of BigInt/BigInt.`

This test case is actually mistaken. We make the conversion because we're trying to convert from satoshi units (10e-8) to other units such as wei (10e-18), but in the code base it's still always converting between integers. You can see in the UnitConverter class that we use the `floor` function to chop off any fractional units/amounts when converting.

sangaman:
After going through this PR again I think it makes most sense for me to try finishing this PR from where I left off, as I'd said it's mainly a matter of getting tests to pass and they're still not passing as-is and it's not apparent why. It'd be easier for me to work from my last commit so I'll force push this branch tomorrow, if I don't get tests passing by the end of the week I can revert back to the current latest commit `bdcf8a2`

rsercano:
> After going through this PR again I think it makes most sense for me to try finishing this PR from where I left off, as I'd said it's mainly a matter of getting tests to pass and they're still not passing as-is and it's not apparent why. It'd be easier for me to work from my last commit so I'll force push this branch tomorrow, if I don't get tests passing by the end of the week I can revert back to the current latest commit `bdcf8a2`

Currently only simulation tests are failing, it's because they work by adding currencies after initialization and `UnitConverter` doesn't update currencies when adding a currency which I fixed by changing orderbook's addCurrency functionality, nonetheless sim tests are still failing from same situation.

So In general I fixed all tests and did some small changes which I mentioned above, wouldn't it be better to continue from this one? Because otherwise you would need to do same changes literally unless you revert back UnitConverter's functions to static calls again. @sangaman 


> This test case is actually mistaken. We make the conversion because we're trying to convert from satoshi units (10e-8) to other units such as wei (10e-18), but in the code base it's still always converting between integers. You can see in the UnitConverter class that we use the `floor` function to chop off any fractional units/amounts when converting.

On the other hand for this one, I still can't see where we chop off fractional units in UnitConverter since we do call `amountToUnits` while withdrawing from Connext currencies `const argUnits = this.unitConverter.amountToUnits({...` 
https://github.com/ExchangeUnion/xud/blob/bigint/lib/utils/UnitConverter.ts#L38
And I don't believe this test case was mistaken, I should be able to `withdraw 0.1 ETH`.

sangaman:
> On the other hand for this one, I still can't see where we chop off fractional units in UnitConverter since we do call `amountToUnits` while withdrawing from Connext currencies `const argUnits = this.unitConverter.amountToUnits({...`
> [https://github.com/ExchangeUnion/xud/blob/bigint/lib/utils/UnitConverter.ts#L38](https://github.com/ExchangeUnion/xud/blob/bigint/lib/utils/UnitConverter.ts?rgh-link-date=2020-12-02T09%3A13%3A55Z#L38)
> And I don't believe this test case was mistaken, I should be able to `withdraw 0.1 ETH`.

The test case was mistaken because amount should not be fractional (also it's currently just not matching the description since it uses `0.0000005` instead of 0.000001`), so if you wanted to express `0.000001` with `amount` it would be `100` since amount represents satoshi units. Here's what it looks like currently, it's just an oversight.

```javascript
    it('withdraws 0.000001 ETH amount', async () => {
      expect.assertions(3);
      const txhash = await connext.withdraw({
        currency: 'ETH',
        destination: DESTINATION_ADDRESS,
        amount: 0.0000005,
      });
```

You can check how I've fixed the test case currently, it works out.

`amountToUnits` uses `Math.floor()` currently to chop off any fractional units, but that would only matter in the case of the base units being larger than satoshis. `amount` will always be an integer.

rsercano:
Ok

> > On the other hand for this one, I still can't see where we chop off fractional units in UnitConverter since we do call `amountToUnits` while withdrawing from Connext currencies `const argUnits = this.unitConverter.amountToUnits({...`
> > [https://github.com/ExchangeUnion/xud/blob/bigint/lib/utils/UnitConverter.ts#L38](https://github.com/ExchangeUnion/xud/blob/bigint/lib/utils/UnitConverter.ts?rgh-link-date=2020-12-02T09%3A13%3A55Z#L38)
> > And I don't believe this test case was mistaken, I should be able to `withdraw 0.1 ETH`.
> 
> The test case was mistaken because amount should not be fractional (also it's currently just not matching the description since it uses `0.0000005` instead of 0.000001`), so if you wanted to express `0.000001`with`amount`it would be`100` since amount represents satoshi units. Here's what it looks like currently, it's just an oversight.
> 
> ```js
>     it('withdraws 0.000001 ETH amount', async () => {
>       expect.assertions(3);
>       const txhash = await connext.withdraw({
>         currency: 'ETH',
>         destination: DESTINATION_ADDRESS,
>         amount: 0.0000005,
>       });
> ```
> 
> You can check how I've fixed the test case currently, it works out.
> 
> `amountToUnits` uses `Math.floor()` currently to chop off any fractional units, but that would only matter in the case of the base units being larger than satoshis. `amount` will always be an integer.

Okay if you say so, @sangaman I hope my work saved you some time here.

sangaman:
This PR is ready for review/testing and I've updated the OP @erkarl @raladev. There *should* be no changes from a user facing perspective given the standard setup using seed currencies and trading pairs, as the main changes are internal to xud. But it should be possible to work with currencies besides the ones that had decimal places hardcoded into xud previously (BTC, LTC, ETH, USDT, WETH, DAI, XUC). I haven't tested that yet but any new currencies/tokens would use the same unit conversion mechanisms as the existing ones.

kilrau:
>  I hope my work saved you some time here

Definitely @rsercano 

Something fresh for extensive testing @raladev

erkarl:
utACK

Definitely needs **extensive** testing by @raladev

raladev (changes requested):
- [ ] xud is crashing with current version, so lets wait for conflict resolving by @sangaman .
```

[webproxy]
disable = true
port = 8080
[xud-backup] Waiting for backup_dir value
/app/dist/Xud.js:164
                    this.swapClientManager?.setLogLevel(level);
                                           ^

SyntaxError: Unexpected token '.'
    at wrapSafe (internal/modules/cjs/loader.js:915:16)
    at Module._compile (internal/modules/cjs/loader.js:963:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/app/bin/xud:3:13)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
```

rsercano:
> * [ ]  xud is crashing with current version, so lets wait for conflict resolving by @sangaman .
> 
> ```
> 
> [webproxy]
> disable = true
> port = 8080
> [xud-backup] Waiting for backup_dir value
> /app/dist/Xud.js:164
>                     this.swapClientManager?.setLogLevel(level);
>                                            ^
> 
> SyntaxError: Unexpected token '.'
>     at wrapSafe (internal/modules/cjs/loader.js:915:16)
>     at Module._compile (internal/modules/cjs/loader.js:963:27)
>     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
>     at Module.load (internal/modules/cjs/loader.js:863:32)
>     at Function.Module._load (internal/modules/cjs/loader.js:708:14)
>     at Module.require (internal/modules/cjs/loader.js:887:19)
>     at require (internal/modules/cjs/helpers.js:74:18)
>     at Object.<anonymous> (/app/bin/xud:3:13)
>     at Module._compile (internal/modules/cjs/loader.js:999:30)
>     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
> ```

This might be related with your node version @raladev which version have you used?

raladev:
> This might be related with your node version @raladev which version have you used?

Yeap, it is because of nodejs version, because our docker images are using 12.20 node  version.
```
/root/.cache/node-gyp/12.20.0/include/node/node.h:737:43: warning: cast between incompatible function types from 'void (*)(Nan::ADDON_REGISTER_FUNCTION_ARGS_TYPE, Nan::ADDON_REGISTER_FUNCTION_ARGS_TYPE)' {aka 'void (*)(v8::Local<v8::Object>, v8::Local<v8::Object>)'} to 'node::addon_register_func' {aka 'void (*)(v8::Local<v8::Object>, v8::Local<v8::Value>, void*)'} [-Wcast-function-type]
  737 |       (node::addon_register_func) (regfunc),                          \
      |                                           ^
/root/.cache/node-gyp/12.20.0/include/node/node.h:771:3: note: in expansion of macro 'NODE_MODULE_X'
  771 |   NODE_MODULE_X(modname, regfunc, NULL, 0)  // NOLINT (readability/null_usage)
      |   ^~~~~~~~~~~~~
```

sangaman:
> Yeap, it is because of nodejs version, because our docker images are using 12.20 node version.

This will be fixed after https://github.com/ExchangeUnion/xud-docker/pull/820 and #2035 are merged. I tested this branch with xud-docker using node v14 and it didn't hit any build/compile issues. You should still be able to try yourself with `xud.sh -b node-lts` - I changed that branch back to using xud master but I most recently built and pushed the image using the `sqlite3-v5` branch rebased on this branch, which was compatible.

I'll rebase this PR after #2035 is merged.

raladev (commented):
- works fine for lnd, waiting for connext on simnet to check it for eth.

raladev (approved):

erkarl (approved):

PR #1905, opened 2020-09-17
fix(connext): avoid scientific notation for amount

sangaman:
This changes the way we convert numbers to strings when building requests to send to the Connext client by using `toLocaleString` instead of `toString` to avoid representing large numbers with scientific notation.

This should help with issues we've faced using toknes that deal with larger quantities like DAI.

An example of how `toLocaleString` compares with `toString` as well as `bigint`:

```typescript
const x = 1725000000000000000000; // 1725 DAI in wei units
const bigX = 1725000000000000000000n;

console.log(x.toString());                                          // 1.725e+21
console.log(x.toLocaleString('fullwide', { useGrouping: false }));  // 1725000000000000000000
console.log(bigX.toString());                                       // 1725000000000000000000

const y = x + 1;
console.log(y.toLocaleString('fullwide', { useGrouping: false }));  // 1725000000000000000000
// but we will never be making payments with wei-level precision, xud only uses up to 8 digits to right of decimal point
// the approach above gives us up to 16 digits of precision, which is on the order of 100 million + 8 decimal places of precision
```

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -351,7 +351,7 @@ class ConnextClient extends SwapClient {
       let secret;
       if (deal.role === SwapRole.Maker) {
         // we are the maker paying the taker
-        amount = deal.takerUnits.toString();
+        amount = deal.takerUnits.toLocaleString('fullwide', { useGrouping: false });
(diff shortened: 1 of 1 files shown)


erkarl (approved):
LGTM :+1:

raladev (approved):

PR #1908, opened 2020-09-20
fix: handling insufficient balance errors for swap clients

### LINKED ISSUE #1558: UnexpectedClientError during swap when client has not enough lnd funds for swap
### Precondition

Alice has 2 LTC channels with Bob. Local balance of first channel is 0.6 and local balance of the second is 0.9.

### Steps to reproduce

1. Place buy 1 LTC/XUC 2000 1 order as Bob
2. Place sell 1 LTC/XUC mkt order as Alice

### Expected behaviour

If swap can not be performed using 2 channels -> No swap + 'insufficient local balance in channel' error in cli and [ORDERBOOK] log + order of Bob still exists in Alice's orderbook.
If swap can be performed using 2 channels -> successful swap

### Actual behaviour

failed swap + UnexpectedClientError + No order of Bob in Alice's orderbook because of fail.

### Screens
[Screenshot logs](https://user-images.githubusercontent.com/29906866/81942658-120afe00-9603-11ea-89c9-e925870ff774.png)
[Screenshot cli](https://user-images.githubusercontent.com/29906866/81942666-133c2b00-9603-11ea-8895-b546f17100de.png)

rsercano:
attempts to resolve #1558 

This's a WIP (Work in Progress) PR, I'm currently having trouble with my local setup and @raladev is helping me with it, apart from that please let me know what else actually need to be done.

I suppose matched orders should return to the orderbook? @kilrau (I couldn't test this yet too)

@sangaman @raladev @kilrau

### DIFF
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -335,6 +335,9 @@ class Swaps extends EventEmitter {
     try {
       route = await swapClient.getRoute(makerUnits, destination, makerCurrency);
     } catch (err) {
+      if (err === errors.INSUFFICIENT_BALANCE) {
+        throw SwapFailureReason.InsufficientBalance;
+      }
       throw SwapFailureReason.UnexpectedClientError;
     }
 
@@ -589,7 +592,7 @@ class Swaps extends EventEmitter {
         deal,
         peer,
         reqId,
-        failureReason: SwapFailureReason.UnexpectedClientError,
+        failureReason: (err === errors.INSUFFICIENT_BALANCE) ? SwapFailureReason.InsufficientBalance : SwapFailureReason.UnexpectedClientError,
         errorMessage: err.message,
         failedCurrency: deal.takerCurrency,
       });
--- a/lib/constants/enums.ts
+++ b/lib/constants/enums.ts
@@ -161,6 +161,8 @@ export enum SwapFailureReason {
   InvalidOrders = 15,
   /** Our payment to the peer was rejected, either deliberately or due to an error. */
   PaymentRejected = 16,
+  /** The swap failed due to insufficient balance. */
+  InsufficientBalance = 17,
 }
 
 export enum DisconnectionReason {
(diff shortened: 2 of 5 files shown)


kilrau:
> I suppose matched orders should return to the orderbook? @kilrau (I couldn't test this yet too)

Not clear what you mean here, could you rephrase?

rsercano:
> > I suppose matched orders should return to the orderbook? @kilrau (I couldn't test this yet too)
> 
> Not clear what you mean here, could you rephrase?

I mean, currently if we get unexpectedclienterror due to insufficient balance, maker orders are being removed?

rsercano:
**EDIT:** nvm @kilrau, thanks to @raladev I managed to test this locally, seems to be working, could you please test and let me know?
@raladev @kilrau @sangaman 

image

lib/swaps/Swaps.ts
```diff
@@ -335,6 +335,9 @@ class Swaps extends EventEmitter {
     try {
       route = await swapClient.getRoute(makerUnits, destination, makerCurrency);
     } catch (err) {
+      if (err.message?.includes('insufficient local balance')) {
```
sangaman, line 338: This is an lnd specific message so it should detected and handled within the LndClient class in the `getRoute` method. I think ideally the LndClient will throw a specific error type (you can add it to swaps/errors.ts) indicating that there's insufficient balance, and then that error should be checked for here. That way Connext/other swap clients can throw the same error later on.

lib/swaps/Swaps.ts
```diff
@@ -589,7 +592,8 @@ class Swaps extends EventEmitter {
         deal,
         peer,
         reqId,
-        failureReason: SwapFailureReason.UnexpectedClientError,
+        failureReason: (err.message?.includes('insufficient local balance')) ?
```
sangaman, line 595: Same thing here, `getRoute` should throw an error indicating insufficient balance, and then we check for that error here when setting the failure reason.

sangaman (changes requested):

rsercano:
Would you please recheck? @sangaman

raladev (approved):

sangaman (approved):
New approach looks good. I'm thinking it might make sense to refactor a bit since ConnextClient has it's own error for insufficient balance and, but that isn't particularly important here, just a thought for down the road. Thanks!

Issue #1526, opened 2020-05-06
feat: Change output of removeorder command/grpc call


raladev:
### Reason 

UX improvements

### Task 1

Change current output of removeorder to more useful. 
As user I want to know status of the operation, but instead of that I got info about amount in hold (it is additional information, not main)
It seems removed qty should be returned in response + remained qty if it was partial removal + qty in hold if it exists:
{
removedQty:5;# qty that was successfully removed
remainingQty:7;# remaining qty after partial removing
quantityOnHold:5; # qty in swap, but it will be removed if swap fails
}

### Task 2

Add Text output to xucli for removeorder command. Now it returns only json.

Draft:

- order "id" with qty "qty" fully removed - if remained = 0, and quantityOnHold = 0;
- order "id" with qty "qty" partially removed, remained qty = "qty" - if remained != 0, and quantityOnHold = 0;
- order "id" with qty "qty" partially removed, qty on hold because of swap process = "qty" - if remained = 0, and quantityOnHold != 0;

kilrau:
Noticed that today too. Currently only two cases:
1. quantityOnHold: 0
```
testnet > removeorder a8d8ffb0-acd0-11ea-9bb4-378737f17cd2
{
  "quantityOnHold": 0
}
```
->
```
testnet > removeorder a8d8ffb0-acd0-11ea-9bb4-378737f17cd2
Order a8d8ffb0-acd0-11ea-9bb4-378737f17cd2 successfully removed.
```

2. quantityOnHold: X
```
testnet > removeorder a8d8ffb0-acd0-11ea-9bb4-378737f17cd2
{
  "quantityOnHold": 123
}
```
->
```
testnet > removeorder a8d8ffb0-acd0-11ea-9bb4-378737f17cd2
Order a8d8ffb0-acd0-11ea-9bb4-378737f17cd2 partially removed, remaining quantity on hold: 123
```

erkarl:
Reverted https://github.com/ExchangeUnion/xud/pull/1985

PR #1910, opened 2020-09-21
fix: added new grpc method to easily cancel all orders

### LINKED ISSUE #1878: Easy way to cancel orders in CLI
As CLI user I am missing an easy way to cancel my own orders. Right now I have to `xucli listorders -j`, find my order and the its ID and copy paste it into `xucli removeorder <id>` which is tedious right now and will be nearly impossible to do once there are more orders on the network

rsercano:
attempts to resolve #1878 

Sorry to implement this prior p1 ones, I had already started this, so wanted to complete first. 

image


lib/orderbook/OrderBook.ts
```diff
@@ -816,6 +816,23 @@ class OrderBook extends EventEmitter {
     return order;
   }
 
+  public removeOwnOrders = () => {
+    const successfullyRemovedOrderIds = [];
+    const failedToRemovedOrderIds = [];
+
+    for (const localId of Array.from(this.localIdMap.keys())) {
+      try {
+        this.logger.debug(`Trying to remove order with id: ${localId}`);
```
sangaman, line 825: I think this logging statement should be removed, or at most be set to `trace`, since there are other logging statements when the orders are actually removed.

proto/xudrpc.proto
```diff
@@ -732,6 +743,15 @@ message RemoveOrderResponse {
   uint64 quantity_on_hold = 1 [json_name = "hold"];
 }
 
+message CancelAllOrdersRequest {}
+
+message CancelAllOrdersResponse {
+  // The order ids that are successfully removed.
+  repeated string removed_orders = 1 [json_name = "removed_orders"];
+  // The order ids that are failed to be removed.
+  repeated string failed_orders = 2 [json_name = "failed_orders"];
```
sangaman, line 752: I don't think this field is necessary. There shouldn't be a reason that we are unable to remove/cancel our own orders, they'll be canceled and if not there is something seriously wrong.

lib/orderbook/OrderBook.ts
```diff
@@ -816,6 +816,23 @@ class OrderBook extends EventEmitter {
...
+        this.removeOwnOrderByLocalId(localId, true);
+        successfullyRemovedOrderIds.push(localId);
+      } catch (err) {
```
sangaman, line 828: I don't think try/catch really makes sense here, since we should expect to be able to remove our own orders. The errors that could arise here would apply only if we try to remove an order that doesn't exist, but if we're removing the set of existing orders that shouldn't be an issue. We shouldn't be hitting exceptions here.

lib/orderbook/OrderBook.ts
```diff
@@ -816,6 +816,23 @@ class OrderBook extends EventEmitter {
     return order;
   }
 
+  public removeOwnOrders = () => {
+    const successfullyRemovedOrderIds = [];
+    const failedToRemovedOrderIds = [];
+
+    for (const localId of Array.from(this.localIdMap.keys())) {
```
sangaman, line 823: `Array.from` is not necessary here.

proto/xudrpc.proto
```diff
@@ -732,6 +743,15 @@ message RemoveOrderResponse {
   uint64 quantity_on_hold = 1 [json_name = "hold"];
 }
 
+message CancelAllOrdersRequest {}
+
+message CancelAllOrdersResponse {
+  // The order ids that are successfully removed.
+  repeated string removed_orders = 1 [json_name = "removed_orders"];
```
sangaman, line 750: Maybe this should be `removed_order_ids`?

sangaman (changes requested):
See comments.

sangaman:
This is also definitely not a "fix" commit/PR, should be a `feat` as it's a new feature.

lib/cli/commands/cancelallorders.ts
```diff
@@ -0,0 +1,21 @@
...
+export const describe = 'removes all orders';
+
+export const builder = (argv: Argv) => argv
+    .example('$0 cancelallorders', describe);
+
+const formatOutput = (response: CancelAllOrdersResponse.AsObject) => {
+  response.removedOrdersList.forEach((removedOrder => console.log(`Cancelled order with id ${removedOrder}`)));
+  response.failedOrdersList.forEach((failedOrder => console.log(`Failed to cancel order with id ${failedOrder}`)));
+
+  console.log('No more orders found');
```
sangaman, line 16: One more afterthought, I don't think we need this message here, since it's a synchronous call that returns immediately. I also think we should handle the case where no orders are canceled, writing "no more orders found" when no orders were found at all doesn't make sense. Instead I think we should print "no orders found" when nothing was canceled, and just the canceled order ids without "no more orders found" when we do have orders that were canceled.

proto/xudrpc.proto
rsercano: Alright removing then

lib/orderbook/OrderBook.ts
rsercano: Oh true, my bad.

rsercano:
> This is also definitely not a "fix" commit/PR, should be a `feat` as it's a new feature.

Yeap sorry, my bad, I literally suck at commit messages.

rsercano:
Fixed review points could you please take a look at again?

raladev:
- [ ] it should be removeallorders,  not cancel, because our command for single order is removeorder.

raladev (changes requested):
Above

rsercano:
>  error throwing because of attempt of removing order in hold interrupts work of cancelallorders command. (Place 3 orders, start to fill one of them by peer, use cancelallorders command during swap = no removed orders)

I actually used behavior of `removeOrder` So in this case I guess it makes sense to add failed order ids into output due to on hold ? @sangaman @kilrau @raladev

kilrau:
Yes. You can simple append the failed cancellations:
```
cancelled order with id X
cancelled order with id Y
failed to cancel order with id Z due to "on hold"
```

sangaman:
> > error throwing because of attempt of removing order in hold interrupts work of cancelallorders command. (Place 3 orders, start to fill one of them by peer, use cancelallorders command during swap = no removed orders)
> 
> I actually used behavior of `removeOrder` except this command always tries to remove entire order quantity. So in this case I guess it makes sense to add failed order ids into output due to on hold ? @sangaman @kilrau @raladev

My apologies, this actually looks like a bug in the existing `removeOrder` logic - if it sees there's a hold on an order then it removes any quantity not on hold and then waits for holds to be resolved (if a swap fails, it'll immediately remove the quantity that's no longer on hold). However, the problem is that if there's no quantity that's not on hold, it attempts to remove 0 quantity and that throws an error. So that just needs be fixed, I'll push a fix for it separately.

With that fix the behavior would be that the call immediately goes through and removes all quantity for all orders that isn't on hold - for any holds it will wait and remove any freed up quantity automatically. Does that sound good to everyone?

kilrau:
> for any holds it will wait and remove any freed up quantity automatically

Will the call not return until all quantity is removed? If so, assuming an order is on hold for several minutes for whatever reason, when will the removeorder call return?

sangaman:
> > for any holds it will wait and remove any freed up quantity automatically
> 
> Will the call not return until all quantity is removed? If so, assuming an order is on hold for several minutes for whatever reason, when will the removeorder call return?

The call returns immediately after its been received, no waiting for holds to expire since those could technically take hours. With the singular remove order call it returns the quantity that's still on hold. Maybe here we could return a boolean indicating whether any orders remain on hold, or possibly a list of orders that are still on hold if we think that may be helpful.

#1921 fixes the bug I mentioned above, once that's merged we can rebase this branch on master and we shouldn't hit that "can't remove 0 quantity" error any longer.

raladev:
https://github.com/ExchangeUnion/xud/pull/1921 is merged, @rsercano please do rebase

raladev:
> it would be good to add info about full or partial removing to output, because order may be removed partially (because another part of order is in hold).
can be done only after https://github.com/ExchangeUnion/xud/issues/1526

So 
>error throwing because of attempt of removing order in hold interrupts work of cancelallorders command. (Place 3 orders, start to fill one of them by peer, use cancelallorders command during swap = no removed orders)
is last that we need and it should work after the rebase

rsercano:
Rebased @raladev thanks for checking this out !

lib/orderbook/OrderBook.ts
```diff
@@ -826,6 +826,22 @@ class OrderBook extends EventEmitter {
...
+  public removeOwnOrders = () => {
+    const removedOrderLocalIds = [];
+    const onHoldOrderLocalIds = [];
+
+    for (const localId of this.localIdMap.keys()) {
+      try {
+        this.removeOwnOrderByLocalId(localId, true);
+        removedOrderLocalIds.push(localId);
+      } catch (ex) {
+        onHoldOrderLocalIds.push(localId);
```
sangaman, line 838: `removeOwnOrderByLocalId` doesn't throw an error when the order has a hold, instead it will return a non-zero number (the method comments explain in more detail). So rather than try/catching here, you should just check whether it returns 0 or not and add it to one of the two lists accordingly.

sangaman (changes requested):

lib/orderbook/OrderBook.ts
rsercano: True that, sorry for overlooking it!

rsercano:
Fixed @sangaman thanks for checking

erkarl (approved):

raladev (approved):

raladev (approved):

PR #1911, opened 2020-09-21
fix(connext): display error message for 400 status code

ghost:
Depends on: https://github.com/ExchangeUnion/xud-docker/pull/729

This PR changes the logic around connext requests and 400 status code. We will now display a nicer error message to the user. Previously, the request would get "stuck" and never return.

xud logs:
```
21/09/2020 14:25:56.714 [RPC] error: call /xudrpc.Xud/OpenChannel errored with code 2: failed to open ETH channel for 1999942865: Signer address Ether balance 19999428650000000000 is less than required amount 19999638650000000000
```

cli:
```
➜  xud git:(fix/openchannel-error) ✗ ./bin/xucli openchannel ETH 19.99942865
Error: 2 UNKNOWN: failed to open ETH channel for 1999942865: Signer address Ether balance 19999428650000000000 is less than required amount 19999638650000000000
```

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -819,11 +819,17 @@ class ConnextClient extends SwapClient {
       this.logger.trace(`sending request to ${endpoint}${payloadStr ? `: ${payloadStr}` : ''}`);
       const req = http.request(options, async (res) => {
         let err: XudError | undefined;
+        let body;
         switch (res.statusCode) {
           case 200:
           case 201:
           case 204:
             resolve(res);
+            break;
+          case 400:
+            body = await parseResponseBody<ConnextErrorResponse>(res);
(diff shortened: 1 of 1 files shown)


sangaman (approved):
Looks good. Only minor thing to perhaps consider is whether we want to create a new internal connext error type for the 400/409 status case and assign it to the `err` variable which gets logged outside the switch. But it's fine as is.

erkarl:
> Looks good. Only minor thing to perhaps consider is whether we want to create a new internal connext error type for the 400/409 status case and assign it to the `err` variable which gets logged outside the switch. But it's fine as is.

Good idea. What error type did you have in mind? I thought about using `connext.errors.INSUFFICIENT_BALANCE`, but didn't because 400/409 can also return other types of errors.

Going to merge this and change the error type in a follow-up PR once we decide which one to use.

PR #1915, opened 2020-09-23
feat(simnet): add DAI

ghost:


erkarl:
Updated with `USDT/DAI` trading pairs.

raladev (approved):

PR #1885, opened 2020-09-10
feat(connext): request collateral for order amount

sangaman:
This adds a check when a new order is receiving tokens via connext that ensures we have sufficient collateral from the connext node to fulfill the order. If not, we fail the order request and request the additional collateral necessary. This involves the following changes:

1. The Connext client now tracks the inbound node collateral for each currency and refreshes this value every time it queries for channel balances and every time a collateral request completes.

2. Before placing a new order that is receiving Connext tokens we check the inbound amount against the available collateral for that currency. In the case of market orders, we use the best available matching price in the order book to calculate inbound & outbound amounts.

3. If the collateral is insufficient, the order will be rejected and a collateral request will be performed for the missing capacity.

4. Any other collateral checks while we're awaiting a response for a collateral request will not perform an additional collateral request so as to prevent duplicate calls.

5. Traders may repeat their order request and it will be accepted once sufficient collateral to complete the trade is acquired.

6. Collateral is no longer requested upon depositing funds to a Connext channel, instead it is only requested upon placing orders.

Closes #1845.

~~Note that this PR does not remove the collateralization that happens after depositing/opening a channel. I figure that can be done in a separate PR, but we may want to reconsider instead. One thing that occurred to me while working on this is that if we only issue market orders, the collateralization may never happen since it's based on requesting a fixed amount based on the price and quantity of orders we place. However, market orders don't have a price and therefore don't have a specific amount of collateral for us to request. Maybe there's a cleaner solution for dealing with this without requesting collateral after opening a channel, but I haven't thought of it yet.~~

Issue #1584, opened 2020-05-27
Reserved Balance (show balance reserved by open orders separately)


kilrau:
Exchanges usually show balance that is reserved by open orders slightly different, e.g. as "frozen", so should we. I prefer the term "reserved" since it's simply balance reserved for a certain open order.

Additionally, this reserved balance should be subtracted from `tradinglimits` to make `placeorder` balance checks account for it.

sangaman:
Do you mean subtracting the quantity from all our open orders from the numbers we return/display for balance calls?

kilrau:
Subtracting and showing it in a separate field or parenthesis e.g.:
```
simnet > getbalance

Balance:
┌──────────┬───────────────┬──────────────────────────────┬────────────────────────────────┐
│ Currency │ Total Balance │ Channel Balance (Tradable)   │ Wallet Balance (Not Tradable)  │
├──────────┼───────────────┼──────────────────────────────┼────────────────────────────────┤
│ BTC      │ 5.92217896    │ 0 (1 reserved | 2.5 pending) │ 1.96102386 (0.4611551 pending) │
├──────────┼───────────────┼──────────────────────────────┼────────────────────────────────┤
```

kilrau:
This feature was requested from two users independently in the last days, so I am bumping prio here. Do you think it makes sense as described? @sangaman

kilrau:
Bumped prio to P1 since this is needed for https://github.com/ExchangeUnion/xud/issues/1896.

kilrau:
As just discussed with @raladev we have no collateralization for `placeorder` with `nobalancechecks` flag (https://github.com/ExchangeUnion/xud/pull/1916#issuecomment-701914861).  So while we want to restrict order placing by default (if e.g. all balance is used up by open orders), one can skip this check by appending `--nobalancechecks` to `placeorder`.

kilrau:
Closing here in favour of https://github.com/ExchangeUnion/xud/issues/1678#issuecomment-708530763

PR #1916, opened 2020-09-24
feat(connext): lazy collateralization

### LINKED ISSUE #1896: Lazy Collateralization
We just discussed to keep the scope of https://github.com/ExchangeUnion/xud/pull/1885 as is and tackle https://github.com/ExchangeUnion/xud/issues/1584 next and expose reserved balance internally in xud.

In a third step, this issue is about triggering a request for collateralization for the orders up to the full amount of `reserved balance`, but let the order pass through `placeorder` as long as the connext channel is sufficiently collateralized for this one single order.

sangaman:
Closes #1896. Branched off of and dependant on #1885. WIP for now but this adds functionality to track reserved outbound & inbound balances each time an own order is added to or removed from the order book. Every time an order is added (and reserved amount increases) we check if the inbound capacity for that currency is sufficient to cover all orders. In Connext's case, if it's not sufficient then we make a collateral request to cover all orders +3% (unless we already have a pending collateral request waiting). We don't do anything on the lnd side since lnd can't dynamically increase its inbound capacity (although we could consider doing something like logging a message).

Related to #1584, which can use the reserved amounts that are being tracked in this PR and expose them over rpc.

Tests still need to be updated to mock the new methods that track reserved amounts and/or add test cases for the reserved amount tracking logic.

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -38,7 +38,7 @@ import {
 } from './types';
 import { parseResponseBody } from '../utils/utils';
 import { Observable, fromEvent, from, combineLatest, defer, timer } from 'rxjs';
-import { take, pluck, timeout, filter, mergeMap, catchError, mergeMapTo } from 'rxjs/operators';
+import { take, pluck, timeout, filter, catchError, mergeMapTo } from 'rxjs/operators';
 import { sha256 } from '@ethersproject/solidity';
 
 interface ConnextClient {
@@ -268,6 +268,28 @@ class ConnextClient extends SwapClient {
     return this._totalOutboundAmount.get(currency) || 0;
   }
 
+  /**
+   * Checks whether we have a pending collateral request for the currency and,
+   * if one doesn't exist, starts a new request for the specified amount. Then
+   * calls channelBalance to refresh the inbound capacity for the currency.
+   */
+  private requestCollateralInBackground = (currency: string, units: number) => {
+    // first check whether we already have a pending collateral request for this currency
+    // if not start a new request, and when it completes call channelBalance to refresh our inbound capacity
+    const requestCollateralPromise = this.requestCollateralPromises.get(currency) ?? this.sendRequest('/request-collateral', 'POST', {
+      assetId: this.tokenAddresses.get(currency),
+      amount: units.toLocaleString('fullwide', { useGrouping: false }),
+    }).then(() => {
+      this.requestCollateralPromises.delete(currency);
+      this.logger.debug(`completed collateral request of ${units} ${currency} units`);
+      return this.channelBalance(currency);
+    }).catch((err) => {
+      this.requestCollateralPromises.delete(currency);
+      this.logger.error(err);
+    });
+    this.requestCollateralPromises.set(currency, requestCollateralPromise);
+  }
+
   public checkInboundCapacity = (inboundAmount: number, currency: string) => {
     const inboundCapacity = this._maxChannelInboundAmount.get(currency) || 0;
     if (inboundCapacity < inboundAmount) {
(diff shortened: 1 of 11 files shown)


sangaman:
I finished adding comprehensive tests across the `OrderBook`, `SwapClientManager`, and `ConnextClient` classes. This should be ready for testing/review.

lib/connextclient/ConnextClient.ts
```diff
@@ -289,15 +289,47 @@ class ConnextClient extends SwapClient {
...
+        reservedInboundAmount * 1.03 - inboundCapacity,
+        ConnextClient.MIN_COLLATERAL_REQUEST_SIZES[currency] ?? 0,
+      );
+      const unitsToRequest = this.unitConverter.amountToUnits({ currency, amount: quantityToRequest });
+
+      // first check whether we already have a pending collateral request for this currency
+      // if not start a new request, and when it completes call channelBalance to refresh our inbound capacity
+      // we don't await this request - instead we allow for "lazy collateralization" to complete since
+      // we don't expect all orders to be filled at once, we can be patient
+      const requestCollateralPromise = this.requestCollateralPromises.get(currency) ?? this.sendRequest('/request-collateral', 'POST', {
```
erkarl, line 320: This logic below seems to be the same as the one defined in `public checkInboundCapacity`. Can we extract this to a function?

raladev:
- we still have order placing collaterization, should it be removed?
```
simnet > buy 0.5 ETH/BTC 0.1 2
9 FAILED_PRECONDITION: channel collateralization in progress, please try again in ~1 minute
```

erkarl (commented):
Looking very good. Left some minor comments + simulation tests are still failing.

erkarl:
>     * we still have order placing collaterization, should it be removed?
> 
> 
> ```
> simnet > buy 0.5 ETH/BTC 0.1 2
> 9 FAILED_PRECONDITION: channel collateralization in progress, please try again in ~1 minute
> ```

No, this feature is complementary to the order placing collateralization.

For example:
1. 0 inbound liquidity
2. buy 0.5 ETH -> placeorder will request collateral and not let the order enter the orderbook
3. buy 0.5 ETH -> placeorder will succeed immediately and request the additional 0.5 ETH collateral in the background (this feature)

raladev (approved):
Approved but need to fix tests

lib/connextclient/ConnextClient.ts
sangaman: Done.

sangaman:
I finally came up with a way to get the simulation instability tests to pass, as connext payments were inexplicably failing for certain cases. I couldn't figure out why the regular collateral calls on `PlaceOrder` calls weren't causing any issues, until eventually I realized that all simulation tests use `nobalancechecks` which skips the collateral on order calls. This PR actually triggers collateral requests in the simulation tests, and for a reason I still don't understand it caused payments to fail in certain cases.

To work around this, I made it so the custom-xud patch code removes the lazy collateral calls logic, and tests are passing now.

lib/connextclient/ConnextClient.ts
```diff
@@ -268,6 +268,28 @@ class ConnextClient extends SwapClient {
...
+   * calls channelBalance to refresh the inbound capacity for the currency.
+   */
+  private requestCollateralInBackground = (currency: string, units: number) => {
+    // first check whether we already have a pending collateral request for this currency
+    // if not start a new request, and when it completes call channelBalance to refresh our inbound capacity
+    const requestCollateralPromise = this.requestCollateralPromises.get(currency) ?? this.sendRequest('/request-collateral', 'POST', {
+      assetId: this.tokenAddresses.get(currency),
+      amount: units.toLocaleString('fullwide', { useGrouping: false }),
+    }).then(() => {
+      this.requestCollateralPromises.delete(currency);
```
erkarl, line 283: nit: can we move this to `.finally` block? That way we can also remove it from the `catch` block.

lib/connextclient/ConnextClient.ts
```diff
@@ -721,18 +757,22 @@ class ConnextClient extends SwapClient {
...
-            );
+            const requestCollateralPromise = this.sendRequest('/request-collateral', 'POST', {
+              assetId,
+              amount: (minCollateralRequestUnits.toLocaleString('fullwide', { useGrouping: false })),
+            }).finally(() => {
+              this.requestCollateralPromises.delete(currency);
+            });
+
+            this.requestCollateralPromises.set(currency, requestCollateralPromise);
+            return from(requestCollateralPromise);
```
erkarl, line 768: Maybe we can also remove the above lines and replace it with `return from(this.requestCollateralInBackground(currency, minCollateralRequestUnits))` as it seems to be doing the same thing (with extra logging)?

kilrau:
> This PR actually triggers collateral requests in the simulation tests, and for a reason I still don't understand it caused payments to fail in certain cases.

Shouldn't we find out why these fail though?

lib/connextclient/ConnextClient.ts
sangaman: I had it like that originally, but I changed it back to this because the `then` block is also doing and returning the `channelBalance` call but I want to delete the promise immediately after the collateral request completes and not have to wait for `channelBalance`. The only way I could think to do that is with two separate `catch` blocks and a `finally` block or like this, and I went with the second option. Let me know if that makes sense or if you have a better idea.

lib/connextclient/ConnextClient.ts
sangaman: `return from(this.requestCollateralInBackground(currency, minCollateralRequestUnits))` doesn't work exactly because requestCollateralInBackground doesn't return a promise, by design the promise is resolved/handled in the background. I could just call the method and return nothing? that should be fine right?

sangaman:
> > This PR actually triggers collateral requests in the simulation tests, and for a reason I still don't understand it caused payments to fail in certain cases.
> 
> Shouldn't we find out why these fail though?

Yes, I reached out to the connext team and I've spent a lot of time trying to figure it out myself but it's still a myster. I was thinking it could be that payments to a node that is in the middle of recollateralization would fail, but that's not supposed to be the case according to connext.

erkarl (approved):

raladev (approved):

PR #1924, opened 2020-09-29
test(db): init & seed value tests

sangaman:
This refactors the database test suite to use jest instead of mocha and adds new test cases to test the database initialization logic that seeds new databases with default values depending on the network and also updates existing databases with new default values that don't exist in the db.


erkarl (approved):
Amazing. Thanks for adding the extra test case for new default values.

PR #1925, opened 2020-10-01
feat(rpc): expose reserved balance for GetBalance 

sangaman:
This adds a field to the `GetBalance` response to return the balance that is reserved for open orders.

Related issue #1584 - before closing it though I think we'll need another PR that enforces the reserved outbound amount before accepting new orders. As is currently it's possible to have more outbound balance reserved than you have available if you keep adding orders in excess of your balance.

This is in draft status until #1916 is merged. I haven't tested this manually yet but it should be ready for testing.


kilrau:
> Related issue #1584 - before closing it though I think we'll need another PR that enforces the reserved outbound amount before accepting new orders.

Agreed and good thinking, I didn't actually mention this in the issue :+1: Can you take care of this PR too?

sangaman:
> Agreed and good thinking, I didn't actually mention this in the issue +1 Can you take care of this PR too?

Yes that's what I'm thinking, but it probably makes sense to keep it separate.

raladev:
@sangaman can u please fix conflicts?

sangaman:
> @sangaman can u please fix conflicts?
> 
>     * [ ]  for now it does not work -_-
>       [Screenshot from 2020-10-05 16-24-18](https://user-images.githubusercontent.com/29906866/95085013-510f2c80-0727-11eb-9d6a-15887e158057.png)
>       [Screenshot from 2020-10-05 16-24-00](https://user-images.githubusercontent.com/29906866/95085018-52405980-0727-11eb-8aa4-a9aea62a4bf8.png)

I fixed the conflicts, but I'm pretty stumped as to why this didn't work for you. I added some test cases as well as a simulation test assertion that verifies that `GetBalance` returns the expected reserved amount after placing an order and then after removing it.

Could you try testing again real quick and if you get the same result could you upload your log? I added some temporary logging statements that might be helpful to figure out what's going on.

raladev:
- i see reserve log lines but nothing changed in getbalance response
[Screenshot from 2020-10-09 16-33-14](https://user-images.githubusercontent.com/29906866/95589443-6baa1400-0a4d-11eb-9244-162fec2422c3.png)
[Screenshot from 2020-10-09 16-33-10](https://user-images.githubusercontent.com/29906866/95589445-6c42aa80-0a4d-11eb-9dd7-e3e93acba011.png)

[reserve.log](https://github.com/ExchangeUnion/xud/files/5355245/reserve.log)

```
09/10/2020 13:31:13.073 [ORDERBOOK] trace: order added: {"pairId":"BTC/USDT","price":11000,"quantity":5000000,"isBuy":false,"localId":"1","id":"b32eca10-0a33-11eb-9376-bbbedc4e2d67","initialQuantity":5000000,"hold":0,"createdAt":1602250273073}
09/10/2020 13:31:13.074 [ORDERBOOK] trace: added 5000000 to BTC reserved amount, new amount is 5000000
09/10/2020 13:31:13.076 [RPC] trace: call /xudrpc.Xud/PlaceOrder succeeded


09/10/2020 13:32:23.268 [ORDERBOOK] trace: order added: {"pairId":"BTC/USDT","price":10000,"quantity":3000000,"isBuy":true,"localId":"2","id":"dd05b240-0a33-11eb-9376-bbbedc4e2d67","initialQuantity":3000000,"hold":0,"createdAt":1602250343268}
09/10/2020 13:32:23.268 [ORDERBOOK] trace: added 30000000000 to USDT reserved amount, new amount is 30000000000

```

raladev (changes requested):
above

sangaman:
OK I figured it out, I was setting the reserved balance properly when a single currency was specified for the `GetBalance` call - but when retrieving the balance for all currencies it wasn't being set. My tests were only using the single currency case and yours was using multi-currency, which is why only you were getting any problems. I've added a test case for fetching reserved balances for multiple currencies and fixed the underlying issue, it should work for you now. Thanks again for the tests @raladev.

raladev (dismissed):
Approved, but in follow up PR with channel balance check we need to:
- add `in order` amount in GetBalance table;
- think about behavoiur with closed channel - should we remove orders if channel was partially or fully closed?

kilrau:
>     * should we remove orders if channel was partially or fully closed?

I would say no for now, sounds like this could lead to unintended behaviour. In the case of arby, the *next* order issuance will account for the reduced balance, so shouldn't be a problem.

sangaman:
>     * add `in order` amount in GetBalance table;

Is it not there for you currently in the parentheses? Or are you suggesting it should be in a separate column? I can definitely change the formatting of the table in this PR if we'd like, I just tried to follow the example in the original issue.

> > ```
> > * should we remove orders if channel was partially or fully closed?
> > ```
> 
> I would say no for now, sounds like this could lead to unintended behaviour. In the case of arby, the _next_ order issuance will account for the reduced balance, so shouldn't be a problem.

I agree, auto cancelling orders would be fairly complicated and I'm not sure it would be desirable.

raladev:
> Is it not there for you currently in the parentheses?

- [ ] for now there is definitely nothing in getbalance table.
Screenshot from 2020-10-12 17-15-05

kilrau:
> > Is it not there for you currently in the parentheses?
> 
>     * [ ]  for now there is definitely nothing in getbalance table.
>       Screenshot from 2020-10-12 17-15-05

As just discussed, we want this to be in for this PR

sangaman:
I suspect this may have to do with the `xucli` you're running in docker not using the same branch as what `xud` is running - could that be the case @raladev? Based on the code and some quick tests it should be outputting like `0.0098 (0.0002 in orders)` in the table.

raladev:
> I suspect this may have to do with the `xucli` you're running in docker not using the same branch as what `xud` is running - could that be the case @raladev? Based on the code and some quick tests it should be outputting like `0.0098 (0.0002 in orders)` in the table.

not sure how it can be in docker env, but now i also checked clean native fresh compiled xud and got same result.
Screenshot from 2020-10-13 14-48-12
Screenshot from 2020-10-13 14-48-01

sangaman:
I finally figured out why it wasn't displaying properly, so I fixed that and as per our discussions I moved the reserved `In Orders` balance to a separate columns. I also added a suite of tests to make sure the `getbalance` formatted CLI output looks proper.

```
┌──────────┬───────────────┬──────────────────────────────────────────┬────────────────────────────┬───────────┐
│ Currency │ Total Balance │ Wallet Balance (Not Tradable)            │ Channel Balance (Tradable) │ In Orders │
├──────────┼───────────────┼──────────────────────────────────────────┼────────────────────────────┼───────────┤
│ BTC      │ 0.005         │ 0.004 (0.00075 pending | 0.001 inactive) │ 0.001 (0.00025 pending)    │ 0.0008    │
└──────────┴───────────────┴──────────────────────────────────────────┴────────────────────────────┴───────────┘
```

raladev (approved):
May be merged after simulation tests passing

PR #1928, opened 2020-10-09
feat: derive child seed for swap clients

### LINKED ISSUE #1576: Seedutil: derive child seed for swap clients (allow creation any time)

sangaman:
This change uses separate mnemonics and seeds for each swap client that are unique to that client, rather than using a single seed and mnemonic across all clients. To accomplish this, a new method is added to the seedutil tool to derive a child mnemonic from a provided mnemonic. This is done by extracting the 16 bytes of entropy from the provided seed, concatenating the entropy with the name of the client (e.g. "LND-BTC") and then hashing the concatenation and taking 16 bytes from the resulting hash.

Closes #1576.

### DIFF
--- a/lib/lndclient/LndClient.ts
+++ b/lib/lndclient/LndClient.ts
@@ -11,7 +11,8 @@ import { LightningClient, WalletUnlockerClient } from '../proto/lndrpc_grpc_pb';
 import * as lndrpc from '../proto/lndrpc_pb';
 import swapErrors from '../swaps/errors';
 import SwapClient, { ChannelBalance, ClientStatus, PaymentState, SwapClientInfo, TradingLimits, WithdrawArguments } from '../swaps/SwapClient';
-import { SwapDeal, CloseChannelParams, OpenChannelParams } from '../swaps/types';
+import { CloseChannelParams, OpenChannelParams, SwapDeal } from '../swaps/types';
+import { deriveChild } from '../utils/seedutil';
 import { base64ToHex, hexToUint8Array } from '../utils/utils';
 import errors from './errors';
 import { Chain, ChannelCount, ClientMethods, LndClientConfig, LndInfo } from './types';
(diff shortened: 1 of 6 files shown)


raladev:
- [ ] xud is crashed on start
```
[webproxy]
disable = true
port = 8080
[xud-backup] Waiting for backup_dir value
Oct 09 10:32:27.000 [notice] Bootstrapped 5% (conn): Connecting to a relay
internal/modules/cjs/loader.js:834
  throw err;
  ^

Error: Cannot find module 'lib/utils/seedutil'
Require stack:
- /app/dist/lndclient/LndClient.js
- /app/dist/swaps/SwapClientManager.js
- /app/dist/Xud.js
- /app/bin/xud
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:831:15)
    at Function.Module._load (internal/modules/cjs/loader.js:687:27)
    at Module.require (internal/modules/cjs/loader.js:903:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/app/dist/lndclient/LndClient.js:48:20)
    at Module._compile (internal/modules/cjs/loader.js:1015:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1035:10)
    at Module.load (internal/modules/cjs/loader.js:879:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Module.require (internal/modules/cjs/loader.js:903:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/app/dist/lndclient/LndClient.js',
    '/app/dist/swaps/SwapClientManager.js',
    '/app/dist/Xud.js',
    '/app/bin/xud'
  ]
}
```

raladev (changes requested):
Above

erkarl:
> xud is crashed on start

Getting the same output as above.

sangaman:
I think the crash had to do with `Error: Cannot find module 'lib/utils/seedutil'` - importing as `lib/utils/seedutil` instead of `../utils/seedutil` doesn't work for reasons I don't fully understand, but vscode automatically uses the former when generating imports occasionally. It should work now.

I also had to tweak the seedutil test case to ensure the deciphered seed doesn't change between runs, since the child mnemonic will change each time it is derived due to the salt being different.

sangaman:
I added a test case for deriving a child seed for an aezeed with a password.

raladev (approved):

erkarl (approved):

PR #1938, opened 2020-10-19
feat: add connext vector support

ghost:
https://github.com/ExchangeUnion/xud-docker/pull/770


lib/connextclient/ConnextClient.ts
```diff
@@ -784,14 +939,16 @@ class ConnextClient extends SwapClient {
     }
 
     const withdrawResponse = await this.sendRequest('/withdraw', 'POST', {
-      recipient: destination,
+      publicIdentifier: this.publicIdentifier,
+      channelAddress: this.channelAddress,
       amount: amount.toLocaleString('fullwide', { useGrouping: false }),
       assetId: this.tokenAddresses.get(currency),
+      recipient: destination,
+      fee: '120', // TODO: estimate fee
```
michael1011, line 947: Can be done with a call to the eth provider: [eth_gasPrice](https://eth.wiki/json-rpc/API#eth_gasprice)

What is the default value when it is not set here?

michael1011 (commented):
And I have one more architectural question: who pays for the gas of the withdrawals from channels?

lib/swaps/SwapClient.ts
```diff
@@ -330,9 +330,12 @@ abstract class SwapClient extends EventEmitter {
     { remoteIdentifier, units, currency, destination, force, fee }: CloseChannelParams,
   ): Promise<string[]>;
 
-  /** Gets a deposit address. */
+  /** Gets a deposit address for channel. */
```
sangaman, line 333: How about "Gets an address for depositing directly to a channel"? I think that'd be a bit clearer in differentiating this from `walletDeposit`

lib/swaps/SwapRecovery.ts
```diff
@@ -16,7 +16,7 @@ interface SwapRecovery {
  * ensuring that we do not lose funds on a partially completed swap.
  */
 class SwapRecovery extends EventEmitter {
-  public static readonly PENDING_SWAP_RECHECK_INTERVAL = 300000;
+  public static readonly PENDING_SWAP_RECHECK_INTERVAL = 15000;
```
sangaman, line 19: What is the reason for changing this constant?

package.json
```diff
@@ -59,7 +59,7 @@
     "test:sim:clean:connext": "(cd test/simulation && ./docker-clean.sh connext)",
     "test:sim:clean:custom-xud": "(cd test/simulation && ./docker-clean.sh custom-xud)",
     "test:sim:logs": "cd test/simulation && ./logs.sh",
-    "test:jest": "jest --unhandled-rejections=strict --rootDir test/jest",
+    "test:jest": "jest --forceExit --unhandled-rejections=strict --rootDir test/jest",
```
sangaman, line 62: Mainly curious but why `--forceExit` here? I've spent some time fiddling with these jest options but didn't consider this one.

proto/xudrpc.proto
```diff
@@ -102,6 +102,15 @@ service Xud {
     };
   }
 
+  /* Gets an address to deposit a given currency into the channel.
```
sangaman, line 105: Should this say "a channel" ? I also think "directly" would be helpful here as in "deposit a given currency directly into a/the channel"

sangaman (commented):
I haven't looked closely through the changes to the test code yet but I can give that a review as well after tests are passing or if we're having trouble getting tests to pass.

package.json
michael1011: `--forceExit` terminates Jest even if there are dangling unresolved Promises. Is there some bug in the changes tests that requires a force exit of Jest?

lib/connextclient/ConnextClient.ts
```diff
@@ -190,13 +213,16 @@ class ConnextClient extends SwapClient {
       tokenAddress: expectedTokenAddress,
     } = expectedIncomingTransfer;
     const currency = this.getCurrencyByTokenaddress(tokenAddress);
+    const blockHeight = await this.getHeight();
+    const timelock = expiry - blockHeight;
+    // The timelock can be up to 10 blocks less than the agreed upon value
+    const TIMELOCK_BUFFER = 10;
```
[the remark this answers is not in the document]
erkarl: It's arbitrary. Can you recommend a better value?

lib/connextclient/types.ts
```diff
@@ -7,6 +7,7 @@ export type ConnextClientConfig = {
   port: number;
   webhookhost: string;
   webhookport: number;
+  nodeIdentifier: string;
```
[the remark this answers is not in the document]
erkarl: It's being used for the `/setup` call to connext: <https://github.com/ExchangeUnion/xud/blob/feat/connext-vector/lib/connextclient/ConnextClient.ts#L946>

lib/connextclient/types.ts
```diff
@@ -18,153 +19,277 @@ export type ConnextInfo = {
...
+  result: string;
+};
+
+type Balance = {
+  amount: string[];
+  to: string[];
+};
+
+/** Connext channel details */
+export type ConnextChannelDetails = {
```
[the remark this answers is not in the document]
erkarl: Yep.

package.json
erkarl: I was experiencing tests passing locally, but running forever in the CI. Didn't want to spend more time debugging this as this seemed to solve the issue.

lib/swaps/SwapRecovery.ts
erkarl: Sorry, forgot to change this back as I was testing the failed swap flows. Changed.

lib/connextclient/ConnextClient.ts
erkarl: Thanks, will address this in a separate PR. Static "120" works for simnet.

lib/connextclient/errors.ts
```diff
@@ -53,9 +55,13 @@ const errors = {
     message: 'connext token address not found',
     code: errorCodes.TOKEN_ADDRESS_NOT_FOUND,
   },
-  PAYMENT_NOT_FOUND: {
-    message: 'connext payment not found',
-    code: errorCodes.PAYMENT_NOT_FOUND,
+  NOT_FOUND: {
```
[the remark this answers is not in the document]
erkarl: Indeed, removed the `TOKEN_ADDRESS_NOT_FOUND` error.

erkarl:
> And I have one more architectural question: who pays for the gas of the withdrawals from channels?

The user, it's deducted from the channel balance.

lib/connextclient/types.ts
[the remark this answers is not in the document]
erkarl: I went ahead and removed it from the configuration options. Instead, now the router node public keys are hardcoded in xud (just like simnet xud seed nodes).

package.json
sangaman: I'll reference this in the running "Tests Flakiness" issue.

sangaman (commented):
The code looks good to me. What is the plan for re-enabling the simulation tests that have been commented out? And would it make sense to comment out the connext related instability tests rather than simply deleting them, if we plan to re-enable them in the near future?

erkarl:
> What is the plan for re-enabling the simulation tests that have been commented out?

We definitely want to get simulation tests going for v2 as well, but not in this PR. I'll create follow up issues from xud-docker/xud side once the current PRs are ready to merge.

> And would it make sense to comment out the connext related instability tests rather than simply deleting them, if we plan to re-enable them in the near future?

I did remove/disable v1 connext simtests in a separate PR (https://github.com/ExchangeUnion/xud/pull/2008) so that if we get around to enable v2 we can easily see the history of what was deleted (instead of going through git history).

sangaman (approved):
Looking forward to having this all work! Thanks.

PR #1939, opened 2020-10-19
fix: rename tradinglimits reserved_inbound json_name

krrprr:
The `tradinglimits` call's `reserved_inbound`'s `json_name` in `xudrpc.proto` is renamed from `reserved_outbound` to `reserved_inbound`.


sangaman (dismissed):

kilrau:
Good catch!

krrprr:
So sorry, my commit wasn't signed before. Signed it now, so that it could be merged to master.

erkarl (approved):

PR #1941, opened 2020-10-20
fix: addpair should prevent adding same base and quote assets (#1559)

### LINKED ISSUE #1559: Same currency pair can be added
### Steps to reproduce

1. Add Pair BTC/BTC
2. Place order to BTC/BTC order book and fill it

### Expected behaviour

No opportunity to add <symbol>/<same_symbol> because:
1. its useless action
2. for now eth/eth swap leads to SwapTimeOut and 500 Connext Error

### Actual behaviour

Pair Added, Swap performed
Note: These Swaps work only for lndbtc and lndltc and dont work for Connext pairs (leads to SwapTimeOut and 500 Connext Error).

rsercano:
resolves #1559 
image

### DIFF
--- a/lib/orderbook/errors.ts
+++ b/lib/orderbook/errors.ts
@@ -16,6 +16,7 @@ const errorCodes = {
   INSUFFICIENT_OUTBOUND_BALANCE: codesPrefix.concat('.12'),
   MIN_QUANTITY_VIOLATED: codesPrefix.concat('.13'),
   QUANTITY_ON_HOLD: codesPrefix.concat('.15'),
+  DUPLICATE_PAIR_CURRENCIES: codesPrefix.concat('.16'),
 };
 
 const errors = {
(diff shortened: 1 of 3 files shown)


raladev (dismissed):

lib/orderbook/errors.ts
```diff
@@ -75,6 +76,10 @@ const errors = {
     message: `order with local id ${localId} has a quantity of ${holdQuantity} satoshis on hold, try again later`,
     code: errorCodes.QUANTITY_DOES_NOT_MATCH,
   }),
+  PAIR_CURRENCIES_CANT_BE_SAME: (baseCurrency: string, quoteCurrency: string) => ({
```
erkarl, line 79: Not sure about the naming of this error. At the same time I don't have a better alternative. Thoughts @sangaman?

erkarl (commented):
Changes look good. Just 1 minor comment/question regarding the error name that we might want to address.

lib/orderbook/errors.ts
sangaman: Yeah it's a bit clunky, how about `DUPLICATE_PAIR_CURRENCIES`?

sangaman (changes requested):
We should add this new error code to `getGrpcError.ts` as one of the `INVALID_ARGUMENT` cases.

rsercano:
Thanks @erkarl @sangaman, fixed

raladev (approved):

sangaman (approved):

PR #1943, opened 2020-10-20
fix: added swap success log entry (#1649)

### LINKED ISSUE #1649: Add `[SWAPS] Successfully executed swap` log entry
From the logs it's clear where a swap procedure begins, but not where it ends. We would want to add an entry along the lines
```
12/06/2020 14:41:23.855 [SWAPS] Successfully executed swap 6e94f2d976bbb2e9fbf380c8644199f7cab513e13c4ace2c9e85c70dabc6100c with peer 02d7348d32c1e97f38c582946e4430a924c7a274d03051865a7e6a7d542cce642b (CheeseMonkey)
```

in the default log level (currently `debug` I believe).

rsercano:
resolves #1649

### DIFF
--- a/lib/swaps/Swaps.ts
+++ b/lib/swaps/Swaps.ts
@@ -1354,6 +1354,7 @@ class Swaps extends EventEmitter {
         peerPubKey: deal.peerPubKey,
         role: deal.role,
       };
+      this.logger.info(`Successfully executed swap ${deal.rHash} with peer ${deal.peerPubKey} (${this.pool.getNodeAlias(deal.peerPubKey)})`);
       this.emit('swap.paid', swapSuccess);
     };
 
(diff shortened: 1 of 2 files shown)


erkarl (dismissed):

lib/swaps/Swaps.ts
```diff
@@ -1354,6 +1354,7 @@ class Swaps extends EventEmitter {
         peerPubKey: deal.peerPubKey,
         role: deal.role,
       };
+      this.logger.info(`Successfully executed swap ${deal.rHash} with peer ${deal.peerPubKey} (${this.pool.getPeer(deal.peerPubKey).alias})`);
```
sangaman, line 1357: This may be minor but I noticed that we're using `getPeer` here which could fail if we've disconnected from the peer. It's somewhat of an edge case but it could happen. Rather than risk hitting an unhandled exception here, I'd do `this.pool.getNodeAlias(deal.peerPubKey)` instead to get the alias which will return a value even if we're no longer connected to our swap partner.

sangaman (changes requested):

rsercano:
Thanks for checking @sangaman, changed it.

raladev (approved):

sangaman (approved):

PR #1949, opened 2020-10-21
feat: reserved capacity checks on PlaceOrder

### LINKED ISSUE #1947: Check total reserved balance before placing new orders
Offshoot from #1678.

> Reject orders that would put our total reserved balance over our total capacity. An exception to this would be for inbound limits for Connext currencies, since we have the ability to dynamically request additional inbound collateral via our "lazy collateral" approach.
> 
> Revisit the logic used for accepting market orders as it pertains to our trading limits, we want to ensure that market orders won't make it so that they don't leave us with insufficient balance to cover our outstanding orders.

sangaman:
This rejects orders that would put our total reserved balance over our total capacity for either the outbound or inbound currency. The sum of the inbound & outbound amounts for a newly placed order are added to the amounts reserved by open orders, and if either of these amounts exceed the corresponding capacity then the request to place the order is rejected.

An exception to this are inbound limits for Connext currencies, since we have the ability to dynamically request additional inbound collateral via our "lazy collateral" approach.

It is still possible for market orders to cause our open orders to exceed our capacity. This is a difficult problem to avoid entirely, as the price that market orders will execute at is unknown until the execution is complete. Even if we simulate the matching routine, we won't know which matches will succeed until we attempt a swap.

Instead, we generously assume that market orders will execute at the best quoted price for purposes of these capacity checks. For users that simultaneously place limit orders and market orders for the same
currencies, it should be made clear that market orders may use up their available balance needed for their limit orders to succeed.

Closes #1947.

### DIFF
--- a/test/jest/SwapClientManager.spec.ts
+++ b/test/jest/SwapClientManager.spec.ts
@@ -5,6 +5,8 @@ import Logger from '../../lib/Logger';
 import SwapClient from '../../lib/swaps/SwapClient';
 import SwapClientManager from '../../lib/swaps/SwapClientManager';
 import { UnitConverter } from '../../lib/utils/UnitConverter';
+import { OwnLimitOrder } from '../../lib/orderbook/types';
+import { errorCodes } from '../../lib/swaps/errors';
 
 jest.mock('../../lib/db/DB', () => {
   return jest.fn().mockImplementation(() => {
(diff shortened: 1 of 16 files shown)


raladev:
- [ ] Question: should we fix message text for negative values? (can be reproduced we partial channel closing and closing and reopening)

```
simnet > sell 2 USDT/DAI mkt
9 FAILED_PRECONDITION: USDT outbound capacity of -4913205400 is not sufficient for order amount of 200000000
```

sangaman:
>     * [ ]  Question: should we fix message text for negative values? (can be reproduced we partial channel closing and closing and reopening)
> 
> 
> ```
> simnet > sell 2 USDT/DAI mkt
> 9 FAILED_PRECONDITION: USDT outbound capacity of -4913205400 is not sufficient for order amount of 200000000
> ```

Yes... I think that makes sense. We would just display `0` in such a case?

raladev:
> We would just display `0` in such a case?

it should be enough, because we dispaly 0 for such case in tradinglimits call

sangaman:
Modified the insufficient capacity messages to not display negative numbers and also rebased this branch to resolve conflicts.

raladev (approved):

test/jest/SwapClientManager.spec.ts
```diff
@@ -297,31 +299,107 @@ describe('Swaps.SwapClientManager', () => {
...
+        localId: 'test',
+      };
+
+      await expect(swapClientManager.checkSwapCapacities(order)).rejects.toHaveProperty('code', errorCodes.SWAP_CLIENT_NOT_FOUND);
     });
   });
+
+  describe('checkSwapCapacities', () => {
+
+  });
```
erkarl, line 404: Remove this block?

sangaman: Thank you, forgot to clean this up.

sangaman:
I just cleaned up the empty block in the tests that karl found, I think otherwise this is ready to merge just needs an approval again.

kilrau:
@raladev @erkarl

raladev (approved):

erkarl (approved):

PR #1959, opened 2020-10-23
fix:incorrect address conversion for connext currencies while adding currencies #1957

### LINKED ISSUE #1957: Incorrect contract address when adding native ETH
Steps:
1. remove pair ETH/BTC
2. removecurrency ETH
3. addcurrency ETH Connext 18 0x0000000000000000000000000000000000000000
4. listcurrencies

Actual result:
Incorrect token address. As a result connext client cant get data for ETH from indra node.
```
ETH    │ 18     │ 0x0                                        │ Connext  
```
Screenshot from 2020-10-22 15-50-54

Expected result:
```
 ETH    │ 18     │ 0x0000000000000000000000000000000000000000 │ Connext
```

rsercano:
resolves #1957

This was a bug coming from yargs, it was converting only number connext addresses into decimals, instead of letting them stay as hexadecimal so I made a small check and fixed it on coerce method.

This was an easy fix but trickier to find.

### DIFF
--- a/lib/cli/commands/addcurrency.ts
+++ b/lib/cli/commands/addcurrency.ts
@@ -29,6 +29,14 @@ export const builder = (argv: Argv) => argv
   .option('token_address', {
     description: 'the contract address for tokens such as ERC20',
     type: 'string',
+    coerce: (address) => {
+      if (address.startsWith('0x')) {
+        return address;
+      }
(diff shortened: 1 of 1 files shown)


raladev (approved):

sangaman (approved):
This is surprising and unusual behavior by yargs for handling `string` arguments, but I confirmed the behavior and your fix works as intended. Thanks for figuring it out.

PR #1960, opened 2020-10-23
fix: removing all listeners for pool and httpserver (#1668)

rsercano:
can you please have a look @moshababo @sangaman @kilrau @raladev

### DIFF
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -1041,6 +1041,7 @@ class Pool extends EventEmitter {
     return new Promise<void>((resolve) => {
       if (this.server) {
         this.server.once('close', resolve);
+        this.server.removeAllListeners();
       } else {
         resolve();
       }


raladev (commented):
Im not sure that it will help with all shutdown cases, but at least it breaks nothing.

Note:
xud turns off fully only when gets EHOSTUNREACH error for connext client (connext client was killed before xud).
 
```
27/10/2020 12:17:27.073 [GLOBAL] info: XUD shutdown gracefully
27/10/2020 12:17:47.937 [P2P] warn: could not open connection to outbound peer (02f894dac96881c972bcc04e0e25ce26cadafbf674a699a2ce87a7aa6ed21dc209@ffnytww5suojkpotlgu2oi7cff4regvskic5dwfxplnztl54u6ne6uid.onion:28885): could not connect to peer at ffnytww5suojkpotlgu2oi7cff4regvskic5dwfxplnztl54u6ne6uid.onion:28885: Proxy connection timed out
27/10/2020 12:17:47.938 [P2P] debug: creating new outbound socket connection to ffnytww5suojkpotlgu2oi7cff4regvskic5dwfxplnztl54u6ne6uid.onion:28885
27/10/2020 12:17:47.939 [P2P] info: Completed start-up connections to known peers
27/10/2020 12:18:34.172 [CONNEXT] error: Error: connect EHOSTUNREACH 10.0.1.6:5040
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1145:16)
27/10/2020 12:18:34.174 [CONNEXT] error: failed to update total outbound capacity: Error: connect EHOSTUNREACH 10.0.1.6:5040
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1145:16)
27/10/2020 12:18:34.175 [CONNEXT] error: Error: connect EHOSTUNREACH 10.0.1.6:5040
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1145:16)
27/10/2020 12:18:34.177 [CONNEXT] error: Error: connect EHOSTUNREACH 10.0.1.6:5040
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1145:16)
```

rsercano:
> xud turns off fully only when gets EHOSTUNREACH error for connext client (connext client was killed before xud).

yes but that's only one of the cases, there're several I'm afraid

lib/http/HttpServer.ts
```diff
@@ -118,6 +118,7 @@ class HttpServer {
    */
   public close = () => {
     return new Promise<void>((resolve) => {
+      this.server.removeAllListeners();
```
sangaman, line 121: I figure if anything it makes more sense for this to be in the `close` callback below.

lib/p2p/Pool.ts
```diff
@@ -1036,6 +1036,7 @@ class Pool extends EventEmitter {
    */
   private unlisten = () => {
     if (this.server && this.server.listening) {
+      this.server.removeAllListeners();
```
sangaman, line 1039: Same here.

sangaman (changes requested):
Does this actually do anything? In both cases, it doesn't sem that we are adding any listeners to the `server` objects, except for an `error` listener when connecting which we start listening which we explicitly remove afterwards.

I also don't believe that having listeners prevents the node js process from exiting. 

Which listeners were you seeing in your testing that prompted this PR?

rsercano:
After using wtfnode as suggested by you @sangaman I saw that those listeners were still there even after shutdown, and after this fix it wasn't there anymore.

kilrau:
If it gets @sangaman's  approval and @raladev can confirm it doesn't break anything it can be merged as is imo.

sangaman:
> After using wtfnode as suggested by you @sangaman I saw that those listeners were still there even after shutdown, and after this fix it wasn't there anymore.

@rsercano Which listeners? Does it say what event they were listening for? Does it help in any cases where xud isn't shutting down gracefully?

I don't expect this to break anything, but I also don't expect it to fix anything. I can't find any documentation that suggests this should be done with nodejs `net` servers. I think at a minimum though we should remove the `removeAllListeners` call into the `close` callback so that we're not removing listeners until after the server has closed, otherwise I could imagine some weird edge cases caused by like an incoming connection before the server has closed but after we've removed the listeners, I don't know what would happen then.

rsercano:
Listeners according to the wtfnode:

image

You're right about moving them into close callback, tho. still I'm unsure wtfnode is helpful enough, it still shows peer sockets, and connext socket after shutdown.

Nonetheless after this fix I didn't see these listeners in the wtfnode, and it worths a shot, also I'll check peer side as far as I can.

rsercano:
With a second thought, we never use http server while working on cmd, so I'm detaching **removeAllListeners** call from there, and keeping the one in the Pool by moving it to callback.

sangaman:
Interestingly this change now breaks a bunch of our p2p network tests, I verified I'm getting the same locally.

```
  P2P Sanity Tests
    ✓ should connect successfully (56ms)
    ✓ should update the node state
    ✓ should fail connecting to the same node
    ✓ should disconnect successfully
    ✓ should fail when connecting to an unexpected node pub key
    ✓ should fail when connecting to an invalid node pub key
    ✓ should fail when connecting to self
    ✓ should fail connecting to a non-existing node
    ✓ should revoke connection retries when connecting to the same nodePubKey
    ✓ should fail when connecting to a node that has banned us
    1) "after all" hook for "should fail when connecting to a node that has banned us"

  P2P Networks Tests
    2) should fail to connect a node from mainnet to a node from simnet
    3) should fail to connect a node from mainnet to a node from testnet
    4) should fail to connect a node from mainnet to a node from regtest
    5) should fail to connect a node from testnet to a node from mainnet
(node:422797) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 SIGINT listeners added to [process]. Use emitter.setMaxListeners() to increase limit
(node:422797) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 SIGTERM listeners added to [process]. Use emitter.setMaxListeners() to increase limit
    6) should fail to connect a node from testnet to a node from simnet
    7) should fail to connect a node from testnet to a node from regtest
    8) should fail to connect a node from simnet to a node from mainnet
    9) should fail to connect a node from simnet to a node from testnet
    10) should fail to connect a node from simnet to a node from regtest
    11) should fail to connect a node from regtest to a node from mainnet
    12) should fail to connect a node from regtest to a node from testnet
    13) should fail to connect a node from regtest to a node from simnet
    14) should successfully connect a node from simnet to a node from simnet
    15) should successfully connect a node from testnet to a node from testnet
    16) should successfully connect a node from simnet to a node from simnet
    17) should successfully connect a node from regtest to a node from regtest
```

I'd like to dig into this a bit more to see what's going on. I'm guessing though that those listeners are added by internal NodeJS code and removing them breaks some stuff, since we're not adding any listeners ourselves.

PR #1961, opened 2020-10-27
feat(rpc): InitSwapClient

### LINKED ISSUE #1929: Handle swap client initialization after initial xud create
Second point of https://github.com/ExchangeUnion/xud/issues/1576
Follow up issue for https://github.com/ExchangeUnion/xud/pull/1928

sangaman:
This allows for xud to initialize a swap client after xud itself has already been initialized using a new rpc command `InitSwapClient`. The swap client must be configured properly and running before it can be initialized. The procedure for initializing is as follows.

1. A swap client type and (optionally) a currency is specified by the `InitSwapClient` request.
2. Xud extracts the deciphered seed from its decrypted node key, this contains the entropy used to generate the master seed/mnemonic.
3. Xud uses the seedutil to convert the deciphered seed back to a mnemonic. For Lnd, it uses this mnemonic to initialize the wallet.

Closes #1929.

Cli usage example: `xucli initclient Lnd LTC`


erkarl:
Looks good code wise - needs manual testing by @raladev

sangaman:
I'm closing this in favor of an alternate approach we discussed to automatically initialize any wallets that aren't initialized upon unlocking xud, aligning with how we currently handle the connext wallet.

PR #1964, opened 2020-10-27
fix: tls certificate check on startup (#1510)

### LINKED ISSUE #1510: bug: ENDENT error if user requests getinfo during initialization of xud
### Background

TS error appeared if user requests getinfo during initialization of xud.
Screen

### Steps to reproduce

1. bash xud.sh 
2. wait for container creation
3. use getinfo after xud container creation

### Expected behaviour

Readable error without error trace (see status output)

rsercano:
resolves #1510 

This's the only thing that I could think of to prevent startup if `tls.cert `doesn't exist. But this spits command help followed by error message.

### DIFF
--- a/lib/cli/command.ts
+++ b/lib/cli/command.ts
@@ -38,6 +38,18 @@ const loadXudConfig = async (argv: Arguments<any>) => {
   }
 };
 
+const getTlsCert = (certPath: string) => {
+  try {
+    return fs.readFileSync(certPath);
+  } catch (err) {
+    if (err.code === 'ENOENT') {
+      throw `tls cert could not be found at ${certPath}, it may take several seconds to be created on xud's first run`;
+    }
+
+    throw err;
+  }
+};
+
 /**
  * A generic function to instantiate an XU client.
  * @param argv the command line arguments
@@ -46,7 +58,7 @@ export const loadXudClient = async (argv: Arguments<any>) => {
   await loadXudConfig(argv);
 
   const certPath = argv.tlscertpath || path.join(argv.xudir, 'tls.cert');
-  const cert = fs.readFileSync(certPath);
+  const cert = getTlsCert(certPath);
   const credentials = grpc.credentials.createSsl(cert);
 
   return new XudClient(`${argv.rpchost}:${argv.rpcport}`, credentials);
@@ -56,7 +68,7 @@ export const loadXudInitClient = async (argv: Arguments<any>) => {
   await loadXudConfig(argv);
 
   const certPath = argv.tlscertpath || path.join(argv.xudir, 'tls.cert');
-  const cert = fs.readFileSync(certPath);
+  const cert = getTlsCert(certPath);
   const credentials = grpc.credentials.createSsl(cert);
 
   return new XudInitClient(`${argv.rpchost}:${argv.rpcport}`, credentials);


raladev (changes requested):
- [x] would be good to change the message to `xud is not initialized, tls.cert is not found` or to `xud is in initialization state, please wait` because `xud ` . 
`could not connect` error does not look correct because we have no connection attempt in if statement, we just checking tls path @rsercano @kilrau

rsercano:
Changed message to `xud is in initialization state, please wait` since this looks more convenient @raladev @kilrau

kilrau:
Can shorten `xud is in initialization state, please wait` -> `xud is initializing, please wait @rsercano

raladev (approved):

sangaman (changes requested):
This prints `xud is initializing` even if we simply have the wrong path for the tls certificate - no matter how many times we retry the call we'd continuously get the same message unless we fix the path. That seems like undesirable behavior to me. We'd also get this message if xud isn't running and has never been run (even if the path is correct), and the error message gives the impression that we just need to wait when in reality xud needs to be manually started as it isn't actually initializing.

Perhaps instead we can wait a few seconds if we detect that the cert does not exist, try again, and if it still doesn't exist proceed with a message that we could not find the tls cert at the provided path?

@kilrau @raladev

raladev:
> Perhaps instead we can wait a few seconds if we detect that the cert does not exist, try again, and if it still doesn't exist proceed with a message that we could not find the tls cert at the provided path?

on slow machines tls creation can take 10-15 sec, so im not sure that timer based solution is good. Maybe we can some flag that says:
0. creation flow was not initialized - print with creation/unlocking suggestion (we already have it)
1. creation flow was initialialized, but not finished - print 'xud is initializing'
2. creation flow was finished successfully/ with error - print err instead of 'xud is initializing'

sangaman:
> > Perhaps instead we can wait a few seconds if we detect that the cert does not exist, try again, and if it still doesn't exist proceed with a message that we could not find the tls cert at the provided path?
> 
> on slow machines tls creation can take 10-15 sec, so im not sure that timer based solution is good. Maybe we can some flag that says:
> 
>     1. creation flow was not initialized - print with creation/unlocking suggestion (we already have it)
> 
>     2. creation flow was initialialized, but not finished - print 'xud is initializing'
> 
>     3. creation flow was finished successfully/ with error - print err instead of 'xud is initializing'

10-15 seconds is a lot more than what I've experienced or would have expected, is that something you've encountered?

I'm not sure how the flag you're suggesting would work, would that be a flag on the xucli command? And how would we know what's going on with the creation flow?

Since the issue appears to arise when running xud through docker, maybe we should handle that there? Docker can detect the `ENOENT` error and if it's the very first time we're running xud handle it appropriately.

raladev:
> 10-15 seconds is a lot more than what I've experienced or would have expected, is that something you've encountered?

Yeap, usually on my VM, not on host machine.

> I'm not sure how the flag you're suggesting would work, would that be a flag on the xucli command? And how would we know what's going on with the creation flow?

for first point, yes, it can be on xucli side, but for third point it seems impossible to get result of create/initialize flow from xud if tls was not created. So it would be better to handle the case on xud-docker side like it was made in the status command.

> Since the issue appears to arise when running xud through docker, maybe we should handle that there? Docker can detect the `ENOENT` error and if it's the very first time we're running xud handle it appropriately.

make sense, we have `xud is starting` case for status command, maybe we can try to use same logic for the check on xud-docker side.

rsercano:
So, what's the current plan for this please?

lib/cli/command.ts
```diff
@@ -46,6 +46,9 @@ export const loadXudClient = async (argv: Arguments<any>) => {
   await loadXudConfig(argv);
 
   const certPath = argv.tlscertpath || path.join(argv.xudir, 'tls.cert');
+  if (!fs.existsSync(certPath)) {
```
sangaman, line 49: Rather than do an `existsSync` call up front. We should wrap the `readFileSync` call and then handle the `ENOENT` error with a custom error message, otherwise rethrow the error.

lib/cli/command.ts
```diff
@@ -56,6 +59,9 @@ export const loadXudInitClient = async (argv: Arguments<any>) => {
...
```
sangaman, line 62: Same here, or better yet create a separate function that attempts to read and return the cert and use the function in both places, so we don't have code duplication.

sangaman (changes requested):
If we don't want to make docker changes, we can print a better error message here but I think it's unusual to print a message that `xud is initializing` when we actually have no indication that that is happening.

The issue of the tls cert path being correct but a cert not being created yet while xud is running would only come up if someone starts xud for the very first time and then within seconds makes an xucli command. I think printing something like `tls cert could not be found at [path], xud may be initializing` or `tls cert could not be found at [path], it may take several seconds to be created on xud's first run` would be more correct and informative.

raladev (dismissed):

erkarl (approved):

rsercano:
There were some unnecessary whitespace changes, removed them @raladev @erkarl

raladev (approved):

sangaman (approved):
Looks very nice, thank you.

erkarl (approved):

PR #1965, opened 2020-10-27
fix(p2p): don't reconnect peers when pool closed

sangaman:
This ensures that we don't attempt to reconnect to peers that have disconnected from us after we have started closing the p2p pool. This may help prevent scenarios where we unintentionally attempt to reconnect to peers after shutting down xud.


erkarl (approved):

kilrau:
Should be tested against https://github.com/ExchangeUnion/xud/issues/1668#issuecomment-684828808 @raladev

raladev (approved):

PR #1966, opened 2020-10-28
feat(lnd): change gRPC client options

ghost:
This "fix" enables @reliveyy to access lnd running on the host machine when xud is running within xud-docker environment.


raladev (approved):

erkarl:
Going to merge this, but we have to keep in mind that this option is only safe when lnds run in the docker internal network. Using lnds outside of the docker network have to assume that their connection to lnds can be MITM attacked.

CC @sangaman

PR #1954, opened 2020-10-21
refactor(rpc): rename reserved TradingLimits fields

ghost:
Reverts ExchangeUnion/xud#1953

PR #1980, opened 2020-11-03
refactor(rpc): rename reserved TradingLimits fields

sangaman:
This renames the `reservedOutbound` and `reservedInbound` fields on the `TradingLimits` call to `reservedSell` and `reservedBuy` respectively.

This is take two for #1946, which was reverted shortly after being merged to prevent conflicts with xud-ui. Whenever the ui project is ready for these new names we should merge this for good.


kilrau:
Nice - let's get it in with the next release then!

raladev:
we already have identical PR for that - https://github.com/ExchangeUnion/xud/pull/1954 )

kilrau:
> we already have identical PR for that - #1954 )

True, closing here.

krrprr (approved):

raladev (approved):

erkarl (approved):

erkarl:
Thanks @sangaman

PR #1986, opened 2020-11-10
fix: changes removeorder output to a more meaningful message (#1526)

### LINKED ISSUE #1526: feat: Change output of removeorder command/grpc call

rsercano:
resolves #1526


erkarl (dismissed):

raladev (dismissed):

lib/orderbook/OrderBook.ts
```diff
@@ -927,7 +934,13 @@ class OrderBook extends EventEmitter {
       this.swaps.on('swap.paid', paidHandler);
     }
 
-    return remainingQuantityToRemove;
+    const remainingFromOrder = orderQuantity - removedQuantity;
+    return {
+      removedQuantity,
+      onHoldQuantity,
+      pairId: order.pairId,
+      remainingQuantity: remainingFromOrder > onHoldQuantity ? remainingFromOrder : 0,
```
sangaman, line 942: This seems off to me. If we're trying to indicate the quantity that will remain on the order after any holds or complete, I figure it should be:

```typescript
const remainingQuantity = orderQuantity - removedQuantity - remainingQuantityToRemove;
```

Or perhaps simpler we just use the current quantity on the order (after we've removed whatever we were able to before holds) in which case I don't think we need the `orderQuantity` variable at all.

```typescript
const remainingQuantity = order.quantity - remainingQuantityToRemove;
```

sangaman (changes requested):
See below about how we're calculating remaining quantity, otherwise this looks good to me.

raladev (approved):

erkarl (approved):

erkarl:
@sangaman can we get this in?

sangaman (approved):
Looks good with my last feedback addressed, thanks.

Issue #1945, opened 2020-10-20
Make `tradinglimits` account for lightning chan reserve


kilrau:
**EDIT:** renamed this issue, new goal [below](https://github.com/ExchangeUnion/xud/issues/1945#issuecomment-713448164).

LND `0.12` will return [inbound balance with the `channelbalance` call](https://github.com/lightningnetwork/lnd/pull/4512/files). Since this is considerably more efficient and accurate than our current "iterate through lnd's `listchannels`" approach, we should move to this call too once lndbtc & lndltc 0.12 are released.

Also I just double checked the following:
- lnd's `channelbalance` is simply all `local_balance` fields of all channels accumulated
- same way does `channelbalance` include all `remote_balance` fields of all channels accumulated with the PR above.

Here is the **but**: lnd's `local_balance` and `remote_balance` subtract the `commit_fee`, but **not** the channel reserve `chan_reserve_sat` which is essentially non-usable balance. Question is: do we want to manually subtract channel reserves from our local and remote balances to give the user a correct picture of what balance is actually usable?

sangaman:
> do we want to manually subtract channel reserves from our local and remote balances to give the user a correct picture of what balance is actually usable?

I think it makes sense to subtract channel reserves since it is non tradeable balance. However, to do so wouldn't we still need to iterate through the list of channels?

Also, do we actually know that this new approach would be more efficient or accurate than iterating through the list of channels? I don't know why our current approach wouldn't be accurate, and my guess would be it's at least *roughly* the same in terms of efficiency - either way it's not computationally expensive to iterate through a list of up to a couple dozen channels.

kilrau:
> I think it makes sense to subtract channel reserves since it is non tradeable balance. However, to do so wouldn't we still need to iterate through the list of channels?

Yes, that's why I brought it up. The new PR doesn't include something like total inbound/outbound chan reserve...

> Also, do we actually know that this new approach would be more efficient or accurate than iterating through the list of channels? I don't know why our current approach wouldn't be accurate, and my guess would be it's at least roughly the same in terms of efficiency - either way it's not computationally expensive to iterate through a list of up to a couple dozen channels.

Yep, for us even more accurate since we'd exclude channel reserve.

Summary of the todo's:
- move tradinglimits' `maxBuy` & `maxSell` to iterating through channels (currently we are using lnd's `channelbalance` for `maxSell` I believe)
- subtract chan reserve from tradinglimits' `maxBuy` & `maxSell` 
[...]

PR #1988, opened 2020-11-10
fix(lnd): don't calculate negative capacities

sangaman:
This fixes a bug in the logic for calculating the inbound & outbound amount/capacity totals. We subtract the channel reserve amounts from the balances when determining how large a payment the channel could support, however we should not end up with a negative number.


kilrau:
I was honestly not aware that `tradinglimits` is already accounting for the channel reserve. I am closing https://github.com/ExchangeUnion/xud/issues/1945#issuecomment-713448164.

michael1011 (approved):

PR #2000, opened 2020-11-20
feat(rpc): GetMnemonic

### LINKED ISSUE #1982: Support displaying mnemonic again
This is needed for https://github.com/ExchangeUnion/xud-ui/issues/39 - we are calling `create` in XUD EXPLORER with some default values, the user "creating" the environment is actually "changing" the password and showing the mnemonic again in this flow.

I guess this should be a separate "show mnemonic" call.

sangaman:
This adds a new rpc call `GetMnemonic` that can be used to retrieve the master seed mnemonic for xud at runtime after an xud node has been created.

Closes #1982.


erkarl:
utACK

sangaman:
The output just looks like this currently, I'm open to suggestions on if we want to include any additional information or formatting.

```
$ ./bin/xucli mnemonic
above moon armor scale valid accuse inject choose unable normal toy exit ostrich doctor laptop desert lady romance ginger start tomato orbit february swim
```

kilrau:
Looking at the current cli commands I'd change to `getmnemonic` to follow the convention

sangaman:
Here's a quick primer on the aezeed format we're using for our mnemonic, and why the mnemonic we get will be different from the original creation of the seed and different from each time we get the mnemonic at runtime.

The deciphered aezeed data is a 1 byte payload, 2 byte birthday/timestamp, and 16 bytes of entropy. We store that to our node key file encrypted on the disk, so all this is preserved, but we don't store the enciphered seed anywhere. The mnemonic represents the 33 byte enciphered seed, which includes the deciphered key plus a salt and checksum.

The salt is going to to change every time we encipher the deciphered seed to recreate the mnemonic. So we'll get a different mnemonic every time, which may appear unusual to people used to conventional bip39 mnemonic seeds, but it's in fact a feature and not a bug.

kilrau:
Thanks for the explainer @sangaman , I'd say let's leave the behavior as is and as just discussed add a hint (at least on the cli) for the user to know that this mnemonic will be different every time but still always recovers the same funds.

kilrau:
Can we wrap this up today? @sangaman 

Missing before @raladev can start testing again:
- https://github.com/ExchangeUnion/xud/pull/2000#issuecomment-731230368
- https://github.com/ExchangeUnion/xud/pull/2000#issuecomment-731266441

kilrau:
https://github.com/ExchangeUnion/xud-docker/pull/795#event-4027542446

sangaman:
New output:


```
$ ./bin/xucli getmnemonic
 1. abstract    2. future      3. flavor      4. aim       
 5. amount      6. okay        7. hidden      8. orange    
 9. combine    10. collect    11. liar       12. dish      
13. fossil     14. sort       15. nest       16. elbow     
17. bachelor   18. combine    19. inherit    20. salon     
21. annual     22. there      23. actress    24. spoil     


Please write down your 24 word mnemonic. It will allow you to recover your xud node key and on-chain funds for the initialized wallets listed above should you forget your password or lose your device. Off-chain funds in channels can NOT be recovered with it and must be backed up and recovered separately. Keep it somewhere safe, it is your ONLY backup in case of data loss.


Note that this mnemonic is not the same as the mnemonic displayed when this xud node was first created. Xud uses aezeed which is salted with random bytes, xud does not store the salt or the mnemonic words. Each time xud regenerates the mnemonic, it uses a new salt so that the mnemonic is different, but it still restores the same xud node key and the same underlying wallet addresses.
```

kilrau:
Can short from

`Note that this mnemonic is not the same as the mnemonic displayed when this xud node was first created. Xud uses aezeed which is salted with random bytes, xud does not store the salt or the mnemonic words. Each time xud regenerates the mnemonic, it uses a new salt so that the mnemonic is different, but it still restores the same xud node key and the same underlying wallet addresses.`

->

`Note that this mnemonic changes each time xud regenerates the mnemonic, but it still restores the same xud node key and the same underlying wallets.`

Apart from that looking good.

kilrau:
*You can leave the details about why the mnemonic changes as in-code comment for that sentence.

sangaman:
Done, kept the message in the shorter form you suggested but tweaked it just a bit more:

`Note that this mnemonic changes each time it is regenerated by xud, but it still restores the same xud node key and underlying wallets`

kilrau:
Please give this another spin @raladev

raladev (approved):

erkarl (approved):

Issue #1970, opened 2020-10-30
Classify peers as `offline`, long-dead peers as `dead`


kilrau:
### Background

Currently logs of an "older" xud environment look like [this](https://paste.ubuntu.com/p/NhwB7FY9s5/) when starting/restarting xud. Busy for minutes with a decent load on the tor process trying to connect to *very* dead peers. It takes a while (in my case 6 minutes) until xud finally reconnected to its previous peers.

### Your environment

* version of `xud`: `1.2.0-d4587458`
* which operating system (`uname -a` on *Nix):
* version of `lndbtc`, `lndltc`, `connext` and others:
* any other relevant environment details:

### Steps to reproduce

Restart xud which has 100+ dead peers in the db.

### Expected behaviour

xud immediately connects to previosly connected peers, leaves classified *dead* peers reconnection attempts to the regular scheduled attempts at runtime, but **not** on startup.

### Actual behaviour

xud tries to connect to all peers on startup, leading to behavior described above.

kilrau:
How this could work:
- a counter records *unsuccessful* connection retries per peer and classifies a peer as `offline` after 288 failed consecutive connection attempts (1 day), to `dead` after `1708` failed consecutive attempts (1 week).
- xud doesn't connect to `offline` peers in the first round of re-connection attempts on startup to avoid above described behavior
- xud will stop attempting to connect to `dead` peers altogether
- once a successful connection to (or from) a `offline`/`dead` peer is established the peers `offline`/`dead` status is erased

LePremierHomme:
A more technical design proposal for the basic feature:

1. Add `numConnFailures` field to `Nodes` xud db table.
2. Add 2 fields to `PoolConfig`: `nodeOfflineThreshold` (default: 288), `nodeDeadThreshold` (default: 1708). (I might as well just start with plain constants in the code.)
3. Increment the node entry `numConnFailures` field on every distinct connection failure (**to any address**).  Reset the counter on a successful connection.
4. on xud launch, filter out nodes which their `numConnFailures` pass `nodeDeadThreshold`, and group the nodes where it passed `nodeOfflineThreshold` to initiate the connection after 5-10 (?) min. (If in the meanwhile the node connected to us, this connection attempt will fail. We can actively abort it instead, but i’m not sure it worth the code complexity.)

erkarl:
> A more technical design proposal for the basic feature:
> 
>     1. Add `numConnFailures` field to `Nodes` xud db table.
> 
[...]

PR #2009, opened 2020-11-29
feat(p2p): Track peers' connection failures and rank accordingly

LePremierHomme:
Closing https://github.com/ExchangeUnion/xud/issues/1970.

Few remarks:
* The `offline`/`dead` terminology wasn't used. Instead, `consecutiveConnFailure` counter was added to known nodes, and a ranking function which works in accordance with some thresholds. 
* Peer session can fail even if raw connection was successful (in handshake or validation). These failures suppose to affect the peer's reputation. The current ranking function doesn't consider that, but rather only the `consecutiveConnFailure` counter (it's possible to change that).
* `consecutiveConnFailure` cannot be correlated to time, so simpler thresholds (200 & 1500) than what was specified in https://github.com/ExchangeUnion/xud/issues/1970#issuecomment-733626904 were used.
* If no potential peers found in the first threshold (200) group, the second's (1500) group is used. A more sophisticated adjustments can be made, but I just went with the simplest.

### DIFF
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -77,7 +77,7 @@ class NodeList extends EventEmitter {
     this.nodes.forEach(callback);
   };
 
-  public rank = (): { primary: NodeInstance[], secondary: NodeInstance[] } => {
+  public rank = (): { primary: NodeInstance[]; secondary: NodeInstance[] } => {
     const primary: NodeInstance[] = [];
     const secondary: NodeInstance[] = [];
 
@@ -93,7 +93,7 @@ class NodeList extends EventEmitter {
       return { primary: secondary, secondary: [] };
     }
     return { primary, secondary };
-  }
+  };
 
   /**
    * Get the internal node id for a given nodePubKey.
@@ -288,7 +288,7 @@ class NodeList extends EventEmitter {
       node.consecutiveConnFailures = node.consecutiveConnFailures + 1;
       await node.save();
     }
-  }
+  };
 
   public resetConsecutiveConnFailures = async (nodePubKey: string) => {
     const node = this.nodes.get(nodePubKey);
@@ -296,7 +296,7 @@ class NodeList extends EventEmitter {
       node.consecutiveConnFailures = 0;
       await node.save();
     }
-  }
+  };
 
   private setBanStatus = (node: NodeInstance, status: boolean) => {
     node.banned = status;
--- a/lib/p2p/NodeList.ts
+++ b/lib/p2p/NodeList.ts
@@ -43,6 +43,9 @@ class NodeList extends EventEmitter {
   private static readonly BAN_THRESHOLD = -50;
   private static readonly MAX_REPUTATION_SCORE = 100;
 
+  private static readonly PRIMARY_PEER_CONN_FAILURES_THRESHOLD = 200;
+  private static readonly SECONDARY_PEER_CONN_FAILURES_THRESHOLD = 1500;
+
   public get count() {
     return this.nodes.size;
   }
@@ -73,6 +76,24 @@ class NodeList extends EventEmitter {
   public forEach = (callback: (node: NodeInstance) => void) => {
     this.nodes.forEach(callback);
   };
+
+  public rank = (): { primary: NodeInstance[], secondary: NodeInstance[] } => {
+    const primary: NodeInstance[] = [];
+    const secondary: NodeInstance[] = [];
+
+    this.nodes.forEach((node) => {
+      if (node.consecutiveConnFailures < NodeList.PRIMARY_PEER_CONN_FAILURES_THRESHOLD) {
+        primary.push(node);
+      } else if (node.consecutiveConnFailures < NodeList.SECONDARY_PEER_CONN_FAILURES_THRESHOLD) {
+        secondary.push(node);
+      }
+    });
+
+    if (primary.length === 0) {
+      return { primary: secondary, secondary: [] };
+    }
+    return { primary, secondary };
+  }
 
   /**
    * Get the internal node id for a given nodePubKey.
@@ -260,6 +281,22 @@ class NodeList extends EventEmitter {
 
     return false;
   };
+
+  public incrementConsecutiveConnFailures = async (nodePubKey: string) => {
+    const node = this.nodes.get(nodePubKey);
+    if (node) {
+      node.consecutiveConnFailures = node.consecutiveConnFailures + 1;
+      await node.save();
+    }
+  }
+
+  public resetConsecutiveConnFailures = async (nodePubKey: string) => {
+    const node = this.nodes.get(nodePubKey);
+    if (node && node.consecutiveConnFailures > 0) {
+      node.consecutiveConnFailures = 0;
+      await node.save();
+    }
+  }
 
   private setBanStatus = (node: NodeInstance, status: boolean) => {
     node.banned = status;
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -203,35 +203,37 @@ class Pool extends EventEmitter {
     this.bindNodeList();
 
     this.loadingNodesPromise = this.nodes.load();
-    this.loadingNodesPromise.then(async () => {
-      if (this.disconnecting) {
+    this.loadingNodesPromise
+      .then(async () => {
+        if (this.disconnecting) {
+          this.loadingNodesPromise = undefined;
+          return;
+        }
+
+        const { primary, secondary } = this.nodes.rank();
+
+        // delay connection attempts to secondary peers - whom we've had trouble connecting to in the past -
+        // to prevent overwhelming xud at startup
+        if (secondary.length > 0) {
+          this.secondaryPeersTimeout = setTimeout(async () => {
+            this.logger.info('Connecting to secondary known / previously connected peers');
+            await this.connectNodes(secondary, true, true);
+            this.logger.info('Completed start-up connections to secondary known peers');
+          }, Pool.SECONDARY_PEERS_DELAY);
+        }
+
+        if (primary.length > 0) {
+          this.logger.info('Connecting to known / previously connected peers');
+          await this.connectNodes(primary, true, true);
+          this.logger.info('Completed start-up connections to known peers');
+        }
+
         this.loadingNodesPromise = undefined;
-        return;
-      }
-
-      const { primary, secondary } = this.nodes.rank();
-
-      // delay connection attempts to secondary peers - whom we've had trouble connecting to in the past -
-      // to prevent overwhelming xud at startup
-      if (secondary.length > 0) {
-        this.secondaryPeersTimeout = setTimeout(async () => {
-          this.logger.info('Connecting to secondary known / previously connected peers');
-          await this.connectNodes(secondary, true, true);
-          this.logger.info('Completed start-up connections to secondary known peers');
-        }, Pool.SECONDARY_PEERS_DELAY);
-      }
-
-      if (primary.length > 0) {
-        this.logger.info('Connecting to known / previously connected peers');
-        await this.connectNodes(primary, true, true);
-        this.logger.info('Completed start-up connections to known peers');
-      }
-
-      this.loadingNodesPromise = undefined;
-    }).catch((reason) => {
-      this.logger.error('Unexpected error connecting to known peers on startup', reason);
-      this.loadingNodesPromise = undefined;
-    });
+      })
+      .catch((reason) => {
+        this.logger.error('Unexpected error connecting to known peers on startup', reason);
+        this.loadingNodesPromise = undefined;
+      });
 
     this.verifyReachability();
     this.connected = true;
--- a/lib/p2p/Pool.ts
+++ b/lib/p2p/Pool.ts
@@ -81,6 +81,7 @@ class Pool extends EventEmitter {
   /** A collection of known nodes on the XU network. */
   private nodes: NodeList;
   private loadingNodesPromise?: Promise<void>;
+  private secondaryPeersTimeout?: NodeJS.Timeout;
   /** A collection of opened, active peers. */
   private peers = new Map<string, Peer>();
   private server?: Server;
@@ -97,6 +98,9 @@ class Pool extends EventEmitter {
   private nodeKey: NodeKey;
   /** The minimum version of xud we accept for peers */
   private minCompatibleVersion: string;
+
+  /** Secondary peers connection attempt delay. */
+  private static readonly SECONDARY_PEERS_DELAY = 600000; // 10 min
 
   constructor({
     config,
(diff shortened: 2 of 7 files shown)


kilrau:
NICE!

>     * Peer session can fail even if raw connection was successful (in handshake or validation). These failures suppose to affect the peer's reputation. The current ranking function doesn't consider that, but rather only the `consecutiveConnFailure` counter (it's possible to change that).

You didn't say why you decided to make the `consecutiveConnFailure` counter account for these connection failures, but not the ranking function. Anyhow, if it adds significant compliexity to make the ranking function account for these failures, we can leave as is. If not we should make it account for it to be consistent imo.

> If no potential peers found in the first threshold (200) group, the second's (1500) group is used.

Is this on startup (first connection attempt) or 2 onwards? The second group (1500) was the one we defined as `dead` and I don't think it should ever be considered for connection. What we *can* do in case the first group is empty: try to connect to [seed nodes](https://github.com/ExchangeUnion/xud/blob/master/lib/db/seeds/mainnet.ts#L4) ignoring their status

kilrau:
Rest sounds good!

LePremierHomme:
> You didn't say why you decided to make the consecutiveConnFailure counter account for these connection failures, but not the ranking function. Anyhow, if it adds significant compliexity to make the ranking function account for these failures, we can leave as is. If not we should make it account for it to be consistent imo.

`consecutiveConnFailure` counts raw connection failures, but not general peer session failures. The ranking function currently considers only `consecutiveConnFailure`, but it can consider also other failures, by using the peer reputation score. It’s easy to implement. (We can also just make the raw connection failures as another reputation event, but I thought that keeping it separate might be worthwhile.)

> Is this on startup (first connection attempt) or 2 onwards? The second group (1500) was the one we defined as dead and I don't think it should ever be considered for connection. What we can do in case the first group is empty: try to connect to seed nodes ignoring their status

The first group is < 200 (good peers), the second group is < 1500 (“offline” peers). The rest (“dead” peers) are ignored. See here: https://github.com/ExchangeUnion/xud/pull/2009/files#diff-1046c174ecc568579870d80a7fb8d8125eca7011b25e73deba9153915e5b4185R77

kilrau:
> It’s easy to implement

Then I'd say let's do it!

> (We can also just make the raw connection failures as another reputation event, but I thought that keeping it separate might be worthwhile.)

Yep, let's keep these separate for now.

> The first group is < 200 (good peers), the second group is < 1500 (“offline” peers). The rest (“dead” peers) are ignored. See here: https://github.com/ExchangeUnion/xud/pull/2009/files#diff-1046c174ecc568579870d80a7fb8d8125eca7011b25e73deba9153915e5b4185R77

Gotcha, all good then. Thx for pointing me at it!

raladev (commented):
I tried to test it but fased some strange behavoir during the testing.
1. when i disabled my wifi, waited ~7 mins and checked listpeers I still got peers (but im not online) - does not look like good behavior, because if i have peers i have orders from them (but i was not able to swap with them because i have no internet);
2. After wifi enabling i waited ~7 mins and checked listpeers after that -> i lost one of four peers - OzoneYellow that connected through clearnet IP instead of tor (xud1 peer in simnet network), i waited 10 additional minutes and checked listpeers after that, it still was not in list (but another my node has it); I got it back only after restart;
[Screenshot from 2020-12-01 16-23-17](https://user-images.githubusercontent.com/29906866/100750561-43101c00-33f7-11eb-81ea-9ab04ca099b2.png)
3. also after restart i got the case when one my node has another my node in listpeers but another one has no this not in the list (after 5 min of wifi enabling);
[Screenshot from 2020-12-01 16-49-58](https://user-images.githubusercontent.com/29906866/100750492-296ed480-33f7-11eb-8613-2de2dec005bf.png)


So can anyone please describe how P2P connection establishing / connection supporting / connection restoring cases work in xud?

erkarl:
Can we rebase this branch against master? Looks like there are some extra commits here (change password).

kilrau:
> Can we rebase this branch against master? Looks like there are some extra commits here (change password).

:+1: @LePremierHomme

kilrau:
> * when i disabled my wifi, waited ~7 mins and checked listpeers I still got peers (but im not online) - does not look like good behavior, because if i have peers i have orders from them (but i was not able to swap with them because i have no internet);
> 
> * After wifi enabling i waited ~7 mins and checked listpeers after that -> i lost one of four peers - OzoneYellow that connected through clearnet IP instead of tor (xud1 peer in simnet network), i waited 10 additional minutes and checked listpeers after that, it still was not in list (but another my node has it); I got it back only after restart;
>   [Screenshot from 2020-12-01 16-23-17](https://user-images.githubusercontent.com/29906866/100750561-43101c00-33f7-11eb-81ea-9ab04ca099b2.png)
> 
> * also after restart i got the case when one my node has another my node in listpeers but another one has no this node in the list (after 5 min of wifi enabling);
>   [Screenshot from 2020-12-01 16-49-58](https://user-images.githubusercontent.com/29906866/100750492-296ed480-33f7-11eb-8613-2de2dec005bf.png)

I agree that we want to fix all of these, but probably none of it is related to this PR. Did your run these test cases on master to verify? @raladev If they reproduce on master just the same, I suggest to design a decent set of "p2p reconnection" test cases. We still have tons of issues with p2p reconnections, but all of them are hard to reproduce and thus hard to fix for @LePremierHomme 

Your thoughts? @LePremierHomme

raladev:
> @raladev If they reproduce on master just the same,

I reproduced them on both versions

>> So can anyone please describe how P2P connection establishing / connection supporting / connection restoring cases work in xud?
> I suggest to design a decent set of "p2p reconnection" test cases. 

I can do that, but i need to know how it should work in ideal way )

kilrau:
> I reproduced them on both versions

Alright, then it's clear that they are not related to this PR and shouldn't be in here. I opened https://github.com/ExchangeUnion/xud/issues/2010 - let's start collecting the test cases there once @LePremierHomme confirmed how it this *should* behave. I'll add my 2cts after this.

lib/Xud.ts
```diff
@@ -119,16 +119,16 @@ class Xud extends EventEmitter {
...
 
       let nodeKey: NodeKey | undefined;
       if (this.config.noencrypt) {
         if (nodeKeyExists) {
           nodeKey = await NodeKey.fromFile(nodeKeyPath);
         } else {
-          nodeKey = await NodeKey.generate();
-          await nodeKey.toFile(nodeKeyPath);
+          nodeKey = await NodeKey.generate(nodeKeyPath);
+          await nodeKey.toFile();
```
sangaman, line 131: These are related to the change password PR, which is now merged. I'd try taking just your most recent commit and rebasing it on master to make this PR easier to review/assess. If git is giving you trouble I can definitely take care of it and update this branch.

lib/p2p/NodeList.ts
```diff
@@ -74,6 +74,24 @@ class NodeList extends EventEmitter {
     this.nodes.forEach(callback);
   }
 
+  public rank = (): { primary: NodeInstance[], secondary: NodeInstance[] } => {
+    const primary: NodeInstance[] = [];
+    const secondary: NodeInstance[] = [];
+
+    this.nodes.forEach((node) => {
+      if (node.consecutiveConnFailures < 200) {
```
sangaman, line 82: I think it would be nice to name the 200 and 1500 values as `const` variables, it can just be somewhere in the header of this file for now. Something like `PRIMARY_PEER_CONN_FAILURE_THRESHOLD` and `SECONDARY_PEER_CONN_FAILURE_THRESHOLD` perhaps?

lib/p2p/NodeList.ts
```diff
@@ -256,6 +274,22 @@ class NodeList extends EventEmitter {
     return false;
   }
 
+  public incrementConsecutiveConnFailures = async (nodePubKey: string) => {
+    const node = this.nodes.get(nodePubKey);
+    if (node) {
```
sangaman, line 279: Should we maybe throw an error here (and in the reset method below) if the node does not exist?

lib/p2p/Pool.ts
```diff
@@ -191,11 +195,26 @@ class Pool extends EventEmitter {
...
     this.loadingNodesPromise.then(async () => {
-      if (this.nodes.count > 0 && !this.disconnecting) {
+      if (this.disconnecting) {
+        this.loadingNodesPromise = undefined;
+        return;
+      }
+
+      const { primary, secondary } = this.nodes.rank();
+      if (secondary.length > 0) {
+        this.secondaryPeersTimeout = setTimeout(async () => {
```
sangaman, line 205: I think comments would be good here to explain what's going on so we don't forget going forward (and also to ensure I understand right now). How about something like

`We delay connection attempts to secondary peers - whom we've had trouble connecting to in the past - to prevent overwhelming xud at startup`

sangaman (commented):
Looks good, just a few comments. I'm also thinking it might make sense to connect to secondary peers in batches, otherwise we may face the same sort of issue of a huge number of connection attempts being made at once (and mostly failing). Maybe connecting to secondary peers ~10 at a time. But that can be a followup PR/issue.

sangaman:
> I tried to test it but faced some strange cases during the testing.
> 
>     1. when i disabled my wifi, waited ~7 mins and checked listpeers I still got peers (but im not online) - does not look like good behavior, because if i have peers i have orders from them (but i was not able to swap with them because i have no internet);
> 
>     2. After wifi enabling i waited ~7 mins and checked listpeers after that -> i lost one of four peers - OzoneYellow that connected through clearnet IP instead of tor (xud1 peer in simnet network), i waited 10 additional minutes and checked listpeers after that, it still was not in list (but another my node has it); I got it back only after restart;
>        [Screenshot from 2020-12-01 16-23-17](https://user-images.githubusercontent.com/29906866/100750561-43101c00-33f7-11eb-81ea-9ab04ca099b2.png)
> 
>     3. also after restart i got the case when one my node has another my node in listpeers but another one has no this node in the list (after 5 min of wifi enabling);
>        [Screenshot from 2020-12-01 16-49-58](https://user-images.githubusercontent.com/29906866/100750492-296ed480-33f7-11eb-8613-2de2dec005bf.png)
> 
> 
> So can anyone please describe how P2P connection establishing / connection supporting / connection restoring cases work in xud? @sangaman @LePremierHomme

I'm opening a separate issue as this is definitely a concern but not related to any of the changes in this PR.

lib/p2p/NodeList.ts
LePremierHomme: I was thinking it will get clunky if we’ll add more logic here. But will fix.

lib/p2p/NodeList.ts
LePremierHomme: Other methods in this class are more graceful. I'm not sure if there's a good reason for that.
But anyway, I've changed the 2 methods i've added to use `assert`.

LePremierHomme: Actually it's possible that the node entry wouldn't exist, if we never had a successful connection/session before.

LePremierHomme:
@kilrau 
> > It’s easy to implement

> Then I'd say let's do it!

See https://github.com/ExchangeUnion/xud/pull/2009/commits/8e7ba3180766589fd93fd43d7e18604130a68917. Feel free to suggest changes.

LePremierHomme:
Will reply about the other p2p issues in a bit.

lib/p2p/NodeList.ts
sangaman: OK, I believe that's right actually, good to leave as is.

lib/p2p/NodeList.ts
sangaman: Yeah I don't think it's more logic per se, just naming variables so it's easier to see what the 200 and 1500 are for (and also a little easier to change or use elsewhere in the future, I believe).

lib/p2p/NodeList.ts
```diff
@@ -43,6 +43,11 @@ class NodeList extends EventEmitter {
   private static readonly BAN_THRESHOLD = -50;
   private static readonly MAX_REPUTATION_SCORE = 100;
 
+  private static readonly PRIMARY_RANK_CONN_FAILURES_THRESHOLD = 200;
```
sangaman, line 46: Minor nit, but I like "Peer" in place of "Rank" here. As is it's not clear by the name alone that this pertains to peers/nodes, although in the context of this file/class (NodeList) that may be clear enough. I'd also be fine with just dropping "Rank" from the names entirely so that it'll be a little shorter. Also OK to keep as is if you like the current names, just thought I'd mention it.

sangaman (commented):
Code & original concept look good to me, however I'm unsure about the newly added logic to make peers secondary (or make us not connect at all). For one thing, we are defacto silently "banning" peers if they are outbound and we don't have listening addresses, because under the secondary threshold we stop automatically connecting to them. But beyond that, I thought the purpose of this PR was to deal with a flood of bad connection attempts, but a low reputation score doesn't necessarily mean we'll have any trouble connecting to the peer.

What I actually thought @kilrau was suggesting originally was to make 200/1500 consecutive failed connection attempts a negative reputation event that lowers a node's score. That makes a little more sense to me (we don't want peers that are habitually unreachable over the long run), although I don't think it's necessary in this PR.

kilrau:
> Code & original concept look good to me, however I'm unsure about the newly added logic to make peers secondary (or make us not connect at all). For one thing, we are defacto silently "banning" peers if they are outbound and we don't have listening addresses, because under the secondary threshold we stop automatically connecting to them. But beyond that, I thought the purpose of this PR was to deal with a flood of bad connection attempts, but a low reputation score doesn't necessarily mean we'll have any trouble connecting to the peer.

We just discussed this on the daily call and this behaviour is fine as-is, since we eventually want to stop connecting to permanently dead peers (and there are quite some) and our default setups all have valid listening addresses, so if a long-dead peer comes back to life after several weeks or months, it will reconnect to **us**.

> What I actually thought @kilrau was suggesting originally was to make 200/1500 consecutive failed connection attempts a negative reputation event that lowers a node's score. That makes a little more sense to me (we don't want peers that are habitually unreachable over the long run), although I don't think it's necessary in this PR.

We discussed this too, hopefully a simple change: remove reputation events from this PR. (Reason: we are not sure if reconnection attempts/down-time should be immediately tied to reputation)

LePremierHomme:
> We discussed this too, hopefully a simple change: remove reputation events from this PR. (Reason: we are not sure if reconnection attempts/down-time should be immediately tied to reputation)

Connection failures are currently *not* tied to reputation, but ranking does consider both connection failures and reputation. Are suggesting to have ranking consider only the connection failures (as before my last commit)?

lib/p2p/NodeList.ts
LePremierHomme: Fixed.

kilrau:
> Are suggesting to have ranking consider only the connection failures (as before my last commit)?

Yes!

erkarl:
Think we can get this in after rebase? @sangaman

sangaman (approved):
utACK

PR #2011, opened 2020-12-02
feat: lock file

### LINKED ISSUE #1989: Use `.lock` file for data directory
When playing with two XUDs on a local regtest network, I noticed that they were using the same database and node public key. This is not a critical issue since I don't think that this could cause data corruption of any files, but it is an inconvenience that XUD does not prevent that from happening. There is no use case for using the same data directory for two nodes at the same time, so we should not let XUD do that since it could cause some weird behaviour.

What I propose is creating an empty `.lock` file for the data directory (or maybe node public key and database file) on startup and removing that file when XUD shuts down. When XUD detects that such a file exists, it should abort the startup and log that there might be an XUD running with that data directory already

sangaman:
This introduces a lock file for xud that is created in the xud data directory any time xud starts up. Anytime a lock file already exists, an error message is logged and xud exits. This prevents multiple xud processes from trying to use the same node key or directory at the same time.

The lock files are unique to each network, so running multiple processes with different networks (e.g. simnet and mainnet) is allowed. They are deleted when xud shuts down.

Closes #1989.

### DIFF
--- a/lib/Xud.ts
+++ b/lib/Xud.ts
@@ -35,9 +35,9 @@ class Xud extends EventEmitter {
   public service!: Service;
   private logger!: Logger;
   private config: Config;
-  private db!: DB;
-  private pool!: Pool;
-  private orderBook!: OrderBook;
+  private db?: DB;
+  private pool?: Pool;
+  private orderBook?: OrderBook;
   private rpcServer?: GrpcServer;
   private httpServer?: HttpServer;
   private grpcAPIProxy?: GrpcWebProxyServer;
@@ -46,6 +46,7 @@ class Xud extends EventEmitter {
   private swapClientManager?: SwapClientManager;
   private unitConverter?: UnitConverter;
   private simnetChannels$?: Subscription;
+  private lockFilePath?: string;
 
   /**
    * Create an Exchange Union daemon.
@@ -88,6 +89,24 @@ class Xud extends EventEmitter {
     }
 
     try {
+      if (this.config.instanceid === 0) {
+        // if we're not running with multiple instances as indicated by an instance id of 0
+        // create a lock file to prevent multiple xud instances from trying to run
+        // with the same node key and/or database
+        const lockFilePath = path.join(this.config.xudir, `${this.config.network}.lock`);
+        try {
+          await (await fs.open(lockFilePath, 'wx')).close();
+          this.lockFilePath = lockFilePath;
+        } catch (err) {
+          if (err.code === 'EEXIST') {
+            this.logger.error(`A lock file exists at ${lockFilePath}. Another xud process is running or a previous process exited ungracefully.`);
+            return;
+          } else {
+            throw err;
+          }
+        }
+      }
+
       if (!this.config.rpc.disable) {
         // start rpc server first, it will respond with UNAVAILABLE error
         // indicating xud is starting until xud finishes initializing
@@ -296,6 +315,10 @@ class Xud extends EventEmitter {
     // TODO: ensure we are not in the middle of executing any trades
     const closePromises: Promise<void>[] = [];
 
+    if (this.lockFilePath) {
+      closePromises.push(fs.unlink(this.lockFilePath).catch(this.logger.error));
+    }
+
     this.simnetChannels$?.unsubscribe();
 
     if (this.swapClientManager) {
@@ -318,7 +341,7 @@ class Xud extends EventEmitter {
     }
     await Promise.all(closePromises);
 
-    await this.db.close();
+    await this.db?.close();
     this.logger.info('XUD shutdown gracefully');
 
     this.emit('shutdown');
--- a/lib/service/Service.ts
+++ b/lib/service/Service.ts
@@ -5,6 +5,7 @@ import { ProvidePreimageEvent, TransferReceivedEvent } from '../connextclient/ty
 import { OrderSide, Owner, SwapClientType, SwapRole } from '../constants/enums';
 import { OrderAttributes, TradeInstance } from '../db/types';
 import Logger, { Level, LevelPriority } from '../Logger';
+import NodeKey from '../nodekey/NodeKey';
 import OrderBook from '../orderbook/OrderBook';
 import { Currency, isOwnOrder, Order, OrderPortion, OwnLimitOrder, OwnMarketOrder, OwnOrder, PeerOrder, PlaceOrderEvent } from '../orderbook/types';
 import Pool from '../p2p/Pool';
@@ -19,7 +20,6 @@ import { checkDecimalPlaces, sortOrders, toEip55Address } from '../utils/utils';
 import commitHash from '../Version';
 import errors from './errors';
 import { NodeIdentifier, ServiceComponents, ServiceOrder, ServiceOrderSidesArrays, ServicePlaceOrderEvent, ServiceTrade, XudInfo } from './types';
-import NodeKey from 'lib/nodekey/NodeKey';
 
 /** Functions to check argument validity and throw [[INVALID_ARGUMENT]] when invalid. */
 const argChecks = {
--- a/test/integration/Service.spec.ts
+++ b/test/integration/Service.spec.ts
@@ -111,7 +111,7 @@ describe('API Service', () => {
   });
 
   it('should remove an order', () => {
-    const tp = xud['orderBook'].tradingPairs.get('LTC/BTC')!;
+    const tp = xud['orderBook']!.tradingPairs.get('LTC/BTC')!;
     expect(tp.ownOrders.buyMap.has(orderId!)).to.be.true;
     const args = {
       orderId: '1',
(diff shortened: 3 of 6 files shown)


ghost (dismissed):
utACK

raladev:
- [ ] does not work for xud-docker env

Steps:
1. `bash xud.sh -b lock`
2. initialize the wallet and unlock it
3. docker rm -f xud_simnet_1
4. `bash xud.sh -b lock`

Actual result:
.lock file still exists
[Screenshot from 2020-12-02 16-13-44](https://user-images.githubusercontent.com/29906866/100877716-399bb800-34ba-11eb-93df-1ff56bc70c3d.png)
[Screenshot from 2020-12-02 16-13-49](https://user-images.githubusercontent.com/29906866/100877710-386a8b00-34ba-11eb-97cb-ab3222b23cf4.png)
[Screenshot from 2020-12-02 16-13-46](https://user-images.githubusercontent.com/29906866/100877711-399bb800-34ba-11eb-8316-d50a5661c2b6.png)

raladev (changes requested):
- [ ] does not work for xud-docker env

Steps:
1. `bash xud.sh -b lock`
2. initialize the wallet and unlock it
3. docker rm -f xud_simnet_1
4. `bash xud.sh -b lock`

Actual result:
.lock file still exists
[Screenshot from 2020-12-02 16-13-44](https://user-images.githubusercontent.com/29906866/100877716-399bb800-34ba-11eb-93df-1ff56bc70c3d.png)
[Screenshot from 2020-12-02 16-13-49](https://user-images.githubusercontent.com/29906866/100877710-386a8b00-34ba-11eb-97cb-ab3222b23cf4.png)
[Screenshot from 2020-12-02 16-13-46](https://user-images.githubusercontent.com/29906866/100877711-399bb800-34ba-11eb-8316-d50a5661c2b6.png)

michael1011 (dismissed):

sangaman:
>     * [ ]  does not work for xud-docker env
> 
> 
> Steps:
> 
>     1. `bash xud.sh -b lock`
> 
>     2. initialize the wallet and unlock it
> 
>     3. docker rm -f xud_simnet_1
> 
>     4. `bash xud.sh -b lock`
> 
> 
> Actual result:
> .lock file still exists
> [Screenshot from 2020-12-02 16-13-44](https://user-images.githubusercontent.com/29906866/100877716-399bb800-34ba-11eb-93df-1ff56bc70c3d.png)
> [Screenshot from 2020-12-02 16-13-49](https://user-images.githubusercontent.com/29906866/100877710-386a8b00-34ba-11eb-97cb-ab3222b23cf4.png)
> [Screenshot from 2020-12-02 16-13-46](https://user-images.githubusercontent.com/29906866/100877711-399bb800-34ba-11eb-8316-d50a5661c2b6.png)

Hm this is likely due to the sudden/ungraceful way of stopping the container, not giving it a chance to delete the lock file. Does "downing" the container give the same behavior? I can try testing that myself. I wonder if there's anything else to do from the xud perspective to delete the lock file with ungraceful shutdowns, but if the process is just killed abruptly then I'm pretty sure nothing can be done. I could maybe put in a prompt on the docker side to delete the lock file if one is encountered.

michael1011:
>  I wonder if there's anything else to do from the xud perspective to delete the lock file with ungraceful shutdowns, but if the process is just killed abruptly then I'm pretty sure nothing can be done.

How does Bitcoin Core handle this? The also have lockfiles for the database files and somehow managed to deal with them gracefully even when you force kill the Bitcoin Core process. But researching that is low prio.

> I could maybe put in a prompt on the docker side to delete the lock file if one is encountered.

In xud-docker we could simply check whether an xud container is running and if not, automatically remove the lockfile.

kilrau:
This could also happen on other ungraceful shutdowns and I believe we should handle it, otherwise we'll have to explain to our users how to delete lockfiles :/ I never had lockfile issues with any other software, even though I killed processes/containers, so there must be a trick to it.

Probably a stupid question, but can a lockfile depend on a process? https://stackoverflow.com/questions/1599459/optimal-lock-file-method/ and https://perl.plover.com/yak/flock/ recommend using `flock()` and the process actually tries to *obtain* the lock on the file, not judging by pure existence of the lock file as far as I understand it.

sangaman:
I'm used to deleting lock files with other applications that crash suddenly (due to power loss, etc). I believe I've had it happen with bitcoind too, and there are plenty of results for people getting `cannot obtain a lock on data directory` with bitcoin core. Conceptually I can't think of any you could delete a file inside a process that is suddenly killed.

Making it depend on a process could work potentially, but I don't know how we would play nice with running multiple processes on one machine which is a valid use case if they're using different networks or perhaps even different data directories.

sangaman:
I can confirm that `down` or `stop xud` in xud-docker handle the lock file gracefully. So I think from docker's perspective it makes sense to prompt the user if it encounters a lock file. I'll look into that separately.

I'll fix this PR to update the xud custom patch.

kilrau:
Thanks again for throwing this up @sangaman , we definitely want and need an advisory lock like this. I still have the following concerns though:

> I'm used to deleting lock files with other applications that crash suddenly (due to power loss, etc). 

I am not tbh, and I killed my bitcoind/lnd multiple times already including sudden power loss. I am just not convinced we are handling things right if a simple container kill, leaves the `.lock` file on disk and then xud as well-behaved process refuses to start on this data dir because the `.lock` file **exists**. I am not challenging the `.lock` file approach in general, just how it's currently handled in this PR by deleting the file on graceful shutdown and judging `.lock` file exists = data dir locked.

I definitely would want to avoid explaining to users how to delete a `.lock` file vs. protection in a case where multiple xud's try to run on the same data dir.

So I did a little test with my bitcoind:
- t0: bitcoind running (note the creation date of the `.lock` file - there were definitely multiple restarts & host reboots in between this date and today):
```
kilrau@beast:/media/SSD/bitcoind$ ls -la
total 47560
drwxr-xr-x 6 d    d        4096 Dec  3 10:23 .
drwxr-xr-x 5 d    d        4096 Nov 10 08:59 ..
-rw------- 1 root root        0 Jul 20 09:54 .lock
```
- t1: stop bitcoind gracefully, `.lock` file still exists
```
kilrau@beast:/media/SSD/bitcoind$ ls -la
total 79612
drwxr-xr-x 6 d    d        4096 Dec  3 10:34 .
drwxr-xr-x 5 d    d        4096 Nov 10 08:59 ..
-rw------- 1 root root        0 Jul 20 09:54 .lock
```
- t2: starting bitcoind again, `.lock` file still there, creation date unchanged:
```
kilrau@beast:/media/SSD/bitcoind$ ls -la
total 78600
drwxr-xr-x 6 d    d        4096 Dec  3 10:37 .
drwxr-xr-x 5 d    d        4096 Nov 10 08:59 ..
-rw------- 1 root root        0 Jul 20 09:54 .lock
```

So it looks like bitcoind does **not** delete the `.lock` file when shutting down, and when coming up judging the lock somehow differently than by pure existence of the file. Reading through https://github.com/bitcoin/bitcoin/issues/19167#issuecomment-698594254 it looks like bitcoind is using `fcntl` which offers advanced [file locks](https://www.informit.com/articles/article.aspx?p=23618&seqNum=4) and locks are placed on the `.lock` file itself without ever changing/deleting it. Since the author proposed a switch to `flock()` and that's what [I previously found too](https://github.com/ExchangeUnion/xud/pull/2011#issuecomment-737313524), can we [look into using it](http://beej.us/guide/bgipc/html/multi/flocking.html)?

michael1011:
One thing to keep in mind though when talking about low level file system calls @kilrau:

We are using Node. Which means we needs to have native C++ bindings to be able to call those things which does not only add a dependency but also might not be compatible with every OS and file system. The first Node `flock` library I found when googling for such things was incompatible with BSD for example

kilrau:
Maybe `fcntl` is the better choice then

michael1011:
> Maybe fcntl is the better choice then

May I remind you: we cannot use native file system calls without a native C++ (or Rust or whatever) dependency which might not be compatible with all file systems, operating systems and architectures

kilrau:
Ok. Reading between the lines, you'd vouch for sticking with the current lockfile exists = locked approach?

michael1011:
I don't know what to suggest tbh. Both approaches have their up and downsides and we should consider carefully

kilrau:
> I don't know what to suggest tbh. Both approaches have their up and downsides and we should consider carefully

Agreed, it's not clear what the best way is right now and at least I have concerns of potentially making things worse with this PR (few will benefit from what the lockfile prevents, more might run into troubles not being able to start xud because lockfile exists).

I converted this PR to draft, we can pick it up again once a good solution evolves.

kilrau:
Just dropping this here as an observation:
```
~$ sudo apt update
Reading package lists... Done
E: Could not get lock /var/lib/apt/lists/lock. It is held by process 2328 (packagekitd)
N: Be aware that removing the lock file is not a solution and may break your system.
E: Unable to lock directory /var/lib/apt/lists/
```

michael1011:
One thing that just came to my mind: what if we write the PID of the XUD process in the lockfile and [check if the PID is still running](https://stackoverflow.com/a/21296291) if the lockfile exists?

kilrau:
Hmm yeah that could work. Also should work on windows as per your link

PR #2013, opened 2020-12-03
style(lint): upgrade from tslint to eslint

ghost:
This PR upgrades our deprecated tslint linter to eslint. In addition to default eslint configuration I had to disable some of the rules in order not to generate huge code differences in this PR.

If we choose to adjust some of the rules we can do so in the follow-up PRs.

EDIT by @kilrau : depends on #2016


sangaman (commented):
I ran `npm install` from this branch, but it seems that linting isn't catching some common errors. I tried putting excess whitespace at the end of lines, and I tried removing the commas from the end of lines, and neither were fixed or detected by running `npm run lint`. Is the same happening for you? If so I think we should try to make sure that at least most of the old rules are preserved.

erkarl:
> I ran `npm install` from this branch, but it seems that linting isn't catching some common errors. I tried putting excess whitespace at the end of lines, and I tried removing the commas from the end of lines, and neither were fixed or detected by running `npm run lint`. Is the same happening for you? If so I think we should try to make sure that at least most of the old rules are preserved.

Indeed, the recommended configuration for eslint does not enforce code style. I'd keep eslint for code intelligence and for enforcing style use prettier instead https://github.com/ExchangeUnion/xud/pull/1676

kilrau:
Then let's get both in ?

sangaman:
> Then let's get both in ?

https://www.youtube.com/watch?v=OawrlVoQqSs

I agree if it's not too much work. I think it'd be nice to preserve the style enforcement so that there aren't PRs in the meantime that violate them that would then need to be cleaned up later.

kilrau:
Since prettier has a separate PR now (cheers @erkarl ), this deserves another review @sangaman

erkarl:
> Since prettier has a separate PR now (cheers @erkarl ), this deserves another review @sangaman

The prettier PR is against this branch. Let's get the prettier one in first, then this one.

sangaman:
> > Since prettier has a separate PR now (cheers @erkarl ), this deserves another review @sangaman
> 
> The prettier PR is against this branch. Let's get the prettier one in first, then this one.

You mean other way around? This first then prettier? Since prettier is based off this.

sangaman (approved):
Looks good.

I think I might enable the `"no-async-promise-executor"` in a follow-up PR. There are only two violations of that rule currently and it does make sense to me not to mix promise.then and async together. Same for `no-useless-escape` and `prefer-const`, maybe `no-useless-escape` as well. I also might look into the [eslint-config-airbnb](https://www.npmjs.com/package/eslint-config-airbnb) package since we were using the tslint airbnb variety before.

erkarl:
> You mean other way around? This first then prettier? Since prettier is based off this.

Oh, yes. Let me fix up the simulation tests in prettier branch so we can get them in at the same time we disable current tslint style enforcement.

sangaman:
> > You mean other way around? This first then prettier? Since prettier is based off this.
> 
> Oh, yes. Let me fix up the simulation tests in prettier branch so we can get them in at the same time we disable current tslint style enforcement.

Great, I'm ready to merge these once that's done, so as to minimize any conflicts/switching back and forth going forward.

PR #2015, opened 2020-12-03
fix(p2p): fix stall timer checks

### LINKED ISSUE #2010: Peer connection test cases
This issue is to collect a set of peer connection test cases with the goal to reliably reproduce issues regarding **connecting/re-connecting to peers** that continue to pop up: https://github.com/ExchangeUnion/xud/pull/2009#issuecomment-736737531

sangaman:
This fixes a regression introduced in #1994 that prevents enabling the stall timer for peers that checks whether peers are not responding to our packets in a reasonable timeframe and disconnecting those that are not. Without these checks, peers that have gone offline may remain in the list of connected peers, resulting in unexpected behavior and also preventing those peers from establishing new connections with us once they come back online.

Closes #2010.

I'm also hoping, but can't say for sure, that this fixes the test flakiness case that arose recently described here https://github.com/ExchangeUnion/xud/issues/1771#issuecomment-732289609.

### DIFF
--- a/lib/p2p/Peer.ts
+++ b/lib/p2p/Peer.ts
@@ -617,7 +617,7 @@ class Peer extends EventEmitter {
   }
 
   private initStall = (): void => {
-    if (this.status !== PeerStatus.Closed) {
+    if (this.status === PeerStatus.Closed) {
       return;
     }
     assert(!this.stallTimer);


michael1011 (approved):

raladev (approved):

LePremierHomme:
Sorry, my bad!

sangaman:
> Sorry, my bad!

It happens! I overlooked it too when reviewing. I'm thinking a bit about a good regression test to add in to prevent such a regression in the future, it's a bit awkward because we'd need to wait 5-10 seconds for the stall timer to kick in which is a bit long of a delay for the test, but maybe I can reduce the delay without messing with any of the key functionality.

PR #2022, opened 2020-12-04
fix(rpc): 0 remaining quantity on full remove 

### LINKED ISSUE #2020: removeorder responce returns partial removal message instead of full
Steps:
1. place order
2. remove it

Actual result:
message says that order was removed partially and full amount of order displayed as remaining
Screenshot from 2020-12-04 18-20-59

Expected result:
`Order fully removed`

Note:
Partial removing works fine (correct message and value)

sangaman:
This fixes a bug where fully removing an order via `RemoveOrder` would return the entire quantity of the order as the remaining quantity instead of zero. The problem was that when removing a quantity in full, we would simply delete it from our collection of orders but not modify its quantity, which is used (since #1986) for informational purposes on the rpc response.

It also tidies up the cli output a bit so as not to display zero amounts for holds or remaining quantity.

Fixes #2020

### DIFF
--- a/lib/cli/commands/removeorder.ts
+++ b/lib/cli/commands/removeorder.ts
@@ -26,7 +26,7 @@ const displayOutput = (orderId: string, removeOrderResponse: RemoveOrderResponse
   } else {
     let message = `Order ${orderId} partially removed`;
     if (remainingQuantity) {
-      message += `, ${remainingQuantity} ${removedCurrency} remaining quantity`;
+      message += `, ${satsToCoinsStr(remainingQuantity)} ${removedCurrency} remaining quantity`;
     }
     if (quantityOnHold) {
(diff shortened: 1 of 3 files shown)


erkarl (dismissed):

raladev (changes requested):
- [ ] would be good to change sats to user values
```
simnet > removeorder 1 0.1
Order 1 partially removed, 90000000 LTC remaining quantity
simnet > removeorder 1 0.8
Order 1 partially removed, 10000000 LTC remaining quantity
```

sangaman:
@raladev thanks for catching, can you check again?

erkarl (approved):

kilrau (approved):

raladev (approved):

PR #2023, opened 2020-12-05
feat: new grpc call for subscribing alerts and low balance alert (#864)

### LINKED ISSUE #864: [Concept] Channel unbalanced alerts

rsercano:
Resolves #864 

This adds a new middleware between alerts that are being thrown during xud flows and grpc level as we discussed @sangaman But I haven't created two separate PRs, it doesn't make sense since we add a single alert instead of two as in the previous PR (#1984)

- Created a new middleware `Alerts` class.
- I added time to alert message let me know if want me to remove it or put it to json output.
- Minimum balance alert threshold to re-throw is 10 seconds, let me know if you want to change.
- I added a TODO to `Alerts` class, it's because we need a clear mechanism there for the map.

This requires a full re-test sorry @raladev


rsercano:
An unrelated test is failing can you please enlighten me https://github.com/ExchangeUnion/xud/pull/2023/checks?check_run_id=1503260138#step:7:351 ?

erkarl:
Thanks for the PR.

> Created a new middleware Alerts class.

Why do we need a a new class `Alerts`? Currently, the class has no public methods and it's just a wrapper event emitter for swap client. Would it make sense to move this to `SwapClientManager`? Thoughts @sangaman @rsercano

rsercano:
> Thanks for the PR.
> 
> > Created a new middleware Alerts class.
> 
> Why do we need a a new class `Alerts`? Currently, the class has no public methods and it's just a wrapper event emitter for swap client. Would it make sense to move this to `SwapClientManager`? Thoughts @sangaman @rsercano?

Since alerts are not specific to swaps/swapclients actually it was @sangaman's idea to create a new class, which I think makes more sense because in a further time we can add more alerts to xud flow's different places.

raladev:
@rsercano can u please rebase it ? (need to get connext changes for vector from master).

rsercano:
> @rsercano can u please rebase it ? (need to get connext changes for vector from master).

rebased @raladev

raladev (changes requested):
- [x] u dont need to throw alert for remote balance of connext currencies because this value is not user controllable, and depends on collateralization 
`(Tue Dec 08 2020 11:20:58 GMT+0000) LowTradingBalance: Remote trading balance (0 ETH) is lower than 10% of trading capacity (15 ETH)`
- we need to decide how it will look for connext or we just should support this alert only for lnd (looks correct for me) @rsercano @sangaman @kilrau 

- [x] text message contains time, but json does not contain timestamp, would be good to add. 
```
(Tue Dec 08 2020 11:21:22 GMT+0000) LowTradingBalance: Remote trading balance (0 ETH) is lower than 10% of trading capacity (15 ETH)
^Csimnet > streamalerts -j
Successfully connected, streaming
{
  "type": "LowTradingBalance",
  "payload": {
    "totalBalance": 1500000000,
    "side": "Remote",
    "sideBalance": 0,
    "bound": 10,
    "currency": "ETH"
  }
}
```

kilrau:
>     * we need to decide how it will look for connext or we just should support this alert only for lnd (looks correct for me) @rsercano @sangaman @kilrau

The user will be able to influence this. At the moment the `requestCollateral` is hidden from the user, but once users need to pay for collateral we will have to show it to the user. We can keep this alert imo.

rsercano:
@raladev added date to json output as [Unix time](https://en.wikipedia.org/wiki/Unix_time) and formatted in formatted output :)

kilrau:
Let's give this another round of review @sangaman @raladev

raladev (dismissed):

test/integration/Service.spec.ts
```diff
@@ -210,4 +213,38 @@ describe('API Service', () => {
...
+        const secondLowBalanceEvent = cold('--a-b|') as Observable<BalanceAlert>;
+        const cancelled = cold('-') as Observable<void>;
+
+        const lowBalanceObservables: Observable<BalanceAlert>[] = [firstLowBalanceEvent, secondLowBalanceEvent];
+        const finalObservable = service['getMergedObservable$'](lowBalanceObservables, cancelled);
+        expectObservable(finalObservable).toBe('-aa-(bb)c---');
+      });
+    });
+
+    it('should cancelled with cancelled$', async () => {
```
erkarl, line 238:
```suggestion
    it('should cancel with cancelled$', async () => {
```

ghost (dismissed):

lib/connextclient/ConnextClient.ts
```diff
@@ -50,13 +51,15 @@ interface ConnextClient {
   on(event: 'htlcAccepted', listener: (rHash: string, amount: number, currency: string) => void): this;
   on(event: 'connectionVerified', listener: (swapClientInfo: SwapClientInfo) => void): this;
   on(event: 'depositConfirmed', listener: (hash: string) => void): this;
+  on(event: 'lowTradingBalance', listener: (alert: BalanceAlert) => void): this;
```
sangaman, line 54: This can use the new `Alert` type, I described above, same for other places the `lowTradingBalance` event is defined.

lib/swaps/SwapClient.ts
```diff
@@ -228,6 +228,35 @@ abstract class SwapClient extends EventEmitter {
...
+  protected checkLowBalance = (
+    remoteBalance: number,
+    localBalance: number,
+    totalBalance: number,
+    alertThreshold: number,
+    currency: string,
+    emit: Function,
+  ) => {
+    if (localBalance < alertThreshold) {
+      emit('lowTradingBalance', {
```
sangaman, line 240: Is there a reason this is taking the `emit` function here as a parameter instead of `this.emit()` with types for the event? I'm thinking the `emit` and `on` functions can be defined in this class (alongside other generic swap client events like `connectionVerified`) rather than in `LndClient` and `ConnextClient` since they seem to be using the same types. Then this can use `this.emit`.

proto/xudrpc.proto
```diff
@@ -416,6 +439,23 @@ message Balance {
...
+  // The total balance.
+  uint64 total_balance = 1 [json_name = "total_balance"];
+  // The side of the low balance.
+  enum Side {
+    REMOTE = 0;
+    LOCAL = 1;
+  }
+  Side side = 2 [json_name = "side"];
+  // The bound of the low balance in percentage.
+  uint32 bound = 3 [json_name = "bound"];
```
sangaman, line 452: Maybe this should be `percent` instead of `bound`? Seems a bit more clear to me. And the comment can say `The percent of the total channel capacity that is on this side of the channel`

proto/xudrpc.proto
```diff
@@ -416,6 +439,23 @@ message Balance {
...
+  // The current side balance.
```
sangaman, line 453: `The balance in satoshis on this side of the channel`

proto/xudrpc.proto
```diff
@@ -416,6 +439,23 @@ message Balance {
   uint64 unconfirmed_wallet_balance = 6 [json_name = "unconfirmed_wallet_balance"];
 }
 
+message BalanceAlert {
+  // The total balance.
+  uint64 total_balance = 1 [json_name = "total_balance"];
```
sangaman, line 444: I think this would be better called `capacity` since that is the term I'm most used to seeing when referring to channel sizes (see for example https://1ml.com/channel/722305472288391168). And the comment can say "The total capacity of the channel this alert is for"

proto/xudrpc.proto
```diff
@@ -416,6 +439,23 @@ message Balance {
...
+  enum Side {
+    REMOTE = 0;
+    LOCAL = 1;
+  }
+  Side side = 2 [json_name = "side"];
+  // The bound of the low balance in percentage.
+  uint32 bound = 3 [json_name = "bound"];
+  // The current side balance.
+  uint64 side_balance = 4 [json_name = "side_balance"];
+  // The currency of the alert.
```
sangaman, line 455: The currency of the channel this alert is for

test/integration/Service.spec.ts
sangaman: This is still an issue (albeit minor).

proto/xudrpc.proto
```diff
@@ -416,6 +439,23 @@ message Balance {
...
+  // The side of the low balance.
+  enum Side {
+    REMOTE = 0;
+    LOCAL = 1;
+  }
+  Side side = 2 [json_name = "side"];
+  // The bound of the low balance in percentage.
+  uint32 bound = 3 [json_name = "bound"];
+  // The current side balance.
+  uint64 side_balance = 4 [json_name = "side_balance"];
```
sangaman, line 454: If we do rename `total_balance` to `capacity` as suggested above, I figure this can just be `balance` instead of `side_balance` since it's clearly implied that it's the balance for this side of the channel.

lib/alerts/Alerts.ts
```diff
@@ -0,0 +1,63 @@
...
+  private listenLowTradingBalanceAlerts(swapClientManager: SwapClientManager) {
+    const lndClients = swapClientManager.getLndClientsMap().values();
+    for (const lndClient of lndClients) {
+      lndClient.on('lowTradingBalance', this.onLowTradingBalance);
+    }
+    swapClientManager.connextClient?.on('lowTradingBalance', this.onLowTradingBalance);
+  }
+
+  private onLowTradingBalance = (balanceAlert: BalanceAlert) => {
+    const stringRepresentation = JSON.stringify(balanceAlert);
```
sangaman, line 38: I think I see the desire to convert to a string here, but I'm concernedthis will cause some issues using plain old JSON.stringify, which isn't guaranteed to give the same string for equivalent objects. It's also a bit expensive of a procedure, and uses up more memory in the map.

Doesn't have to be addressed in this PR since it's not that simple, but maybe it can be addressed at the same time as cleanup. I'm thinking it'd make sense to add a timestamp property to the `Alert` object, and then convert alert's to a string identifier like `[type]-[channel_id]` rather than converting the entire object to a json string.

For now I think add a TODO here and it's ok with me.

sangaman (changes requested):
The middleware design is nice, but I made a few comments/suggestions. I still question the value of the channel balance alerts, per my prior comments in related PRs, but I won't stand in the way of it. If we decide it's not helpful in the future, we can remove the low balance alert but keep this alerts framework.

test/integration/Service.spec.ts
rsercano: my bad, failed to commit 🤦

lib/alerts/Alerts.ts
```diff
@@ -0,0 +1,63 @@
+import { EventEmitter } from 'events';
+import { BalanceAlert } from './types';
+import SwapClientManager from '../swaps/SwapClientManager';
+import { MIN_BALANCE_ALERT_THRESHOLD_IN_MS } from './consts';
+import Logger from '../Logger';
+import { AlertType, ChannelSide } from '../constants/enums';
+import { satsToCoinsStr } from '../cli/utils';
+
+interface Alerts {
+  on(event: 'alert', listener: (alert: any) => void): this;
```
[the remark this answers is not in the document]
rsercano: That's brilliant thank you

proto/xudrpc.proto
rsercano: I changed comment as `The percent of the total trading balance.` since this's actually trading balance alert, but changed the field name as `percent`

proto/xudrpc.proto
rsercano: But we show total trading balance, instead of channel balances?

proto/xudrpc.proto
rsercano: As for total balance, we actually trigger alert for whole trading balance.

lib/alerts/Alerts.ts
rsercano: Yes, you're right, alerts already have timestamps attached within BaseAlert, adding `TODO` there, but I guess we should prepare **cleanup** as soon as possible to avoid memory issues.

rsercano: Maybe even within this PR, wdyt? @sangaman

lib/swaps/SwapClient.ts
rsercano: Yes actually, I was thinking the same, but if I do the what you said, then I have to define `BaseAlert` fields such as alert `type, message, date` which are being defined in Alerts middleware class.

raladev (dismissed):

lib/swaps/SwapClient.ts
sangaman: You're avoiding that not by passing around the `emit` function but by defining it as just a `Function` - which loses all type checking entirely.  This could be done by just hard casting whatever object to the desired type, but that still robs us of the benefits of type checking and could lead to bugs down the road.

I started writing up a proposal for how to refactor this but it wound up being easier just to refactor it myself, so I'll push a commit here to be squashed. Basically you can define a `BalanceAlertEvent` type that only contains the fields specific to balance alerts, and then use that on the swap client level while keeping the `BalanceAlert` within the Alerts middleware. This way we can preserve all the type checking.

proto/xudrpc.proto
sangaman: Maybe I was confused then, this field is the total balance that the node has across all channels? If so then it's definitely not capacity, but that would seem out of place to me.

proto/xudrpc.proto
sangaman: I'm confused, what exactly is `side_balance` and `total_balance` supposed to represent? If it's not clear to me it definitely won't be clear to people trying to make sense of our API.

lib/alerts/Alerts.ts
sangaman: Up to you, it's not a trivial effort.

sangaman:
I pushed a commit to do some refactoring and to preserve type checking. As a general rule I really think we ought to preserve types whenever possible, which are lost with declaring `emit: Function`.

My last concern is in regards to the proto file field naming & comments/description for purposes of API documentation. See my comments on the unresolved issues above.

kilrau:
also keep an eye on the failing builds

sangaman:
> also keep an eye on the failing builds

Thanks, just a linting issue. Fixed.

proto/xudrpc.proto
rsercano: @raladev @kilrau

rsercano:
I guess @sangaman already pushed a commit for this review: https://github.com/ExchangeUnion/xud/pull/2023#discussion_r547105133 Is there anything to do for me here? @raladev @sangaman

raladev:
> I guess @sangaman already pushed a commit for this review: [#2023 (comment)](https://github.com/ExchangeUnion/xud/pull/2023#discussion_r547105133) Is there anything to do for me here? @raladev @sangaman

- [ ] Resolve conflicts
- [ ] Check tests, they contain JS errors

raladev (dismissed):

proto/xudrpc.proto
sangaman: I think we still need an answer here and I'm not sure what it is.

lib/grpc/GrpcService.ts
```diff
@@ -991,6 +992,35 @@ class GrpcService {
...
+    const cancelled$ = getCancelled$(call);
+    this.service.subscribeAlerts((serviceAlert: Alert) => {
+      const alert = new xudrpc.Alert();
+      alert.setType(serviceAlert.type as number);
+      alert.setMessage(serviceAlert.message);
+      alert.setDate(serviceAlert.date);
+      if (serviceAlert.type === AlertType.LowTradingBalance) {
+        const balanceServiceAlert = serviceAlert as Alert;
+        const balanceAlert = new xudrpc.BalanceAlert();
+        balanceAlert.setBound(balanceServiceAlert.bound);
```
sangaman, line 1012: Bound is not in the xudrpc.proto definition currently. Regenerating the protos results in compile errors currently. Do we want this field?

proto/xudrpc.proto
sangaman: Can you change this or would you like me to?

sangaman (changes requested):
The proto file still needs work unfortunately. The `bound` field is not in the proto definition currently, but it's being used in the xud code. I also still have some open questions for the naming and commenting of the balance alert fields in the proto file, we need to make sure they are clear.

Once that's done we can squash and merge everything. I'm taking care of the merge conflicts, but we need another commit to address the proto file concerns. See if my questions make sense to you @kilrau, I can also take care of the final changes once it's clear to me which exactly which fields we want on the balance alert message.

lib/grpc/GrpcService.ts
rsercano: fixed

rsercano:
https://github.com/ExchangeUnion/xud/pull/2023#discussion_r551751553 @kilrau @raladev

kilrau:
I synced with @michael1011 about this and we will not use this information for a channel rebalancer any time soon, the UI will rather use it to display a notification about a general imbalance on a certain asset (not even a specific channel). I think the fields in this PR deliver this, so if if passed your review with that in mind, we can merge as is @sangaman

raladev (approved):

Issue #2019, opened 2020-12-04
Connext followups


erkarl:
- [ ] estimate fee for withdrawals
- [ ] support on-chain transfers from signer address
- [ ] monitor channel address and reconcile deposit on demand instead of using interval
- [ ] support for non-ETH currencies
- [ ] setup testnet router
- [ ] setup mainnet router
- [ ] simulation tests vector v2

kilrau:
>  arm64 images

Tested this today, almost there, just needs switch to manifests.

kilrau:
Reminder to add "nicer error message" for xud to the list (from the connext call, I believe it was for insufficient/no funds)

PR #2026, opened 2020-12-07
feat(connext): estimate fee for closechannel

ghost:
This PR adds fee estimation logic to `ConnextClient.closeChannel`. Previously, a static fee was used.

Related issue: https://github.com/ExchangeUnion/xud/issues/2019

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -42,6 +42,7 @@ import {
   ConnextBlockNumberResponse,
   ConnextChannelDetails,
   GetBlockByNumberResponse,
+  EthproviderGasPriceResponse,
 } from './types';
 import { parseResponseBody } from '../utils/utils';
 import {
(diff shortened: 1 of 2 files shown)


raladev (dismissed):

lib/connextclient/ConnextClient.ts
```diff
@@ -900,6 +901,21 @@ class ConnextClient extends SwapClient {
     return parseInt(blockNumberResponse.result, 16);
   }
 
+  private getGasPrice = async () => {
```
sangaman, line 904: A comment here would be helpful too in regards to what exactly this returns - is this getting the most recent average gas price? Or is this derived from some fee estimation logic built into geth?

kilrau: Shouldn't break anything, here a heads up for this upcoming change in geth how gas prices are calculated (moving from the current auction model/fee market to a moving average): https://github.com/ethereum/go-ethereum/pull/21971 

thanks for pointing me at it @michael1011

michael1011: The method call Karl is using returns data from GETHs internal fee estimation logic.

What Kilian mentioned is EIP-1559 which will be a breaking change in the way Ethereum handles gas prices and eventually be activated in a hardfork. But that won't matter for us since the REST endpoints of GETH won't change because of that.

erkarl: Comment added.

sangaman (approved):

PR #2027, opened 2020-12-07
feat(connext): reconcile deposit when channel balance changes

ghost:
This PR changes the logic around `/deposit` calls. Previously, we were creating new commitment transaction every 5 seconds. With this PR we'll only create new state updates when the channel balance of ETH changes.

Related issue: https://github.com/ExchangeUnion/xud/issues/2019

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -41,7 +41,7 @@ import {
 } from './types';
 import { parseResponseBody } from '../utils/utils';
 import { Observable, fromEvent, from, defer, Subscription, throwError, interval, timer, combineLatest, of } from 'rxjs';
-import { take, pluck, timeout, filter, mergeMap, catchError, mergeMapTo } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


raladev (dismissed):

lib/connextclient/ConnextClient.ts
```diff
@@ -464,18 +465,20 @@ class ConnextClient extends SwapClient {
     if (this._reconcileDepositSubscriber) {
       this._reconcileDepositSubscriber.unsubscribe();
     }
-    // TODO: increase interval after testing to 30+ sec
-    this._reconcileDepositSubscriber = interval(5000)
+    const ethBalance$ = interval(30000).pipe(
```
sangaman, line 468: I think it may be helpful in the future to comment what this is doing, just reusing some of the text from this PR, because it's not obvious what this is doing particularly if you're not familiar with `distinctUntilChanged` and such:

`// only create new state updates when the channel balance of ETH changes`

sangaman (dismissed):

erkarl:
@sangaman comments added - going to merge this. If the comments are not clear, I'll add more in a follow-up PR.

PR #2034, opened 2020-12-09
style: add airbnb-typescript eslint rules

sangaman:
This adds and enforces airbnb typescript linting rules to eslint to approximate the rules from the prior tslint implementation. Several new rules are enforced compared to tslint, and several new rules have been disabled if they introduced too much change from the existing code.

This also adds the eslint prettier config rules to prevent overlap and conflicts with prettier for strictly stylistic/formatting rules.

The linting for tests is still pretty limited, but it was limited before as well.


erkarl:
Linter is not passing for the latest commibt build.

erkarl (approved):

erkarl:
Awesome. Thanks.

PR #2036, opened 2020-12-11
feat(connext): increase channel timeout

erkarl:


erkarl:
Going to merge this to see if it fixes an issue with swaps in https://github.com/ExchangeUnion/xud-docker/pull/821

erkarl:
The issue was that our current channel timeout was smaller than the default transfer timeout. This was causing transfers to fail because it did not pass the validations that were added in 0.0.33 connext version.

PR #2039, opened 2020-12-12
fix: removeallorders won't be interrupted by a removeorder error 

### LINKED ISSUE #2030: removeallorder command may be interrupted by any valid removeorder error
Steps to reproduce:
1. place `buy 0.1 ETH/BTC 0.1 1`
2. make self-match - `buy 0.1 ETH/BTC 0.1 2`
3. place several other orders  - `buy 0.1 ETH/BTC 0.1`
4. `removeallorders`

Actual result:
removeallorders call was interrupted by removing unexisted order (https://github.com/ExchangeUnion/xud/pull/1557) and as a result, other orders were not removed

```
simnet > sell 0.1 ETH/BTC 0.1 1
remaining 0.1 ETH entered the order book as cd7b7f10-38de-11eb-b217-d34be1dc0c45
no matches found
simnet > buy 0.1 ETH/BTC 0.1 1
6 ALREADY_EXISTS: order with local id 1 already exists
simnet > buy 0.1 ETH/BTC 0.1 2
9 FAILED_PRECONDITION: channel collateralization in progress, please try again in ~1 minute
simnet > buy 0.1 ETH/BTC 0.1 2
matched 0.1 ETH @ 0.1 with own order cd7b7f10-38de-11eb-b217-d34be1dc0c45
simnet > buy 0.1 ETH/BTC 0.1 
remaining 0.1 ETH entered the order book as ddf7c5b0-38de-11eb-b217-d34be1dc0c45
no matches found
simnet > buy 0.1 ETH/BTC 0.1
remaining 0.1 ETH entered the order book as de99ab00-38de-11eb-b217-d34be1dc0c45
no matches found
simnet > buy 0.1 ETH/BTC 0.1
remaining 0.1 ETH entered the order book as df19fe90-38de-11eb-b217-d34be1dc0c45
no matches found
simnet > listorders

Trading pair: ETH/BTC
│ 0.1           │ 0.1         │ TellPotato       │               │             │                  │
│ 0.1           │ 0.1         │ TellPotato       │               │             │                  │
│ 0.1           │ 0.1         │ TellPotato       │               │             │                  │

Trading pair: LTC/BTC
simnet > removeallorders
Error: 5 NOT_FOUND: order with id cd7b7f10-38de-11eb-b217-d34be1dc0c45 could not be found
simnet > listorders

Trading pair: ETH/BTC
│ 0.1           │ 0.1         │ TellPotato       │               │             │                  │
│ 0.1           │ 0.1         │ TellPotato       │               │             │                  │
│ 0.1           │ 0.1         │ TellPotato       │               │             │                  │

Trading pair: LTC/BTC
simnet > 
```

Expected result:
removeallorders call can not be interrupted by valid removeorder call errors, but only by assertions.

rsercano:
resolves #2030 

- This adds a new field to GRPC for errored orders while removing, and prints them out formatted.

But I'm still wondering why a self-match doesn't remove the order? 🤔


lib/orderbook/OrderBook.ts
```diff
@@ -946,17 +946,22 @@ class OrderBook extends EventEmitter {
...
-        onHoldOrderLocalIds.push(localId);
+      try {
+        const { onHoldQuantity } = this.removeOwnOrderByLocalId(localId, true);
+        if (onHoldQuantity === 0) {
+          removedOrderLocalIds.push(localId);
+        } else {
+          onHoldOrderLocalIds.push(localId);
+        }
+      } catch (ex) {
+        erroredOrdersByLocalId.set(localId, ex.message);
```
erkarl, line 960: @sangaman what do you think of this approach? Shouldn't `removeallorders` work in any case or is there a genuine case when we cannot remove some specific orders?

rsercano: Since removeorder throws exception in any case of failures, maybe we can extend to catch some specific errors here, will wait @sangaman 's feedback

raladev (approved):

lib/orderbook/OrderBook.ts
sangaman: Yes I don't think this is the right approach. I suspect there may be orders that aren't getting removed from `localIdMap` even when they are removed from the rest of the order book, therefore we may sometimes be removing orders that no longer exist. That's what throws an error and causes the issue.

I'm interested in figuring out *why* we get not found errors here in the first place, but that can be a separate PR. Fow now, I think all that needs to be done is to log a `debug` message if the error has the code for `ORDER_NOT_FOUND` saying that we tried to remove an order that no longer exists, and log an error message in any other case. There *shouldn't* be any cases where we fail to remove an order, and if the reason we can't remove it is because it doesn't exist, I don't think we need to return anything via rpc about that order since it is indeed removed and the end state will be that there are no orders remaining, which is what we want.

sangaman (changes requested):
See below, I think we should only add the `catch` block in this PR and drop the rpc/cli changes. That will allow the loop to continue removing other orders so that the end state is no orders remaining in the order book.

raladev (approved):

sangaman (approved):
Good as a quick solution to the issue. I'll add some error logging to the `catch` block here later.

PR #2043, opened 2020-12-15
feat(connext): multi currency support

ghost:
This PR adds support for withdrawing from on-chain address. In the process, `ethers` dependency is added to make development flow with the eth provider considerably faster. Previously, we were using the json-rpc API directly.

I'm planning to extract Connext swap client into a separate module so that the dependency will be opt-in (only for those who activate the swap client).


lib/connextclient/ConnextClient.ts
```diff
@@ -1045,13 +1084,33 @@ class ConnextClient extends SwapClient {
...
+      this.getTokenAddress(currency),
+      destination
+    );
+    broadcastTransaction$.subscribe({
+      // portal pack to imperative world
+      next: (transaction) => {
+        this.logger.info(`on-chain transfer broadcasted, transaction hash: ${transaction.hash}`);
+      },
+      error: (e) => {
+        this.logger.error(`could not broadcast on-chain transfer: ${JSON.stringify(e)}`);
```
michael1011, line 1106: Should we really use the broadcast terminology? Why not just `on-chain transfer sent, transaction hash: ` and `could not send onchain transfer`?

lib/connextclient/ethprovider.ts
```diff
@@ -0,0 +1,60 @@
...
+      name: 'simnet',
+      chainId,
+    },
+  );
+  const erc20abi = ['function balanceOf(address) view returns (uint)', 'function transfer(address to, uint amount)'];
+  const signer = ethers.Wallet.fromMnemonic(seed).connect(provider);
+  const erc20 = new ethers.Contract(contractAddress, erc20abi, signer);
+  // convert promises to observables
+  const erc20balance$ = from(erc20.balanceOf(signer.address)) as Observable<BigNumber>;
+  const ethBalance$ = from(provider.getBalance(signer.address));
```
michael1011, line 43: Why do we need the Ether balance for sending ERC20 tokens?

lib/connextclient/ethprovider.ts
```diff
@@ -0,0 +1,60 @@
...
+      url: `http://${host}:${port}/ethprovider/${chainId}`,
+    },
+    {
+      name: 'simnet',
+      chainId,
+    },
+  );
+  const erc20abi = ['function balanceOf(address) view returns (uint)', 'function transfer(address to, uint amount)'];
+  const signer = ethers.Wallet.fromMnemonic(seed).connect(provider);
+  const erc20 = new ethers.Contract(contractAddress, erc20abi, signer);
```
michael1011, line 40: Call my old fashioned but I like classes with constructors that initialize objects like the `provider` and simple function that only have a couple arguments. But you do you :shrug:

lib/connextclient/ConnextClient.ts
```diff
@@ -1045,13 +1084,33 @@ class ConnextClient extends SwapClient {
...
+
+    // This is a bridge between the imperative and functional code. First step is to get rid of all of the external state.
+    const broadcastTransaction$ = onChainSendERC20(
+      this.host,
+      this.port,
+      CHAIN_IDENTIFIERS[this.network],
+      this.seed,
+      this.getTokenAddress(currency),
+      destination
+    );
```
michael1011, line 1099: Aren't you missing the `units: string` argument here?

lib/connextclient/ethprovider.ts
```diff
@@ -0,0 +1,60 @@
+import { Observable, from, combineLatest } from 'rxjs';
+import { ethers, BigNumber } from 'ethers';
+import { mergeMap } from 'rxjs/operators';
+
+type OnChainTransaction = {
```
michael1011, line 5: Ethers has great TypeScript support in my experience. They don't expose a type for this?

michael1011 (commented):
Concept ACK. Just some minor comments form my side

lib/connextclient/ethprovider.ts
erkarl: This in preparation for adding balance verification logic that currently resides in `ConnextClient.withdraw`.

lib/connextclient/ethprovider.ts
erkarl: I'm planning to refactor this function so that it receives signer (with provider) as an argument.

lib/connextclient/ConnextClient.ts
erkarl: Yes, indeed :)

lib/connextclient/ethprovider.ts
erkarl: No, since it's a contract function.
```
readonly [key: string]: ContractFunction | any;
```

lib/connextclient/ConnextClient.ts
erkarl: I was thinking we might want to wait for the transfer event for emitting `complete` and `next` when broadcasted. But since complete is not implemented, yet - I'll change this to sent terminology.

erkarl: Changed.

lib/connextclient/ethprovider.ts
erkarl: I went ahead and removed it right now to keep things easier. The checks are being done before this function is called.

lib/connextclient/ConnextClient.ts
michael1011: Just fyi: those events are emitted when the transaction is included in a block and not when it is accepted in the mempool

erkarl: Yup - good to know. We don't really care about that right now. The swapclient interface will expect `Promise<transactionId: string>`.

lib/connextclient/ethprovider.ts
michael1011: What about [ContractTransaction](https://github.com/ethers-io/ethers.js/blob/master/packages/contracts/src.ts/index.ts#L90) or one of the types it extends from?

erkarl: Yeap, good idea. Changed to `ContractTransactions` for the transfer call, but for balance calls I'm sticking with `Observable<ethers.BigNumber>` type since it's more accurate.

erkarl:
This PR has evolved into:
- adding support for `xucli walletwithdraw` for ETH/ERC20
- monitoring channel for ERC20 deposits -> creating a new commitment transaction so that ERC20 tokens show up under channel balance"

Everything in the `ethprovider.ts` will be converted to a separate module. This in preparation for extracting `ConnextClient` into a separate module.

The PR is reviewable commit-by-commit so I'm not squashing the commits, yet.

jest.config.js
```diff
@@ -1,8 +1,10 @@
 module.exports = {
   preset: 'ts-jest',
+  testMatch: ["<rootDir>/test/jest/**/*.ts"],
   globals: {
     'ts-jest': {
-      tsConfig: '<rootDir>/tsconfig.json'
+      tsconfig: '<rootDir>/tsconfig.json'
+
```
michael1011, line 7: Why blank line?

test/jest/Connext.spec.ts
```diff
@@ -105,25 +108,28 @@ describe('ConnextClient', () => {
...
-            amount: 123,
-            fee: 1,
-          });
-        } catch (e) {
-          expect(e).toMatchSnapshot();
-        }
-      });
+    });
+
+    it('fails with custom fee', async () => {
```
michael1011, line 120: Why are we not allowing custom fees?

michael1011 (commented):
utACK

Fixing those lint warnings would be nice though

test/jest/Connext.spec.ts
erkarl: Not implemented, yet. This is porting over the functionality that we had for V1 to V2. We'll of course support this in the future.

jest.config.js
erkarl: Removed. Thanks.

sangaman:
> utACK
> 
> Fixing those lint warnings would be nice though

I left those warnings in when working on the eslint rules, they weren't trivial to fix but they also seemed like things that made sense to change (like redeclaring variables, which can be confusing and lead to some hard to find bugs imo), so I figured the warnings would be a good reminder to address later and to prevent new code from violating the rule. I didn't realize how annoying they'd be in the PR code diffs, but I guess that's a good motivator to actually fix these soonish.

That said, it doesn't look like this PR is introducing any new warnings, so I wouldn't let that hold this up.

.eslintrc.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   },
   rules: {
     'no-empty': 'off',
+    "object-curly-newline": 0,
```
sangaman, line 9: What is this for? It doesn't seem to cause any linting issues if I remove the line.

jest.config.js
```diff
@@ -1,8 +1,9 @@
 module.exports = {
   preset: 'ts-jest',
+  testMatch: ["<rootDir>/test/jest/**/*.ts"],
```
sangaman, line 3: I think this would conflict with the npm scripts which specify `testMatch` on a per-npm-script basis but share this same jest config file. Were you having a problem running jest tests?

sangaman (commented):
Looks good so far. I'm not familiar with ramda (but I take it it will move out of xud eventually) and only superficially familiar with currying functions so it's not exactly apparent to me what's going on in a few places. I think more comments/method block comments in the ethProvider file would be good to explain what everything is does, so it's more comprehensible in the future for others who might not be so familiar either.

jest.config.js
```diff
@@ -1,8 +1,9 @@
...
   globals: {
     'ts-jest': {
-      tsConfig: '<rootDir>/tsconfig.json'
+      tsconfig: '<rootDir>/tsconfig.json'
```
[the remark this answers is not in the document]
erkarl: `tsConfig` is deprecated in favour of `tsconfig`. After this change the deprecation warning disappeared.

.eslintrc.js
erkarl: Indeed, think it was something related to my environment. Can't get any linting diff when removing it. Removed the change.

jest.config.js
erkarl: > Were you having a problem running jest tests?

Indeed. It was the `npm run test:jest:watch` script that needed `testMatch` params. Removed this line `jest.config.js`

jest.config.js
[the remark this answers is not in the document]
sangaman: Ah yes after some more googling it looks like you're right: https://github.com/kulshekhar/ts-jest/pull/2127

If you want to leave this as a separate commit when you're squashing (perhaps along with the change to `npm run test:jest:watch`) I think that would make sense.

sangaman:
Let me know if/when I should review this again @erkarl .

erkarl:
> Let me know if/when I should review this again @erkarl .

Will do. I'm waiting for a new Connext release to test that everything is working correctly in this branch.

erkarl:
> That said, it doesn't look like this PR is introducing any new warnings, so I wouldn't let that hold this up.

Think it's not worth addressing them right now because once we've extracted `ConnextClient` into a separate module all code from current `ConnextClient.ts` should be ported over. Provided I don't introduce new warnings - these warning should disappear.

test/jest/Connext.spec.ts
erkarl: Created an issue for this https://github.com/ExchangeUnion/xud/issues/2046

erkarl:
> I think more comments/method block comments in the ethProvider file would be good to explain what everything is does, so it's more comprehensible in the future for others who might not be so familiar either.

@sangaman I added comments to ethProvider file - hope it is clearer now. Think this is ready for review again. I won't be adding more functionality to this PR.

michael1011 (approved):
utACK

lib/connextclient/ethprovider.ts
```diff
@@ -0,0 +1,119 @@
...
+const getERC20Balance = curry(
+  (address: string, contract: ethers.Contract): Observable<ethers.BigNumber> => {
+    return from(contract.balanceOf(address)) as Observable<ethers.BigNumber>;
+  },
+);
+
+// This is the main function that has to be called before this package exposes more functions.
+// Think of it as a constructor where we create the interal state of the module before
+// we export more functionality to the consumer.
+const getEthprovider = (host: string, port: number, name: string, chainId: number, seed: string) => {
```
sangaman, line 88: The smallest of nits, I'm seeing mixed casing of `Ethprovider` and `EthProvider`. I don't know which one is "correct" but might make sense to stick to one.

sangaman (approved):
Looks really good and comments are very helpful, thanks.

lib/connextclient/ethprovider.ts
erkarl: Thanks. Will address this in a follow-up PR.

PR #2049, opened 2020-12-22
fix(orderbook): remove local order id mappings

### LINKED ISSUE #1556: ALREADY_EXISTS error for id of maker order after self-match

### LINKED ISSUE #2045: Local order Ids still tracked after order is fully matched
Internally matched orders are removed from the trading pair queues but not from the global order book mapping between local order ids and global order ids. This is a memory leak and can also lead to some unexpected behavior, for example it's the underlying cause of #2030.

Related PR #1557.

sangaman:
This fix ensures that we remove the mapping between local id and global order id for orders that have been fully matched or fully removed from the order book. Previously, internally matched orders would be removed from the trading pair queues but not from the global order book mapping between local order ids and global order ids.

Fixes #2045. Fixes #1556.

### DIFF
--- a/lib/orderbook/OrderBook.ts
+++ b/lib/orderbook/OrderBook.ts
@@ -356,15 +356,9 @@ class OrderBook extends EventEmitter {
   private addTradingPair = (pairId: string) => {
     const tp = new TradingPair(this.logger, pairId, this.nomatching);
     this.tradingPairs.set(pairId, tp);
-    tp.on('peerOrder.dust', (order) => {
-      this.removePeerOrder(order.id, order.pairId);
-    });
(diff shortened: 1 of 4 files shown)


erkarl (approved):
Nice. utACK

raladev (approved):

PR #2050, opened 2020-12-28
feat(connext): change simnet connext router node public key

ghost:


erkarl:
Merging this as the simnet is currently broken without it and the change is minor.

CC @sangaman @kilrau @raladev

PR #2053, opened 2021-01-06
feat(rpc): OrderBook rpc call

### LINKED ISSUE #2048: Move `orderbook` construction logic to grpc
XUD UI needs to present the order book the same way `xucli orderbook` currently does. To avoid duplicate logic, we'd want to move the logic constructing the order book from `xucli` into xud to expose the order book directly via the grpc.

Wireframe screenshot to give an idea how the final result should look like on XUD UI side:
image

For more, see https://github.com/ExchangeUnion/xud-ui/issues/32.

sangaman:
This creates a new `OrderBook` rpc call to return a depth chart for the xud order book. This repurposes the logic that previously existed only in the `xucli orderbook` call which groups orders into buckets according to rounded price and moves that to the rpc layer, so that consumers of the API can get an xud depth chart without needing to reimplement the depth chart/bucket logic.

Closes #2048.

I haven't tested this yet so I'm leaving it in draft state for now.


raladev:
@sangaman can u please fix LINTing mistakers?

sangaman:
> @sangaman can u please fix LINTing mistakers?

Yes, I'm going to run some quick tests myself just didn't have the time to last night. I realize I didn't open this as a draft either so I'll just leave it for now, I'll ping you when I update the PR.

raladev (changes requested):
- [ ] buy and sell sides have incorrect sorting, it should work vice versa because i want to see top of buy orders (it is most expensive buy order) and top of sell orders (it is most cheapest sell order). For now these orders are not displayed in gui console table because of incorrect sort.
Screenshot from 2021-01-06 19-17-59
- [ ] price rounding does not work well for smallest and biggest orders, I was able to get order with 0 price and level with same price on both sides:
- smallest price of sell/buy order should always be rounded to top
- biggest price of buy orders should be always rounded to bottom
- smallest price of sell orders should be always rounded to top
Screenshot from 2021-01-06 19-50-08
Screenshot from 2021-01-06 19-25-29
- [ ] also i noticed that grpc request has no level bound parameter, would be good to add it because it may reduce size of data that should be transfered for the request. For example, for our xud-ui app we only need ~10 levels on each side, there is no reason in 0.00001, 0.00002 levels.

proto/xudrpc.proto
```diff
@@ -683,6 +694,29 @@ message Order {
   uint64 hold = 10 [json_name = "hold"];
 }
 
+message OrderBookRequest {
+  // The trading pair for which to retrieve orders.
+  string pair_id = 1;
+  // Tshe number of digits following the decimal point
```
krrprr, line 700: A little typo (`Tshe` -> `The`)

proto/xudrpc.proto
```diff
@@ -203,6 +203,17 @@ service Xud {
     };
   }
 
+  /* Gets an order book depth chart where orders are grouped into "buckets"
+   * according to their price rounded to a given level of precision.
+   * shell: xucli orderbook [pair_id] [precision] */
+  rpc OrderBook(OrderBookRequest) returns (OrderBookResponse) {
+    option (google.api.http) = {
+      post: "/v1/orderbook"
```
krrprr, line 211: Would it make sense as a get request instead?

krrprr (commented):
Looks good to me!
Found just a minor typo in a comment and left a question.

proto/xudrpc.proto
sangaman: I thought this too but the proto definition wouldn't accept `get` here since the request has a payload.

sangaman:
Sorry for the delay, but I've addressed the feedback and added tests, including a suite of performance tests that one can run via `npm run test:perf` - these aren't configured to run as part of GitHub actions but they can be used to quickly benchmark xud functions on a given machine. Organizing the order book depth chart with 20k orders is quite fast for me, below are typical results:

```
 PASS  test/perf/OrderBook.spec.ts (5.975 s)
  Order Book Performance Tests
    ✓ adds 10000 buy and 10000 sell orders to the order book (581 ms)
    ✓ listOrders with 20000 orders (49 ms)
    ✓ orderbook depth chart with 20000 orders (58 ms)
```

I fixed the sorting issue with the order book. However the rounding issue doesn't have a clear solution, the current and preexisting behavior is to round all prices to the nearest precision level. That's how a bucket's price can round down to zero, for example. 

I'm thinking that it actually makes sense to always round down buy orders and always round up sell orders. The thinking being that the order book depth chart should show how much volume there is to a certain price point. If the buy side shows price of 0.01 and quantity of 0.5, to me that means I should be able to sell at a price of 0.01 and get 0.05 filled. So when a buy bucket shows a price of zero, that just means that if I market sold at any price that's how much would get filled. On the sell side of things, the logic works in reverse. Let me know if that makes sense to you and I'll make this change to the bucket rounding logic (which would be a fix/change compared to the existing functionality).

raladev:
>I'm thinking that it actually makes sense to always round down buy orders and always round up sell orders. 
makes sense

> So when a buy bucket shows a price of zero, that just means that if I market sold at any price that's how much would get filled. 

Can we just change `0` to `<0.01` (for --precision 2). It definetly would be better then 0. if it hard to do on grpc side because of var type, we can just handle it in UI and cli.

sangaman:
> Can we just change `0` to `<0.01` (for --precision 2)? It definitely would be better then 0. if it hard to do on grpc side because of var type, we can just handle it in UI and cli.

Yeah that's a good idea, I'll put in that change now and squash the commits.

kilrau:
Sounds good to me!

sangaman:
I fixed the rounding to work as I described and made the cli display zero values as "smaller than the smallest bucket" such as `<0.01`. While I was working on that it occurred to me that we'll also want to round precision to the left of the decimal point sometimes, like with BTC/USDT for instance, so I made `precision` a signed variable on the rpc request.

raladev (approved):

krrprr (approved):
good to go for the UI :+1:

PR #2058, opened 2021-01-07
fix(connext): fix ethprovider network name

erkarl:
This PR ensures that we pass in `rinkeby` as the network name when running in testnet mode. Otherwise `ethers.providers.JsonRpcProvider` would fail to initialize.

Can be tested using https://github.com/ExchangeUnion/xud-docker/pull/819

### DIFF
--- a/lib/connextclient/ConnextClient.ts
+++ b/lib/connextclient/ConnextClient.ts
@@ -66,7 +66,7 @@ const getRouterNodeIdentifier = (network: string): string => {
       return 'vector5yw6vEx4VJPWMC4Vf48WSfY4bzgfJxAt4EGGFGT1o1mmQnRz35';
     case 'testnet':
       // public key of our testnet router node
-      return 'vector6XKj2XPrVEYfTEdxEjgVVnFCtZVXa4RrfzQLiXpjfcj4GJMdpR';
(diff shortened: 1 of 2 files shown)


michael1011 (commented):
Why is setting the chain id and network name required in the first place? If it does not work without that, it indicates that the connext endpoint is missing some information. Ethers should fetch that data on its own automatically

raladev (approved):

erkarl:
> Why is setting the chain id and network name required in the first place? If it does not work without that, it indicates that the connext endpoint is missing some information. Ethers should fetch that data on its own automatically

It seems to be a required type for `Network`:

```
export declare type Network = {
    name: string;
    chainId: number;
    ensAddress?: string;
    _defaultProvider?: (providers: any, options?: any) => any;
};
```

sangaman (approved):

michael1011:
> It seems to be a required type for Network:

Yeah but why do you set the network in the first place? In my experience, ethers fetches the values it needs automatically.

erkarl:
> > It seems to be a required type for Network:
> 
> Yeah but why do you set the network in the first place? In my experience, ethers fetches the values it needs automatically.

You're right. Removed it and it still works. Thanks.

michael1011 (approved):

michael1011 (approved):

PR #2061, opened 2021-01-11
refactor(p2p): JSON encoding to use protobuf instead

LePremierHomme:
Closing https://github.com/opendexnetwork/opendex/issues/34.

Changes:
* p2p packets' checksum to utilize their protobuf encoding instead of generating a JSON string.
* `SessionInit` p2p packet fields needed for signing to be encoded via the same protobuf's scheme of the entire packet, instead of JSON. Using a subset of the fields isn't a problem with protobuf - the used fields will be encoded while the remaining ones won't exist on the binary representation.

BREAKING CHANGE: JSON encoding to use protobuf instead


raladev (dismissed):

lib/p2p/packets/Packet.ts
```diff
@@ -104,10 +99,10 @@ abstract class Packet<T = any> implements PacketInterface {
   };
 
   /**
-   * Calculating the packet checksum using its JSON representation hash first 4 bytes.
+   * Calculating the packet checksum using its binary representation hash first 4 bytes.
    */
-  public checksum = (): number => {
-    return createHash('sha256').update(this.toJSON()).digest().readUInt32LE(0);
+  public checksum = (bytes: Uint8Array): number => {
```
sangaman, line 104: This is no longer using `this`, it's just a plain function that accepts a parameter. As such I'm thinking it might make sense to move this to a utils function.

Either that or it might make sense to remove the `bytes` parameter and just  do `.update(this.serialize())` to get the `UintArray` for this packet. Whichever makes more sense to you.

package.json
```diff
@@ -215,6 +214,7 @@
     "grpc-tools": "^1.10.0",
     "grpc_tools_node_protoc_ts": "^5.1.0",
     "jest": "^26.6.3",
+    "json-stable-stringify": "^1.0.1",
```
sangaman, line 217: It looks like this is only being used for the Parser.spec.ts suite, but it doesn't make much sense to me to use it for tests when it's not used in the actual functioning of xud. I think we can just change the line that expects the `stringify` outputs to match to instead check that the packet checksums match, and then get rid of this package entirely, right? If you'd like though I could do this in a follow-up PR instead.

sangaman (commented):
Looks good, just a few questions.

package.json
LePremierHomme: I was thinking of this as well, but there’s a small problem - protobuf scheme (in which checksum now relies on) is fixed (unlike JSON), and so it will only compare between the define fields, while in some of the negative test cases these are just a subset, because an additional (undefined in protobuf) field is added (e.g., `reqId`). 
Also, there’s a small benefit in comparing between the constructed POJO, and not the binary representation, which is an intermediate stage. I tried to use the built-in`JSON.stringify`, but the output isn’t deterministic and so many tests fail.

So the usage of `stringify` here isn’t related to its previous usage in xud. It’s just a tool for comparing between nested objects reliably. 
Let me know what you agree. If not, I can just change some of the negative tests.

lib/p2p/packets/Packet.ts
LePremierHomme: > This is no longer using `this`, it's just a plain function that accepts a parameter. As such I'm thinking it might make sense to move this to a utils function.

Good idea.

> Either that or it might make sense to remove the `bytes` parameter and just  do `.update(this.serialize())` to get the `UintArray` for this packet. Whichever makes more sense to you.

I didn't want to trigger `.update(this.serialize())` because in both use-cases we already have it ready, so i'd rather not re-generate it.

package.json
sangaman: Thanks for the explanation. That makes sense and I think it's OK to leave it for now, it's not a big deal to have as a dev dependency. I might take another look at this test suite some time to see if there's a sensible way to run it without this stringify function but no need to hold this PR up for that.

sangaman (approved):
Looks good, just need to squash the commit and merge whenever we are ready for the breaking change this introduces.

